PR #69, opened 2025-07-29
fix curvine-cli report inode/block error

201341:
Signed-off-by: swj <1186093704@qq.com>

### DIFF
--- a/curvine-server/src/master/fs/master_filesystem.rs
+++ b/curvine-server/src/master/fs/master_filesystem.rs
@@ -474,12 +474,21 @@ impl MasterFilesystem {
         for peer in &wm.conf.journal.journal_addrs {
             info.journal_nodes.push(peer.to_string())
         }
+        let files = match self.list_status("/") {
+            Ok(v) => v.len(),
+            Err(e) => {
+                warn!("Failed to list root directory: {}", e);
+                0
+            }
+        };
+        info.inode_num = files as i64;
 
         for (_, worker) in wm.worker_map.workers() {
             info.capacity += worker.capacity;
             info.available += worker.available;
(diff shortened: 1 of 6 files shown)


curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -434,12 +434,15 @@ impl MasterFilesystem {
         for peer in &wm.conf.journal.journal_addrs {
             info.journal_nodes.push(peer.to_string())
         }
+        let files = self.list_status("/").unwrap();
```
bigbigxu: Don't use unwrap here

curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -28,6 +28,7 @@ pub struct VfsDataset {
     worker_id: u32,
     ctime: u64,
     dir_list: DirList,
+    dir_block_nums: HashMap<u32, i64>,
```
bigbigxu: In fact, we can get the total number of blocks by directly counting the number of block_maps. dir_block_nums should not be needed, and this data is not updated when adding or deleting blocks.

szbr9486:
Great！

xuen ***@***.***> 于 2025年7月29日周二 19:50写道：

curvine-server/src/master/fs/master_filesystem.rs
201341: fix

curvine-server/src/worker/storage/vfs_dataset.rs
201341: fix

PR #70, opened 2025-07-29
feat: Add gcs and azblob support by OpenDAL

Xuanwo:
This PR adds GCS and AzBlob support via OpenDAL. It primarily serves as a demonstration of how OpenDAL can work with Curvine.

- More features like logging, trace, metrics, retry could be added in the future.
- More services like cos, obs could be added in the future too.


Xuanwo:
> The new interface, implemented using a macro and impl Future , offers better performance. The trait is defined in curvine_common::fs::FileSystem .

Thank you for the guidance. I'll give it a try.

---

Out of curiosity, could you elaborate a bit more on the performance issues? The difference in dynamic dispatch should be almost unnoticeable compared to IO.

PR #76, opened 2025-07-31
fix: listxattr return empty to avoid EIO (#79)

201341:
Signed-off-by: swj <1186093704@qq.com>


lzjqsdd:
Thank you very much for the improvement ！
         
We have just added a specification for commit information, which can be referenced https://github.com/CurvineIO/curvine/wiki/Contribute-convention

lzjqsdd:
At present, some cicd errors may be confusing, and there are still some problems with CICD submission validation, which are still being optimized

PR #91, opened 2025-08-01
fix: unsupported msg typo

cxzl25:


lzjqsdd:
Thanks!   The commit message format now needs to follow https://github.com/CurvineIO/curvine/blob/main/COMMIT_CONVENTION.md

Issue #37, opened 2025-07-21
Enhance Fuse interface implement


szbr9486:
Currently, Curvine only implements some of the Fuse interfaces. We now need to clarify which interfaces are already supported and which are pending support.

bigbigxu:
## FUSE Interface Support Status

### ✅ Supported Interfaces
- FUSE_INIT
- FUSE_LOOKUP
- FUSE_GETATTR
- FUSE_OPENDIR
- FUSE_STATFS
- FUSE_MKDIR
- FUSE_RELEASEDIR
- FUSE_READDIR
- FUSE_READDIRPLUS
- FUSE_FORGET
- FUSE_READ
- FUSE_FLUSH
- FUSE_OPEN
- FUSE_WRITE
- FUSE_CREATE
- FUSE_RELEASE
- FUSE_UNLINK
- FUSE_RMDIR
- FUSE_RENAME

### ❌ Unsupported Interfaces
- FUSE_ACCESS
- FUSE_GETXATTR
- FUSE_GETXATTR
- FUSE_SETATTR
- FUSE_REMOVEXATTR
- FUSE_FALLOCATE
- FUSE_MKNOD
- FUSE_BATCH_FORGET
- FUSE_INTERRUPT
- FUSE_LISTXATTR
- FUSE_FSYNC

lzjqsdd:
now support symlink and readlink

PR #135, opened 2025-08-06
feat: add access permission check (#37)

lzjqsdd:
Now supports permission validation for user groups/users

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -245,6 +245,140 @@ impl CurvineFileSystem {
         Ok(res)
     }
 
+    /// Check if the current user has the requested access permissions
+    fn check_access_permissions(
+        &self,
+        status: &FileStatus,
+        current_uid: u32,
+        current_gid: u32,
+        mask: u32,
+    ) -> bool {
+        let file_uid = self.resolve_file_uid(&status.owner);
+        let file_gid = self.resolve_file_gid(&status.group);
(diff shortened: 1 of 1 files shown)


lzjqsdd:
need fmt😅

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -245,6 +245,84 @@ impl CurvineFileSystem {
         Ok(res)
     }
 
+    /// Check if the current user has the requested access permissions
```
szbr00, line 248: The check_access_permissions function is too long. Could you split it to few of small functions？

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -245,6 +245,84 @@ impl CurvineFileSystem {
...
+
+        // Check each requested permission
+        let mut has_permission = true;
+
+        // Check read permission (R_OK = 4)
+        if (mask & libc::R_OK as u32) != 0 {
+            let has_read = (permission_bits & 0o4) != 0;
+            has_permission = has_permission && has_read;
+            debug!("Read permission check: requested={}, granted={}", (mask & libc::R_OK as u32) != 0, has_read);
+        }
```
szbr00, line 357: Add else...

lzjqsdd: This is a filter chain of multiple permission checks , and eles will continue to proceed downwards

szbr9486 (approved):

PR #142, opened 2025-08-07
Xuen

bigbigxu:
feat: add mount query interface configure (#132)


curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
```diff
@@ -199,4 +211,15 @@ public CurvineFsStat getFsStat() throws IOException {
         GetMasterInfoResponse info = GetMasterInfoResponse.parseFrom(bytes);
         return new CurvineFsStat(info);
     }
+
+    public MountPointInfo getMountPoint(Path path) throws IOException {
```
szbr00, line 215: getMountPointInfo ？

curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
```diff
@@ -199,4 +211,15 @@ public CurvineFsStat getFsStat() throws IOException {
...
+        System.out.println(path.toString());
```
szbr00: Use regular log？

curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
bigbigxu: Keep the same code as before

szbr9486 (approved):

PR #146, opened 2025-08-07
refactor: remove unnecessary lock operations in MountManager (#143)

bigbigxu:


curvine-common/examples/raft_node2.rs
```diff
@@ -39,7 +39,13 @@ fn main() -> CommonResult<()> {
 
     let log_store = RocksLogStorage::from_conf(&conf, true);
     let app_store: RocksAppStorage<String, String> = RocksAppStorage::new("../target/data/raft-2");
-    let raft = RaftJournal::new(rt.clone(), log_store, app_store.clone(), conf.clone());
+    let raft = RaftJournal::new(
+        rt.clone(),
+        log_store,
+        app_store.clone(),
+        RoleMonitor::new(),
```
szbr00: Suggest add the tail of method params if you add new param

curvine-common/src/raft/raft_journal.rs
zuston, line 1: Is this file modification related to this PR?

zuston (approved):
LGTM. However, could you please add a more detailed explanation about the reason behind this change, specifically why it is safe to proceed without using a lock?

lzjqsdd (approved):

PR #151, opened 2025-08-07
fix: enable ufs mount bucket directly(#152)

201341:


201341:
> The commit msg format should be: fix: enable ufs mount bucket directly(#issue_id) I did't find the issue no 151. Could create the issue about this pr, thanks.

done

PR #162, opened 2025-08-10
fix: Add Master readiness check in mini-cluster to prevent startup race conditions in test environment

jlon:
This PR fixes a race condition in the mini-cluster test environment where Workers would attempt to connect to the Master before its RPC service was fully initialized, causing "Failed to connect" warnings and test instability.

**Changes:**
- Added `wait_master_ready()` function to ensure Master is fully operational before starting Workers
- Implemented two-phase readiness check: Raft leader election + RPC service availability
- Added proper timeout and retry mechanisms with detailed logging

**Impact:**
- Eliminates connection refused errors during cluster startup
- Improves test reliability and stability
- Provides better visibility into cluster initialization process

### DIFF
--- a/curvine-server/src/test/mini_cluster.rs
+++ b/curvine-server/src/test/mini_cluster.rs
@@ -20,6 +20,7 @@ use curvine_common::conf::ClusterConf;
 use curvine_common::raft::{NodeId, RaftPeer};
 use curvine_common::FsResult;
 use dashmap::DashMap;
+use log::info;
 use orpc::client::RpcClient;
 use orpc::common::LocalTime;
 use orpc::io::net::{InetAddr, NetUtils};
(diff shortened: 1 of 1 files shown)


curvine-server/src/test/mini_cluster.rs
```diff
@@ -75,24 +76,112 @@ impl MiniCluster {
...
+        let mut retry_count = 0;
+
+        info!(
+            "Waiting for all Workers to register, expected Worker count: {}",
+            self.worker_conf.len()
+        );
+
         while LocalTime::mills() <= wait_time {
+            retry_count += 1;
             let info = fs.get_master_info().await?;
```
szbr00, line 163: change info to master_info

szbr9486 (approved):

PR #163, opened 2025-08-10
chore: get current dir to execute script

QuantumBear:
restart-all.sh: based on curr_dir to run scripts, avoid error when run on a different directory.

add shebang


bigbigxu:
Using curr_dir="$(pwd)" to get the path might not be a good idea. If you want to execute restart.sh from any directory, it would be better to get the current path using "$(cd "$(dirname "$0")"/..; pwd)".

PR #164, opened 2025-08-10
fix: remove credential of cicd compile images

lzjqsdd:
change curvine-compile to public and remove credential to avoid community members can't get secret of curvine to pull image when trigger github actions


lzjqsdd:
remove subject-case always low-case commitlint

szbr9486 (approved):

PR #170, opened 2025-08-11
fix: failed to load so library

bigbigxu:


curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java
```diff
@@ -148,7 +148,6 @@ public static void load() {
             return;
         } catch (UnsatisfiedLinkError e) {
             // pass
-            LOGGER.warn("xxx {} {}", libraryName, e);
```
szbr00, line 151: why delete this warning log info

curvine-libsdk/src/lib.rs
```diff
@@ -25,4 +25,4 @@ mod lib_fs_reader;
 pub use self::lib_fs_reader::LibFsReader;
 
 pub mod java;
-pub mod python;
+// pub mod python;
```
szbr00: delete

curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java
bigbigxu: This is the debug log, which was submitted by mistake before.

curvine-libsdk/src/lib.rs
bigbigxu: fix it

curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java
szbr00: It is true that some error logs need to be printed here

curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java
```diff
@@ -172,7 +171,6 @@ public static void load() {
             LOGGER.info("Loaded {} by System.loadLibrary", filename);
             System.load(filename);
         } catch (Throwable e) {
-            LOGGER.warn("xxx {} {}", libraryName, e);
```
szbr00, line 175: Need log info

curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java
bigbigxu: fix it

szbr9486 (approved):

Issue #117, opened 2025-08-05
[FEATURE] Optimize the mount point query interface


bigbigxu:
1. Support querying for both UFS and Curvine paths simultaneously
2. Split the path and query possible mount points

PR #172, opened 2025-08-11
feat: add more options on mount

zuston:


zuston:
Could you help take a look? @bigbigxu I think I will add some tests to validate this. Do you have some existing tests that I can extend

curvine-server/src/master/mount/mount_entry.rs
```diff
@@ -28,12 +29,18 @@ pub struct MountPointEntry {
...
+    pub(crate) block_size: u64,
+    pub(crate) replicas: u32,
+    pub(crate) storage_type: StorageType,
 }
 
 impl MountPointEntry {
     pub fn new(id: u32, curvine_uri: String, ufs_uri: String, mnt_opt: MountOptions) -> Self {
         let consistency_conf = mnt_opt.consistency_config.unwrap_or_default();
         let strategy = ConsistencyStrategy::from(consistency_conf);
+        let block_size = mnt_opt.block_size.unwrap_or_default();
```
bigbigxu, line 41: unwrap_or_default should not be used here, option should still be used.

curvine-server/src/common/ufs_manager.rs
```diff
@@ -60,7 +60,18 @@ impl UfsManager {
         Ok(ufs_conf)
     }
 
-    async fn get_mount_point(&mut self, ufs_base_uri: &String) -> FsResult<MountPointInfo> {
+    pub async fn get_mount_point_with_uri(
```
bigbigxu, line 63: This interface has been reimplemented in https://github.com/CurvineIO/curvine/issues/117

curvine-server/src/worker/load/ufs_connector.rs
```diff
@@ -211,6 +218,11 @@ impl CurvineFsWriter {
             opts_builder = opts_builder.ttl_action(action);
         }
 
+        opts_builder = opts_builder
```
bigbigxu, line 221: When parameters such as block_size are None, the configuration in the curvine-cluster.toml file should be used.

curvine-server/src/master/mount/mount_entry.rs
zuston: If so, users should always specify this option when mounting

curvine-server/src/worker/load/ufs_connector.rs
zuston: From my perspective, the curvine-cluster.toml file specified at startup may be inconsistent with the master’s configuration, especially if some workers start later than others. According to the central control design, workers should always retrieve their configurations from the master.

Based on the above assumption, block_size should never be None; this is ensured by the master when retrieving the mount point information.

What do you think?

bigbigxu: However, the block size of the mount point can be left unset. If not set, the client configuration of the current node should be used.

zuston: yes. I think the block size should always be set by the master side. If user don't specify this when mounting, master will set this according to the client conf in this master node.

szbr9486:
BTW,  could you add the usage in doc

curvine-server/src/worker/load/ufs_connector.rs
zuston: > yes. I think the block size should always be set by the master side. If user don't specify this when mounting, master will set this according to the client conf in this master node.

Is this OK? @bigbigxu

lzjqsdd:
> Aha, this PR haven't add test cases (but merged). I will add tests and doc in the later PRs. @bigbigxu

maybe we can use [WIP] state in pr title to if this item havn't completed 😁

PR #179, opened 2025-08-12
fix: illegal option hostname on macos

201341:


etc/curvine-env.sh
```diff
@@ -18,9 +18,9 @@
 
 export CURVINE_HOME="$(cd "$(dirname "$0")"/..; pwd)"
 
-# Get the IP address from hostname, taking the last network interface address
+# Get the IP address from ifconfig, taking the last network interface address
 LOCAL_HOSTNAME="localhost"
-LOCAL_IP=$(hostname -I | awk '{print $NF}')
+LOCAL_IP=$(ifconfig | grep "inet " | awk '{print $2}' | grep -v 127.0.0.1 | tail -n 1)
```
szbr00, line 23: Is it more concise replaced as 
hostname -i | awk 'print $NF'

201341: on macos, hostname doen't have the -i option.

szbr9486 (approved):

etc/curvine-env.sh
```diff
@@ -18,9 +18,9 @@
...
 
 # master bound host name
```
lzjqsdd, line 25: it would be better to check the system version to select the detection method according to different systems.

201341: get, I will improve this in next pr.

PR #191, opened 2025-08-14
feat: support additional parameters when mounting fuse

jlon:
support additional parameters when mounting fuse


lzjqsdd (approved):

curvine-fuse/src/bin/curvine-fuse.rs
```diff
@@ -105,6 +286,10 @@ impl FuseArgs {
         Ok(conf)
```
szbr00, line 286: Record conf info into log file

PR #203, opened 2025-08-15
feat: add replica num check logical implemention

szbr9486:


curvine-server/src/master/fs/replica_checker.rs
```diff
@@ -0,0 +1,147 @@
...
+        let fs_dir = fs_dir_lock.read();
+        let rocks = fs_dir.get_rocks_store();
+
+        // Iterate all inodes; count under-replicated blocks
+        let mut under_replicated_blocks: i64 = 0;
+        let mut under_replicated_files: i64 = 0;
+        let mut scanned_files: i64 = 0;
+        let mut scanned_blocks: i64 = 0;
+
+        let iter = match rocks.inodes_iter() {
```
bigbigxu, line 59: To retrieve the lost replicas, I think we should first obtain the lost worker, then find all the files corresponding to the replicas on this worker, and then replicate them.

szbr00: The pattern you are talking about is worker-level fault tolerance checking, and this feature focuses on copy checking of file blocks.

szbr9486:
> In this PR, the checker currently only verifies block replicas and does not take further action.在此 PR 中，检查器目前仅验证块副本，不采取进一步作。 Looking ahead, could it be extended to handle missing blocks once their replication factor is satisfied?展望未来，一旦满足复制因子，是否可以扩展到处理丢失的块？ If so, should we consider renaming it to BlockReplicationManager to better reflect its expanded responsibilities?如果是这样，我们是否应该考虑将其重命名为 BlockReplicationManager 以更好地反映其扩展的职责？

OK. Make sense

zuston:
I want to try this to re-implement this.

PR #215, opened 2025-08-20
fix: invalid deleting when removing by worker_id in rocks store

zuston:


curvine-server/src/master/meta/store/rocks_inode_store.rs
```diff
@@ -59,9 +58,9 @@ impl RocksInodeStore {
     }
 
     // Get all location information for all block ids.
-    pub fn get_locations(&self, id: i64) -> CommonResult<Vec<BlockLocation>> {
-        let prefix = RocksUtils::i64_to_bytes(id);
-        let iter = self.db.prefix_scan(Self::CF_LOCATION, prefix)?;
```
zuston, line 64: To ensure compatibility, I kept `CF_LOCATION` and `CF_BLOCK` unchanged, instead of renaming them to `CF_LOCATION_TO_BLOCKS` and `CF_BLOCK_TO_LOCATIONS.`

bigbigxu (approved):

PR #218, opened 2025-08-21
feat(1/N): add support of block replication

zuston:


bigbigxu:
> This is a major feature. I will keep the development progress open in this PR for reviewers. cc @bigbigxu

Fix the Clippy check errors. You can use the following command to check and attempt to fix them:
```
cargo clippy --release --fix  --allow-dirty --allow-staged --allow-no-vcs -- --deny warnings --allow clippy::uninlined-format-args
```
Some errors may need to be fixed manually.

zuston:
> There are still errors: https://github.com/CurvineIO/curvine/actions/runs/17123516579/job/48569767112?pr=218

Yes. After all ready, I will fix all

curvine-common/src/state/worker_address.rs
```diff
@@ -49,3 +50,27 @@ impl Display for WorkerAddress {
         )
     }
 }
+
+impl From<WorkerAddress> for WorkerAddressProto {
```
bigbigxu: Use the functions of ProtoUtils to handle address conversion.
Strictly speaking, Proto data structures should only appear in one place and should not be used elsewhere.

zuston: done

bigbigxu (approved):

Issue #227, opened 2025-08-29
[FEATURE] Data loading module code optimization


bigbigxu:
1. Load has been upgraded to a job to support subsequent expansion of other task types, such as refreshing data to ufs.
2. Load data, reading parameters such as TTL and replicas from the mount point configuration file.

PR #228, opened 2025-08-29
feat: data loading module code optimization (#227)

bigbigxu:


curvine-common/proto/job.proto
```diff
@@ -0,0 +1,83 @@
...
+package proto;
+
+option java_package = "io.curvine.proto";
+option java_multiple_files = true;
+option java_outer_classname = "LoadProto";
+
+import "common.proto";
+
+// job and task status enumeration
+enum WorkStateProto {
```
szbr00: WorkState->JobState maybe better？

curvine-common/proto/job.proto
```diff
@@ -0,0 +1,83 @@
...
+  WORK_STATE_PROTO_PENDING = 0;
```
szbr00: PENDING is Ok

curvine-common/proto/job.proto
```diff
@@ -0,0 +1,83 @@
...
+  WORK_STATE_PROTO_LOADING = 1;
```
szbr00: Ditto.  LOADING is Ok

curvine-common/proto/job.proto
szbr00: State enum number start with 1 maybe better? Zero means abnormal state.

curvine-common/src/conf/job_conf.rs
```diff
@@ -0,0 +1,86 @@
...
+impl Default for JobConf {
+    fn default() -> Self {
+        Self {
+            job_life_ttl: Default::default(),
+            job_life_ttl_str: "6h".to_string(),
+
+            job_cleanup_ttl: Default::default(),
+            job_cleanup_ttl_str: "10m".to_string(),
+
+            job_max_files: 100000,
```
szbr00: 100000 define as a constant variable

curvine-common/src/conf/job_conf.rs
```diff
@@ -0,0 +1,86 @@
...
+
+            task_timeout: Default::default(),
+            task_timeout_str: "1h".to_string(),
+
+            task_report_interval: Default::default(),
+            task_report_interval_str: "5s".to_string(),
+
+            worker_max_concurrent_tasks: 1000,
```
szbr00: As before

curvine-server/src/worker/task/task_store.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::ops::Deref;
+use std::sync::Arc;
+
+use curvine_common::state::LoadTaskInfo;
+use orpc::sync::FastDashMap;
+
+use crate::worker::task::TaskContext;
+
+#[derive(Clone)]
+pub struct TaskStore {
```
szbr00, line 24: What is the difference between job and task？

curvine-server/src/worker/task/task_manager.rs
```diff
@@ -0,0 +1,98 @@
...
+    rt: Arc<Runtime>,
+    fs: CurvineFileSystem,
+    tasks: TaskStore,
+    sender: AsyncSender<Arc<TaskContext>>,
+    progress_interval_ms: u64,
+    task_timeout_ms: u64,
+}
+
+impl TaskManager {
+    pub fn with_rt(
```
szbr00, line 45: Add annotations for the with_rt function

curvine-server/src/worker/task/task_store.rs
bigbigxu: The job will be split into multiple tasks and distributed to different workers for execution.

curvine-common/proto/job.proto
bigbigxu: Both job and task have state, so a compromise name is used.

PR #234, opened 2025-09-01
feat: implements s3-compatible api gateway for curvine

jlon:
## Summary
Implements a complete S3-compatible API gateway that provides AWS S3 API access to Curvine.

### Bucket Operations
- **CreateBucket**: Create new storage buckets
- **HeadBucket**: Check bucket existence and permissions  
- **ListBuckets**: List all available buckets
- **DeleteBucket**: Remove empty buckets
- **GetBucketLocation**: Retrieve bucket region information

### Object Operations  
- **PutObject**: Upload files with custom metadata support
- **GetObject**: Download files with streaming support
- **HeadObject**: Retrieve object metadata without downloading
- **DeleteObject**: Remove objects from buckets
- **ListObjects**: List objects in bucket with prefix filtering

### Advanced Features
- **Range Requests**: Support for partial content downloads
  - Normal ranges: `bytes=0-100`
  - Open ranges: `bytes=100-`  
  - Suffix ranges: `bytes=-100`
- **Custom Metadata**: Store and retrieve `x-amz-meta-*` headers
- **Streaming I/O**: Memory-efficient processing of large files (tested up to 100MB)
- **Concurrent Operations**: Thread-safe parallel upload/download

## Performance & Compatibility
- **Memory Efficient**: Fixed 4KB buffer usage regardless of file size
- **High Throughput**: ~96MB/s upload, ~213MB/s download performance  
- **Error Handling**: Proper HTTP status codes (404, 409, 416) with S3-compliant responses
- **AWS Tool Compatibility**: Works with AWS CLI, boto3, and other S3 clients

This implementation enables Curvine to serve as a drop-in replacement for AWS S3 in existing applications and tools.


build/bin/curvine-gateway.sh
szbr00, line 1: Change file name as curvine-s3-gateway.sh

build/bin/curvine-gateway.sh
```diff
@@ -0,0 +1,371 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Source environment variables
+. "$(cd "`dirname "$0"`"; pwd)"/../conf/curvine-env.sh
+
+# Service configuration
+SERVICE_NAME="curvine-gateway"
```
szbr00: curvine-s3-gateway

curvine-gateway/src/utils/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+                    *head = Some(parse_chunk_line(&content[0..pos]).map_err(|_| {
+                        log::warn!("parse content line error\n{}", unsafe {
+                            std::str::from_utf8_unchecked(&content[0..pos])
+                        });
+                        ChunkParseError::IllegalContent
+                    })?);
+                    if let Some(hdr) = head {
+                        if hdr.content_size == 0 {
+                            let next=circle_hasher.next(
+                                    "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
```
szbr00: Here is the question of why will have this fixed string

curvine-common/src/conf/cluster_conf.rs
```diff
@@ -247,6 +250,34 @@ impl Display for ClusterConf {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(default)]
+pub struct S3GatewayConf {
+    /// Listen address for S3 gateway (host:port)
+    pub listen: String,
+    /// Region to report via GetBucketLocation
+    pub region: String,
+    /// Temporary directory for multipart uploads
+    pub multipart_temp: String,
+    /// S3 access key (optional, falls back to environment variables)
```
szbr00: There is no need to annotate every variable

curvine-s3-gateway/src/auth/mod.rs
szbr00, line 1: Streamline the annotation！

build/bin/curvine-gateway.sh
jlon: yes, i will fix it

curvine-common/src/conf/cluster_conf.rs
jlon: yes, i will fix it

curvine-gateway/src/utils/mod.rs
jlon: This string is the SHA256 hash of the empty string. In S3 chunked uploads, when an empty chunk (content_size = 0) is encountered, the SHA256 hash of the empty string is used as the payload hash.

jlon: This is a standard requirement of the AWS S3 protocol. Each block transfer must end with an empty block, which is part of the protocol specification and is used to mark the completion of the transfer and perform final signature verification.

jlon:
First of all, why not use MinIO Gateway + FUSE directly? Because there is a performance bottleneck. The FUSE layer adds additional system call overhead. Each S3 operation must go through: S3 client → MinIO Gateway → FUSE → kernel → Curvine. The multi-layer conversion leads to increased latency and CPU consumption. In addition, the file system semantics of FUSE do not fully match the object storage semantics of S3, and it is difficult to ensure the consistency of concurrent access. The suggestion about Curvine worker s3 Agent is indeed valuable, and support will be considered in scenarios with higher performance requirements in the future. It is still necessary to provide a centralized gateway at present. Many machine learning and AI training directly use the s3 client. Curvine provides a standard s3 gateway to make it more convenient for users to access.

curvine-gateway/src/utils/mod.rs
szbr00: It is recommended to use this string as a constant and add an annotation

szbr9486 (approved):

Issue #229, opened 2025-08-29
[FEATURE] Added client metrics timing reporting mechanism


bigbigxu:
1. Regularly report client metrics to the master.
2. When the FileSystem is closed, metrics are reported within a limited timeout.

PR #236, opened 2025-09-02
feat: added client metrics timing reporting mechanism (#229)

bigbigxu:


curvine-client/src/client_metrics.rs
```diff
@@ -12,28 +12,94 @@
...
         };
+
         Ok(cm)
     }
 
     pub fn text_output(&self) -> CommonResult<String> {
         Metrics::text_output()
     }
+
+    pub fn encode() -> CommonResult<Vec<MetricValue>> {
```
szbr00, line 51: Will this function be called frequently？
There is an nested loop in this function.  Maybe that will be an performance black hole.

curvine-client/src/file/curvine_filesystem.rs
```diff
@@ -37,6 +40,8 @@ pub struct CurvineFileSystem {
 }
 
 impl CurvineFileSystem {
+    pub const CLOSE_TIMEOUT_SECS: u64 = 5;
```
szbr00: todo: make this variable into configuration

curvine-client/src/unified/mount_cache.rs
```diff
@@ -11,23 +11,30 @@ use std::time::Duration;
 pub struct MountValue {
     pub info: Arc<MountInfo>,
     pub ufs: UfsFileSystem,
+    pub mount_id: String,
```
szbr00, line 14: This variable duplicates the mount id in MountInfo

curvine-server/src/master/master_metrics.rs
```diff
@@ -98,6 +102,35 @@ impl MasterMetrics {
 
         Metrics::text_output()
     }
+
+    pub fn get_or_register(&self, value: &MetricValue) -> CommonResult<CounterVec> {
```
szbr00, line 106: The meaning of this function name is a bit ambiguous

orpc/src/common/metrics.rs
```diff
@@ -12,46 +12,62 @@
...
-            Some(v) => {
-                match v.register(c) {
-                    Ok(_) => (),
-                    Err(e) => warn!("register {}", e),
-                }
-                Ok(())
-            }
-
-            None => err_box!("Prometheus registry not init"),
+    pub fn name(&self) -> &str {
```
szbr00: get_name maybe better?

curvine-client/src/client_metrics.rs
bigbigxu: This is to obtain the metrics that need to be reported. The call is not frequent and is sent once every 10 seconds by default.

curvine-client/src/unified/mount_cache.rs
bigbigxu: The default mount id is u32 type, which is used to reduce memory copying.

curvine-server/src/master/master_metrics.rs
bigbigxu: Here, the metric reported by the client will be automatically registered and the metric execution value will be updated.

szbr9486 (approved):

PR #242, opened 2025-09-04
fix: ufs filetype miss and get_status error

201341:


curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -315,14 +315,24 @@ impl FileSystem<UnifiedWriter, UnifiedReader, ClusterConf> for UnifiedFileSystem
     async fn get_status(&self, path: &Path) -> FsResult<FileStatus> {
         match self.get_mount(path).await? {
             None => self.cv.get_status(path).await,
-            Some((ufs_path, mount)) => mount.ufs.get_status(&ufs_path).await,
+            Some((ufs_path, mount)) => {
+                if mount.info.cv_path == path.path() {
```
bigbigxu, line 319: This destroys the semantics of transparent transmission. Even if you want to access the root directory of the mount point, you still need to access ufs directly.

curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -315,14 +315,24 @@ impl FileSystem<UnifiedWriter, UnifiedReader, ClusterConf> for UnifiedFileSystem
...
+                mount.ufs.get_status(&ufs_path).await
+            }
         }
     }
 
     async fn get_status_bytes(&self, path: &Path) -> FsResult<BytesMut> {
         match self.get_mount(path).await? {
             None => self.cv.get_status_bytes(path).await,
-            Some((ufs_path, mount)) => mount.ufs.get_status_bytes(&ufs_path).await,
+            Some((ufs_path, mount)) => {
```
bigbigxu, line 330: This destroys the semantics of transparent transmission. Even if you want to access the root directory of the mount point, you still need to access ufs directly.

curvine-client/src/unified/unified_filesystem.rs
201341: For ufs mount point, it is created on the cv filesystem, when detecting the mount point path, get status results should  come from cv, not ufs. It's a bug fix

curvine-client/src/unified/unified_filesystem.rs
201341: ditto

curvine-client/src/unified/unified_filesystem.rs
bigbigxu: In Curvine, a mount point is simply a virtual directory mapped directly to the UFS. All metadata access must be transparently passed to the UFS.

If you need to view files in Curvine, you can do so by setting enable_unified to false.

201341: When I set enable_unified to true, the ufs mount point is created on cv filesystem, For fuse, need to lookup the ufs mount point first, and then fuse can lookup next level. When I mount the s3 bucket on the cv path, the ufs root dir can't be found through ufs directly, so there must be a special handle.

bigbigxu: you are right

bigbigxu (approved):

PR #246, opened 2025-09-08
feat: cli Add mv command

201341:


bigbigxu (approved):

curvine-ufs/src/s3/s3_filesystem.rs
```diff
@@ -348,7 +371,27 @@ impl FileSystem<S3Writer, S3Reader> for S3FileSystem {
     }
 
     async fn rename(&self, _src: &Path, _dst: &Path) -> FsResult<bool> {
-        err_ufs!("Not supported")
+        if _src.is_root() || _dst.is_root() {
```
lzjqsdd: _src，_dst  no longer need _ prefix if parameter is used

201341: fix

PR #256, opened 2025-09-10
refactor: optimize data loading code to solve the problem that block_on may cause deadlock

bigbigxu:


curvine-cli/src/cmds/fs/blocks.rs
```diff
@@ -0,0 +1,123 @@
...
+                    "json" => Self::print_json(&file_blocks),
+                    _ => Self::print_table(&file_blocks),
+                }
+
+                Ok(())
+            }
+        }
+    }
+
+    fn print_table(file_blocks: &FileBlocks) {
```
szbr00, line 50: print table？ what’s the table means

curvine-common/proto/master.proto
```diff
@@ -135,13 +135,14 @@ message GetMasterInfoResponse {
...
     repeated string journal_nodes = 2;
 
-    required int64 inode_num = 3;
-    required int64 block_num = 4;
-    required int64 capacity = 5;
-    required int64 available = 6;
-    required int64 fs_used = 7;
-    required int64 non_fs_used = 8;
-    required int64 reserved_bytes = 9 [default = 0];
+    required int64 inode_dir_num = 3;
```
szbr00, line 138: To maintain protocol compatibility, it's best to put the new fields at the end

curvine-common/src/conf/job_conf.rs
```diff
@@ -58,7 +58,7 @@ impl JobConf {
     pub const DEFAULT_JOB_MAX_FILES: usize = 100000;
     pub const DEFAULT_TASK_TIMEOUT: &'static str = "1h";
     pub const DEFAULT_TASK_REPORT_INTERVAL: &'static str = "10s";
-    pub const DEFAULT_WORKER_MAX_CONCURRENT_TASKS: usize = 1000;
```
szbr00, line 61: what's the difference  between 1k and 100

curvine-cli/src/cmds/fs/blocks.rs
bigbigxu: The output format is in table form and also supports json format

curvine-common/src/conf/job_conf.rs
bigbigxu: Loading 1k files at the same time may be too much and may easily use up the bandwidth of the worker node.

PR #266, opened 2025-09-12
feat: added support for falling back to ufs when reading curvine data fails

bigbigxu:


curvine-libsdk/java/src/main/java/io/curvine/CurvineFallbackInputStream.java
```diff
@@ -0,0 +1,148 @@
...
+    public long getPos() throws IOException {
+        if (ufsInputStream != null) {
+            return ufsInputStream.getPos();
+        } else {
+            return cvInputStream.getPos();
+        }
+    }
+
+    @Override
+    public boolean seekToNewSource(long l) throws IOException {
```
szbr00, line 136: Add some log info？

PR #285, opened 2025-09-17
feat: add s3a transparent proxy (#276)

bigbigxu:


curvine-client/src/unified/mount_cache.rs
szbr00, line 1: Add necessary annotations

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,196 @@
...
+public class CurvineFilesystemProvider implements Closeable {
+    public static final Logger LOGGER = LoggerFactory.getLogger(CurvineFilesystemProvider.class);
+
+    public static String ZK_SERVER = "fs.cv.proxy.zk.server";
+    public static String ZK_PATH = "fs.cv.proxy.zk.path";
+    public static String CACHE_ENABLE = "fs.cv.proxy.enable";
+    public static String CACHE_TIMEOUT = "fs.cv.proxy.timeout";
+    public static String CLUSTER_NAME = "fs.cv.proxy.cluster.name";
+
+    public static String ZK_ENABLE_VALUE = "1";
```
szbr00: What's the ZK_ENABLE_VALUE means? Why used string value

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,196 @@
...
+            try {
+                nodeCache.close();
+            } catch (IOException e) {
+                // pass
+            }
+            nodeCache = null;
+        }
+
+        if (zkClient != null) {
+            zkClient.close();
```
szbr00: Catch possible exceptions

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,196 @@
...
+    private NodeCache nodeCache;
+    private final AtomicBoolean enable;
+
+    private CurvineFileSystem fs;
+
+    public CurvineFilesystemProvider(Configuration conf) {
+        zkServer = conf.get(ZK_SERVER, "");
+        zkPath = conf.get(ZK_PATH, "/curvine/proxy_enable");
+        proxyEnable = conf.getBoolean(CACHE_ENABLE, true);
+        proxyTimeout = (int) conf.getTimeDuration(CACHE_TIMEOUT, "60s", TimeUnit.MILLISECONDS);
```
szbr00: Is it possible to throw an exception here?

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,196 @@
...
+        } else {
+            setupZkWatcher();
+        }
+    }
+
+    private void setupZkWatcher() {
+        try {
+            zkClient = CuratorFrameworkFactory.newClient(
+                    zkServer,
+                    proxyTimeout,
```
szbr00, line 92: Both parameters are the same, maybe should make them difference

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,196 @@
...
+                enable.set(ZK_ENABLE_VALUE.equals(value));
+            }
+        } catch (Exception e) {
+            enable.set(false);
+            LOGGER.warn("Error handling ZooKeeper data change", e);
+        }
+    }
+
+    public CurvineFileSystem getFs() {
+        if (!enable.get()) {
```
szbr00: Print warning log info

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,196 @@
...
+                LOGGER.warn("createCurvineFileSystem", e);
+            }
+        }
+        return fs;
+    }
+
+    private CurvineFileSystem createCurvineFileSystem() throws Exception {
+        CompletableFuture<CurvineFileSystem> future = CompletableFuture.supplyAsync(() -> {
+            try {
+                String scheme = "cv";
```
szbr00: Maybe the filesystem prefix scheme could be mutable

szbr00: Anyway, the scheme should be a constants variable

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: Indicates whether the function is enabled. ZK stores bytes, so using strings is simpler and more direct.

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: If the configuration is wrong, it is reasonable to throw an exception during initialization

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: They should be identical, hopefully failing fast on ZooKeeper exceptions.

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: This method needs to be called every time a file is opened, which will result in a large amount of invalid logs.

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
szbr00: It's easy to encounter problems when throwing exceptions in the constructor

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
szbr00: You mean，enable is false normally？

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,209 @@
...
+    private final String zkPath;
+    private final boolean proxyEnable;
+    private final int proxyTimeout;
+    private final String clusterName;
+
+    private final Configuration conf;
+
+    private CuratorFramework zkClient;
+    private NodeCache nodeCache;
+    private final AtomicBoolean enable;
```
szbr00: enable -> enabled

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class CurvineFilesystemProvider implements Closeable {
+    public static final Logger LOGGER = LoggerFactory.getLogger(CurvineFilesystemProvider.class);
+
+    public static String ZK_SERVER = "fs.cv.proxy.zk.server";
+    public static String ZK_PATH = "fs.cv.proxy.zk.path";
+    public static String CACHE_ENABLE = "fs.cv.proxy.enable";
```
szbr00: CACHE_ ENABLE->CACHE_ENABLED
fs.cv.proxy.enable->fs.cv.proxy.enabled

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
szbr00: Could it be int type?

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: This has nothing to do with the default value, it depends on the user behavior

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: This is essentially a judgment value, converted to boolean, it seems unnecessary to convert it again?

curvine-libsdk/java/src/main/java/io/curvine/CurvineFilesystemProvider.java
bigbigxu: What should I do if the configuration is wrong?

szbr9486 (approved):

PR #291, opened 2025-09-18
feat: add hostname validation for master

jlon:
feat: add hostname validation for master


curvine-common/src/conf/cluster_conf.rs
```diff
@@ -79,11 +79,32 @@ impl ClusterConf {
         // Check the environment variable configuration.
         if let Ok(v) = env::var(Self::ENV_MASTER_HOSTNAME) {
             conf.master.hostname = v.to_owned();
+            let hostname_exists = conf
```
bigbigxu, line 82: The hostname value configured in the master should be checked; the user may not have configured the environment variable.

lzjqsdd (approved):

PR #301, opened 2025-09-22
feat: add cache-only options for cv command

ziqonghuang:


curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -333,6 +337,9 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
     }
 
     async fn get_status(&self, path: &Path) -> FsResult<FileStatus> {
+        if !self.enable_read_ufs {
```
bigbigxu: Whether the unified access interface is enabled is controlled by the enable_unified variable, not enable_read_ufs.
This check is already done in the get_mount method, so it's unnecessary here.

curvine-cli/src/cmds/fs/cat.rs
```diff
@@ -10,13 +10,19 @@ pub enum CatCommand {
...
         path: String,
+
+        #[clap(short = 'c', long, help = "Query only Curvine cached data")]
+        cache_only: bool,
     },
 }
 
 impl CatCommand {
-    pub async fn execute(&self, client: UnifiedFileSystem) -> CommonResult<()> {
+    pub async fn execute(&self, mut client: UnifiedFileSystem) -> CommonResult<()> {
```
bigbigxu: I think we should add a global parameter to the fs command to control whether to enable the unified access interface, instead of adding one parameter to each command.

bigbigxu (approved):

PR #332, opened 2025-10-15
fix: vfs test fail

lzjqsdd:


curvine-server/src/worker/storage/vfs_dir.rs
```diff
@@ -121,16 +121,12 @@ impl VfsDir {
...
-        let disk_used = self.stats.used_space() as i64;
         let disk_available = self.stats.available_space() as i64;
 
         let capacity = self.capacity();
         let fs_used = self.fs_used();
         let reserved_bytes = self.reserved_bytes;
-        let non_fs_used = disk_used - fs_used;
-
-        let calculated_available = capacity - fs_used - non_fs_used - reserved_bytes;
+        let calculated_available = capacity - fs_used - reserved_bytes;
```
bigbigxu, line 129: Calculating non_fs_used is necessary and should not be removed, especially in shared disk scenarios, as this will result in incorrect free space calculations.

PR #334, opened 2025-10-15
fix: fixed the error of adding subtasks infinitely when loading folders on ufs

jlon:
fixed the error of adding subtasks infinitely when loading folders on ufs

### DIFF
--- a/curvine-server/src/master/job/job_runner.rs
+++ b/curvine-server/src/master/job/job_runner.rs
@@ -28,7 +28,6 @@ use futures::future;
 use log::{error, info, warn};
 use orpc::common::{ByteUnit, FastHashMap, FastHashSet, LocalTime, Utils};
 use orpc::err_box;
-use std::collections::LinkedList;
 use std::sync::Arc;
 
 pub struct LoadJobRunner {
@@ -175,8 +174,8 @@ impl LoadJobRunner {
             .into_iter()
             .map(|(id, task)| async move {
                 let client = self.factory.get_worker_client(&task.task.worker).await?;
-                client.submit_load_task(task.task).await?;
-                info!("Submit sub-task {}", id);
+                client.submit_load_task(task.task.clone()).await?;
+                info!("Submit sub-task {} for file: {}", id, task.task.source_path);
                 Ok::<(), FsError>(())
             })
             .collect();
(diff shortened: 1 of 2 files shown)


curvine-server/src/master/job/job_runner.rs
```diff
@@ -195,36 +194,38 @@ impl LoadJobRunner {
         job.update_state(JobTaskState::Pending, "Assigning workers");
         let block_size = job.info.block_size;
 
-        let mut total_size = 0;
-        let mut stack = LinkedList::new();
-        let mut task_index = 0;
-        stack.push_back(source_status);
-        while let Some(status) = stack.pop_front() {
+        let (mut total_size, mut stack) = (0i64, vec![source_status]);
```
bigbigxu, line 197: Vec is not suitable for use as a stack, as it will generate a large number of memory copy operations.

curvine-server/src/master/job/job_runner.rs
```diff
@@ -195,36 +194,38 @@ impl LoadJobRunner {
...
+        let (mut task_index, mut dir_count, mut file_count) = (0usize, 0u32, 0u32);
+
+        while let Some(status) = stack.pop() {
             if status.is_dir {
-                let dir_path = Path::from_str(status.path)?;
-                let childs = ufs.list_status(&dir_path).await?;
-                for child in childs {
-                    stack.push_back(child);
+                dir_count += 1;
+                let normalized_path = Path::from_str(&status.path)?
```
bigbigxu, line 204: This step is unnecessary, Path::from_str(&status.path) has already completed the path standardization.

curvine-server/src/master/job/job_runner.rs
```diff
@@ -195,36 +194,38 @@ impl LoadJobRunner {
...
+                    .normalize_uri()
+                    .unwrap_or_else(|| status.path.clone());
+
+                if !visited_paths.insert(normalized_path) {
```
bigbigxu, line 208: I think the core problem is why ufs returns the same path instead of judging it here.

curvine-server/src/master/job/job_runner.rs
jlon: Here I am not using Vec, but LinkedList, which uses push_back/pop_front each time. It is essentially a queue structure. Only FileStatus is stored inside the node, and block data is not copied repeatedly, so the large number of memcpy problems you are worried about will not occur.

PR #341, opened 2025-10-16
feat: add process metrics collecting by default

zuston:


zuston:
> There are conflicts in the code and there is already merged code.

Done @bigbigxu

PR #343, opened 2025-10-16
fix: fixed the error of adding subtasks infinitely when loading folders on ufs

jlon:


curvine-ufs/src/opendal.rs
```diff
@@ -474,17 +474,151 @@ impl OpendalFileSystem {
...
             if trimmed.is_empty() {
                 Ok("/".to_string())
             } else {
-                // Keep the trailing slash for directories if it exists
                 Ok(trimmed.to_string())
             }
         }
     }
+
+    async fn process_entry_simple(&self, entry: opendal::Entry) -> Option<FileStatus> {
```
lzjqsdd: remove it

lzjqsdd (approved):

PR #344, opened 2025-10-16
fix: fuse cp file to s3 and hdfs error number handle, fix s3 writer s…

lzjqsdd:
1. Enhance the conversion processing from curvine error to fuse error, ensuring that fuse can recognize more sys error no
2. Fixed S3 write status error and segmented upload complete logic
3. Fixing the issue of incorrect return due to HDFS permission errors


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -192,12 +192,13 @@ impl CurvineFileSystem {
     }
 
     async fn fs_get_status(&self, path: &Path) -> FuseResult<FileStatus> {
+        info!("fs_get_status: path={}", path);
```
lzjqsdd: remove

PR #348, opened 2025-10-17
feat(1/N): Initialize pipeline write

zuston:


zuston:
Still in a early stage, could you help take a look? @bigbigxu

PR #363, opened 2025-10-24
fix: journal restore test fail

lzjqsdd:


lzjqsdd:
1. fix mkdir journal log
- mkdir store failed when `mkdir -p` replay,    log mkdir need log correct parent path。 before that, log mkdir record two same mkdirEntry, 
  - log mkdir `/journal/a`
  - log mkdir `journal/a`

Correct is 
  - log mkdir `/journal`
  - log mkdir `/journal/a`
  
2. mount and umount replay should't proposal log again, this cause race condition
    In JournalLoader:
    UnMount used mnt_mgr.unmount_by_id(...) → fs_dir.unmount(...) → journal_writer.log_unmount(...).
    Mount used mnt_mgr.unprotected_add_mount(...) + fs_dir.store_mount(info, false) but still risked confusion; both needed clean “no-log” replay paths.
    Replay must be side-effect-free; it must never emit new journal entries.

szbr9486 (approved):

PR #370, opened 2025-10-30
feat: add OpenFile interface

bigbigxu:
add `open_file` interface to implement POSIX open semantics, supporting read, write, and readwrite modes with random read/write capabilities)


curvine-common/src/fs/writer.rs
```diff
@@ -104,15 +104,7 @@ pub trait Writer {
...
-            if pos < 0 {
-                return err_box!(format!("Cannot seek to negative position: {}", pos));
-            }
-            // Default implementation: flush buffer, update position
-            self.flush_chunk().await?;
-            *self.pos_mut() = pos;
-            Ok(())
-        }
+    fn seek(&mut self, _pos: i64) -> impl Future<Output = FsResult<()>> {
+        async move { err_box!("Not support") }
```
szbr00, line 108: Will this method be supported in the future?

curvine-common/src/state/block_info.rs
```diff
@@ -164,14 +184,96 @@ impl SearchFileBlocks {
...
+    pub last_commit: Option<CommitBlock>,
+}
+
+impl WriteFileBlocks {
+    pub fn new(file_blocks: FileBlocks) -> Self {
+        let mut search_off = Vec::new();
+        let mut off = 0;
+        for _ in &file_blocks.block_locs {
+            let start = off;
+            let end = off + file_blocks.status.block_size;
```
szbr00: off += file_blocks.status.block_size;
search_off.push(Range { start, off });

remove end

Is it better ?

curvine-common/src/state/block_info.rs
```diff
@@ -164,14 +184,96 @@ impl SearchFileBlocks {
             err_box!("Not found block for pos {}", file_pos)
         }
     }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct WriteFileBlocks {
+    pub status: FileStatus,
+    pub block_locs: Vec<LocatedBlock>,
+    search_off: Vec<Range<i64>>,
```
szbr00, line 193: In the future, skip list structures can be used to store Range objects, improving retrieval efficiency.

curvine-common/src/fs/writer.rs
bigbigxu: Curvine already supports random writes, while UFS generally does not support random writes after seek, so its default implementation will directly throw an error.

curvine-common/src/state/block_info.rs
bigbigxu: Skip lists may not be needed here. The data processed by Search_off is sorted, and binary search is used to locate the data block to which the current position belongs.

szbr9486 (approved):

PR #374, opened 2025-11-03
feat: implement OpenFile interface on client side with random write support

bigbigxu:


curvine-client/src/block/block_writer_local.rs
```diff
@@ -91,25 +85,25 @@ impl BlockWriterLocal {
...
-        let res = self
-            .rt
+        self.rt
             .spawn_blocking(move || {
                 file.as_mut().write_all(chunk.as_slice())?;
-                Ok(())
+                Ok::<(), FsError>(())
             })
-            .await?;
+            .await??;
```
szbr00, line 93: Why is there two question marks？

curvine-client/src/file/curvine_filesystem.rs
```diff
@@ -125,51 +126,16 @@ impl CurvineFileSystem {
...
+    pub async fn open_with_opts(
+        &self,
+        path: &Path,
+        opts: CreateFileOpts,
+        flags: OpenFlags,
+    ) -> FsResult<FileBlocks> {
+        if flags.read_write() {
+            return err_box!("Cannot open {} for read and write.", path);
+        }
+        self.fs_client.open_with_opts(path, opts, flags).await
```
szbr00, line 138: Question mark missed？

curvine-client/src/file/curvine_filesystem.rs
```diff
@@ -189,17 +155,55 @@ impl CurvineFileSystem {
     }
 
     pub async fn open(&self, path: &Path) -> FsResult<FsReader> {
-        let file_blocks = self.fs_client.get_block_locations(path).await?;
+        self.open_for_read(path).await
```
szbr00, line 158: Question mark missed？

bigbigxu: The return type of open_for_read is FsResult<FsReader>, so no question is needed.

curvine-client/src/file/curvine_filesystem.rs
bigbigxu: The return type of open_for_read is FsResult<FileBlocks>, so no question is needed.

curvine-client/src/block/block_writer_local.rs
bigbigxu: The `spawn_blocking` function returns a `Result<Result<(), FsError>>`, so two `questions` are needed to return the error.

PR #379, opened 2025-11-06
feat: optimize  fuse client hang issue during multi-master failover

jlon:
Fix FUSE client hang during master failover by optimizing RPC timeout and retry logic.

**Changes:**
- Reduce RPC timeout: 120s → 10s
- Reduce retry duration: 300s → 40s  
- Add concurrent RPC on NotLeaderMaster error
- Implement immediate node switching (no retry delay within round)

**Result:** Recovery time reduced from 250-360s to 12-40s (12-23x faster).

### DIFF
--- a/curvine-common/src/error/fs_error.rs
+++ b/curvine-common/src/error/fs_error.rs
@@ -422,6 +422,10 @@ impl ErrorExt for FsError {
     fn should_retry(&self) -> bool {
         self.retry_master()
     }
+
+    fn should_continue(&self) -> bool {
+        matches!(self, FsError::NotLeaderMaster(_))
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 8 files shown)


curvine-common/src/conf/client_conf.rs
```diff
@@ -297,13 +297,13 @@ impl Default for ClientConf {
             conn_retry_min_sleep_ms: 300,
             conn_retry_max_sleep_ms: 10 * 1000,
 
-            rpc_retry_max_duration_ms: 5 * 60 * 1000,
-            rpc_retry_min_sleep_ms: 300,
-            rpc_retry_max_sleep_ms: 30 * 1000,
+            rpc_retry_max_duration_ms: 40 * 1000, // 40s: optimized for fast master failover
```
bigbigxu: The modified configuration size is very insecure.
Under network fluctuations, rapid RPC timeouts can lead to the following:

1. RPC requests fail, and workers quickly lose their jobs.
2. Normal business requests fail.
3. Frequent job selection/selection issues occur.

PR #393, opened 2025-11-11
fix: fuse recv umount signal

lzjqsdd:


curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -241,4 +253,8 @@ impl<T: FileSystem> FuseReceiver<T> {
         res?;
         Ok(())
     }
+
+    pub fn raw_fd(&self) -> orpc::sys::RawIO {
```
lzjqsdd: unused, remove it

szbr9486 (approved):

PR #414, opened 2025-11-14
feat: add write/read metrics

ziqonghuang:


curvine-server/src/worker/worker_metrics.rs
```diff
@@ -24,10 +24,14 @@ pub struct WorkerMetrics {
 
     pub(crate) write_bytes: Counter,
     pub(crate) write_time_us: Counter,
+    pub(crate) write_count: Counter,
+    pub(crate) write_avg_time_ms: Gauge,
```
lzjqsdd: This metric is unnecessary; it can be calculated directly in the external system using count and time

PR #423, opened 2025-11-18
fix: add nlink for dir

201341:


curvine-server/src/master/meta/inode/inode_view.rs
```diff
@@ -394,7 +402,7 @@ impl InodeView {
 
             Dir(_, d) => {
                 status.file_type = FileType::Dir;
-                status.len = d.children_len() as i64;
+                status.len = 4 << 10;
```
bigbigxu: This default value does not need to be changed.， In Fuse, this value will be set according to the file type as follows:
    pub fn fuse_st_size(status: &FileStatus) -> u64 {
        match status.file_type {
            FileType::Link => status.target.as_ref().map(|x| x.len()).unwrap_or(0) as u64,
            FileType::Dir => FUSE_DEFAULT_PAGE_SIZE as u64,
            _ => status.len as u64,
        }
    }

201341: Got it

PR #425, opened 2025-11-19
fix: update nlink when create/unlink dir

201341:


lzjqsdd:
> > dir is not support link cause dir hardlink will cause infinite path. maybye we should forbidden this operation. Reference: https://askubuntu.com/questions/210741/why-are-hard-links-not-allowed-for-directories.
> 
> dir desn't support link, But when create the subdir on the dir. there will be an ".." link to dir. So we need to add the link number.

you are right, now i see what you means

Issue #402, opened 2025-11-13
[BUG]: The Fuse client cannot manipulate files in the mounted HDFS/s3 directory.


jlon:
I used the command shown below to mount s3 to /mnt/s3 in the curvine path.
```
./dist/bin/cv mount s3://flink/user /mnt/s3  -c s3.endpoint_url=http://xxxx-access-test.wanyol.com   -c s3.region_name=xxxx   -c s3.credentials.access=xxxx  -c s3.credentials.secret=xxxxxxx
```
Next, as the root user, I started the mount point using the following command: `./build/dist/bin/curvine-fuse.sh start`. Then, I went to the mount point directory `/curvine-fuse/mnt/s3` and tried to manipulate files using the following command, but it failed.
```
root@localhost:/curvine-fuse/mnt/s3# echo "1" >> test.log
-bash: test.log: Input/output error
root@localhost:/curvine-fuse/mnt/s3# touch test.log
touch: setting times of 'test.log': No such file or directory
```

szbr9486:
Thank you for your feedback. We have fixed this issue. If you encounter any other problems while using it in the future, please let us know anytime. Thank you again for your patience and support.

201341:
@szbr9486  this issue hasn't been fixed, And still can't append in ufs.

jlon:
Resolved by #431 and #687. UFS-backed file creation was fixed, and the later write path handles UFS-only append through the normal open-with-options flow; #687 includes append regression coverage.

PR #431, opened 2025-11-21
fix: create file error for ufs

201341:
related issue: https://github.com/CurvineIO/curvine/issues/402

### DIFF
--- a/curvine-ufs/src/opendal.rs
+++ b/curvine-ufs/src/opendal.rs
@@ -215,7 +215,6 @@ impl Writer for OpendalWriter {
                 bytes::Bytes::copy_from_slice(slice)
             }
         };
-
         let len = data.len() as i64;
 
         let writer = self.writer.as_mut().unwrap();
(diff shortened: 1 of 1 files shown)


curvine-ufs/src/opendal.rs
```diff
@@ -701,8 +700,27 @@ impl FileSystem<OpendalWriter, OpendalReader> for OpendalFileSystem {
...
+        let exist = self.exists(path).await?;
+        if exist && !_overwrite {
+            return Err(FsError::common(format!(
+                "File already exists: {}",
+                path.full_path()
+            )));
+        }
+
+        self.operator
+            .write(&object_path, opendal::Buffer::new())
```
bigbigxu, line 712: Why do we need to create an empty file here? The writer will create it automatically:
async fn write_chunk(&mut self, chunk: DataSlice) -> FsResult {
if self.writer.is_none() {
self.writer = Some(
self.operator
.writer(&self.object_path)
.await
.map_err(|e| FsError::common(format!("Failed to create writer: {}", e)))?,
);
}

201341: As #402 shows. when I execute touch cmd, I need to get an empty file.

bigbigxu: Yes, that's a problem. However, we shouldn't create an empty file first. Instead, we should modify the `writer` property of `OpendalWriter`:
change `writer: Option<opendal::Writer>` to `writer: opendal::Writer`

This should be more standardized and avoid additional metadata overhead.

201341: This method can't show the file info when calling ls.

bigbigxu: That's true. However, creating the file beforehand might affect other types of storage, such as HDFS. Perhaps it would be better to check if the writer is none and create an empty object within the `complete` method?

Issue #404, opened 2025-11-13
[FEATURE] File Resize and Sparse File Support


bigbigxu:
## Overview

This update adds file resize interface to support sparse file creation and provides limited support for space preallocation.

### 1. Add Resize Interface

A new `resize` interface has been added to support file size adjustment and sparse file creation. This interface uniformly supports the following semantics:

- **truncate**: POSIX `truncate()` system call semantics
- **ftruncate**: POSIX `ftruncate()` system call semantics
- **fallocate**: Linux `fallocate()` system call semantics

#### Implementation Details

- **Master Side**: `curvine-server/src/master/meta/inode/inode_file.rs`
  - `resize()`: Core resize logic supporting extension and truncation
  - `extend()`: Extend file and allocate new blocks
  - `truncate()`: Truncate file and remove blocks beyond target length

- **Client Side**: `curvine-client/src/file/fs_writer_base.rs`
  - `resize()`: Client-side resize coordination logic ensuring atomicity

- **FUSE Side**: `curvine-fuse/src/fs/curvine_file_system.rs`
  - `set_attr()`: Handle FUSE `FATTR_SIZE` attribute changes
  - `fs_resize()`: FUSE layer resize handling

#### Feature Description

- **File Extension**: When target size is larger than current file size, extend the file and create file holes
- **File Truncation**: When target size is smaller than current file size, remove blocks beyond the target
- **Sparse Files**: Support creating sparse files with holes that do not consume actual storage space

#### Technical Implementation

- When extending files, only create new unallocated blocks (sparse blocks) at the end
- Block allocation uses lazy strategy - disk space is only allocated when data is actually written
- Support resize operations across block boundaries

#### Test Coverage

- `build/tests/fuse-test.sh`: `test_truncate()` function
  - Test file extension (200MB → 400MB)
  - Test file truncation (400MB → 300MB)

#### Feature Description

Curvine's support for `fallocate` uses a **lazy allocation strategy**:

- **Preallocate Metadata**: `fallocate` calls update file metadata, marking the space to be allocated
- **Deferred Disk Allocation**: Actual disk space allocation is deferred until data is written
- **Block-level Allocation**: Disk space is only allocated for a block when data is actually written to that block

#### Supported fallocate Operations

- ✅ **Basic Allocation**: `fallocate -l <size> <file>` - Allocate space of specified size
- ✅ **Extend File**: Extend file to specified size
- ✅ **Reduce File**: Reduce file to specified size
[...]

PR #437, opened 2025-11-27
feat: added a resize interface to LocalFile to modify file size and p…

bigbigxu:
…re-allocated space (#404)

### DIFF
--- a/orpc/src/sys/sys_libc.rs
+++ b/orpc/src/sys/sys_libc.rs
@@ -16,11 +16,13 @@
 
 use crate::err_box;
 use crate::handler::RpcFrame;
-use crate::io::IOResult;
+use crate::io::{IOResult, LocalFile};
 use crate::sys::*;
 use crate::{err_io, err_msg, try_err, try_option};
+use fs2::FileExt;
 use std::ffi::{CStr, CString};
 use std::fs;
+use std::fs::Metadata;
 use std::io::IoSlice;
 use std::path::Path;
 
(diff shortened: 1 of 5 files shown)


orpc/src/sys/sys_libc.rs
```diff
@@ -611,3 +613,79 @@ pub fn get_groupname_by_gid(gid: u32) -> Option<String> {
...
+        let actual_size = metadata.st_blocks() * ST_BLOCK_SIZE;
+
+        // If actual size is less than logical size, the file has holes
+        // 1g > 4k
+        if actual_size < logical_size {
+            Ok(actual_size)
+        } else {
+            Ok(logical_size)
+        }
+    }
```
szbr00, line 685: add else to indicate that it is not supported now

orpc/src/sys/sys_libc.rs
```diff
@@ -611,3 +613,79 @@ pub fn get_groupname_by_gid(gid: u32) -> Option<String> {
...
+    #[cfg(target_os = "linux")]
+    {
+        unsafe {
+            let fd = get_raw_io(file)?;
+            let result = libc::fallocate64(fd, mode, off as libc::off64_t, len as libc::off64_t);
+
+            err_io!(result)?;
+            Ok(())
+        }
+    }
```
szbr00, line 656: add else to indicate that it is not supported now

orpc/src/sys/sys_libc.rs
```diff
@@ -611,3 +613,79 @@ pub fn get_groupname_by_gid(gid: u32) -> Option<String> {
...
+    #[cfg(target_os = "linux")]
+    {
+        unsafe {
+            let fd = get_raw_io(file)?;
+            let result = libc::ftruncate64(fd, len as libc::off64_t);
+
+            err_io!(result)?;
+            Ok(())
+        }
+    }
```
szbr00, line 631: add else to indicate that it is not supported now

orpc/src/sys/sys_libc.rs
bigbigxu: Linux uses the operating system's API, resulting in better performance and predictable outcomes; non-Linux operating systems use Rust's own API, which results in poorer performance, depending on the underlying implementation.

Issue #380, opened 2025-11-06
[FEATURE] Implement write caching data


szbr9486:
**Is your feature request related to a problem? Please describe.**
Currently, Curvine only supports reading data from UFS for caching. Writing data to UFS bypasses Curvine and involves writing directly to UFS; The data cannot be written to the Curvine cache.

**Describe the solution you'd like**
We believe that write caching is necessary in certain scenarios, and we hope to implement write caching capabilities.

PR #478, opened 2025-12-11
feat: add write_type mount param, update TTL, and add cache tests (cl…

bigbigxu:
…ose #380)

### DIFF
--- a/curvine-common/src/state/mount.rs
+++ b/curvine-common/src/state/mount.rs
@@ -16,6 +16,7 @@ use crate::conf::ClientConf;
 use crate::fs::Path;
 use crate::state::{CreateFileOpts, CreateFileOptsBuilder, StorageType, TtlAction};
 use num_enum::{FromPrimitive, IntoPrimitive};
+use orpc::common::DurationUnit;
 use orpc::{err_box, CommonError, CommonResult};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
@@ -215,7 +216,9 @@ pub struct MountOptionsBuilder {
 impl MountOptionsBuilder {
     pub fn new() -> Self {
         Self {
-            write_type: WriteType::Through,
+            write_type: WriteType::AsyncThrough,
+            ttl_ms: Some(7 * DurationUnit::DAY as i64),
+            ttl_action: Some(TtlAction::Delete),
             ..Default::default()
         }
     }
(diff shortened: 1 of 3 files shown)


curvine-common/src/state/mount.rs
```diff
@@ -215,7 +216,9 @@ pub struct MountOptionsBuilder {
 impl MountOptionsBuilder {
     pub fn new() -> Self {
         Self {
-            write_type: WriteType::Through,
+            write_type: WriteType::AsyncThrough,
+            ttl_ms: Some(7 * DurationUnit::DAY as i64),
```
szbr00, line 220: i64 -> i32 is ok？

curvine-tests/tests/write_cache_test.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[test]
+fn test_mount_write_cache() {
+    // Check if UFS configuration is available, if not, skip the test
+    if env::var("UFS_TEST_PATH").is_err() {
+        println!("⚠️  UFS_TEST_PATH is not set, skipping test");
+        println!(
+            "   Set UFS_TEST_PATH and UFS_TEST_PROPERTIES environment variables to run this test"
+        );
+        println!("   Example: export UFS_TEST_PATH=hdfs://192.168.108.129:9000");
```
szbr00: 192.168.108.129 -> 127.0.0.1 ?

curvine-tests/tests/write_cache_test.rs
```diff
@@ -0,0 +1,187 @@
...
+        write(&fs, WriteType::CacheThrough, true).await;
+        write(&fs, WriteType::AsyncThrough, false).await;
+        write(&fs, WriteType::AsyncThrough, true).await;
+        write(&fs, WriteType::CacheThrough, true).await;
+        write(&fs, WriteType::AsyncThrough, true).await;
+    })
+}
+
+async fn write(fs: &UnifiedFileSystem, write_type: WriteType, random_write: bool) {
+    let chunk_size = 64 * 1024;
```
szbr00, line 61: Is there an default CHUNK_SIZE val?

bigbigxu: There is no default value. Each write operation is 64KB, with a total write volume of 1MB. Multiple writes are primarily used to check for data consistency after random writes.

curvine-common/src/state/mount.rs
bigbigxu: ttl_ms represents the data expiration time. If it is i32, the maximum value can be 24 days, which is relatively small.

PR #479, opened 2025-12-11
fix: page cache support in FUSE Filesystem (close #443)

### LINKED ISSUE #443: [BUG]: Fix Page Cache Support in FUSE Filesystem
## Summary

This PR fixes and improves page cache support in the FUSE filesystem implementation by addressing notification message handling and cache validity detection logic.

## Background

The FUSE filesystem needs to properly manage kernel page cache to ensure data consistency and performance. This involves:
1. Sending cache invalidation notifications to the kernel when file metadata changes
2. Correctly determining when to keep or invalidate cache based on file modification time

### 1. FUSE Notification Message Handling

**Problem:**
- Notification messages (`unique=0`) were incorrectly formatted
- The `error` field in notification messages was set incorrectly (using `FUSE_SUCCESS` instead of notification type)
- Notification messages could not be sent properly, causing "Invalid argument (os error 22)" errors

**Solution:**
- Implemented `FuseNotifyCode` enum with proper notification types
- Created `send_notify()` method that correctly formats notification messages
- Fixed notification message header: `error` field now contains the notification type (opcode) as a positive value
- Ensured `unique` field is set to `0` for all notification messages

**Files Changed:**
- `curvine-fuse/src/session/fuse_notify_code.rs` - New file with notification code enum
- `curvine-fuse/src/session/fuse_response.rs` - Added `send_notify()` method and updated `send_inode_out()`

### 2. Cache Validity Detection Logic

**Problem:**
- The `should_keep_cache()` method used unreliable heuristics (`mtime == 0 && len == 0`) to detect first access
- Empty files (`len == 0`) are valid and should not be treated as uninitialized
- The logic could incorrectly keep cache when files were modified

**Solution:**
- Use explicit `cache_valid` flag from `NodeAttr` to track initialization state
- Removed `len` comparison from change detection (only check `mtime` now)
- Set `cache_valid = true` immediately after checking to mark node as initialized

**Files Changed:**
- `curvine-fuse/src/fs/state/node_state.rs` - Refactored `should_keep_cache()` method

### Notification Message Format

According to FUSE protocol, notification messages have a special format:
- `unique` field must be `0`
- `error` field contains the notification type (opcode) as a positive value
- Notification messages should be sent directly to the FUSE device

### Cache Validity Logic

The new logic flow:
1. Check `cache_valid` flag to determine if this is the first access
2. Mark node as initialized by setting `cache_valid = true`
3. Compare `mtime` values to detect changes
4. Keep cache if first access OR no change detected
5. Invalidate cache if change detected

### New Files
- `curvine-fuse/src/session/fuse_notify_code.rs` - FUSE notification code enumeration

### Modified Files
- `curvine-fuse/src/session/fuse_response.rs`
  - Added `send_notify()` method for sending notification messages
  - Updated `send_inode_out()` to use `send_notify()`
  - Fixed notification message formatting

- `curvine-fuse/src/fs/state/node_state.rs`
  - Refactored `should_keep_cache()` to use `cache_valid` flag
  - Removed `len` comparison from change detection
  - Improved initialization detection logic

- `curvine-fuse/src/session/mod.rs`
  - Added `fuse_notify_code` module export

### Test Cases
1. **Notification Message Sending**
   - Verify notification messages are sent correctly when cache needs invalidation
   - Ensure no "Invalid argument" errors occur

2. **Cache Validity Detection**
   - First access to a file should keep cache
   - File modification should invalidate cache
   - Empty files should be handled correctly

3. **Edge Cases**
   - Files with `mtime = 0` (legitimate timestamp)
   - Empty files (`len = 0`)
   - Rapid file modifications

### Performance
- Improved cache hit rate by correctly detecting when cache is still valid
- Reduced unnecessary cache invalidations

### Reliability
- Fixed notification message errors that could cause cache inconsistencies
- More accurate cache validity detection prevents stale data issues

bigbigxu:


curvine-common/src/conf/fuse_conf.rs
```diff
@@ -225,22 +223,20 @@ impl FuseConf {
     }
 
     pub fn set_fuse_opts(&self, mount_options: &mut String) {
-        self.fuse_opts.iter().for_each(|opt| {
-            match opt.as_str() {
-                "default_permissions" => {
-                    mount_options.push_str(",default_permissions");
-                }
-                // "auto_unmount" => {
```
lzjqsdd, line 233: why removed auto_umount

bigbigxu: This parameter is not supported by /dev/fuse.

curvine-fuse/src/bin/curvine-fuse.rs
```diff
@@ -211,18 +204,10 @@ impl FuseArgs {
...
-        }
-
         if let Some(direct_io) = self.direct_io {
             conf.fuse.direct_io = direct_io;
         }
 
-        if let Some(kernel_cache) = self.kernel_cache {
-            conf.fuse.kernel_cache = kernel_cache;
-        }
-
```
szbr00, line 225: What are the effects of deleting auto_cache and kernel_cache?

curvine-fuse/src/session/fuse_response.rs
```diff
@@ -12,14 +12,17 @@
...
 // limitations under the License.
 
-use crate::raw::fuse_abi::fuse_out_header;
-use crate::session::FuseTask;
-use crate::{FuseError, FuseResult, FuseUtils};
+#![allow(unused)]
+
+use crate::raw::fuse_abi::{fuse_notify_inval_inode_out, fuse_out_header};
+use crate::session::{FuseNotifyCode, FuseOpCode, FuseTask};
+use crate::{FuseError, FuseResult, FuseUtils, FUSE_NOTIFY_UNIQUE};
```
szbr00: Put the parameter FUSE_NOTIFY_UNIQUE on the line below.

curvine-fuse/src/bin/curvine-fuse.rs
bigbigxu: Initially designed with reference to libfuse, these two parameters were actually ineffective and were therefore removed. Now, enabling page cache simply requires setting direct_io = false, which is simpler and conforms to POSIX semantics.

curvine-fuse/src/session/fuse_response.rs
bigbigxu: FUSE_NOTIFY_UNIQUEis a FUSE constant. Should it be defined with the other FUSE constants for consistency?

szbr9486 (approved):

PR #481, opened 2025-12-12
perf: improve s3 gateway performance by removing async-trait boxing

jlon:
**Summary**
- Replace `#[async_trait]` handlers with static dispatch (RPITIT + generic functions) and enum-based I/O dispatch to eliminate dynamic trait objects.
- Inline futures and remove vtable lookups along the S3 request path (HEAD/GET/LIST), enabling compiler inlining and cache-friendly state machines.
- Introduce `PollReaderEnum` / `AsyncReadEnum` to remove boxed I/O futures for PUT/multipart flows.

**Key Changes**
- `lib.rs` / `router.rs`: drop 11 trait-object extensions; inject concrete `Arc<S3Handlers>` directly.
- `s3_api.rs` / `handlers.rs`: convert core handlers to generic functions returning `impl Future`; add direct methods on `S3Handlers` to avoid boxing.
- `utils/mod.rs`: add enum dispatch for readers/writers to replace `dyn` dispatch in hot I/O paths.


curvine-s3-gateway/src/auth/sig_v4.rs
```diff
@@ -25,7 +25,7 @@ pub trait VHeader {
     fn rng_header(&self, cb: impl FnMut(&str, &str) -> bool);
 }
 
-use crate::s3::error::Error;
+use crate::s3::error_code::Error;
```
szbr00, line 28: There is no need to change it to error_code

szbr9486 (approved):

Issue #405, opened 2025-11-13
[FEATURE] File Lock Operations Implementation


bigbigxu:
# File Lock Operations Implementation

## Overview

Curvine implements three FUSE file lock operations: `get_lk`, `set_lk`, and `set_lkw`. These operations support both POSIX locks (fcntl) and BSD locks (flock) with automatic expiration and cleanup.

### `get_lk` - Query Lock Status

Queries the lock status for a file region without acquiring a lock.

- Checks for conflicting locks and returns the first conflicting lock if found
- Returns lock details if conflict exists, otherwise returns unlocked status
- Non-blocking operation

### `set_lk` - Non-blocking Lock

Acquires a file lock without blocking. Returns immediately if lock cannot be acquired.

- Attempts to acquire lock atomically
- Returns `EAGAIN` error if lock conflicts with existing lock
- Updates file handle lock state on successful acquisition
- Suitable for applications that can handle lock failures immediately

### `set_lkw` - Blocking Lock with Retry

Acquires a file lock with automatic retry mechanism.

- Blocks and retries lock acquisition until successful
- Uses exponential backoff retry strategy
- Logs wait status periodically for monitoring
- Suitable for applications that must wait for lock availability

### POSIX Locks (Plock)
- Range-based locking supporting byte ranges within a file
- Multiple locks can be held by the same owner on different ranges
- Compatible with `fcntl` system call

### BSD Locks (Flock)
- Whole-file locking
- Single lock per file owner
- Compatible with `flock` system call

### Automatic Lock Expiration
- Locks automatically expire after a configurable timeout (default: 300 seconds)
- Expired locks are automatically removed during conflict checks
- Prevents stale locks from blocking operations indefinitely
- Handles process crashes gracefully by cleaning up orphaned locks

### Conflict Detection
- Write locks conflict with both read and write locks
- Read locks only conflict with write locks
- Multiple read locks can coexist
- Locks from the same owner don't conflict with each other
- Range overlap detection for POSIX locks

### Lock Cleanup
- Expired locks are automatically cleaned up during lock operations
- Empty lock entries are removed to maintain efficiency
- Cleanup happens transparently without additional overhead

PR #482, opened 2025-12-12
feat: implement file lock metadata management with expiration (#405)

bigbigxu:


curvine-server/src/master/meta/lock_meta.rs
```diff
@@ -0,0 +1,463 @@
...
+    fn is_lock_expired(lock: &FileLock, expire_ms: u64) -> bool {
+        if lock.acquire_time == 0 {
+            return true;
+        }
+        let now = LocalTime::mills();
+        now.saturating_sub(lock.acquire_time) > expire_ms
+    }
+
+    pub fn len(&self) -> usize {
+        self.plock.len() + self.flock.len()
```
szbr00: Do the elements in plock need to be summed up?

PR #502, opened 2025-12-18
curvine gpu cluster auto deploy playbook (auto deploy scripts for ansible)

yiminghub2024:
curvine gpu cluster auto deploy playbook (auto deploy scripts for ansible)
add function :
Introduction of Advantages Automated Component Deployment: Automatically deploys Master, Worker, and Fuse services onto their respective functional nodes.

Smart Disk Detection: Automatically detects both used and unused NVMe drives across every GPU server within your cluster.

Automated Storage Initialization: Automatically formats unused disks and mounts them to your specified paths.

Dynamic Configuration Management: Automatically adds specified paths and all other NVMe mount points (excluding the system partition) to the configuration file.

Service-Oriented Management: Integrates Master, Worker, and Fuse services with systemd, allowing management via standard systemctl commands.

Flexible Lifecycle Management: Supports idempotent deployments (repeated runs), deployment repairs, and incremental configuration updates. Provides automated batch management for service start, restart, and stop operations.

Environment Automation: Automatically injects and activates required environment variables based on your specifications.

Self-Correction Logic: Automatically detects configuration errors and applies corrections to ensure deployment integrity.


lzjqsdd:
This PR contains a large number of AI-generated summaries that need to be removed. In addition, the relevant deployment scripts should be written in English to ensure broader universality.

yiminghub2024:
update 
1.delete all .md .txt
2.change  all language from chinese to english 
done

ansible/group_vars/all.yml
```diff
@@ -0,0 +1,34 @@
+---
+# 全局变量配置
+# 这里的变量会应用到所有主机
+
+# Curvine安装配置
```
gongxun0928: comments in EN?

ansible/deploy_all.sh
```diff
@@ -0,0 +1,1385 @@
...
+    else
+        # Disk itself has a mount point, subtract 1
+        children_mounted=$((children_mounted > 0 ? children_mounted - 1 : 0))
+    fi
+    
+    # Output debug info to stderr (does not affect results)
+    # echo "DEBUG: $disk has_children=$has_children in_df=$in_df children_mounted=$children_mounted" >&2
+    
+    # If no child devices, not in df, and no mount points (including child devices), then it's unformatted
+    if [ "$has_children" -eq 0 ] && [ "$in_df" -eq 0 ]; then
```
szbr00, line 407: **Critical:** This predicate does not establish that the disk is empty: an unmounted whole-disk filesystem (or a disk carrying LUKS/LVM/other signatures) can have no children and no mountpoints, then be passed to `mkfs.ext4 -F` below. Please fail closed by checking `blkid`/`lsblk` FSTYPE, PTTYPE, and signatures, and require an explicit device allowlist before formatting.

ansible/deploy_curvine.yml
```diff
@@ -0,0 +1,532 @@
...
+          [Unit]
+          Description=Curvine Worker Service
+          After=network.target
+          
+          [Service]
+          Type=forking
+          User=root
+          Group=root
+          WorkingDirectory={{ curvine_install_dir }}
+          Environment="CURVINE_MASTER_HOSTNAME=localhost"
```
szbr00, line 455: **Major:** Worker processes must not receive `CURVINE_MASTER_HOSTNAME=localhost`. `ClusterConf` treats this variable as the local master/journal identity and rejects it when it is absent from `journal_addrs`, which is the normal case after this deployment writes the real master addresses. Please omit/unset this variable for worker and FUSE services, while keeping `CURVINE_WORKER_HOSTNAME` for worker binding.

ansible/deploy_all.sh
```diff
@@ -0,0 +1,1385 @@
...
+    exit 0
+fi
+
+# Execute deployment (only deploy successfully connected nodes)
+log ""
+log "========================================" 
+log "Starting deployment..."
+log "========================================"
+if [ -n "$success_hosts" ]; then
+    log_cmd ansible-playbook deploy_curvine.yml --limit "$success_hosts" || true
```
szbr00, line 613: **Major:** The trailing `|| true` makes `deploy_result` unconditionally zero, so a failed Ansible deployment is treated as successful and the script proceeds to configuration/service startup. Please capture `log_cmd`'s status before handling it, and stop or report a failed deployment instead of continuing as if it succeeded.

ansible/hosts.ini
```diff
@@ -0,0 +1,37 @@
+[master]
+# Add master node IP addresses below, one per line
+# Example:
+10.128.131.219
```
szbr00, line 4: **Major:** The checked-in inventory contains active host entries, so following the documented one-click command immediately targets those machines as root (with host-key checking disabled). Please ship only commented placeholders or require the user to generate a separate local inventory before any deployment command can run.

ansible/update_config_after_format.yml
```diff
@@ -0,0 +1,90 @@
...
+        src: "{{ curvine_config_file }}"
+        dest: "{{ curvine_config_file }}.before_mount_update"
+        remote_src: yes
+      when: host_mountpoints is defined and host_mountpoints | length > 0
+
+    - name: Update meta_dir for master nodes
+      replace:
+        path: "{{ curvine_config_file }}"
+        regexp: '^(\s*)meta_dir\s*=\s*"([^"]*)"'
+        replace: '\1meta_dir = "{{ host_mountpoints[0] }}/\2" | regex_replace("/data/", "/")'
```
szbr00, line 43: **Major:** This replacement writes the `| regex_replace(...)` text literally because the filter is outside the Jinja expression, and it also prefixes the full captured `/data/...` path, producing invalid TOML. Please compute the final path inside a Jinja expression (and apply the same fix to `journal_dir`) or use the Python updater consistently.

ansible/deploy_curvine.yml
```diff
@@ -0,0 +1,532 @@
...
+        state: directory
+        mode: '0755'
+      when: not install_dir_check.stat.exists
+
+    - name: Copy dist.tar.gz to master nodes
+      copy:
+        src: "{{ master_dist_file }}"
+        dest: /root/dist.tar.gz
+        mode: '0644'
+      when: not install_dir_check.stat.exists
```
szbr00, line 118: **Major:** Using only the existence of `/root/dist` as the install guard means every subsequent deployment silently skips the supplied package, even when it contains a repair or newer binaries. Please compare package/version/checksum and perform an idempotent staged update (with rollback), rather than treating any existing directory as an already-correct installation; the worker path needs the same handling.

szbr00 (commented):
Found six actionable deployment-safety and correctness issues, including one destructive disk-formatting risk.

PR #531, opened 2025-12-25
fix: mkdir error and change error info

201341:


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -656,11 +656,7 @@ impl fs::FileSystem for CurvineFileSystem {
             | "security.selinux"
             | "system.posix_acl_access"
             | "system.posix_acl_default" => {
-                // Silently ignore system extended attributes without logging
-                // Return ENODATA error to indicate the attribute doesn't exist
-                // This is the correct FUSE protocol response for non-existent attributes
-                // Empty error message to avoid any logging overhead
-                return err_fuse!(libc::EOPNOTSUPP, "not support get_xattr {}", name);
+                return err_fuse!(libc::ENODATA, "get_xattr {}", name);
```
bigbigxu, line 659: According to the POSIX specification, EOPNOTSUPP should be returned here.

curvine-fuse/src/fuse_error.rs
```diff
@@ -74,7 +74,7 @@ impl From<FsError> for FuseError {
             FsError::InvalidPath(_) => Some(libc::EINVAL),
             FsError::DiskOutOfSpace(_) => Some(libc::ENOSPC),
             FsError::Timeout(_) => Some(libc::ETIMEDOUT),
-            FsError::Unsupported(_) => Some(libc::ENOSYS),
+            FsError::Unsupported(_) => Some(libc::EOPNOTSUPP),
```
bigbigxu: Returning to ENOSYS is correct.

curvine-fuse/src/session/fuse_response.rs
```diff
@@ -116,7 +116,7 @@ impl FuseResponse {
 
             Err(e) => {
                 let e = e.into();
-                if self.debug || e.errno != libc::ENOENT {
+                if self.debug {
```
bigbigxu: For non-libc::ENOENT errors, log output is required.

curvine-fuse/src/fs/curvine_file_system.rs
201341: No, If the kernel send the getxattr request, should return ENODATA if we don't support. otherwise the kernel will return operatioin not support error.

curvine-fuse/src/fuse_error.rs
201341: For the fuse error, should return syscall errno, But now curvine just return libc error?

curvine-fuse/src/session/fuse_response.rs
201341: for the libc error code, it is normal business response code, it shouldn't append to the log.

curvine-fuse/src/fuse_error.rs
bigbigxu: libc::xx is the error number.

curvine-fuse/src/fs/curvine_file_system.rs
bigbigxu: The disk file system returns an EOPNOTSUPP error for these attribute pairs.

curvine-fuse/src/session/fuse_response.rs
bigbigxu: More logs are helpful for troubleshooting. The Fuse API documentation typically only contains very basic error messages, which are not helpful for locating and troubleshooting problems.

curvine-fuse/src/fs/curvine_file_system.rs
201341: > The disk file system returns an EOPNOTSUPP error for these attribute pairs.

If return ENOSUPPORT, when lookup nobody:nobody dir, the filesystem will report Operation not support error, it is not ok

bigbigxu: Yes, you're right.

curvine-fuse/src/session/fuse_response.rs
201341: Ok, Generally  metaserver or client should add access logs for troubleshooting, I will revert the change

PR #537, opened 2025-12-26
feat(fuse): add directory handle for large directory listing performance

bigbigxu:


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -380,6 +380,51 @@ impl NodeState {
...
+
+    pub fn find_dir_handle(&self, ino: u64, fh: u64) -> FuseResult<Arc<DirHandle>> {
+        let lock = self.dir_handles.read();
+        if let Some(v) = lock.get(&ino) {
+            if let Some(handle) = v.get(&fh) {
+                return Ok(handle.clone());
+            }
+        }
+
+        err_fuse!(libc::EBADF, "Ino {} fh {}  not found dir handle", ino, fh)
```
szbr00: What's the meaning of Ino and fh？

PR #539, opened 2025-12-26
feat(client): implement block connection pool

bigbigxu:


curvine-client/src/block/block_client_pool.rs
szbr00, line 1: Add necessary comment information.

szbr9486 (approved):

PR #565, opened 2026-01-08
feat(client): enable block connection pool for large file I/O

bigbigxu:


curvine-client/src/file/fs_context.rs
```diff
@@ -130,17 +129,17 @@ impl FsContext {
     pub async fn acquire_write(
         &self,
         addr: &WorkerAddress,
-        block: &ExtendedBlock,
+        _block: &ExtendedBlock,
```
lzjqsdd: Why not remove it directly

bigbigxu: Removed

szbr9486 (approved):

Issue #433, opened 2025-11-26
[FEATURE] Small files performance enhancement


szbr9486:
**Is your feature request related to a problem? Please describe.**
The current architecture is not very friendly for the performance of reading and writing small files.
In particular，the performance of writing a large number of small files is not as expected.

**Describe the solution you'd like**
1. Optimizing the number of RPCs written to small files, which currently requires three RPC：create_file, add_block, complete
    Can you do batch operations on small files to reduce the number of rpc calls?
2.Fixed data block size will cause resource waste in small file scenarios.
 For example, a large number of small files blocks can lead to metadata explosion, worker-side file inodes waste, etc
3. Small files batch writing: when a large number of small files are written, the client cache will become invalid. 
Example 1, if the client cache is 1MB, and a large number of files smaller than 1MB are written, each one will trigger a flush operation.
Example 2, on the worker side, when writing a large number of small files, there is no batch writing; instead, fsync is called once for each file, resulting in a large number of random writes and impacting performance.

PR #570, opened 2026-01-08
feat: introduce batch writing feature

thuongle2210:
# Description 
Resolve part of #433: [Small files performance enhancement](https://github.com/CurvineIO/curvine/issues/433)
Introduce batch writing feature suitable for batches of small files

# Key Changes
1. Introduce batch writing feature
2. Add unit tests to demonstrate batch writing in both local and remote worker modes

# Performance Impact
1. Reduce RPCs from 3 per file (3×N for N files) to 3 per batch where total size ≤ chunk size

### DIFF
--- a/curvine-server/src/worker/handler/context.rs
+++ b/curvine-server/src/worker/handler/context.rs
@@ -28,9 +28,9 @@ pub struct WriteContext {
     pub block_size: i64,
 }
 
-impl Default for WriteContext {
+impl WriteContext {
     #[inline(always)]
-    fn default() -> Self {
+    pub fn place_holder() -> Self {
         Self {
             block: ExtendedBlock::default(),
             req_id: 0,
@@ -40,9 +40,6 @@ impl Default for WriteContext {
             block_size: 0,
         }
     }
-}
-
-impl WriteContext {
     pub fn from_req(msg: &Message) -> FsResult<Self> {
         let req: BlockWriteRequest = msg.parse_header()?;
 
--- a/orpc/src/io/local_file.rs
+++ b/orpc/src/io/local_file.rs
@@ -36,9 +36,26 @@ pub struct LocalFile {
     buf: BytesMut,
 }
 
-impl Default for LocalFile {
+impl LocalFile {
+    pub fn new<T: AsRef<str>>(path: T, mut inner: fs::File) -> IOResult<Self> {
+        let is_tmpfs = sys::is_tmpfs(path.as_ref())?;
+
+        let len = inner.metadata()?.len() as i64;
+        let pos = inner.stream_position()? as i64;
+        let file = Self {
+            inner,
+            path: path.as_ref().to_string(),
+            is_tmpfs,
+            len,
+            pos,
+            buf: BytesMut::new(),
+        };
+
+        Ok(file)
+    }
+
     #[inline(always)]
-    fn default() -> Self {
+    pub fn place_holder() -> Self {
         use std::env;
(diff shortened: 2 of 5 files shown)


orpc/src/message/rpc_message.rs
```diff
@@ -54,6 +54,12 @@ pub enum RequestStatus {
     Running = 3,  //Streaming request data delivery
     Cancel = 4,   //Cancel request
     Complete = 5, //Request complete
+
+    // batch variants
+    OpenBatch = 6,
```
bigbigxu, line 59: Batch mode is also a type of streaming request and does not require additional types.

orpc/src/io/local_file.rs
```diff
@@ -36,6 +36,30 @@ pub struct LocalFile {
     buf: BytesMut,
 }
 
+impl Default for LocalFile {
```
bigbigxu: LocalFile should not implement Default

thuongle2210: I implemented it using the mem::take() function. However, I can create a placeholder function and use mem::replace() for a similar purpose, then remove the Default requirement.

orpc/src/message/rpc_message.rs
thuongle2210: I updated it.

orpc/src/io/local_file.rs
thuongle2210: I updated it!

orpc/src/io/local_file.rs
```diff
@@ -54,6 +54,29 @@ impl LocalFile {
         Ok(file)
     }
 
+    #[inline(always)]
```
bigbigxu, line 57: This is not standard practice. Normally, you should use Option<LocalFile> instead of LocalFile.

orpc/src/message/rpc_message.rs
bigbigxu, line 59: This is also unnecessary, as the rpc_code in batch mode and normal mode are different, which is already sufficient to distinguish them.

curvine-server/src/master/master_handler.rs
```diff
@@ -292,6 +292,83 @@ impl MasterHandler {
...
+                req.last_block.map(ProtoUtils::extend_block_from_pb),
+            )?;
+            results.push(ProtoUtils::located_block_to_pb(located_block));
+        }
+
+        let rep_header = AddBlocksBatchResponse { blocks: results };
+        ctx.response(rep_header)
+    }
+
+    pub fn complete_files_batch(&mut self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
```
bigbigxu, line 345: Please add audit logs for new request types.

curvine-server/src/master/master_handler.rs
```diff
@@ -292,6 +292,83 @@ impl MasterHandler {
         ctx.response(rep_header)
     }
 
+    pub fn create_files_batch(&mut self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
```
bigbigxu, line 295: Please add audit logs for new request types.

curvine-server/src/worker/handler/context.rs
```diff
@@ -29,6 +29,17 @@ pub struct WriteContext {
 }
 
 impl WriteContext {
+    #[inline(always)]
```
bigbigxu, line 32: Use Option instead where needed.

szbr9486 (approved):

Issue #573, opened 2026-01-12
[BUG]: Copy duplication issue fix


bigbigxu:
1. Using `block_on` with `accept_job` can lead to deadlock.

   pub fn accept_job(&self, job: ReplicationJob) -> CommonResult<()> {
        // step1: check the block_id existence (todo)
        // step2: push into the queue
        self.runtime
            .block_on(async move { self.jobs_queue_sender.send(job).await })?;
        Ok(())
    }

2. After replication is complete, if the new block information is not committed to the master, the block will not be recognized.

PR #575, opened 2026-01-12
feat(retry): fix replication deadlock and data consistency issues (#573)

jlon:
## Problem

This PR fixes two critical bugs reported in issue #573:

1. **Deadlock with `block_on`**: Using `block_on` in `accept_job` can lead to deadlock when the RPC handler thread blocks waiting for an async operation, especially under high concurrency.

2. **Data inconsistency**: After replication completes, if the block information is not committed to the master due to transient network failures, the block will not be recognized.

### 1. Fix deadlock in `WorkerReplicationManager::accept_job`

**Before:**
```rust
pub fn accept_job(&self, job: ReplicationJob) -> CommonResult<()> {
    self.runtime
        .block_on(async move { self.jobs_queue_sender.send(job).await })?;
    Ok(())
}
```

**After:**
```rust
pub fn accept_job(&self, job: ReplicationJob) -> CommonResult<()> {
    if let Err(e) = self.jobs_queue_sender.try_send(job) {
        return err_box!("{}: {}", QUEUE_FULL, e);
    }
    Ok(())
}
```

**Changes:**
- Replaced `block_on` with `try_send` to avoid blocking the RPC handler thread
- Returns an error immediately if the queue is full, allowing the caller to retry

### 2. Add retry mechanism for replication result reporting

**In `WorkerReplicationManager::report_job`:**
- Replaced `fs_client.rpc()` with `fs_client.retry_rpc()` using framework's built-in `TimeBondedRetry` policy
- Handles transient network errors with exponential backoff

**In `MasterReplicationManager::report_under_replicated_blocks`:**
- Added retry loop with exponential backoff when sending block IDs to staging queue
- Prevents tasks from being silently dropped if the queue is temporarily full

### 3. Add retry mechanism for replication job submission

**In `MasterReplicationManager::replicate_block`:**
- Added retry mechanism when submitting replication jobs to workers
- Specifically handles "queue is full" errors with exponential backoff (up to 18 retries over 120 seconds)
- Uses framework's `io_retry_policy()` for consistent retry behavior

### 4. Improve error handling with structured error types

**Created `curvine-common/src/error/replication_error.rs`:**
- Defined `QUEUE_FULL` constant and `is_retryable_error()` function
- Replaced string matching with structured error checking
- Follows DRY and SOLID principles

## Technical Details

- **Retry Policy**: Uses `TimeBondedRetry` with default configuration:
  - Max duration: 120 seconds
  - Min sleep: 100ms
  - Max sleep: 10 seconds
  - Exponential backoff strategy

- **Concurrency Control**: 
  - Worker side: Queue-based with `try_send` (non-blocking)
  - Master side: Semaphore-based (default: 1000 concurrent replications)

### DIFF
--- a/curvine-server/src/master/replication/master_replication_manager.rs
+++ b/curvine-server/src/master/replication/master_replication_manager.rs
@@ -27,7 +27,7 @@ use orpc::io::net::InetAddr;
 use orpc::message::{Builder, RequestStatus};
 use orpc::runtime::{AsyncRuntime, RpcRuntime};
 use orpc::sync::FastDashMap;
-use orpc::{err_box, try_log, try_option, CommonResult};
+use orpc::{err_box, try_option, CommonResult};
 use std::sync::Arc;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
@@ -41,7 +41,6 @@ pub struct MasterReplicationManager {
     worker_manager: SyncWorkerManager,
 
     replication_semaphore: Arc<Semaphore>,
-    runtime: Arc<AsyncRuntime>,
 
     staging_queue_sender: Arc<Sender<BlockId>>,
     inflight_blocks: Arc<FastDashMap<BlockId, InflightReplicationJob>>,
@@ -75,7 +74,6 @@ impl MasterReplicationManager {
             worker_manager: worker_manager.clone(),
             replication_semaphore: Arc::new(semaphore),
             staging_queue_sender: Arc::new(send),
-            runtime: rt.clone(),
             inflight_blocks: Default::default(),
             worker_client_factory: Arc::new(Default::default()),
             replication_enabled: conf.master.block_replication_enabled,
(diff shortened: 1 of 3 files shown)


curvine-server/src/worker/replication/worker_replication_manager.rs
```diff
@@ -103,9 +99,16 @@ impl WorkerReplicationManager {
...
+        let master_client = try_option!(self.master_client.get());
+
+        let context = master_client.fs_client.context();
+        let rpc_conf = context.rpc_conf();
+        let policy = rpc_conf.io_retry_policy();
+
+        let response: ReportBlockReplicationResponse = master_client
             .fs_client
-            .rpc(RpcCode::ReportBlockReplicationResult, request)
+            .retry_rpc(policy, RpcCode::ReportBlockReplicationResult, request)
```
bigbigxu: The rpc() function already includes retry logic, so no additional retry logic needs to be added.

curvine-server/src/master/replication/master_replication_manager.rs
```diff
@@ -213,11 +248,40 @@ impl MasterReplicationManager {
         if !self.replication_enabled {
             return Ok(());
         }
+
+        let retry_builder = self.worker_client_factory.conf().io_retry_builder();
+
         self.runtime.block_on(async move {
```
bigbigxu: block_on cannot be used here either.

PR #579, opened 2026-01-13
fix: invalid uri path when path container special character

lzjqsdd:


curvine-common/src/fs/path.rs
```diff
@@ -37,42 +46,89 @@ impl Path {
...
+
+        Ok(Self {
+            full_path: normalized,
+            parsed: OnceLock::new(),
+        })
+    }
+
+    // Parse "scheme://authority/path" or "/path"
+    fn parse_components(full_path: &str) -> ParsedComponents {
+        if let Some(scheme_end) = full_path.find("://") {
```
szbr00: Is this compatible with Windows system path style?

curvine-common/src/fs/path.rs
```diff
@@ -37,42 +46,89 @@ impl Path {
...
-        let full_path = uri.to_string();
+    // Normalize input path with scheme
+    fn normalize_path_input(input: &str) -> String {
+        if let Some(scheme_end) = input.find("://") {
+            // Has scheme: "s3://bucket/path///" -> "s3://bucket/path"
+            let scheme_part = &input[..scheme_end + 3];
+            let after_scheme = &input[scheme_end + 3..];
 
-        Ok(Self { uri, full_path })
+            if let Some(path_start) = after_scheme.find('/') {
```
bigbigxu: The scheme and authority parsing is already completed and stored, so we can use them directly when needed. It might not be necessary to add a separate parse_componentsmethod right now

curvine-common/src/fs/path.rs
```diff
@@ -37,42 +46,89 @@ impl Path {
...
+                    scheme,
+                    authority,
+                    path,
+                }
+            } else {
+                // "s3://bucket" without path
+                ParsedComponents {
+                    scheme,
+                    authority: Some(after_scheme.to_string()),
+                    path: "/".to_string(),
```
szbr00: Can a static variable be used instead of to_string function?

curvine-common/src/fs/path.rs
szbr00: Use a constant variable instead of ”://“，such as schema_separator ?

curvine-common/src/fs/path.rs
```diff
@@ -37,42 +46,89 @@ impl Path {
...
+    fn components(&self) -> &ParsedComponents {
+        self.parsed
+            .get_or_init(|| Self::parse_components(&self.full_path))
+    }
 
-        let uri = builder.path_and_query(path).build()?;
-        let full_path = uri.to_string();
+    // Normalize input path with scheme
+    fn normalize_path_input(input: &str) -> String {
+        if let Some(scheme_end) = input.find("://") {
```
szbr00: Same as above

curvine-common/src/fs/path.rs
lzjqsdd: yes, normalize fn handles windows path style,  after normalize will parsing path

lzjqsdd: :// already using constant variable

curvine-common/src/fs/path.rs
lzjqsdd: done

curvine-common/src/fs/path.rs
```diff
@@ -161,15 +182,52 @@ impl Path {
...
-        match path.rfind(Self::SEPARATOR).map(|v| &path[..v]) {
-            None => Ok(None),
-
-            Some("") => Path::from_str(Self::SEPARATOR).map(Some),
-
-            Some(v) => {
-                let res = Path::from_str(v)?;
-                Ok(Some(res))
+        // Find parent path from the path field
+        match self.path.rfind(Self::SEPARATOR) {
```
bigbigxu: This should be full_path; in fact, this method shouldn't need to be modified.

curvine-common/src/fs/path.rs
```diff
@@ -25,98 +25,130 @@ static SLASHES: Lazy<Regex> = Lazy::new(|| Regex::new(r"/+").unwrap());
...
+                (Some(after_scheme.to_string()), Self::SEPARATOR.to_string())
+            };
 
-        Ok(Self { uri, full_path })
+            // Build full_path: "s3://bucket/path"
+            let full_path = format!(
+                "{}{}{}{}",
+                scheme.as_ref().unwrap(),
+                Self::SCHEME_DELIMITER,
+                authority.as_ref().unwrap(),
```
bigbigxu: Unwrap should not be used.

curvine-common/src/fs/path.rs
```diff
@@ -25,98 +25,130 @@ static SLASHES: Lazy<Regex> = Lazy::new(|| Regex::new(r"/+").unwrap());
...
         self.full_path.clone()
     }
 
-    pub fn authority_path(&self) -> &str {
-        let full_path = self.full_path();
-
-        let auth_path = if let Some(scheme_end) = full_path.find("://") {
-            let start_index = scheme_end + 2;
-            &full_path[start_index..]
+    pub fn authority_path(&self) -> String {
```
bigbigxu, line 102: This should return &str.

curvine-common/src/fs/path.rs
lzjqsdd: You are right, but there's no issue with the path here, not full_path; it will be concatenated based on the parsed content below. However, it's obviously too complex of a modification, and the original logic is sufficient,  i'll revert it

curvine-common/src/fs/path.rs
lzjqsdd: i'll fix it

bigbigxu (approved):

PR #586, opened 2026-01-15
fix: raft for upstream

shuoranliu:


curvine-common/src/raft/snapshot/download_handler.rs
```diff
@@ -63,9 +69,15 @@ impl SnapshotDownloadHandler {
     pub fn read_chunk(&mut self, req: &Message) -> RaftResult<Message> {
         let reader = try_option_mut!(self.reader);
         let builder = if !reader.has_remaining() {
+            let checksum = reader.checksum();
+            info!(
+                "[Sender] Snapshot file transfer complete, checksum: {}, len: {}",
```
szbr00: Log messages should all begin with a lowercase letter.

szbr00: Change the log level to debug.

shuoranliu: done

szbr9486 (approved):

Issue #583, opened 2026-01-14
[BUG]:Read immediately after write gets zero length


shuoranliu:
**Describe the bug**
- Read the file immediately after write, and the length is zero.

**To Reproduce**
- mount S3 using `async_through` write_type
- execute the following shell command

```bash
python3 -c "
import os
f = open('test_consistency.bin', 'wb')
f.write(b'x' * 10000)
f.close()
print('exists:', os.path.exists('test_consistency.bin'))
print('read test:', len(open('test_consistency.bin', 'rb').read()))
os.remove('test_consistency.bin')
print('done')
"
```

- Result:

```
exists: True
read test: 0
done
```

**Expected behavior**
Get the correct size.

**Screenshots**

**OS Version (please complete the following information):**
 

**Additional context**

bigbigxu:
Please provide the following information, which will help troubleshoot the issue:

Command: `bin/cv fs stat /spark/a/curvine1/stress.csv`. Pay attention to the value of the `UFS Modification time` parameter.

Check the fuse logs to see if file reads are executed as "read from Curvine(cache)" or "read from ufs".

shuoranliu:
```
Name: test_consistency.bin
Id: 0
Is dir: false
Mtime: 1768452812000
Atime: 0
Children num: 0
Is complete: true
Len: 10000
Replicas: 1
Block size: 4194304
File type: File
Xattr: {}
Storage policy: StoragePolicy { storage_type: Disk, ttl_ms: 0, ttl_action: None, ufs_mtime: 0 }
Owner: 
File size: 9.8KB
Block size: 4.0MB
Replication: 1
Ttl: 0ms
Ttl action: None
Type: file
Modification time: 2026-01-15 04:53:32
Ufs Modification time: 1970-01-01 00:00:00
Permission: -rwxr-xr-x
Owner: 
Group: 
Nlink: 0
```

shuoranliu:
```
26/01/15 04:53:31.186 INFO fs_writer.rs:42 Create writer, path=/cube-studio-eu/curvine/lsr-test/test_consistency.bin, pos=0, len = 0, block_size=128.0MB, chunk_size=1
31072, chunk_number=8, replicas=1
26/01/15 04:53:31.206 INFO cache_sync_writer.rs:66 submit(init) job successfully for /cube-studio-eu/curvine/lsr-test/test_consistency.bin, job id job_7832facb6d62d2d
[...]

PR #594, opened 2026-01-19
fix: correct cache status validation logic in get_cache_status (#583)

bigbigxu:


curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -116,20 +116,31 @@ impl UnifiedFileSystem {
...
+        }
+
+        // Return Some if:
+        // 1. File only exists in Curvine (cv_only), or
+        // 2. File is complete (cached from UFS and ready to use)
+        if status.is_cv_only() || status.is_complete {
+            Ok(Some(status))
+        } else {
+            // File is being cached but not complete yet
+            Ok(None)
```
szbr00, line 143: Replace None with a specific message or status.

szbr9486 (approved):

PR #595, opened 2026-01-19
feat: add worker register version compatible check

### LINKED ISSUE #588: [FEATURE]  Ensures that all components are version-consistent when deploying a cluster
If the git versions used by the Master and Worker are inconsistent, it may lead to discrepancies in their communication protocols, resulting in unexpected errors.

Master <-> Worker
During cluster startup, the Master must verify the git version of connecting Workers to ensure compatibility. If a version mismatch is detected, the Master will reject that Worker from joining the cluster.

Master <-> Client
Similarly, when a client requests information from the Master, it must also verify the git version to ensure consistency.

lzjqsdd:


curvine-common/proto/worker.proto
szbr00, line 1: We need to check the Git code versions of each component; this implementation is for release versions.

lzjqsdd: This PR is to address the version control issue for registration, preventing non-compatible versions from being registered. The search for Git Version in terms of operations and maintenance has already been implemented in PR #592, which can check all components.

szbr00: Also check Fuse and Client.

lzjqsdd: For simplicity, the client's version control adopts a trust-master method to check. Once approved by the master, the client can communicate directly with the worker

PR #605, opened 2026-01-21
fix: file truncate and resize for PyTorch checkpoint overwrites

shuoranliu:


curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -942,6 +942,20 @@ impl FsDir {
         if file.len == opts.len {
             return Ok(DeleteResult::new());
         }
+
+        // Clear ufs_mtime when truncating a file that was cached from UFS.
+        // This ensures the file is treated as "cv_only" until the new sync job
+        // completes, preventing stale data reads from UFS during the sync window.
+        let is_truncating = opts.len < file.len;
```
bigbigxu, line 949: Truncate does not require initiating a resize request.

PR #611, opened 2026-01-26
feat: add pipeline replication protocol and errors

jlon:
introduce the protocol, config, and error foundations required for pipeline replication, so later client/worker/master changes can rely on stable RPC messages and error types

### DIFF
--- a/curvine-client/src/block/block_client.rs
+++ b/curvine-client/src/block/block_client.rs
@@ -103,7 +103,12 @@ impl BlockClient {
         seq_id: i32,
         chunk_size: i32,
         short_circuit: bool,
+        pipeline_stream: Vec<WorkerAddress>,
     ) -> FsResult<CreateBlockContext> {
+        let pipeline_stream = pipeline_stream
+            .iter()
+            .map(ProtoUtils::worker_address_to_pb)
+            .collect();
         let header = BlockWriteRequest {
             block: ProtoUtils::extend_block_to_pb(blk.clone()),
             off,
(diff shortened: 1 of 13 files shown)


curvine-client/src/block/block_client.rs
```diff
@@ -103,14 +103,20 @@ impl BlockClient {
         seq_id: i32,
         chunk_size: i32,
         short_circuit: bool,
+        downstream: Vec<WorkerAddress>,
```
szbr00: downstream->pipeline_stream

curvine-common/proto/worker.proto
```diff
@@ -16,6 +16,14 @@ message BlockWriteRequest {
     required bool short_circuit = 4 [default = false];
     required string client_name = 5 [default = ""];
     required int32 chunk_size = 6;
+    repeated WorkerAddressProto downstream = 7;
```
szbr00: Same as above.

curvine-server/src/worker/handler/batch_write_handler.rs
```diff
@@ -89,6 +89,7 @@ impl BatchWriteHandler {
                 short_circuit: header.short_circuit,
                 client_name: header.client_name.clone(),
                 chunk_size: header.chunk_size,
+                downstream: Vec::new(),
             };
```
szbr00, line 93: Same as above.

curvine-client/src/block/block_client.rs
jlon: Thank you for your comment, I have fixed the issue.

PR #613, opened 2026-01-26
feat: unify config reading behavior

ziqonghuang:


curvine-fuse/src/bin/curvine-fuse.rs
```diff
@@ -162,116 +165,9 @@ pub struct FuseArgs {
...
-            conf.fuse.fuse_opts = Self::default_mnt_opts();
-        }
-
-        Ok(conf)
-    }
-
-    fn create_default_conf() -> ClusterConf {
-        ClusterConf::default()
+        let conf = ClusterConf::from(&self.conf, Some(&self), Some("fuse"));
+        println!("Loaded configuration from {}", &self.conf);
```
lzjqsdd, line 211: It seems that the command-line parameters of fuse have not been initialized into the conf?

curvine-common/src/conf/cluster_conf.rs
```diff
@@ -75,37 +75,110 @@ impl ClusterConf {
...
 
-        // Apply client hostname from environment variable
-        if let Ok(v) = env::var(Self::ENV_CLIENT_HOSTNAME) {
-            conf.client.hostname = v;
+    pub fn from<T: AsRef<str>, U: Serialize>(
+        path: T,
+        args: Option<&U>,
+        args_name: Option<&str>,
+    ) -> CommonResult<Self> {
+        let cli_source = if let (Some(name), Some(args_value)) = (args_name, args) {
```
lzjqsdd: add cli source as paramter

curvine-common/src/conf/cluster_conf.rs
```diff
@@ -75,23 +74,45 @@ impl ClusterConf {
...
+                    "Warning: Failed to load config file '{}': {}",
+                    path.as_ref(),
+                    e
+                );
+                eprintln!("Using default configuration");
+                conf = ClusterConf::default();
+            }
         }
 
         conf.master.init()?;
```
lzjqsdd: cli/gateway also need init?

szbr00 (commented):
Found six actionable configuration-loading regressions, including one data-loss risk.

curvine-common/src/conf/cluster_conf.rs
```diff
@@ -75,38 +74,71 @@ impl ClusterConf {
...
+        let config = Config::builder()
+            .add_source(File::new(path.as_ref(), FileFormat::Toml))
+            .add_source(Environment::with_prefix("CURVINE").separator("_"))
+            .add_source(File::from_str(&cli_source, FileFormat::Json));
+
+        let mut conf;
+        match config.build() {
+            Ok(config) => {
+                conf = Self {
+                    format_master: true,
```
szbr00, line 90: **Critical:** This hard-codes both formatting flags to `true` instead of reading the merged top-level values. The checked-in config sets them to `false`, and these flags are consumed by the RocksDB/raft stores and by `VfsDataset`, which recursively deletes existing worker storage when `format_worker` is true. Please deserialize the top-level fields from the merged config and add a restart-persistence test; otherwise a normal restart can destroy persisted cluster data.

curvine-common/src/conf/cluster_conf.rs
```diff
@@ -75,38 +74,71 @@ impl ClusterConf {
...
+                    format_worker: true,
+                    testing: false,
+                    cluster_id: "curvine".to_string(),
+                    master: config.get::<MasterConf>("master").unwrap_or_default(),
```
szbr00, line 94: **Major:** `unwrap_or_default()` silently replaces the entire section when any field fails to deserialize. A single invalid value can therefore discard all otherwise valid settings and start the service with unrelated defaults. Please propagate the `config.get` error with section context (and add a malformed-value test) instead of treating invalid configuration as absent.

curvine-cli/src/main.rs
```diff
@@ -48,58 +49,60 @@ pub struct CurvineArgs {
...
-                }
-                let hostname = tmp[0].to_string();
-                let port: u16 = tmp[1]
-                    .parse()
-                    .map_err(|_| format!("Invalid port number in master_addrs: '{}'", tmp[1]))?;
-                vec.push(InetAddr::new(hostname, port));
+        let args_json = json!({
+            "cli": args_value,
+            "client": {
+                "master_addrs": master_addrs
```
szbr00, line 100: **Major:** When `--master-addrs` is absent, this still injects an empty array at the highest-priority source, erasing `client.master_addrs` from the config file (the FUSE path does the same). Only add this override when the CLI option is present; otherwise multi-master clients can be redirected to reconstructed/default addresses.

curvine-fuse/src/bin/curvine-fuse.rs
```diff
@@ -162,123 +166,50 @@ pub struct FuseArgs {
...
-            for node in master_addrs.split(",") {
-                let tmp: Vec<&str> = node.split(":").collect();
-                if tmp.len() != 2 {
-                    return err_box!("wrong format master_addrs {}", master_addrs);
-                }
-                let hostname = tmp[0].to_string();
-                let port: u16 = tmp[1].parse()?;
-                vec.push(InetAddr::new(hostname, port));
+        let args_json = json!({
+            "fuse": args_value,
```
szbr00, line 203: **Major:** Serializing the whole argument struct makes absent flags override file configuration: `debug = false` and `options = []` are still emitted despite `skip_serializing_none`. This also bypasses the previous `default_mnt_opts()` fallback, so a normal mount loses its default FUSE options. Please build the override object only from explicitly supplied arguments and preserve the default-options behavior.

curvine-s3-gateway/src/bin/curvine-s3-gateway.rs
```diff
@@ -81,41 +86,26 @@ pub enum CredentialAction {
...
-            println!(
-                "Warning: Failed to load config file '{}': {}. Using default configuration",
-                args.conf, e
-            );
-            ClusterConf::default()
-        }
-    };
+    let args_value = serde_json::to_value(&args).unwrap();
+    let args_json = json!({
+        "s3_gateway": args_value,
```
szbr00, line 91: **Major:** Because Clap always populates `listen` and `region` with their defaults, serializing the full struct makes those defaults override the config file even when the user supplied neither flag. Please distinguish explicit CLI values (for example with `Option` fields or value-source checks) so the TOML listen address and region remain effective.

curvine-cli/src/main.rs
```diff
@@ -48,58 +49,60 @@ pub struct CurvineArgs {
...
-        conf.client.init()?;
-
-        Ok(conf)
-    }
+        })
+        .to_string();
 
-    fn create_default_conf() -> ClusterConf {
-        ClusterConf::default()
+        ClusterConf::from(conf_path.unwrap(), Some(&args_json))
```
szbr00, line 105: **Major:** Both configuration sources are optional, but this `unwrap()` panics when neither `--conf` nor `CURVINE_CONF_FILE` is set. Please return a descriptive error or preserve the previous default-config behavior instead of crashing.

PR #614, opened 2026-01-26
feat: add rpc shutdown handle support for testing with minicluster

jlon:
add rpc shutdown handle support for use minicluster test


orpc/src/server/rpc_server.rs
```diff
@@ -31,6 +43,7 @@ pub struct RpcServer<S> {
     addr: InetAddr,
     monitor: ServerMonitor,
     shutdown_hook: Mutex<Vec<Box<dyn FnOnce() + Send + Sync + 'static>>>,
+    shutdown_tx: watch::Sender<bool>,
```
bigbigxu: ServerMonitor already includes broadcast notifications; this code is redundant.

jlon: yes, you are right

PR #618, opened 2026-01-27
fix: master panic messages not shown

shuoranliu:


curvine-server/src/master/master_server.rs
```diff
@@ -196,14 +196,14 @@ impl Master {
         Self::new(conf)
     }
 
-    pub async fn start(mut self) -> ServerStateListener {
+    pub async fn start(mut self) -> CommonResult<ServerStateListener> {
```
bigbigxu, line 199: The startup should trigger pain instead of returning an error, which helps to obtain more stack information.

curvine-common/src/raft/raft_group.rs
```diff
@@ -77,7 +77,19 @@ impl RaftGroup {
             .find(|x| x.1.hostname == addr.hostname && x.1.port == addr.port);
 
         match find {
-            None => err_box!("Not a master role, address {}", addr),
+            None => {
```
bigbigxu, line 80: This check is repeated; a similar check is performed in curvine-server.rs.

PR #619, opened 2026-01-27
feat: enable worker-side pipeline write handling and status reporting

jlon:
enable worker-side pipeline write handling and status reporting with a temporary single-hop writer fallback for build isolation.
 
Core Functionality:
 - Worker-side support for pipeline writes: Parsing `pipeline_stream`, establishing downstream links, and sending back pipeline status.
- Extended write links: `WriteHandler` adds asynchronous pipeline establishment, forwarding, and failure handling.
- Enhanced monitoring and testing: Updates worker metrics and worker_test.

### DIFF
--- a/curvine-server/src/worker/handler/block_handler.rs
+++ b/curvine-server/src/worker/handler/block_handler.rs
@@ -13,50 +13,109 @@
 // limitations under the License.
 
 use crate::worker::block::BlockStore;
-use crate::worker::handler::BlockHandler::{BatchWriter, Reader, Writer};
-use crate::worker::handler::{BatchWriteHandler, ReadHandler, WriteHandler};
+use crate::worker::handler::BlockHandler::{BatchWriter, PipelineWriter, Reader, Writer};
+use crate::worker::handler::WriteContext;
+use crate::worker::handler::{BatchWriteHandler, PipelineWriteHandler, ReadHandler, WriteHandler};
+use curvine_client::file::FsContext;
 use curvine_common::error::FsError;
 use curvine_common::fs::RpcCode;
 use curvine_common::FsResult;
+use log::warn;
+use orpc::error::ErrorExt;
 use orpc::handler::MessageHandler;
-use orpc::message::Message;
+use orpc::message::RequestStatus as ProtoRequestStatus;
+use orpc::message::{Builder, Message, ResponseStatus};
 use orpc::{err_box, CommonResult};
+use std::sync::Arc;
 
 pub enum BlockHandler {
     Writer(WriteHandler),
+    PipelineWriter(PipelineWriteHandler),
     Reader(ReadHandler),
     BatchWriter(BatchWriteHandler),
 }
 
 impl BlockHandler {
-    pub fn new(code: RpcCode, store: BlockStore) -> CommonResult<Self> {
-        let handler = match code {
-            RpcCode::WriteBlock => Writer(WriteHandler::new(store)),
+    pub fn new(msg: &Message, store: BlockStore, fs_context: Arc<FsContext>) -> CommonResult<Self> {
+        let code = RpcCode::from(msg.code());
+        if code != RpcCode::WriteBlock {
+            let handler = match code {
+                RpcCode::ReadBlock => Reader(ReadHandler::new(store)),
+                RpcCode::WriteBlocksBatch => BatchWriter(BatchWriteHandler::new(store, fs_context)),
+                code => return err_box!("Unsupported request type: {:?}", code),
+            };
+            return Ok(handler);
+        }
 
-            RpcCode::ReadBlock => Reader(ReadHandler::new(store)),
+        if msg.request_status() == ProtoRequestStatus::Open {
+            let ctx = WriteContext::from_req(msg)?;
+            let pipeline_enabled = fs_context.cluster_conf().client.enable_pipeline_write;
+            if ctx.has_pipeline_stream() && pipeline_enabled {
+                let queue_size = fs_context.cluster_conf().worker.pipeline_queue_size;
+                let pipeline_write_delay_ms =
+                    fs_context.cluster_conf().worker.pipeline_write_delay_ms;
+                return Ok(PipelineWriter(PipelineWriteHandler::new(
+                    store,
+                    fs_context,
+                    crate::worker::Worker::get_metrics(),
+                    queue_size,
+                    pipeline_write_delay_ms,
+                )));
+            }
+            if ctx.has_pipeline_stream() && !pipeline_enabled {
+                warn!(
+                    "BlockHandler::new_with_request: pipeline_stream ignored because pipeline write is disabled, block_id={}",
+                    ctx.block.id
+                );
+            }
+        }
 
-            RpcCode::WriteBlocksBatch => BatchWriter(BatchWriteHandler::new(store)),
-
-            code => return err_box!("Unsupported request type: {:?}", code),
-        };
-
-        Ok(handler)
+        Ok(Writer(WriteHandler::new(store)))
     }
 }
 
 impl MessageHandler for BlockHandler {
     type Error = FsError;
 
+    fn is_sync(&self, msg: &Message) -> bool {
+        match self {
+            Writer(h) => h.is_sync(msg),
+            PipelineWriter(h) => h.is_sync(msg),
+            Reader(_) => true,
+            BatchWriter(_) => true,
+        }
+    }
+
     fn handle(&mut self, msg: &Message) -> FsResult<Message> {
-        let response = match self {
+        let result = match self {
             Writer(h) => h.handle(msg),
+            PipelineWriter(h) => h.handle(msg),
             Reader(h) => h.handle(msg),
             BatchWriter(h) => h.handle(msg),
         };
+        result.or_else(|e| Ok(msg.error_ext(&e)))
+    }
 
-        match response {
-            Ok(v) => Ok(v),
-            Err(e) => Ok(msg.error_ext(&e)),
-        }
+    async fn async_handle(&mut self, msg: Message) -> FsResult<Message> {
+        let error_ctx = (msg.code(), msg.request_status(), msg.req_id(), msg.seq_id());
+
+        let result = match self {
+            Writer(h) => h.async_handle(msg).await,
+            PipelineWriter(h) => h.async_handle(msg).await,
+            Reader(h) => h.async_handle(msg).await,
+            BatchWriter(h) => h.async_handle(msg).await,
+        };
+
+        result.or_else(|e| {
+            let err_msg = Builder::new()
+                .code(error_ctx.0)
+                .request(error_ctx.1)
+                .response(ResponseStatus::Error)
+                .req_id(error_ctx.2)
+                .seq_id(error_ctx.3)
+                .data(orpc::sys::DataSlice::Buffer(e.encode()))
+                .build();
+            Ok(err_msg)
+        })
     }
 }
--- a/curvine-server/src/worker/handler/pipeline_writer.rs
+++ b/curvine-server/src/worker/handler/pipeline_writer.rs
@@ -1,0 +1,215 @@
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::worker::handler::WriteContext;
+use curvine_client::block::BlockWriterRemote;
+use curvine_client::file::FsContext;
+use curvine_common::error::FsError;
+use curvine_common::state::WorkerAddress;
+use curvine_common::FsResult;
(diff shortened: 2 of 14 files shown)


curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -34,69 +36,115 @@ pub struct WorkerHandler {
...
 }
 
 impl MessageHandler for WorkerHandler {
     type Error = FsError;
 
+    fn is_sync(&self, msg: &Message) -> bool {
+        let code = RpcCode::from(msg.code());
+        match code {
+            RpcCode::SubmitTask | RpcCode::CancelJob | RpcCode::SubmitBlockReplicationJob => true,
+            RpcCode::WriteBlock => {
```
lzjqsdd, line 96: make it an independent function.

curvine-server/src/worker/handler/block_handler.rs
```diff
@@ -29,34 +34,58 @@ pub enum BlockHandler {
...
+    async fn async_handle(&mut self, msg: Message) -> FsResult<Message> {
+        let error_ctx = (msg.code(), msg.request_status(), msg.req_id(), msg.seq_id());
+
+        let result = match self {
+            Writer(h) => h.async_handle(msg).await,
+            Reader(h) => h.async_handle(msg).await,
+            BatchWriter(h) => h.async_handle(msg).await,
+        };
+
+        result.or_else(|e| {
```
bigbigxu, line 109: Please call msg.error_ext(e) to return the error message.

curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -15,38 +15,190 @@
...
+
+            Ok(PipelineStream::new(
+                Box::new(LockedDownstreamWriter::new(writer)),
+                remaining,
+            ))
+        })
+    }
+}
+
+pub struct PipelineStream {
```
bigbigxu: Use a separate .rs file

curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -15,38 +15,190 @@
...
+use orpc::sys::DataSlice;
 use orpc::{err_box, ternary, try_option_mut};
+use std::collections::HashSet;
+use std::env;
 use std::mem;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+use tokio::time::{sleep, Duration};
+
+pub trait DownstreamWriter: Send + Sync {
```
bigbigxu: Use a separate .rs file

curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -15,38 +15,190 @@
...
+    fn worker_id(&self) -> u32;
+    fn write<'a>(&'a mut self, data: DataSlice) -> BoxFuture<'a, FsResult<()>>;
+    fn complete<'a>(&'a mut self) -> BoxFuture<'a, FsResult<()>>;
+    fn cancel<'a>(&'a mut self) -> BoxFuture<'a, FsResult<()>>;
+}
+
+struct LockedDownstreamWriter {
```
bigbigxu: Use a separate .rs file

curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -15,38 +15,190 @@
...
+        }
+    }
+}
+
+impl DownstreamWriter for LockedDownstreamWriter {
+    fn worker_id(&self) -> u32 {
+        self.worker_id
+    }
+
+    fn write<'a>(&'a mut self, data: DataSlice) -> BoxFuture<'a, FsResult<()>> {
```
bigbigxu: In principle, dyn Futures should not be used for high-frequency operations, as this will lead to a significant performance degradation.

curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -34,69 +36,115 @@ pub struct WorkerHandler {
...
-
-        if need_new_handler {
-            let handler = BlockHandler::new(code, self.store.clone())?;
-            let _ = self.handler.replace(handler);
-        }
-
-        match self.handler.as_mut() {
-            None => err_box!("The request is not initialized"),
-            Some(v) => Ok(v),
+        match code {
```
bigbigxu: `is_sync` already determines whether it's synchronous or asynchronous; the code here is redundant.

curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -34,69 +36,115 @@ pub struct WorkerHandler {
     pub task_manager: Arc<TaskManager>,
     pub rt: Arc<Runtime>,
     pub replication_handler: WorkerReplicationHandler,
+    pub fs_context: Arc<FsContext>,
 }
 
 impl MessageHandler for WorkerHandler {
     type Error = FsError;
 
+    fn is_sync(&self, msg: &Message) -> bool {
```
bigbigxu, line 45: The `is_sync` function determines whether the processing is synchronous or asynchronous based on the `code`; no additional logic is needed here.

curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -34,69 +36,115 @@ pub struct WorkerHandler {
...
-
-                if matches!(
-                    msg.request_status(),
-                    RequestStatus::Cancel | RequestStatus::Complete
-                ) {
-                    let _ = self.handler.take();
-                };
-
-                res
+                h.handle(msg)
```
bigbigxu: if matches!(
                    msg.request_status(),
                    RequestStatus::Cancel | RequestStatus::Complete
                ) {
                    let _ = self.handler.take();
                };

Please keep this.

curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -240,9 +387,306 @@ impl WriteHandler {
...
+
+        info!(
+            "WriteHandler::establish_pipeline_stream: SUCCESS block_id={}, next_worker={}",
+            ctx.block.id, next_worker.worker_id
+        );
+
+        Ok(pipeline_stream)
+    }
+
+    pub async fn async_write(&mut self, mut msg: Message) -> FsResult<Message> {
```
bigbigxu: File operations cannot use async, as this will cause the tokio io thread to block, resulting in a significant performance drop under high concurrency and IO conditions.

curvine-server/src/worker/handler/block_handler.rs
jlon: Using `msg.error_ext` with `async_handle` is semantically equivalent in the current implementation. In the asynchronous `async_handle()`, `msg` is moved into `h.async_handle(msg)`, and `msg.error_ext(&e)` cannot be called later because the Message is not cloned. If you want to use `msg.error_ext`, use the signature of `MessageHandler::async_handle`. This is too intrusive to the underlying implementation.

curvine-server/src/worker/handler/worker_handler.rs
jlon: `WorkerHandler::is_sync` is not redundant. During the Open phase, the pipeline needs to determine whether to use asynchronous or synchronous methods based on the request.

bigbigxu: Whether to use synchronous or asynchronous operation is determined by the rpc_code, not by the request status.

curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -34,69 +36,134 @@ pub struct WorkerHandler {
...
+                }
+                res
+            }
+        }
+    }
 
+    async fn async_handle(&mut self, msg: Message) -> FsResult<Message> {
+        let code = RpcCode::from(msg.code());
+        match code {
+            RpcCode::SubmitTask => self.task_submit(&msg),
```
bigbigxu, line 72: Synchronization requests should not be handled here.

curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -34,69 +36,134 @@ pub struct WorkerHandler {
...
+                }
                 res
             }
         }
     }
 }
 
 impl WorkerHandler {
-    fn get_handler(&mut self, msg: &Message) -> FsResult<&mut BlockHandler> {
+    fn is_sync_request(&self, msg: &Message) -> bool {
```
bigbigxu, line 92: Based on the rpc_code, such complex logic is unnecessary.

curvine-server/src/worker/handler/block_handler.rs
```diff
@@ -13,50 +13,104 @@
...
 impl BlockHandler {
-    pub fn new(code: RpcCode, store: BlockStore) -> CommonResult<Self> {
-        let handler = match code {
-            RpcCode::WriteBlock => Writer(WriteHandler::new(store)),
-
-            RpcCode::ReadBlock => Reader(ReadHandler::new(store)),
-
-            RpcCode::WriteBlocksBatch => BatchWriter(BatchWriteHandler::new(store)),
+    pub fn new(msg: &Message, store: BlockStore, fs_context: Arc<FsContext>) -> CommonResult<Self> {
+        let code = RpcCode::from(msg.code());
```
bigbigxu, line 40: pipe write uses a separate rpc_code

curvine-server/src/worker/handler/pipeline_writer.rs
```diff
@@ -0,0 +1,207 @@
...
+    worker_id: u32,
+    data_sender: AsyncSender<DataSlice>,
+    task_sender: AsyncSender<PipelineTask>,
+    err_monitor: Arc<ErrorMonitor<FsError>>,
+}
+
+impl PipelineChannel {
+    fn new(writer: BlockWriterRemote) -> Self {
+        let worker_id = writer.worker_address().worker_id;
+        let (data_sender, data_receiver) = AsyncChannel::new(0).split();
```
bigbigxu: The queue size should be configurable.

curvine-server/src/worker/handler/block_handler.rs
jlon: In sync paths we already use msg.error_ext. In async_handle the Message is moved into the handler, so msg.error_ext cannot be called afterward. We construct an equivalent error response with the same code/status/req_id/seq_id, which is semantically identical without requiring intrusive trait changes.

jlon: Agree on equivalence. Changing MessageHandler::async_handle signature or making Message cloneable is too intrusive for this PR, so we keep the builder-based error response which preserves the same protocol fields and error payload.

curvine-server/src/worker/handler/worker_handler.rs
jlon: Yes, optimization has been done.

curvine-server/src/worker/handler/write_handler.rs
jlon: Yes, optimization has been done.

curvine-server/src/worker/handler/write_handler.rs
jlon: Yes, optimization has been done.

curvine-server/src/worker/handler/worker_handler.rs
jlon: For WriteBlock, rpc_code alone is not enough. We need to decide sync vs async based on whether the request includes pipeline_stream. The same rpc_code can represent both non-pipeline (sync) and pipeline (async) writes, so request content must be part of the decision.

curvine-server/src/worker/handler/worker_handler.rs
jlon: WriteBlock can be either pipeline or non‑pipeline. The rpc_code is the same, so is_sync must consider request content; otherwise pipeline requests would incorrectly go through the sync path.

curvine-server/src/worker/handler/worker_handler.rs
jlon: We still need to inspect the request for WriteBlock Open; rpc_code alone does not indicate pipeline vs non‑pipeline. That is why the extra logic is necessary.

jlon: Agree. We keep is_sync to decide sync vs async for WriteBlock Open based on pipeline_stream.

curvine-server/src/worker/handler/worker_handler.rs
jlon: rpc_code is not sufficient for WriteBlock; it can represent both pipeline and non‑pipeline writes. We must inspect the request to choose sync vs async.

curvine-server/src/worker/handler/block_handler.rs
jlon: Pipeline write does not have a separate rpc_code today. Adding one would be a protocol change; to keep compatibility we use the existing WriteBlock code and inspect pipeline_stream.

curvine-server/src/worker/handler/worker_handler.rs
jlon: Those branches (SubmitTask/CancelJob/Replication) are short and non‑blocking. Handling them here keeps routing simple and avoids extra dispatch layers. Async entrypoints can still run sync logic safely because no blocking IO happens in these cases.

curvine-server/src/worker/handler/pipeline_writer.rs
jlon: Implemented: queue size is now configurable via worker config (pipeline_queue_size), default 0 to keep current behavior. Pipeline write delay is also config‑driven (pipeline_write_delay_ms).

PR #625, opened 2026-01-29
feat(common): implement state read/write functionality for persisting…

bigbigxu:
… and restoring application state

### DIFF
--- a/curvine-common/src/fs/state_file.rs
+++ b/curvine-common/src/fs/state_file.rs
@@ -1,0 +1,164 @@
+//  Copyright 2025 OPPO.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


curvine-common/src/fs/state_file.rs
szbr00, line 1: What state is being recorded here?

curvine-common/src/fs/state_file.rs
```diff
@@ -0,0 +1,163 @@
...
+
+pub struct StateWriter {
+    path: String,
+    inner: BufWriter<File>,
+}
+
+impl StateWriter {
+    pub fn new<T: AsRef<str>>(path: T) -> IOResult<Self> {
+        let file = File::create(path.as_ref())?;
+        let inner = BufWriter::with_capacity(128 * 1024, file);
```
szbr00: Use a constant to indicate the cache size for later reuse.

curvine-common/src/fs/state_file.rs
bigbigxu: The state information of the fuse is serialized and saved to a file. When the fuse is hot-rebooted, the state is read from the file to restore the state, thus ensuring a seamless switch.

PR #627, opened 2026-01-29
feat(common): add SignalWatch and reload_param for graceful shutdown …

bigbigxu:
…and hot reload

### DIFF
--- a/orpc/src/sys/signal.rs
+++ b/orpc/src/sys/signal.rs
@@ -1,0 +1,165 @@
+//  Copyright 2025 OPPO.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+#![allow(unused)]
+
+use crate::err_box;
+use crate::io::IOResult;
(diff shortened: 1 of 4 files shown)


orpc/src/sys/signal.rs
```diff
@@ -0,0 +1,165 @@
...
+use crate::io::IOResult;
+use futures::future::{select_all, BoxFuture};
+use futures::FutureExt;
+use std::future::Future;
+
+/// Signal kinds that can be received by the process
+/// Values match the standard Linux signal numbers
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(i32)]
+pub enum SignalKind {
```
szbr00, line 27: SignalKind -> SignalType

orpc/src/sys/signal.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// SIGHUP (1) - Hangup detected on controlling terminal or death of controlling process
+    Hangup = 1,
+
+    /// SIGINT (2) - Interrupt from keyboard (Ctrl+C)
+    Interrupt = 2,
+
+    /// SIGQUIT (3) - Quit from keyboard (Ctrl+\)
+    Quit = 3,
+
```
szbr00, line 36: The numbers in the middle are unrelated; should we leave some reserved types in between?

orpc/src/sys/signal.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// SIGUSR1 (10) - User-defined signal 1
+    User1 = 10,
+
+    /// SIGUSR2 (12) - User-defined signal 2
+    User2 = 12,
+
+    /// SIGTERM (15) - Termination signal
+    Terminate = 15,
```
szbr00, line 44: Terminate = 2x ？

orpc/src/sys/signal.rs
bigbigxu: This is to maintain consistency with Linux singal values.

Issue #610, opened 2026-01-26
[FEATURE] Hot upgrade support for curvine-fuse


shuoranliu:
**Is your feature request related to a problem? Please describe.**

Currently, upgrading curvine-fuse requires unmounting the filesystem and restarting the FUSE process, which causes service interruption for all applications accessing the mount point. In production environments, especially for long-running training jobs or data pipelines, this interruption can be costly:

Applications receive I/O errors (EIO/ENOTCONN) during the upgrade window
Jobs may fail and need to be restarted from scratch
Coordinating upgrade windows across multiple nodes is operationally complex
The 30-second termination grace period plus pod startup time (~60s) results in minutes of downtime per node
This makes it difficult to deploy bug fixes or performance improvements without scheduling maintenance windows.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Reference: Similar hot upgrade mechanisms exist in other FUSE implementations and proxies (e.g., Envoy hot restart, Nginx binary upgrade, Cubefs fuse client hot upgrade)

PR #628, opened 2026-01-29
feat(fuse): implement state persistence and restoration for FUSE sess…

bigbigxu:
…ion (#610)


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -439,6 +452,137 @@ impl NodeState {
...
+                .insert(handle.fh, Arc::new(handle));
+        }
+        info!(
+            "node_state::restore: {} dir_handles restored",
+            dir_handles_count
+        );
+
+        let fh_creator_value = reader.read_len()?;
+        self.fh_creator.set(fh_creator_value);
+
```
szbr00: Add final log info

PR #634, opened 2026-02-03
feat(client): optimize metadata requests for UFS-mounted file reads

bigbigxu:


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -81,16 +81,14 @@ impl CurvineFileSystem {
...
-            (status.atime / 1000) as u64
-        } else {
-            (status.mtime / 1000) as u64
-        };
-        let ctime_nsec = if status.atime > 0 {
-            ((status.atime % 1000) * 1000) as u32
-        } else {
-            ((status.mtime % 1000) * 1000) as u32
-        };
+        let mtime_sec = (status.mtime.max(0) / 1000) as u64;
```
szbr00, line 84: Make status.mtime.max a separate variable to reduce one calculation.

PR #638, opened 2026-02-04
feat(fuse): add configurable permission check and refactor permission…

bigbigxu:
… validation

- Add check_permission flag in FuseConf to enable/disable permission checks
- Refactor permission checking logic into check_permissions_for_path method
- Replace duplicate permission check code in lookup, access, and open_dir
- Root user (uid=0) and disabled permission check bypass validation
- Default check_permission is true to maintain POSIX compliance

### DIFF
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -131,6 +131,8 @@ pub struct FuseConf {
 
     pub non_seekable: bool,
 
+    pub check_permission: bool,
+
     pub state_dir: String,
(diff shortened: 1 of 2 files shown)


curvine-server/src/master/meta/inode/inode_file.rs
```diff
@@ -307,6 +308,7 @@ impl InodeFile {
             meta.commit(block);
         }
 
+        info!("blocks: {:?}", self.blocks);
```
szbr00: debug

PR #650, opened 2026-02-10
fix: persist mount metadata across master switch

jlon:
## Summary
- flush mount/unmount journal entries before returning success to avoid mount metadata loss during leader switch
- fix mount update-mode existence check to use curvine mount-path index
- expose standby master fs in mini cluster test helper for failover assertions
- add mount regression and coverage tests in curvine-tests (failover, mount manager, mount table)

## Verification
- cargo test -p curvine-tests --test mount_failover_test --test mount_manager_test --test mount_table_test -- --nocapture

### DIFF
--- a/curvine-server/src/master/mount/mount_table.rs
+++ b/curvine-server/src/master/mount/mount_table.rs
@@ -102,6 +102,10 @@ impl MountTable {
         inner.mountpath2id.contains_key(cv_path)
     }
 
+    pub fn mount_path_exists(&self, cv_path: &str) -> bool {
+        self.mount_point_inuse(cv_path)
+    }
+
     pub fn unprotected_add_mount(&self, info: MountInfo) -> FsResult<()> {
         info!("add mount: {:?}", info);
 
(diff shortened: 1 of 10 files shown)


curvine-server/src/master/mount/mount_table.rs
```diff
@@ -49,11 +49,64 @@ impl MountTable {
...
+
+        let total = mounts.len();
+        info!(
+            "mount restore started: {} entries loaded from metadata store",
+            total
+        );
+        if total == 0 {
+            info!("mount restore completed: no entries found");
+            return;
+        }
```
szbr00, line 67: else {
 info!(
            "mount restore started: {} entries loaded from metadata store",
            total
        );
}

curvine-server/src/master/mount/mount_table.rs
```diff
@@ -49,11 +49,64 @@ impl MountTable {
...
+                error!(
+                    "mount restore failed: unable to load mount table from metadata store, err={}",
+                    e
+                );
+                return;
+            }
+        };
+
+        let total = mounts.len();
+        info!(
```
szbr00: delete

curvine-server/src/master/mount/mount_table.rs
```diff
@@ -49,11 +49,64 @@ impl MountTable {
...
+            return;
+        }
+
+        let mut restored = 0usize;
+        let mut failed = 0usize;
+
+        for mnt in mounts {
+            let mount_id = mnt.mount_id;
+            let cv_path = mnt.cv_path.clone();
+            let ufs_path = mnt.ufs_path.clone();
```
szbr00: Deleting the three lines above only prints in the logs, so it's unnecessary.

jlon: I am keeping these logs on purpose. They provide low-cost but important startup visibility for master failover: load success/failure, empty-table case, and restore progress are distinct states in production troubleshooting.

curvine-server/src/master/mount/mount_table.rs
jlon: Keeping this outside `else` is intentional. We want a clear restore lifecycle log even when metadata load succeeds with zero entries, so operators can distinguish "restore executed and found 0 mounts" from "restore never ran".

curvine-server/src/master/mount/mount_table.rs
jlon: `total` is kept to support restore observability and summary correlation. It is used to report how many entries were loaded from metadata before per-entry restore, which helps diagnose partial-restore cases.

jlon: Updated in 0b5a30e. Removed redundant temp vars in the restore loop and simplified logging flow.

curvine-server/src/master/mount/mount_table.rs
jlon: Updated in 0b5a30e. The start log now runs only when `total > 0`, so the empty-table path emits only a single completion line.

curvine-server/src/master/mount/mount_table.rs
jlon: Updated in 0b5a30e. Empty restore now logs once (`no entries found`) and avoids duplicate-looking startup lines.

PR #654, opened 2026-02-11
fix(mount): preserve default TTL for legacy mount requests

jlon:
## Problem
Mount requests coming from legacy clients (or compatibility paths) may omit `ttl_ms` and `ttl_action` in `MountOptionsProto`.

In the previous flow, missing TTL fields were deserialized as `None`, then converted into `MountInfo` as `ttl_ms=0` and `ttl_action=none`. This disabled `auto_cache` and caused read miss paths to fall back to UFS instead of triggering async cache loading.

## Fix Intent
Ensure mount TTL behavior is stable and backward-compatible when optional TTL fields are missing, while keeping default handling centralized to avoid drift across modules.

## Implementation Details
- Added centralized mount default constants in `curvine-common/src/state/mount.rs`:
  - `DEFAULT_MOUNT_TTL_STR` (`7d`)
  - `DEFAULT_MOUNT_TTL_ACTION_STR` (`delete`)
  - `DEFAULT_MOUNT_TTL_MS`
  - `DEFAULT_MOUNT_TTL_ACTION`
- Updated `MountOptionsBuilder::new()` and `MountOptions::to_info()` to use these constants.
- Added compatibility backfill at protobuf decode boundary in `ProtoUtils::mount_options_from_pb()`:
  - missing `ttl_ms` => `DEFAULT_MOUNT_TTL_MS`
  - missing `ttl_action` => `DEFAULT_MOUNT_TTL_ACTION`
- Updated CLI mount defaults to reference shared constants instead of hardcoded literals.
- Updated client config defaults (`auto_cache_ttl`, `default_cache_ttl`) to reference the same shared mount TTL string constant.

## Tests
- Added regression test: `curvine-tests/tests/mount_ttl_compat_test.rs`
  - Reproduces legacy payload with missing TTL fields
  - Verifies decode-layer backfill and final `MountInfo` defaults
  - Verifies `auto_cache()` remains enabled

## Verification
- `RUSTC_WRAPPER= cargo check -p curvine-cli -p curvine-common`
- `RUSTC_WRAPPER= cargo test -p curvine-tests --test mount_ttl_compat_test -- --nocapture`

### DIFF
--- a/curvine-common/src/state/mount.rs
+++ b/curvine-common/src/state/mount.rs
@@ -21,6 +21,11 @@ use orpc::{err_box, CommonError, CommonResult};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+pub const DEFAULT_MOUNT_TTL_STR: &str = "7d";
+pub const DEFAULT_MOUNT_TTL_ACTION_STR: &str = "delete";
+pub const DEFAULT_MOUNT_TTL_MS: i64 = 7 * DurationUnit::DAY as i64;
+pub const DEFAULT_MOUNT_TTL_ACTION: TtlAction = TtlAction::Delete;
+
 #[repr(i32)]
 #[derive(
     Debug,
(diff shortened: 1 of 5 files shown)


curvine-common/src/state/mount.rs
```diff
@@ -236,8 +241,8 @@ impl MountOptions {
             ufs_path: ufs_path.to_string(),
             mount_id,
             properties: self.add_properties,
-            ttl_ms: self.ttl_ms.unwrap_or(0),
-            ttl_action: self.ttl_action.unwrap_or(TtlAction::None),
+            ttl_ms: self.ttl_ms.unwrap_or(DEFAULT_MOUNT_TTL_MS),
```
bigbigxu, line 244: If not set, the default value should be 0, meaning the data will not expire.

Issue #646, opened 2026-02-09
[BUG]:read after write failed


shuoranliu:
**Describe the bug**

`python3 test_write_read_consistency.py --path /mnt/curvine/lsr-test/fs-test/`

```python
#!/usr/bin/env python3
"""
Test script to verify write-read consistency issue in curvine-fuse.
This reproduces the issue seen in YOLO training where results.csv read fails
after being written due to S3 InvalidRange error.
"""

import os
import sys
import time
import tempfile
import argparse

def test_write_then_read(base_path: str, file_name: str = "test_consistency.csv"):
    """
    Test write-then-read consistency.
    Simulates what YOLO does: write to a file, then immediately read it.
    """
    file_path = os.path.join(base_path, file_name)

    print(f"[TEST] Testing write-read consistency on: {file_path}")

    # Clean up if exists
    if os.path.exists(file_path):
        print(f"[CLEANUP] Removing existing file: {file_path}")
        os.remove(file_path)
        time.sleep(0.5)

    # Test 1: Write new file, then read
    print("\n=== Test 1: Create new file and read immediately ===")
    test_data_1 = "epoch,loss,accuracy\n1,0.5,0.8\n"

    print(f"[WRITE] Writing {len(test_data_1)} bytes...")
    with open(file_path, 'w') as f:
        f.write(test_data_1)
    print("[WRITE] Write completed")

    print("[READ] Reading file immediately after write...")
    try:
        with open(file_path, 'r') as f:
            content = f.read()
        print(f"[READ] Read {len(content)} bytes successfully")
        if content == test_data_1:
            print("[PASS] Content matches!")
        else:
            print(f"[FAIL] Content mismatch!")
            print(f"  Expected: {repr(test_data_1)}")
            print(f"  Got: {repr(content)}")
            return False
    except Exception as e:
        print(f"[FAIL] Read failed: {e}")
        return False

    # Test 2: Append to file, then read (simulates results.csv append mode)
    print("\n=== Test 2: Append to file and read immediately ===")
    test_data_2 = "2,0.4,0.85\n"

    print(f"[APPEND] Appending {len(test_data_2)} bytes...")
    with open(file_path, 'a') as f:
        f.write(test_data_2)
    print("[APPEND] Append completed")

    print("[READ] Reading file immediately after append...")
    try:
        with open(file_path, 'r') as f:
            content = f.read()
        expected = test_data_1 + test_data_2
        print(f"[READ] Read {len(content)} bytes successfully")
[...]

PR #655, opened 2026-02-11
fix(ufs): avoid EIO on concurrent append writes in OpenDAL (issue #646)

jlon:
## Summary
- Fixes a reproducible `EIO` (`[Errno 5]`) during concurrent append/read workloads on mounted UFS paths when `write_type=through`.
- Keeps existing non-append semantics unchanged: random writes are still rejected for OpenDAL writers.
- Adds focused unit tests to lock in both behaviors.

## Problem
Issue: https://github.com/CurvineIO/curvine/issues/646

In FUSE workloads that append to the same file from multiple file descriptors/threads, users can hit intermittent `Input/output error` during the concurrent phase.

This was not reproducible on my side with `async_through`, but was consistently reproducible with the same mounted path configured as `write_type=through`.

## Root Cause
OpenDAL writer currently rejects any seek offset mismatch as random write:
- `OpendalWriter::seek` returned `not support random write` when `self.pos != pos`.

In append workloads (`O_APPEND`), the kernel-provided offset can be stale across FDs under concurrency. That stale offset is expected for append semantics, but we treated it as random write and returned an error, which surfaced as `EIO` in FUSE.

### 1) Append-aware seek behavior in OpenDAL writer
`curvine-ufs/src/opendal.rs`

- Added `append_mode: bool` to `OpendalWriter`.
- `seek` now:
  - returns `Ok(())` in append mode (tolerates stale append offsets);
  - keeps existing rejection for non-append random writes.

### 2) Set append mode at writer creation
`curvine-ufs/src/opendal.rs`

- `create(...)` -> `append_mode: false`
- `append(...)` -> `append_mode: true`

### 3) Added regression-focused unit tests
`curvine-ufs/src/opendal.rs`

- `append_like_seek_should_tolerate_stale_offset`
- `non_append_seek_should_reject_random_offset`

## Why this is safe
- Behavior change is scoped to append writers only.
- Non-append path still rejects random writes exactly as before.
- No protocol/state/journal format changes.
- No additional sync I/O or extra allocations in the hot path.

### Reproduction before fix
- Mounted `s3://flink/user` to `/mnt/s3` with `--mnt-type orch --write-type through`.
- Ran concurrent append/read script under FUSE path `/mnt/data/curvine/build/dist/testing/fuse/mnt/s3/...`.
- First round consistently produced many `OSError(5, 'Input/output error')`.
- FUSE log showed: `not support random write(curvine-ufs/src/opendal.rs:250)`.

### After fix
- Same runtime path and same script after rebuilding and restarting dist binaries.
- 5 consecutive rounds: `err=0`, `TOTAL=0`, no `EIO`.

### Automated checks
- `cargo fmt --all`
- `RUSTC_WRAPPER= cargo test -p curvine-ufs --features opendal-s3 opendal::tests -- --nocapture`
- `RUSTC_WRAPPER= cargo check --release -p curvine-fuse --no-default-features --features curvine-fuse/fuse3,curvine-client/opendal-s3`

## Scope
- This PR is focused only on issue #646 behavior in OpenDAL append/write path.
- Mount metadata/TTL persistence work is intentionally out of scope for this PR.


jlon:
Follow-up for #646 (stale dentry truncate-open miss)

I found and fixed an additional failure mode in the FUSE write path.

Problem
- Symptom: `open(..., 'wb')` intermittently failed with `ENOENT` on mounted paths after heavy delete/recreate cycles.
- Root cause: Linux can route an `O_CREAT|O_TRUNC` user intent through `OPEN` without `CREAT` when a stale positive dentry is hit; backend then sees `WT` (no create) and returns not found.

Fix (commit d7b7d64)
- In `NodeState::new_writer`, if `open_with_opts` fails with `FileNotFound` under `WT` (write + truncate, no create), retry once with `CREAT` enabled.
- Added unit test predicate coverage: `retry_open_as_create_predicate` in `curvine-fuse/src/fs/state/node_state.rs`.

Validation
- `cargo clippy --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args`
- `cargo test -p curvine-fuse node_state::test::retry_open_as_create_predicate -- --nocapture`
- Strict mounted-path reruns:
  - `test_write_read_consistency.py` x10 rounds: all pass
  - `test_concurrent_io.py` x3 rounds: all pass

Performance impact
- No fast-path behavior change.
- The retry executes only on the `FileNotFound + WT` error path.

curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -312,10 +312,14 @@ impl UnifiedFileSystem {
                 }
 
                 _ => {
+                    let mut cv_flags = flags;
                     if flags.overwrite() {
                         mount.ufs.create(&ufs_path, true).await?;
+                        if !cv_flags.create() {
```
bigbigxu, line 1151: This cannot be modified; it does not conform to POSIX semantics.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -258,11 +259,36 @@ impl NodeState {
             return Ok(writer);
         }
 
-        let writer = self.fs.open_with_opts(path, opts, flags).await?;
+        let writer = match self.fs.open_with_opts(path, opts.clone(), flags).await {
+            Ok(writer) => writer,
+
+            Err(e) if Self::should_retry_open_as_create(flags, &e) => {
```
bigbigxu, line 287: This cannot be modified; it does not conform to POSIX semantics.

bigbigxu:
For concurrent read/write tests, if the interval is very long, set attr_timeout and entry_timeout to 0. This will ensure data visibility, but performance may decrease.

jlon:
Follow-up based on @bigbigxu comments (visibility/perf tradeoff and through-mode limits), now in commit 5dd9d28.

What changed
- Clarified `cv mount --write-type` help text to explicitly state that `through` mode writes directly to UFS and concurrent-write behavior depends on UFS capabilities.
- Clarified FUSE CLI help for `--entry-timeout` and `--attr-timeout`: use `0`/`0` for strict read-after-write visibility tests.
- Added an opt-in runtime switch in `build/bin/curvine-fuse.sh`:
  - `CURVINE_FUSE_STRICT_VISIBILITY=1`
  - This appends `--entry-timeout 0 --attr-timeout 0` unless user already sets them explicitly.

Why this form
- No default behavior change, so no performance regression on normal deployments.
- Gives a deterministic strict-visibility mode for long-interval concurrent test runs.
- Makes through-mode boundary explicit to users/operators.

Validation
- cargo clippy -p curvine-cli -p curvine-fuse --all-targets -- --deny=warnings --allow clippy::uninlined-format-args

PR #678, opened 2026-03-01
refactor(client): simplify write types to CacheMode and FsMode only

bigbigxu:
## Summary

Unify the previous four write types (Cache, Through, AsyncThrough, CacheThrough) into two modes: **CacheMode** and **FsMode**. This reduces complexity in the unified client and CLI while keeping the main use cases: writing through to UFS only (CacheMode) and writing to Curvine with optional sync to UFS (FsMode).

## Write type semantics

| Write type   | Behavior |
|-------------|----------|
| **CacheMode** | Write goes directly to the underlying storage (UFS), bypassing Curvine cache. |
| **FsMode**    | Write goes to Curvine (cache) only. Data can be synced to UFS later (e.g. on close/complete or via async cache job). |

- **CLI**: `--write-type` accepts only `cache_mode` and `fs_mode` (default `fs_mode`).
- **TTL action**: No longer a separate CLI flag. It is derived from write type: **FsMode** → `Flush`, otherwise `Delete`. Mount must have a non-zero TTL.

### Client & CLI

- **curvine-cli (mount)**  
  - Removed `--ttl-action`. TTL action is set from write type; TTL must be non-zero.
  - `--write-type` default `fs_mode`, help text updated to the two options.

- **job_master_client**  
  - `wait_job_complete(job_id, mark)` → `wait_job_complete(job_id, fail_if_not_found: bool)`.
  - If `get_job_status` returns `JobNotFound`: when `fail_if_not_found` is true, return `Err`; when false, sleep and retry (keeps waiting).

- **cache_sync_writer**  
  - Removed the check that write type must be AsyncThrough or CacheThrough.
  - Calls `wait_job_complete(..., true)`.
  - FsMode: wait for job complete on close (sync-to-UFS behavior).

- **unified_filesystem**  
  - `CacheValidity::Invalid` no longer carries `Option<FileStatus>`.
  - New `get_mount_checked(path)`: returns `Some(mount)` only for cache-mode mounts; used for symlink/link and other ops that should run only on cache mounts.
  - Create/open and cache validity logic simplified for the two write types.

### Common (state, proto, traits)

- **WriteType enum**: Only `CacheMode` and `FsMode`; proto and display helpers updated.
- **MountInfo**: `is_cache_mode()`, `is_fs_mode()`; removed `is_write_through`, `has_ufs_data`.
- **MountOptionsBuilder**: Default write type `CacheMode`.
- **FileBlocks**: `cv_exists()` added; `get_read_block()` removed.
- **FileStatus**: `ufs_exists()` added; `is_cv_only()` removed.
- **Writer trait**: Default `write_string()` implementation added.

### Tests

- **write_cache_test**: Only mounts and writes for CacheMode and FsMode; removed multi-mode matrix and async wait logic tied to old write types.
- **ufs_test**: Uses `WriteType::CacheMode` instead of `Through`.

## Motivation

- Fewer write types make the API and CLI easier to understand and maintain.
- Clear split: “write to UFS only” (CacheMode) vs “write to Curvine, maybe sync to UFS” (FsMode).
- `fail_if_not_found` on `wait_job_complete` makes “must exist” vs “poll until found” explicit and avoids overloading a string `mark`.

## Files changed

| Area        | Files |
|------------|--------|
| CLI        | `curvine-cli/src/cmds/mount.rs` |
| Client     | `curvine-client/src/rpc/job_master_client.rs`, `unified/cache_sync_writer.rs`, `unified/unified_filesystem.rs` |
| Common     | `curvine-common/proto/common.proto`, `src/fs/writer.rs`, `state/block_info.rs`, `state/file_status.rs`, `state/mount.rs`, `utils/display.rs` |
| Tests      | `curvine-tests/tests/ufs_test.rs`, `write_cache_test.rs` |

## Breaking changes

- **WriteType**: Old variants (Cache, Through, AsyncThrough, CacheThrough) and CLI values (`cache`, `through`, `async_through`, `cache_through`) are removed. Callers must switch to **CacheMode** / **FsMode** and `cache_mode` / `fs_mode`.
- **wait_job_complete**: Second argument is now `fail_if_not_found: bool` instead of `mark: &str`.
- **Mount**: `ttl_action` is no longer a CLI option; it is derived from write type. TTL must be set (non-zero).
- **MountInfo**: `is_write_through()` and `has_ufs_data()` removed; use `is_cache_mode()` / `is_fs_mode()` and logic as needed.
- **FileBlocks**: `get_read_block()` removed; **FileStatus**: `is_cv_only()` removed; use `cv_exists()` / `ufs_exists()` where applicable.

### DIFF
--- a/curvine-cli/src/cmds/mount.rs
+++ b/curvine-cli/src/cmds/mount.rs
@@ -50,13 +50,6 @@ pub struct MountCommand {
     #[arg(long, default_value = "7d")]
     ttl_ms: String,
 
-    #[arg(
-        long,
-        default_value = "delete",
-        help = "TTL expiration action when file expires:\n  none - No action\n  delete - Delete file\n  persist - Export to UFS (skip if exists), keep CV cache\n  evict - Export to UFS (skip if exists), delete CV cache\n  flush - Force export to UFS (overwrite), delete CV cache"
-    )]
-    ttl_action: String,
-
     #[arg(long)]
     replicas: Option<i32>,
 
@@ -68,8 +61,8 @@ pub struct MountCommand {
 
     #[arg(
         long,
-        default_value = "async_through",
-        help = "Write type: cache, through, async_through, cache_through"
+        default_value = "fs_mode",
+        help = "Write type: cache_mode, fs_mode"
     )]
     write_type: String,
 
@@ -243,17 +236,24 @@ impl MountCommand {
     }
 
     pub fn to_mnt_opts(&self) -> CommonResult<MountOptions> {
+        let write_type = WriteType::try_from(self.write_type.as_str())?;
         let mnt_type = MountType::try_from(self.mnt_type.as_str())?;
         let consistency_strategy =
             ConsistencyStrategy::try_from(self.consistency_strategy.as_str())?;
         let ttl_ms = DurationUnit::from_str(self.ttl_ms.as_str())?.as_millis() as i64;
-        let ttl_action = TtlAction::try_from(self.ttl_action.as_str())?;
         let conf_map = self.get_config_map()?;
+
+        let ttl_action = if matches!(write_type, WriteType::FsMode) {
+            TtlAction::Flush
+        } else {
+            TtlAction::Delete
+        };
 
         let mut opts = MountOptions::builder()
             .update(self.update)
             .set_properties(conf_map)
             .mount_type(mnt_type)
+            .write_type(write_type)
             .consistency_strategy(consistency_strategy)
             .ttl_ms(ttl_ms)
             .ttl_action(ttl_action);
@@ -267,9 +267,6 @@ impl MountCommand {
         if let Some(storage_type) = self.storage_type.as_ref() {
             opts = opts.storage_type(StorageType::try_from(storage_type.as_str())?);
         }
-
-        let write_type = WriteType::try_from(self.write_type.as_str())?;
-        opts = opts.write_type(write_type);
 
         if let Some(provider_str) = self.provider.as_ref() {
             let provider = Provider::try_from(provider_str.as_str())?;
--- a/curvine-client/src/rpc/job_master_client.rs
+++ b/curvine-client/src/rpc/job_master_client.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use curvine_common::error::FsError;
 use log::info;
 use std::sync::Arc;
 use tokio::time;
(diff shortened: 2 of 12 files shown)


curvine-cli/src/cmds/mount.rs
```diff
@@ -243,17 +237,28 @@ impl MountCommand {
...
     pub fn to_mnt_opts(&self) -> CommonResult<MountOptions> {
+        let write_type = WriteType::try_from(self.write_type.as_str())?;
         let mnt_type = MountType::try_from(self.mnt_type.as_str())?;
         let consistency_strategy =
             ConsistencyStrategy::try_from(self.consistency_strategy.as_str())?;
         let ttl_ms = DurationUnit::from_str(self.ttl_ms.as_str())?.as_millis() as i64;
-        let ttl_action = TtlAction::try_from(self.ttl_action.as_str())?;
         let conf_map = self.get_config_map()?;
 
+        if ttl_ms.is_zero() {
```
szbr00: Why the ttl_ms must been set？

szbr9486 (approved):

bigbigxu:
> What is the read behaviors of these two modes? Are there any differences suppose we have UFS as backend storage?

# Cache Mode vs FS Mode

This document describes the difference between **cache_mode** and **fs_mode** write types for Curvine mounts. The write type is specified per mount (e.g. via `--write-type` in the mount CLI) and controls where data is written and how the unified layer treats the path.

## Summary

| Aspect | cache_mode | fs_mode |
|--------|------------|---------|
| **Write target** | Underlying storage (UFS) directly; bypasses Curvine cache. | Curvine filesystem (cache) only. |
| **Read source** | Prefer Curvine cache if valid; otherwise read from UFS (if enabled). | Prefer Curvine when data exists there; otherwise UFS read or async load. |
| **Unified “mount” handling** | Path is treated as under mount: create/append/mkdir/delete/rename go to UFS. | Path is not treated as mount for those ops: create/append/mkdir/delete/rename go to Curvine. |
| **Default TTL action** | Delete. | Free (or Flush, depending on config). |
| **Typical use** | UFS as source of truth; cache is optional for reads. | Curvine as primary storage; data may be synced to UFS later (e.g. load job). |

---

## cache_mode

- **Writes** go **directly to UFS**. Create, append, and overwrite are performed on the underlying storage (e.g. S3, HDFS). The Curvine namespace may still be updated for metadata, but the main data path bypasses Curvine cache.
- **Reads** may be served from **Curvine cache** when the cache is considered valid (e.g. consistency check passes); otherwise the client reads from UFS (if `enable_read_ufs` is set) or errors.
- Under the unified filesystem, paths under a **cache_mode** mount are treated as “under mount”: operations such as create, append, mkdir, delete, rename, and exists are executed **on UFS** (or UFS + cache invalidation where applicable).
- Default TTL action is typically **Delete** (remove from Curvine when TTL expires; UFS remains source of truth).

**Use case:** UFS is the source of truth; Curvine acts as an optional read cache. Good when you want to avoid writing through Curvine and keep durability and semantics on UFS.

---

## fs_mode

- **Writes** go **only to Curvine** (the cache layer). Create, append, and overwrite are performed on the Curvine filesystem; data may later be synced to UFS via load/export jobs (e.g. on complete or TTL).
- **Reads** are served from **Curvine when data exists there**; if only UFS has the data, the unified layer may return no cache reader (and optionally trigger async cache load and/or read from UFS if enabled).
- Under the unified filesystem, paths under an **fs_mode** mount are **not** treated as “under mount” for create/append/mkdir/delete/rename: those operations go to **Curvine**. Thus the logical path is managed and stored in Curvine; UFS is updated by background sync (e.g. load job) rather than by direct writes from these APIs.
- Default TTL action is typically **Free** (or Flush): release cache blocks or sync to UFS while keeping file metadata/length; UFS may be updated by load job completion.
- On file complete (e.g. in a wrapper writer), **fs_mode** may **wait for the load job** to finish so that data is synced to UFS before the write is considered done.

**Use case:** Curvine is the primary storage for the mount; applications read and write through Curvine, and UFS is kept in sync by background jobs. Good when you want a single namespace and write path on Curvine with eventual sync to UFS.

---

Issue #676, opened 2026-02-28
[BUG]: Client retry after master failover produces duplicate JournalEntry, causing follower apply loop to crash


thuongle2210:
**Describe the bug**

Assume we have three nodes: `node1` (master), `node2` (follower), and `node3` (follower).  
Curvine uses raft consensus algorithm, so a client sends a `mkdir` entry and then a `delete` on the same folder to `node2` and `node3`.  

- `node2` and `node3` receive the `mkdir` → `node1` receives an ack message.  
- `node2` and `node3` receive the `delete`, but `node1` crashes before receiving the ack.
- A new election occurs → `node2` becomes the leader and the client retries sending the `delete` entry to `node2` (because `node2` does not persist the `req_id`, which is only cached and lost when `node1` fails).
- `node2` replicates the `delete` operation to `node3` → `node3` applies the same logical operation twice.  
→ This raises an issue.

**To Reproduce**

run the test (just a simulation)

```#[test]
fn test_delete_duplicate_journal_on_leader_crash_retry() -> CommonResult<()> {
    Master::init_test_metrics();

    let mut conf = ClusterConf {
        testing: true,
        ..Default::default()
    };
    let worker = WorkerInfo::default();

    // node1 (initial leader)
    conf.change_test_meta_dir("raft-delete-idem-1");
    let js1 = JournalSystem::from_conf(&conf)?;
    let fs1 = MasterFilesystem::with_js(&conf, &js1);
    fs1.add_test_worker(worker.clone());

    // node1: mkdir /data => capture MkdirEntry
    fs1.mkdir("/data", false)?;
    let mkdir_entries = js1.fs().fs_dir.read().take_entries();
    assert_eq!(mkdir_entries.len(), 1, "Expected 1 mkdir entry from node1");
    let mkdir_entry_from_node1 = mkdir_entries.into_iter().next().unwrap();

    // node1: delete /data => capture DeleteEntry
    fs1.delete("/data", false)?;
    let delete_entries1 = js1.fs().fs_dir.read().take_entries();
    assert_eq!(
        delete_entries1.len(),
        1,
        "Expected 1 delete entry from node1"
    );
    let delete_entry_from_node1 = delete_entries1.into_iter().next().unwrap();

    //  node2 (new leader after node1 crash)
    // node2 received node1's mkdir but NOT node1's delete
    conf.change_test_meta_dir("raft-delete-idem-2");
    let js2 = JournalSystem::from_conf(&conf)?;
[...]

PR #679, opened 2026-03-01
fix: handling the idempotency issue by persisting req_id (1/N)

thuongle2210:
# Description
This PR ensures idempotency for `delete` operations. It guarantees that when a client interacts with the cluster via Raft, each node executes client-sent entries **exactly once**.

Resolves part of [pr-676](https://github.com/CurvineIO/curvine/issues/676)

# Key Changes
- Introduces `InodeStore::is_duplicate_req()` and `InodeStore::mark_req_applied()` to handle duplicate entries for operations applied to each node.
- Persists `req_id` unchanged from the client side and adds RPC functions to pass `req_id` through the call chain.
- Adds a unit test to verify correctness.

# Impact
- Handles idempotency for client and Raft nodes interactions.

# Future Work
- Ensure idempotency for other operations.
- Implement TTL cleanup for `PREFIX_REQ_ID` entries in `CF_COMMON`.

### DIFF
--- a/curvine-client/src/file/curvine_filesystem.rs
+++ b/curvine-client/src/file/curvine_filesystem.rs
@@ -31,6 +31,7 @@ use curvine_common::FsResult;
 use log::info;
 use log::warn;
 use orpc::client::ClientConf;
+use orpc::common::Utils;
 use orpc::runtime::{RpcRuntime, Runtime};
 use orpc::{err_box, err_ext};
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


thuongle2210:
> No need to add a separate req_id for RPC requests — Curvine already includes one in the protocol control block for every request, which can be reused directly. Storing req_id in RocksDB is not ideal due to the lack of a cleanup mechanism. Using FsRetryCache on follower nodes is a simpler and cleaner approach.

Point 2: I'm curious because FsRetryCache stores data in memory, which can be lost when a node restarts. Assume node1 sends an entry with req_id=X to all nodes and reaches quorum, creating a snapshot on each node. However, it **doesn't successfully ACK** to the client. The client will then retry and find the new leader to send req_id=X again.

But between the retries, node2 and node3 could restart, lose their memory, and lose the information in FsRetryCache. As a result, when the client finds the new leader. For example, in this case, node2 can't handle the duplication situation.

PR #716, opened 2026-03-17
refactor: add build-server ut-cov for github daily regression actions

lzjqsdd:
## Problem
Daily UT and coverage runs were done with raw `cargo test` / `cargo llvm-cov`; results were not fully reflected in the Actions Step Summary and failed tests were hard to track.

## Design
- Use `build-server.py` with `--run-once-ut-cov` so UT + coverage run through the same pipeline that produces `test_summary.json`.
- Add `--summary-to-markdown` to convert that JSON into Markdown tables and append to `GITHUB_STEP_SUMMARY` (and step log via `tee`).
- Workflow gets `permissions: contents: read`; utcov polling only checks `completed` / `failed` (no cancel path).

## Key Changes
- `.github/workflows/daily-ut-coverage.yml`: run build-server `--run-once-ut-cov`, write Step Summary from `test_summary.json`, upload `regression_result` artifact; add `permissions: contents: read`.
- `curvine-tests/regression/build-server.py`: `/utcov/run`, `/utcov/status`, `run_utcov_script()`, `--run-once-ut-cov`, `--summary-to-markdown`, `summary_json_to_markdown()`; failed tests table sorted by package/file/case; package field lookups simplified.
- `curvine-tests/README.md`: document `--run-once-ut-cov` and `--summary-to-markdown`.

### DIFF
--- a/curvine-tests/regression/build-server.py
+++ b/curvine-tests/regression/build-server.py
@@ -79,9 +79,23 @@ ltp_status = {
     'report_url': ''
 }
 
+# UT + Coverage only (for CI one-shot, no FIO/FUSE/LTP)
+utcov_status = {
+    'status': 'idle',  # idle, testing, completed, failed
+    'message': '',
+    'test_dir': '',
+    'report_url': ''
+}
+
+# Run-once mode: main thread waits on this instead of polling HTTP status
+run_once_done_event = threading.Event()
+run_once_final_status = None
+run_once_final_data = None
+
 # Create lock objects
 build_lock = threading.Lock()
 dailytest_lock = threading.Lock()
+utcov_lock = threading.Lock()
 coverage_lock = threading.Lock()
 regression_lock = threading.Lock()
 fuse_lock = threading.Lock()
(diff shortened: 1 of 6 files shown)


curvine-tests/regression/build-server.py
```diff
@@ -682,6 +691,106 @@ def run_dailytest_script():
...
+                process = subprocess.Popen(
+                    [script_path, project_path, TEST_RESULTS_DIR],
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    universal_newlines=True,
+                    start_new_session=True,
+                )
+                test_processes['regression'] = process
+                try:
+                    for line in process.stdout:
```
szbr00, line 754: This subprocess captures both `stdout` and `stderr`, but the loop only drains `stdout` before calling `wait()`. If `daily_regression_test.sh` writes enough output to stderr (for example from a failing `cargo nextest` run), the stderr pipe can fill up and block the child process, which would make the UT+coverage job hang indefinitely in CI. Could we either merge stderr into stdout (`stderr=subprocess.STDOUT`) or consume both streams concurrently here?

lzjqsdd: Thanks for catching this. I updated both the UT+cov path and the dailytest unittest path to use `stderr=subprocess.STDOUT` so stderr is merged into the same pipe we drain, avoiding pipe-buffer deadlocks. On failure, the summary now stores the last ~500 chars of the merged tail (we keep the last 80 lines while streaming) instead of `process.stderr.read()`.

szbr9486 (approved):

Issue #717, opened 2026-03-17
[FEATURE] Add CLI `free` command and support recursive free on master


bigbigxu:
## Description

Support releasing Curvine cache space from the command line and extend the master free API to allow recursive operation on directories.

### 1. CLI: add `free` command to release Curvine cache space

- Add a new subcommand (e.g. `curvine free` or `curvine cache free`) that triggers release of Curvine cache space.
- The command should call the master free API to free space for the given path or scope.
- Document usage and options (e.g. path, recursive flag) in CLI help.

### 2. Master free API: support recursive operation

- Extend the master free RPC/API so that freeing a directory can be done **recursively** (free cache for the directory and all descendants).
- When recursive is enabled, release cache for the given path and all paths under it; when disabled, only the given path.
- Document recursive vs non-recursive behavior in the API.

## Acceptance criteria

- [ ] CLI has a `free` command that releases Curvine cache for specified path/scope.
- [ ] Master free API accepts a recursive flag and performs recursive free when set.
- [ ] Docs or help text describe the free command and recursive free behavior.

PR #721, opened 2026-03-18
feat(server): add recursive flag and FreeResult; TTL-only free (#717)

bigbigxu:
## Summary

Extend the `free` API with a **recursive** option and a structured **FreeResult** return value. Free semantics are aligned with TTL: only inodes that are eligible for TTL free (e.g. `ttl_action == Free`, UFS synced, CV has data) are actually freed.

### API and Proto

- **FreeRequest** (master.proto): add `required bool recursive = 2`.
- **FreeResponse**: add `required FreeResultProto res = 1` (replaces empty message).
- **FreeResultProto** (common.proto): `inodes`, `bytes` (and internally `blocks` used for worker block removal).
- **curvine-common**: new `FreeResult` in `state::result` with `add(inodes, bytes, blocks)`; `ProtoUtils::free_res_from_pb` / `free_res_to_pb`.

### Client

- **CurvineFileSystem.free**: `free(path, recursive) -> FsResult<FreeResult>`.
- **FsClient.free**: sends `recursive` in `FreeRequest`, returns `FreeResult` from `FreeResponse`.
- **UnifiedFileSystem.free**: always calls `cv.free(path, recursive)` (removed branch that used `delete(path, false)` for non-fs_mode).

### Server

- **MasterFilesystem.free**: `free(path, recursive) -> FsResult<FreeResult>`; builds `FreeResult` from `DeleteResult` and returns it; still calls `worker_manager.remove_blocks` with the freed blocks.
- **FsDir.free**: takes `recursive: bool`, forbids free on root, resolves last inode and calls `unprotected_free(inode, mtime, recursive)`; journals path, mtime, and recursive.
- **unprotected_free**: rewritten to use an explicit stack (no recursion). Resolves `FileEntry` from store; for `Dir`, only pushes children when `recursive == true`; for `File`, calls `file.free(mtime)` and only applies free when it returns true. Aggregates freed inodes and blocks into a single result and calls `store.apply_free(change_inodes)`.
- **InodeFile.free**: returns `bool`; returns `true` only when `ttl_action == Free` and `ufs_exists() && cv_exists()`, then clears mtime and blocks; otherwise no-op and returns false.
- **InodeStore.apply_free**: signature changed from `&[&InodeView]` to `Vec<InodeView>`.
- **Journal**: `FreeEntry` and `log_free` carry `recursive`; replay calls `unprotected_free(inode, mtime, entry.recursive)`.
- **MasterHandler**: free RPC passes `recursive` and returns `FreeResult` in the response.
- **TTL executor**: all `filesystem.free(...)` calls updated to `filesystem.free(..., false)`.

### Tests

- **master_fs_test**: `test_idempotent_free` uses `fs.free("/file.log", false)`.
- **write_cache_test**: free calls updated to `fs.free(&path, false)` / `fs.free(path, false)`.

## Behavior

- **recursive == true**: directory inodes are expanded (children pushed and processed); all eligible files under the tree can be freed.
- **recursive == false**: only the given inode is considered; if it is a file and eligible, it is freed; if it is a directory, no children are pushed (no-op for that dir).
- **Eligibility**: a file is freed only when `InodeFile::free(mtime)` returns true (TTL free action, UFS synced, CV has data). Otherwise the inode is left unchanged and not written back as “freed”.
- **FreeResult**: callers get back inodes count, bytes count, and the block map used for worker cleanup.

### DIFF
--- a/curvine-client/src/unified/unified_filesystem.rs
+++ b/curvine-client/src/unified/unified_filesystem.rs
@@ -21,8 +21,8 @@ use curvine_common::conf::ClusterConf;
 use curvine_common::error::FsError;
 use curvine_common::fs::{FileSystem, FsKind, Path, Reader, Writer};
 use curvine_common::state::{
-    CreateFileOpts, FileAllocOpts, FileLock, FileStatus, JobStatus, LoadJobCommand, MasterInfo,
-    MkdirOpts, MkdirOptsBuilder, MountInfo, MountOptions, OpenFlags, SetAttrOpts,
+    CreateFileOpts, FileAllocOpts, FileLock, FileStatus, FreeResult, JobStatus, LoadJobCommand,
+    MasterInfo, MkdirOpts, MkdirOptsBuilder, MountInfo, MountOptions, OpenFlags, SetAttrOpts,
 };
 use curvine_common::utils::CommonUtils;
 use curvine_common::FsResult;
@@ -167,16 +167,17 @@ impl UnifiedFileSystem {
         self.cv.clone_runtime()
     }
 
-    pub async fn free(&self, path: &Path) -> FsResult<()> {
+    pub async fn free(&self, path: &Path, recursive: bool) -> FsResult<FreeResult> {
         match self.get_mount(path).await? {
             None => err_box!(
                 "the current file is not mounted to ufs, so the `free` command cannot be executed."
             ),
             Some((_, mnt)) => {
                 if mnt.info.is_fs_mode() {
-                    self.cv.free(path).await
+                    self.cv.free(path, recursive).await
                 } else {
-                    self.cv.delete(path, false).await
+                    self.cv.delete(path, false).await?;
+                    Ok(FreeResult::default())
                 }
             }
         }
--- a/curvine-common/proto/master.proto
+++ b/curvine-common/proto/master.proto
@@ -91,9 +91,12 @@ message DeleteResponse {}
 
 message FreeRequest {
     required string path = 1;
+    required bool recursive = 2;
 }
 
-message FreeResponse {}
+message FreeResponse {
+    required FreeResultProto res = 1;
+}
 
 message GetFileStatusRequest {
     required string path = 1;
--- a/curvine-server/src/master/master_handler.rs
+++ b/curvine-server/src/master/master_handler.rs
@@ -24,7 +24,7 @@ use curvine_common::fs::Path;
 use curvine_common::fs::RpcCode;
 use curvine_common::proto::*;
 use curvine_common::state::{
-    CreateFileOpts, FileBlocks, FileStatus, HeartbeatStatus, OpenFlags, RenameFlags,
+    CreateFileOpts, FileBlocks, FileStatus, FreeResult, HeartbeatStatus, OpenFlags, RenameFlags,
 };
 use curvine_common::utils::ProtoUtils;
 use curvine_common::FsResult;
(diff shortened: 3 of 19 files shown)


curvine-server/src/master/master_handler.rs
```diff
@@ -222,16 +222,18 @@ impl MasterHandler {
...
+        ctx.response(FreeResponse {
+            res: ProtoUtils::free_res_to_pb(res),
+        })
     }
 
-    pub fn free0(&self, req_id: i64, header: FreeRequest) -> FsResult<()> {
+    pub fn free0(&self, req_id: i64, header: FreeRequest) -> FsResult<FreeResult> {
         if self.check_is_retry(req_id)? {
-            return Ok(());
+            return Ok(FreeResult::default());
```
szbr00, line 233: On the retry path this now returns `FreeResult::default()`, so a retried `free` RPC will lose the original `{inodes, bytes}` result even if the first attempt already succeeded. That makes the new return value non-idempotent for clients. Could we cache/reconstruct the previous `FreeResult` here, or avoid exposing a structured result on retry until the retry cache can preserve it?

curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -167,16 +167,17 @@ impl UnifiedFileSystem {
...
             None => err_box!(
                 "the current file is not mounted to ufs, so the `free` command cannot be executed."
             ),
             Some((_, mnt)) => {
                 if mnt.info.is_fs_mode() {
-                    self.cv.free(path).await
+                    self.cv.free(path, recursive).await
                 } else {
-                    self.cv.delete(path, false).await
+                    self.cv.delete(path, recursive).await?;
```
szbr00: For non-`fs_mode` mounts this still delegates `free(path, recursive)` to `delete(path, recursive)` and then returns an empty `FreeResult`. With `recursive = true`, a caller can turn a `free` request into a recursive delete, which is much stronger than the TTL-aligned free semantics described in this PR, and the zero result hides that destructive behavior from the caller. Consider rejecting this case or routing it through a dedicated cache-eviction path instead of `delete`.

curvine-common/proto/master.proto
```diff
@@ -91,9 +91,12 @@ message DeleteResponse {}
 
 message FreeRequest {
     required string path = 1;
+    required bool recursive = 2;
```
szbr00, line 94: This changes an existing proto2 RPC message by adding new `required` fields (`FreeRequest.recursive` and `FreeResponse.res`). That is wire-incompatible across mixed versions: an older client will not send `recursive`, and an older server will not populate `res`, so decoding can fail during rolling upgrade or version skew. Could these be `optional` with a default (for example `recursive = false`) or be introduced through a new versioned request/response instead?

curvine-tests/tests/write_cache_test.rs
```diff
@@ -169,7 +169,7 @@ fn test_cache_mode_free() {
         let _ = fs.open(&path).await.unwrap();
         fs.wait_job_complete(&path, false).await.unwrap();
 
-        fs.free(&path).await.unwrap();
+        fs.free(&path, false).await.unwrap();
```
szbr00, line 172: The updated tests only adapt the new signature, but they still exercise `recursive = false` only and do not assert the returned `FreeResult`. Since this PR changes both the semantics (`recursive` traversal) and the API contract (`inodes` / `bytes` result), it would be useful to add at least one end-to-end case for freeing a directory with `recursive = true` and one assertion that the returned `FreeResult` matches what was actually freed.

curvine-server/src/master/master_handler.rs
bigbigxu: The `free` command does not guarantee that the result is idempotent; it simply refuses to retry the request.

szbr9486 (approved):

PR #722, opened 2026-03-18
feat(cli): adds the free command (#close 717)

bigbigxu:


curvine-cli/src/cmds/fs/free.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl FreeCommand {
+    pub async fn execute(&self, client: UnifiedFileSystem) -> CommonResult<()> {
+        match self {
+            FreeCommand::Free { path, recursive } => {
+                if path.trim().is_empty() {
+                    eprintln!("Error: Path cannot be empty");
+                    std::process::exit(1);
+                }
+
```
szbr00, line 34: This CLI now exposes `free` as a user-facing cache-release command, but the underlying `UnifiedFileSystem::free(path, recursive)` still delegates non-`fs_mode` mounts to `delete(path, recursive)` and returns an empty `FreeResult`. That means `cv fs free --recursive ...` can still behave like a recursive delete for some mounts while printing a normal-looking `inodes/space` summary. Could this command reject non-`fs_mode` mounts (or at least reject `--recursive` there) until the backend semantics are fully aligned?

szbr9486 (approved):

PR #724, opened 2026-03-19
fix(server): when set_attr is set recursively, the directory node is …

### LINKED ISSUE #718: [BUG]: When `set_attr` is set recursively, the directory node is set to the expected attribute.

bigbigxu:
…set to the expected attribute (fix #718)

- Add `recursive: bool` to free RPC and API; non-recursive skips expanding directory children. Journal FreeEntry and replay carry recursive.
- Return `FreeResult` (inodes, bytes, blocks) from free instead of (). Proto: FreeRequest.recursive, FreeResponse.res (FreeResultProto).
- FsDir: unprotected_free uses stack-based traversal; only free inodes that satisfy TTL free (ttl_action==Free, ufs_exists, cv_exists).

### DIFF
--- a/curvine-server/src/master/meta/fs_dir.rs
+++ b/curvine-server/src/master/meta/fs_dir.rs
@@ -803,42 +803,44 @@ impl FsDir {
     }
 
     pub fn unprotected_set_attr(&mut self, inode: InodePtr, opts: SetAttrOpts) -> FsResult<()> {
+        if inode.is_file_entry() {
+            return err_box!("set_attr is not supported on unresolved FileEntry inodes; resolve/load the full inode before calling set_attr");
+        }
+
+        let mut change_inodes = vec![];
+        let mut stack = LinkedList::new();
+        stack.push_back(inode.clone());
         let child_opts = opts.child_opts();
-        let recursive = opts.recursive;
-        let parent_inode_id = inode.id();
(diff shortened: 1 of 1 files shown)


curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -803,42 +803,38 @@ impl FsDir {
...
+                    opts.clone()
+                };
+                cur_inode.as_mut().set_attr(set_opts);
+                change_inodes.push(cur_inode.as_ref().clone());
+            }
+
+            match cur_inode.as_mut() {
+                Dir(_, dir) if opts.recursive => {
+                    for child in dir.children_iter() {
+                        stack.push_back(InodePtr::from_ref(child));
```
szbr00, line 828: This rewrites the recursive traversal to walk through `FileEntry` placeholders, but I do not see a regression test that applies `set_attr(recursive = true)` on a directory containing hard-link entries. The existing tests cover direct `set_attr` on a hard link and recursive `set_attr` on a normal subtree separately, but not the combination this change is fixing. Could we add one case that recursively updates a directory with `FileEntry` children and asserts the linked file inherits only the child attrs (for example owner/group/mode, but not ttl/xattr)?

bigbigxu: The recursive setting test is already included in fs_test::set_attr_recursive.

PR #726, opened 2026-03-19
feat(ufs): add local filesystem (file://) as UFS backend

bigbigxu:
## Summary

This PR adds support for using the local filesystem as a unified filesystem (UFS) backend via the `file://` scheme. Tests can run without external storage (e.g. S3/HDFS) by using a local directory under the test run path.

### curvine-client (unified)

- **`file` scheme with Opendal**: Include `"file"` in the list of schemes that use the Opendal provider (both explicit `Provider::Opendal` and `Provider::Auto`), so `file://` paths are handled by the OpenDAL-based UFS implementation.

### curvine-common (fs)

- **`FsKind::File`**: New variant `File` and constant `SCHEME_FILE`; `from_scheme` / `to_scheme` and `support_rename` updated so local file is treated like other UFS kinds that support rename.

- **`Path` (path.rs)**:
  - **file URL parsing**: For scheme `file`, authority is always empty; the path is taken from the substring after `file://` and normalized (e.g. `file:///x/y` → path `/x/y`, `file://./a` → `./a`, `file://../a` → `../a`).
  - **`authority_path()`**: When the path after `://` starts with `//` (e.g. `file:///path`), one leading slash is stripped before returning, so the result is `/path` instead of `//path`.
  - **`get_possible_mounts()`**: For `file` (and other schemes with empty authority), the root mount is emitted as `file:///` (three slashes). For schemes with authority (e.g. `s3://bucket/...`), the root is still emitted as `scheme://authority/`. Added tests for `file:///` and relative `file://` paths.
  - **Tests**: New `test_file_url()` covering `file:///`, `file:///path`, `file://./a`, `file://../a`, and `file://rel/path`; extended `get_possible_mounts` and `authority_path` tests for file URLs.

### curvine-tests

- **Testing harness**:
  - **Temp path**: `new_tmp_path()` now returns an **absolute** path: `current_dir().parent()/testing/curvine-tests/{ts}-{pid}-{rand}`.
  - **UFS path**: Replaced `s3_conf_path` with `ufs_path`. The cluster is configured to use a local UFS at `file://{base_path}/ufs`; that directory is created during `build()`. Removed `get_s3_conf()` and `with_s3_conf_path()`.
- **unified_test.rs**: Removed; it depended on S3 configuration and is no longer needed for the default local-UFS flow.
- **write_cache_test.rs**: If `UFS_TEST_PATH` is not set, it is set to `testing.ufs_path` (the local `file://` path) so the test runs without external UFS env vars.

### curvine-ufs (OpenDAL)

- **Cargo.toml**: The `opendal` feature now enables `opendal/services-fs` so the OpenDAL `Fs` backend is available for `file://`.
- **opendal.rs**:
  - **file scheme**: When scheme is `file`, an OpenDAL `Fs` operator is created with `root("/")`.
  - **`get_object_path()`**: If the path does not start with `/`, it is returned as-is instead of erroring, so relative `file://` paths (e.g. `file://./a`) work.

## Testing

- Existing path and mount tests extended for `file://` in `curvine-common`.
- `write_cache_test` runs with local UFS when `UFS_TEST_PATH` is unset.
- Manual verification: mount and read/write via unified FS using a `file://` UFS path.

## Notes

- Local UFS is intended for development and integration tests; production typically uses object or HDFS backends.
- Rename is supported for `FsKind::File` in the same way as for other UFS kinds that support rename.


curvine-ufs/src/opendal.rs
```diff
@@ -276,6 +276,15 @@ impl OpendalFileSystem {
             .to_string();
 
         let operator = match scheme {
+            FsKind::SCHEME_FILE => {
```
szbr00, line 279: The new path logic and tests explicitly accept relative file URLs like `file://./a` and `file://../a`, but this backend still creates the OpenDAL `Fs` operator with `root("/")` and then passes those relative object paths through unchanged. That means the runtime resolution is no longer relative to the caller’s intended base path; `./a` / `../a` will be interpreted under the filesystem root (or whatever OpenDAL normalizes them to), not relative to the current working directory as the new tests imply. Could we either reject relative `file://` mounts at runtime, or normalize them to an absolute base before constructing the operator and add an end-to-end mount/read test for `file://./...`?

bigbigxu: In OpenDal, the root directory must be set to `/` to use relative and absolute paths. If set to any other value, all paths will be appended to that value. For example, if root is set to `/a`, accessing `/a` will actually access `/a/a`. Setting `/` is necessary for proper access.

curvine-tests/tests/write_cache_test.rs
```diff
@@ -26,22 +26,17 @@ use std::time::Duration;
...
     if env::var("UFS_TEST_PATH").is_err() {
-        println!("⚠️  UFS_TEST_PATH is not set, skipping test");
-        println!(
-            "   Set UFS_TEST_PATH and UFS_TEST_PROPERTIES environment variables to run this test"
-        );
-        println!("   Example: export UFS_TEST_PATH=hdfs://127.0.0.1:9000");
-        println!("   Example: export UFS_TEST_PROPERTIES=\"hdfs.namenode=hdfs://127.0.0.1:9000,hdfs.user=root\"");
-        panic!("UFS_TEST_PATH is not set")
+        env::set_var("UFS_TEST_PATH", testing.ufs_path.clone());
     }
```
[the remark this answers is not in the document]
bigbigxu: This is a complex problem; simply deleting the code is not enough. The next step is to figure out how to run multiple tests concurrently.

PR #727, opened 2026-03-19
refactor: introduce StorageState (Cv/Ufs/Both) for fs_mode and cache_…

### LINKED ISSUE #719: [BUG]: Under fs_mode, the files being written do not have the correct TTL-related attributes set.

bigbigxu:
## Summary

**Core change:** introduce explicit **StorageState** (Cv, Ufs, Both) in `StoragePolicy` so "where data lives" is no longer inferred from `ufs_mtime` alone. This underpins both **fs_mode** (unified FS read/write and cache validity) and **cache_mode** (TTL eviction and load jobs): validity checks, `free()`, and post-copy metadata all use state. Additionally, mount TTL/storage options are merged into create options so fs_mode writes get the correct policy.

## Why introduce StorageState

Previously, "whether data exists in UFS" was inferred only from `ufs_mtime > 0`. That was ambiguous and led to bugs:

- **Reopen / truncate / overwrite**: We set `ufs_mtime = 0` to mean "cache invalid", but there was no explicit notion of "CV-only" vs "synced with UFS". TTL `free()` and cache validity logic had to guess from a single field.
- **After load job**: Both UFS→CV and CV→UFS need to record that the CV file is now in sync with UFS (`ufs_mtime` set). Without a clear state, it was easy to miss updating the right side or to apply TTL incorrectly.
- **free() semantics**: We only want to free CV blocks (and drop to "UFS-only") when data actually exists in both places and TTL allows it. Inferring "both exist" from `ufs_mtime > 0` plus block presence was fragile.

**StorageState** makes the lifecycle explicit:

| State   | Meaning |
|--------|---------|
| **Cv**  | Data exists only in Curvine (e.g. just written, or after reopen/truncate; not yet synced to UFS). |
| **Ufs** | Data exists only in UFS (e.g. after TTL `free()` evicted CV cache). |
| **Both**| Data is present in both CV and UFS and considered in sync (`ufs_mtime` valid). |

Transitions:

- **detach_ufs()**: On write start, reopen, truncate → state = Cv, ufs_mtime = 0.
- **save_ufs(ufs_mtime)**: After a copy (load job or sync) completes → state = Both, ufs_mtime set.
- **free()**: If state is Both and ttl_action allows → state = Ufs, CV blocks freed.

So `ufs_exists()` / `cv_exists()` / `both_exists()` are derived from state instead of heuristics, and TTL assignment on fs_mode writes is fixed by merging mount options and updating state only at well-defined points (create, copy completion, free).

### curvine-common

- **Proto**
  - Add `StorageStateProto` (Cv / Ufs / Both) and `state` field to `StoragePolicyProto` (default Cv).
- **StoragePolicy**
  - Introduce `StorageState` (Cv, Ufs, Both) and store it in `StoragePolicy`; see **Why introduce StorageState** above.
  - Add `with_cv(policy)` for new files (state=Cv, ufs_mtime=0).
  - Add `detach_ufs()` (state→Cv, ufs_mtime=0), `save_ufs(ufs_mtime)` (state→Both when mtime>0), `overwrite(new_policy)` (merge and detach_ufs).
  - Add `ufs_exists()` / `cv_exists()` / `both_exists()` based on state; `free()` uses state and ttl_action.
- **FileStatus**
  - `ufs_exists()` / `cv_exists()` delegate to `storage_policy`.
  - Add `cv_valid(ufs_status)` for cache validity: CV exists, not expired, UFS exists; if `ufs_status` is provided, also checks len and mtime match.
- **MountInfo**
  - Add `merge_create_opts(opts)`: merge mount’s ttl_ms, ttl_action, storage_type, replicas, block_size into `CreateFileOpts` so fs_mode writes get correct TTL/storage attributes.
- **FileBlocks**
  - Rename `cv_exists()` → `data_exists()`: for len==0 returns true; otherwise `cv_exists() && !block_locs.is_empty()`.

### curvine-client

- **CurvineFileSystem**
  - Remove `check_read_status` from `open()` (no read-side expired check before opening).
- **UnifiedFileSystem**
  - **Cache validity**: Use `cv_status.cv_valid(ufs_status)`; only run UFS verification when `read_verify_ufs`; otherwise use `cv_valid(None)`.
  - **Read path**: When valid, set `mtime = storage_policy.ufs_mtime` once; remove redundant `cv_exists` / `ufs_exists` branches.
  - **fs_mode create**: Call `mount.info.merge_create_opts(opts)` so TTL and storage options from the mount are applied on create; use `data_exists()` instead of `cv_exists()` for the writer branch.
  - **copy_from_ufs**: Apply `merge_create_opts` so target file gets mount TTL/storage.

### curvine-server

- **InodeFile**
  - New files use `StoragePolicy::with_cv(opts.storage_policy)`.
  - Replace `invalid_cache()` with `storage_policy.detach_ufs()`; replace direct `ufs_mtime = 0` with `storage_policy.overwrite()` where appropriate.
  - `reopen` / truncate / set_attr use `detach_ufs`; `free()` delegates to `storage_policy.free()`.
  - `ufs_exists()` / `cv_exists()` delegate to `storage_policy`; add `data_exists()` (same semantics as `FileBlocks`).
- **InodeView (set_attr)**
  - When setting `ufs_mtime`, call `storage_policy.save_ufs(ufs_mtime)` for files so state becomes Both.
  - Remove atime handling from set_attr.
- **FsDir (set_attr)**
  - Clearer error when set_attr is used on unresolved FileEntry; clone `child_opts` once; log warning when FileEntry inode is missing.
- **LoadTaskRunner**
  - After copy (both UFS→CV and CV→UFS), set `ufs_mtime` on the CV path via `set_attr` so storage state is updated correctly.
  - `create_unified(path)` no longer takes `read_status`; remove `ufs_mtime` from CreateFileOptsBuilder (state is updated after copy).

### Other

- **curvine-cli/stat**: Indentation fix.
- **rust-toolchain.toml**: Comment out explicit channel (optional / env-specific).

## Testing

- Verify fs_mode write: new file gets mount TTL and storage policy.
- Verify load job (UFS→CV and CV→UFS): after copy, CV inode has correct `ufs_mtime` and state (Both).
- Verify cache validity and read path use `cv_valid` and single mtime assignment.

## Issue

Fixes #719 (incorrect TTL attribute assignment on file writes in fs_mode).

---

### DIFF
--- a/curvine-client/src/file/curvine_filesystem.rs
+++ b/curvine-client/src/file/curvine_filesystem.rs
@@ -30,8 +30,8 @@ use curvine_common::FsResult;
 use log::info;
 use log::warn;
 use orpc::client::ClientConf;
+use orpc::err_box;
 use orpc::runtime::{RpcRuntime, Runtime};
-use orpc::{err_box, err_ext};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::time::timeout;
@@ -136,21 +136,8 @@ impl CurvineFileSystem {
         self.fs_client.exists(path).await
     }
 
-    fn check_read_status(path: &Path, file_blocks: &FileBlocks) -> FsResult<()> {
-        // if !file_blocks.status.is_complete {
-        //     return err_box!("Cannot read from {} because it is incomplete.", path);
-        // }
-
-        if file_blocks.status.is_expired() {
-            return err_ext!(FsError::file_expired(path.path()));
-        }
-
-        Ok(())
-    }
-
     pub async fn open(&self, path: &Path) -> FsResult<FsReader> {
         let file_blocks = self.fs_client.get_block_locations(path).await?;
-        Self::check_read_status(path, &file_blocks)?;
 
         let reader = FsReader::new(path.clone(), self.fs_context.clone(), file_blocks)?;
         Ok(reader)
--- a/curvine-client/src/unified/unified_filesystem.rs
+++ b/curvine-client/src/unified/unified_filesystem.rs
@@ -203,26 +203,17 @@ impl UnifiedFileSystem {
         ufs_path: &Path,
         mount: &MountValue,
     ) -> FsResult<CacheValidity> {
-        if cv_status.is_expired() {
-            return Ok(CacheValidity::Invalid(None));
-        }
-
-        if !cv_status.is_complete() || !cv_status.ufs_exists() {
-            return Ok(CacheValidity::Invalid(None));
-        }
-
-        if !mount.info.read_verify_ufs {
-            return Ok(CacheValidity::Valid);
-        }
-
-        let ufs_status = mount.ufs.get_status(ufs_path).await?;
-        if cv_status.len == ufs_status.len
-            && cv_status.storage_policy.ufs_mtime != 0
-            && cv_status.storage_policy.ufs_mtime == ufs_status.mtime
-        {
+        if mount.info.read_verify_ufs {
+            let ufs_status = mount.ufs.get_status(ufs_path).await?;
+            if cv_status.cv_valid(Some(&ufs_status)) {
+                Ok(CacheValidity::Valid)
+            } else {
+                Ok(CacheValidity::Invalid(Some(ufs_status)))
+            }
+        } else if cv_status.cv_valid(None) {
             Ok(CacheValidity::Valid)
         } else {
-            Ok(CacheValidity::Invalid(Some(ufs_status)))
+            Ok(CacheValidity::Invalid(None))
         }
     }
 
@@ -256,18 +247,12 @@ impl UnifiedFileSystem {
                 err_box!("path {} data lost", cv_path)
             }
         } else {
-            if !blocks.cv_exists() {
-                return Ok(None);
-            }
-
             match self
                 .check_cache_validity(&blocks.status, ufs_path, mount)
                 .await?
             {
                 CacheValidity::Valid => {
-                    if blocks.status.ufs_exists() {
-                        blocks.status.mtime = blocks.status.storage_policy.ufs_mtime;
-                    }
+                    blocks.status.mtime = blocks.status.storage_policy.ufs_mtime;
                     let cv_reader = FsReader::new(cv_path.clone(), self.cv.fs_context(), blocks)?;
                     Ok(Some(FallbackFsReader::new(
                         cv_reader,
@@ -338,6 +323,7 @@ impl UnifiedFileSystem {
         opts: CreateFileOpts,
         cv_len: i64,
     ) -> FsResult<()> {
+        let opts = mnt.info.merge_create_opts(opts);
         let ufs_path = mnt.get_ufs_path(path)?;
         let mut reader = mnt.ufs.open(&ufs_path).await?;
         if reader.len() != cv_len {
@@ -385,8 +371,9 @@ impl UnifiedFileSystem {
             }
 
             Some((_, mount)) if mount.info.is_fs_mode() => {
+                let opts = mount.info.merge_create_opts(opts);
                 let mut writer = self.cv.open_with_opts(path, opts.clone(), flags).await?;
-                if writer.file_blocks().cv_exists() || flags.overwrite() {
+                if writer.file_blocks().data_exists() || flags.overwrite() {
                     Ok(UnifiedWriter::Cv(writer))
                 } else {
                     writer.complete().await?;
@@ -526,7 +513,14 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
 
     async fn open(&self, path: &Path) -> FsResult<UnifiedReader> {
         let (ufs_path, mount) = match self.get_mount(path).await? {
-            None => return Ok(UnifiedReader::Cv(self.cv.open(path).await?)),
+            None => {
+                let reader = UnifiedReader::Cv(self.cv.open(path).await?);
+                return if reader.status().is_expired() {
+                    return err_ext!(FsError::file_expired(path.path()));
+                } else {
+                    Ok(reader)
+                };
+            }
             Some(v) => v,
         };
 
@@ -631,9 +625,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
             Some((ufs_path, mnt)) => match self.cv.get_status(path).await {
                 Ok(mut v) => match self.check_cache_validity(&v, &ufs_path, &mnt).await? {
                     CacheValidity::Valid => {
-                        if v.ufs_exists() {
-                            v.mtime = v.storage_policy.ufs_mtime;
-                        }
+                        v.mtime = v.storage_policy.ufs_mtime;
                         Ok(v)
                     }
                     CacheValidity::Invalid(Some(ufs_status)) => Ok(ufs_status),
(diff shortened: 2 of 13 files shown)


curvine-common/src/state/storage_policy.rs
```diff
@@ -13,16 +13,97 @@
...
+    Both = 3,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct StoragePolicy {
     pub storage_type: StorageType,
     pub ttl_ms: i64,
     pub ttl_action: TtlAction,
     pub ufs_mtime: i64,
+    pub state: StorageState,
```
szbr00, line 46: Adding `state` here also changes the bincode layout of every persisted `InodeFile` / `InodeDir` and journal entry that embeds `StoragePolicy`. `RocksInodeStore::get_inode()` and journal replay still deserialize old bytes directly into the new type, so an upgrade can fail while loading existing metadata even before any proto-level fallback runs. Could we add an explicit serde migration path (for example `#[serde(default)]` plus legacy-state inference, or version the on-disk structs) before persisting this new field?

curvine-common/src/utils/proto_utils.rs
```diff
@@ -209,6 +210,7 @@ impl ProtoUtils {
             ttl_ms: policy.ttl_ms,
             ttl_action: TtlAction::from(policy.ttl_action),
             ufs_mtime: policy.ufs_mtime,
+            state: StorageState::from(policy.state as i8),
```
szbr00, line 213: Even if we make missing/unknown `state` decode safely, defaulting legacy payloads to `StorageState::Cv` is still a migration bug. Before this PR, `ufs_mtime > 0` was the persisted signal that a file had already been synced to UFS; after upgrade, the same old record/message will now report `ufs_exists() == false`, which changes `free()` eligibility, cache-validity checks, and load-job skip decisions. Please infer the initial state from legacy fields (`ufs_mtime`, and on inode-side whether CV blocks still exist) instead of treating all pre-state records as CV-only.

curvine-client/src/file/curvine_filesystem.rs
```diff
@@ -136,21 +136,8 @@ impl CurvineFileSystem {
...
-
-        if file_blocks.status.is_expired() {
-            return err_ext!(FsError::file_expired(path.path()));
-        }
-
-        Ok(())
-    }
-
     pub async fn open(&self, path: &Path) -> FsResult<FsReader> {
         let file_blocks = self.fs_client.get_block_locations(path).await?;
```
szbr00, line 140: Removing the read-status check here changes the behavior of plain `CurvineFileSystem` reads, not just the unified fallback path. Callers like `cv fs cat/get`, FUSE/node-state reads, and other direct `fs.open()` users can now read an expired CV file successfully instead of getting the previous `file_expired` error, so TTL is no longer enforced on those paths. Could we keep the expiry guard for direct CV reads and only bypass it in the unified path that revalidates against UFS?

curvine-server/src/master/meta/inode/inode_view.rs
```diff
@@ -346,15 +346,6 @@ impl InodeView {
...
-        // Handle time modifications
-        if let Some(atime) = opts.atime {
-            match self {
-                File(_, f) => f.atime = atime,
-                Dir(_, d) => d.atime = atime,
-                _ => (),
-            }
-        }
-
         if let Some(mtime) = opts.mtime {
```
szbr00, line 358: `SetAttrOpts` still carries `atime`, and the FUSE setattr path still populates it for `utimens`/`touch -a`, but this method no longer applies `opts.atime` at all. That changes runtime behavior silently: access-time updates now succeed from the caller’s point of view while Curvine keeps the old `atime`, and TTL expiry is still computed from that stale value. Could we keep the old `atime` handling here (or explicitly reject unsupported access-time updates) so mounted files do not expire based on a timestamp that users just tried to refresh?

curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -256,18 +247,12 @@ impl UnifiedFileSystem {
...
-
             match self
                 .check_cache_validity(&blocks.status, ufs_path, mount)
                 .await?
             {
                 CacheValidity::Valid => {
-                    if blocks.status.ufs_exists() {
-                        blocks.status.mtime = blocks.status.storage_policy.ufs_mtime;
-                    }
+                    blocks.status.mtime = blocks.status.storage_policy.ufs_mtime;
```
szbr00, line 255: This fs-mode read path still checks `blocks.cv_exists()`, but after this refactor that resolves to `FileStatus::cv_exists()` and only looks at `StorageState`. The old “non-empty file must also have block locations” guard was moved to `FileBlocks::data_exists()`, and the writer path was updated to use it, but the reader path was not. That means a positive-length file in `Cv`/`Both` state with missing block locations can now be treated as cache-present and handed to `FsReader`, instead of cleanly falling back to UFS or reporting the cache as lost. Could this branch switch to `data_exists()` as well?

curvine-common/src/state/storage_policy.rs
```diff
@@ -13,16 +13,97 @@
...
+    Eq,
+    Default,
+    Hash,
+)]
+pub enum StorageState {
+    #[default]
+    Cv = 1,
+    Ufs = 2,
+    Both = 3,
+}
```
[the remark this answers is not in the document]
bigbigxu: num_enum is now compatible with the #[default] macro.

curvine-common/src/state/file_status.rs
```diff
@@ -87,6 +87,26 @@ impl FileStatus {
...
+            return false;
+        }
+        if self.is_expired() || !self.ufs_exists() {
+            return false;
+        }
+        if let Some(ufs_status) = ufs_status {
+            self.len == ufs_status.len && self.storage_policy.ufs_mtime == ufs_status.mtime
+        } else {
+            true
+        }
```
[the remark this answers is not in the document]
bigbigxu: This check is no longer necessary; the cache is valid only if ufs_mtime is not 0. This check is based on accuracy.

curvine-client/src/unified/unified_filesystem.rs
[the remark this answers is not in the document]
bigbigxu: The cache is valid only if ufs_time is not zero, but this check is no longer needed.

bigbigxu: The cache is valid only if ufs_time is not zero, but this check is no longer needed.

curvine-common/src/state/mount.rs
```diff
@@ -182,13 +204,18 @@ impl MountOptions {
...
         MountInfo {
             cv_path: cv_path.to_string(),
             ufs_path: ufs_path.to_string(),
             mount_id,
             properties: self.add_properties,
             ttl_ms: self.ttl_ms.unwrap_or(0),
-            ttl_action: self.ttl_action.unwrap_or(TtlAction::None),
+            ttl_action,
             read_verify_ufs: self.read_verify_ufs,
             storage_type: self.storage_type,
```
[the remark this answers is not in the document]
bigbigxu: `write_type` is now bound to `ttl_action`. For example, if `fs_mode` is set, `ttl_action` must be `Free` and cannot be modified.

szbr9486 (approved):

szbr9486 (approved):

PR #731, opened 2026-03-20
fix(test): stabilize journal replay and rpc setup

lzjqsdd:
## Problem
Journal replay tests could pass without exercising the real apply path, and full
`JournalSystem` reopen could keep RocksDB handles alive in the same process.
The local ORPC pool test also depended on startup timing instead of actual server
readiness.

## Design
Route replay-related tests through the real `AppStorage::apply` path, add an
explicit shutdown handshake for the journal apply loop, and make hardlink replay
assertions validate inode and `nlink` semantics directly. Keep the RPC pool test
focused on pool behavior by waiting for the local test server to become ready.

## Key Changes
- add a shutdown message/ack flow so `JournalSystem` tests release journal
  resources before reopen
- update journal and master-fs tests to replay entries through
  `AppStorage::apply` instead of a test-only helper path
- replace the hardlink replay hash assertion with direct inode and `nlink`
  validation
- replace fixed sleep in the block client pool test with active readiness polling

### DIFF
--- a/curvine-server/src/master/journal/journal_loader.rs
+++ b/curvine-server/src/master/journal/journal_loader.rs
@@ -68,7 +68,7 @@ impl JournalLoader {
         let client = RaftClient::from_conf(rt.clone(), conf);
         let journal_writer = Arc::new(JournalWriter::new(true, client, conf));
         let log_store = RocksLogStorage::from_conf(conf, false);
-        Self::new(
+        Self::build(
             rt,
             fs_dir,
             mnt_mgr,
@@ -81,7 +81,29 @@ impl JournalLoader {
     }
 
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub(crate) fn new(
+        rt: Arc<Runtime>,
+        fs_dir: SyncFsDir,
+        mnt_mgr: Arc<MountManager>,
+        conf: &JournalConf,
+        job_manager: Arc<JobManager>,
+        log_store: RocksLogStorage,
+        journal_writer: Arc<JournalWriter>,
+    ) -> Self {
+        Self::build(
+            rt,
+            fs_dir,
+            mnt_mgr,
+            conf,
+            job_manager,
+            log_store,
+            journal_writer,
+            false,
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn build(
         rt: Arc<Runtime>,
         fs_dir: SyncFsDir,
         mnt_mgr: Arc<MountManager>,
@@ -328,6 +350,11 @@ impl JournalLoader {
                     }
                 }
 
+                ApplyMsg::Shutdown(tx) => {
+                    let _ = tx.send(());
+                    break;
+                }
+
                 msg => match self.apply_msg(is_leader, &msg).await {
                     Ok(_) => retry_num = 0,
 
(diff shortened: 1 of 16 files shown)


curvine-server/src/master/journal/journal_loader.rs
```diff
@@ -427,6 +432,18 @@ impl JournalLoader {
         }
     }
 
+    pub fn replay_entry(&self, entry: JournalEntry) -> CommonResult<()> {
```
szbr00: `replay_entry()` now updates `op_id` and `last_inode_id`, but it still does not advance `fsm_state.applied` / `journal_applied` the way `apply0()` does on the real replay path. That means manual replay helpers can still diverge from production state if a test (or future code) inspects applied index, creates a snapshot, or depends on replay progress after calling this helper. Could we either update the applied state here as well, or make the helper name/docs explicit that it only replays metadata mutations and does not keep Raft replay state fully in sync?

orpc/src/client/client_factory.rs
```diff
@@ -172,6 +173,23 @@ mod tests {
         let rt = Arc::new(AsyncRuntime::single());
         server.start(0);
 
+        let conf = ClientConf::default();
+        let rt_wait = rt.clone();
+        let wait_addr = addr.clone();
+        rt.block_on(async move {
```
szbr00, line 179: This makes the `client_factory` pool test much less flaky, but the same `SimpleServer::default() + start(0)` pattern still exists in `curvine-client/tests/block_client_pool_test.rs`, where it continues to rely on a fixed `sleep(100ms)` before creating clients. If the goal of this PR is to stabilize local RPC test setup more broadly, that suite can still fail on slow runners for the same reason. Could we switch that test to the same active readiness check as well?

curvine-server/tests/master_fs_test.rs
```diff
@@ -51,7 +50,42 @@ fn new_fs(format: bool, name: &str) -> (MasterFilesystem, JournalSystem) {
...
+            meta_dir: Utils::test_sub_dir(format!("master-fs-test/meta-{}", name)),
+            ..Default::default()
+        },
+        journal: JournalConf {
+            enable: false,
+            journal_dir: Utils::test_sub_dir(format!(
+                "master-fs-test/journal-{}-{}",
+                journal_name,
+                Utils::rand_str(6)
+            )),
```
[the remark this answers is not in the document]
lzjqsdd: Thanks for catching this. Updated in the latest patch to reuse the same journal directory within a single test run, so the reopen regression now exercises the real RocksDB reopen path instead of a fresh directory.

curvine-server/src/master/journal/journal_loader.rs
lzjqsdd: Thanks for the feedback. I removed the test-only `replay_entry()` helper and switched the replay tests to go through the real `AppStorage::apply` path, so applied index and replay state stay aligned with production behavior.

orpc/src/client/client_factory.rs
lzjqsdd: Thanks for the suggestion. Updated in the latest patch: `curvine-client/tests/block_client_pool_test.rs` now uses an active readiness check instead of the fixed sleep, so the pool test waits for the local server to become reachable before asserting pool behavior.

szbr9486 (approved):

PR #732, opened 2026-03-20
fix(server): avoid test-mode journal apply channel closure

lzjqsdd:
## Problem
Journal tests could fail with `channel closed` when snapshot or apply operations
were sent through the async apply channel while the test-mode apply worker was
disabled.

## Design
Keep the existing async path for normal runtime behavior, but use a synchronous
fallback when the journal loader is created without an apply worker in test
mode. This keeps the fix minimal and avoids reintroducing background-task
resource retention in tests.

## Key Changes
- Track whether `JournalLoader` has an async apply worker.
- Execute apply operations synchronously when the worker is disabled.
- Run snapshot creation and snapshot application directly in test mode instead
  of enqueueing them to a closed channel.

### DIFF
--- a/curvine-server/src/master/journal/journal_loader.rs
+++ b/curvine-server/src/master/journal/journal_loader.rs
@@ -54,6 +54,7 @@ pub struct JournalLoader {
     batch_size: u64,
     retry_interval: Duration,
     metrics: &'static MasterMetrics,
+    has_apply_worker: bool,
 }
 
 impl JournalLoader {
@@ -107,6 +108,7 @@ impl JournalLoader {
             batch_size: conf.scan_batch_size,
             retry_interval: Duration::from_secs(conf.retry_interval_secs),
             metrics: Master::get_metrics(),
+            has_apply_worker: !testing,
         };
 
         if !testing {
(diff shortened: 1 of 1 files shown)


curvine-server/src/master/journal/journal_loader.rs
```diff
@@ -715,7 +717,7 @@ impl JournalLoader {
 
 impl AppStorage for JournalLoader {
     async fn apply(&self, wait: bool, msg: ApplyMsg) -> RaftResult<()> {
-        if wait {
+        if wait || !self.has_apply_worker {
```
szbr00, line 720: This avoids the closed-channel failure, but the no-worker path still diverges from real `JournalLoader` semantics: `apply()` now always calls `apply_msg(false, ...)`, and `role_change()` becomes a no-op when `has_apply_worker == false`. In the normal worker path, `RoleChange(Leader)` flips `is_leader` and schedules a `Scan` from `ufs_applied`, so `apply0()` can take the leader branch (`ufs_loader.apply_entry`, leader-only `Snapshot` handling, and `set_applied` behavior). With this fallback, test-mode replay still always runs as follower semantics and never performs that leader catch-up. Could we preserve the current role and trigger the same synchronous `Scan`/leader path when the worker is disabled, so tests exercise the same replay behavior instead of only avoiding the channel error?

szbr9486 (approved):

PR #740, opened 2026-03-25
feat(fs): add list stream interface and protocol support

bigbigxu:
### Summary
- Add a new `ListStream` abstraction in `curvine-common` for streaming file list results.
- Extend protobuf definitions and utility conversions to support list-stream related fields.
- Update client/common integration points (macros, modules, and options) to consume the new interface.
- Refresh workspace dependency manifests/lockfile for the new additions.

### Why
This change establishes a dedicated stream-based listing contract, reducing coupling between listing producers and consumers and preparing the codebase for consistent, paginated, and extensible unified listing behavior.

### DIFF
--- a/curvine-common/proto/master.proto
+++ b/curvine-common/proto/master.proto
@@ -331,3 +331,12 @@ message SetLockRequest {
 message SetLockResponse {
     optional FileLockProto conflict = 1;
 }
+
+message ListOptionsRequest {
+    required string path = 1;
+    required ListOptionsProto options = 2;
+}
+
+message ListOptionsResponse {
+    repeated FileStatusProto statuses = 1;
+}
--- a/curvine-common/src/fs/filesystem.rs
+++ b/curvine-common/src/fs/filesystem.rs
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::fs::{FsKind, Path};
-use crate::proto::{GetFileStatusResponse, ListStatusResponse};
-use crate::state::{FileStatus, SetAttrOpts};
+use crate::fs::{FsKind, ListStream, Path};
+use crate::proto::{GetFileStatusResponse, ListOptionsResponse, ListStatusResponse};
+use crate::state::{FileStatus, ListOptions, SetAttrOpts};
 use crate::utils::ProtoUtils;
 use crate::FsResult;
+use orpc::err_box;
 use prost::bytes::BytesMut;
 use std::future::Future;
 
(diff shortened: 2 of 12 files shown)


curvine-common/src/fs/filesystem.rs
```diff
@@ -65,4 +66,38 @@ pub trait FileSystem<Writer, Reader> {
...
+            let statuses = self
+                .list_options(path, opts)
+                .await?
+                .into_iter()
+                .map(ProtoUtils::file_status_to_pb)
+                .collect();
+
+            let rep = ListStatusResponse { statuses };
+            Ok(ProtoUtils::encode(rep)?)
+        }
```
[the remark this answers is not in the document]
bigbigxu: fixed

curvine-common/proto/master.proto
```diff
@@ -330,4 +330,13 @@ message SetLockRequest {
 // SetLock response
 message SetLockResponse {
     optional FileLockProto conflict = 1;
+}
+
+message ListOptionsRequest {
```
szbr00, line 335: This adds dedicated `ListOptionsRequest/Response` messages, but I don’t see the rest of the RPC path in this PR: `RpcCode` is unchanged, `FsClient` has no `list_options` request method, and `MasterHandler` has no branch to parse or serve this request. So the new wire contract exists in proto only and cannot actually be called. Could we either wire the full RPC end-to-end in the same change, or defer these messages until there is a reachable request path?

curvine-common/src/fs/filesystem.rs
```diff
@@ -65,4 +66,38 @@ pub trait FileSystem<Writer, Reader> {
     }
 
     fn set_attr(&self, path: &Path, opts: SetAttrOpts) -> impl Future<Output = FsResult<()>>;
+
+    fn list_options(
```
szbr00, line 70: This adds public `list_options` / `list_stream` methods to the shared `FileSystem` trait, but the main concrete implementations in this branch (`CurvineFileSystem` and `UnifiedFileSystem`) do not override them yet, so they still inherit the default `not supported` behavior. The enum forwarding in `curvine-client/src/unified/macros.rs` makes the API look wired, but any caller that tries to use it will still fail at runtime. Could we either add real implementations here before exposing the methods, or keep the new API internal until there is at least one working backend?

curvine-common/proto/master.proto
bigbigxu: The current pull request only implements the interface specification; the code will be implemented in a subsequent pull request.

curvine-common/src/fs/filesystem.rs
bigbigxu: The current pull request only implements the interface specification; the code will be implemented in a subsequent pull request.

szbr9486 (approved):

PR #744, opened 2026-03-26
feat(client): add ListOptions RPC, list_stream, and FileSystem wiring

bigbigxu:
### Summary

- **`FsClient` / `CurvineFileSystem`**
  - Add **`list_options`** and **`list_options_bytes`** using `RpcCode::ListOptions` with protobuf request/response conversion.
  - Add **`list_stream`**: async stream that repeatedly calls **`list_options`** with the same `limit`, advancing **`start_after`** from the last entry name until a short page or empty result ends iteration.
  - Implement **`FileSystem<FsWriter, FsReader>`** for **`CurvineFileSystem`** (including the new listing APIs) so the client matches the common **`FileSystem`** trait surface.

- **`UnifiedFileSystem`**
  - Implement **`list_options`**, **`list_options_bytes`**, and **`list_stream`**, dispatching to **`CurvineFileSystem`** or the mounted UFS backend according to path resolution (with metrics on the typed list paths).

- **`curvine-ufs` (`OpendalFileSystem`)**
  - Implement **`list_stream`** using OpenDAL **`lister_options`**, mapping `ListOptions` into OpenDAL list options (including `start_after` prefixed with the directory object path) and filtering out the synthetic “self” directory entry when it matches the listed path.

- **`curvine-common`**
  - Small adjustment in **`list_stream`** (supporting the new usage pattern).

- **Tests**
  - **`fs_test`**: coverage for **`list_options`** (`limit`, `start_after`, combined) and **`list_stream`** pagination vs full directory order; empty-directory **`list_stream`** behavior.
  - **`ufs_test`**: **`list_stream`** over unified UFS-backed paths (non-empty and empty dirs).

### Motivation

Expose **paginated directory listing** and a **streaming list API** consistently across Curvine (RPC), unified layout (mount-aware), and OpenDAL-backed UFS, so upper layers (e.g. FUSE, tests, tools) can use **`FileSystem::list_stream`** / **`list_options`** without ad hoc pagination logic.

### Test plan

- `cargo test -p curvine-tests --test fs_test` (or run the new list tests selectively).
- `cargo test -p curvine-tests --test ufs_test` (or run UFS list_stream coverage).
- `cargo check -p curvine-client -p curvine-ufs -p curvine-common`.

### DIFF
--- a/curvine-client/src/file/curvine_filesystem.rs
+++ b/curvine-client/src/file/curvine_filesystem.rs
@@ -15,11 +15,12 @@
 use crate::block::BatchBlockWriter;
 use crate::file::{FsClient, FsContext, FsReader, FsWriter, FsWriterBase};
 use crate::ClientMetrics;
+use async_stream::stream;
 use bytes::BytesMut;
 use curvine_common::conf::ClusterConf;
 use curvine_common::error::FsError;
-use curvine_common::fs::{Path, Reader, Writer};
-use curvine_common::state::{CommitBlock, FreeResult};
+use curvine_common::fs::{FileSystem, FsKind, ListStream, Path, Reader, Writer};
+use curvine_common::state::{CommitBlock, FreeResult, ListOptions};
 use curvine_common::state::{
     CreateFileOpts, CreateFileOptsBuilder, FileAllocOpts, FileBlocks, FileLock, FileStatus,
     MasterInfo, MkdirOpts, MkdirOptsBuilder, MountInfo, MountOptions, OpenFlags, SetAttrOpts,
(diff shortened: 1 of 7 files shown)


curvine-ufs/src/opendal.rs
```diff
@@ -918,4 +919,46 @@ impl FileSystem<OpendalWriter, OpendalReader> for OpendalFileSystem {
...
         err_ufs!("SetAttr operation is not supported by OpenDAL file system")
     }
+
+    async fn list_stream(&self, path: &Path, opts: ListOptions) -> FsResult<ListStream> {
+        let fs = self.clone();
+        let path = path.clone();
+        let dir_path = fs.get_dir_path(&path)?;
+        let opts = opendal::options::ListOptions {
+            limit: opts.limit,
+            start_after: opts.start_after.map(|x| format!("{}{}", dir_path, x)),
```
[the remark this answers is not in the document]
bigbigxu: In curvine, start_after accepts a file name. In opendal, start_after requires the full object path. Therefore, dir_path needs to be appended here.

curvine-client/src/file/curvine_filesystem.rs
```diff
@@ -195,6 +196,55 @@ impl CurvineFileSystem {
...
+        let path = path.clone();
+        let (limit, mut start_after) = (options.limit, options.start_after);
+
+        let stream = stream! {
+            loop {
+                let options = ListOptions {
+                    limit,
+                    start_after: start_after.clone(),
+                };
+                let list = match fs.list_options(&path, options).await {
```
szbr00, line 214: This loop assumes `list_options()` will eventually return an empty or short page once `start_after` advances, but that is not true for file targets on the current server side: the `File(..)` / top-level `FileEntry(..)` branches still return the same single status without honoring `limit` or `start_after`. In that case `last_name` never makes progress and this stream becomes non-terminating for file paths. Could we special-case non-directory targets here (or otherwise guard against repeated cursors) before exposing `list_stream` as a general API?

curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -666,6 +667,37 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
...
 
+    async fn list_options(&self, path: &Path, options: ListOptions) -> FsResult<Vec<FileStatus>> {
+        let _timer = TimeSpent::timer_counter_vec(
+            Arc::new(FsContext::get_metrics().metadata_operation_duration.clone()),
+            vec!["list_options".to_string()],
+        );
+
+        match self.get_mount_checked(path).await? {
+            None => self.cv.list_options(path, options).await,
+            Some((ufs_path, mount)) => mount.ufs.list_options(&ufs_path, options).await,
```
szbr00, line 678: The mounted-UFS branch now forwards `list_options`/`list_options_bytes` into `mount.ufs`, but `OpendalFileSystem` in this PR only implements `list_stream`; it still inherits the default `FileSystem::list_options()` implementation that returns `not supported list_options`. So `UnifiedFileSystem::list_options()` will still fail for mounted paths even though the API now looks wired end-to-end. Could we either implement paged `list_options` on the UFS backend in the same change, or avoid exposing the unified mounted-path entry points until that backend exists?

curvine-client/src/file/curvine_filesystem.rs
bigbigxu: This is indeed a serious bug. When limit 1 is set to 1, listing a file may cause the loop to fail to terminate. I will fix it.

curvine-client/src/unified/unified_filesystem.rs
bigbigxu: `list_options` is used to implement `list_stream`. OpenDal does not have an interface for `list_options`; it's implemented here simply to maintain interface consistency.

szbr9486 (approved):

PR #745, opened 2026-03-26
feat(fuse): stream-based readdir, list_options fixes, and env defaults

bigbigxu:
### Summary

- **FUSE `readdir` / `DirHandle`**
  - Directory handles now wrap a **`ListStream`** with an internal **`VecDeque`** buffer and configurable **`list_limit`** (from **`FuseConf`**).
  - **`get_batch(offset)`** fills the buffer from the stream, supports **resume after FUSE offset** when inner index was reset (e.g. restore), and **`set_buf`** returns unconsumed entries when the reply buffer is full.
  - **`opendir`** creates handles via **`NodeState::new_dir_handle(ino, path)`** instead of materializing a full **`Vec<FileStatus>`** up front.

- **State persistence**
  - Serialized **`DirHandle`** stores **`path`**; **`stream`** is skipped and **recreated on `restore`** via **`list_stream(path)`**, keeping behavior consistent after restart.

- **FUSE filesystem glue**
  - **`read_dir_common`** uses **`add_dirent`** (plus/non-plus), **`arg.offset`** as the running cookie, and passes **`&FileStatus`** into lookup helpers where appropriate.
  - **`fuse_dirent_list`**: new **`add_dirent`** helper to branch **`add` / `add_plus`**.

- **Session / observability**
  - **`FuseResponse`**: shared **`rep_log`** for errors; suppress routine noise for **`ENOENT` / `ENODATA` / `ENOSYS`** unless debug is on.
  - **FUSE web server**: expose **`FsContext::get_metrics()`** for **`/metrics`**; use **`log::info`** for bind message.

- **Configuration**
  - **`ClusterConf::DEFAULT_FUSE_WEB_PORT`** (**`9002`**); **`FuseConf`** default **`web_port`** uses it and adds **`list_limit`** (default **1000**).

- **Common / server**
  - **`ListStream::collect`** renamed to **`collect_vec`** to avoid confusion with **`StreamExt::collect`**.
  - **`FsDir::list_options`**: for **file** / **`FileEntry`** targets, apply the same **`start_after`** (exclusive) and **`limit`** semantics as directory children via **`list_single_file`**.

- **Tests**
  - **`fs_test`**: **`list_options` / `list_stream`** on a **file** path (including **`start_after`** empty result).
  - **`fuse_test`**: persist/restore updated for **path-based** **`DirHandle`** (mkdir setup, assert **`path`** instead of list length).

- **Tooling / deps**
  - **`etc/curvine-env.sh`**: fixed **`localhost`** hostnames, set **`ORPC_BIND_HOSTNAME=0.0.0.0`**, drop previous OS-specific IP discovery and **`:-` defaults** for Curvine env vars (stricter, dev-oriented).
  - **`curvine-fuse`**: add **`futures`**; trim unused deps from lockfile where applicable.

### Motivation

Large directories should not require loading the full listing into memory for **`opendir`**. Streaming aligns FUSE with **master/UFS `list_stream`**, improves **offset/cookie** handling for partial **`readdir`** replies, and fixes **pagination edge cases** when **`list_options`** targets a single file.

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -254,28 +254,32 @@ impl CurvineFileSystem {
     ) -> FuseResult<FuseDirentList> {
         let handle = self.state.find_dir_handle(header.nodeid, arg.fh)?;
 
-        let mut map = self.state.node_write();
         let mut res = FuseDirentList::new(arg);
-        for (index, status) in handle.get_list(arg.offset as usize) {
-            let attr = if status.name != FUSE_CURRENT_DIR && status.name != FUSE_PARENT_DIR {
-                if self.conf.enable_meta_cache {
-                    let path = Path::from_str(&status.path)?;
-                    self.state.meta_cache().put_status(&path, status.clone());
-                }
-                map.do_lookup(header.nodeid, Some(&status.name), status)?
-            } else {
-                Self::status_to_attr(&self.conf, status)?
-            };
-            let entry = Self::create_entry_out(&self.conf, attr);
+        let mut index = arg.offset;
+        let mut batch = handle.get_batch(arg.offset as usize).await?;
+        {
+            let mut map = self.state.node_write();
+            while let Some(status) = batch.pop_front() {
+                let attr = if status.name != FUSE_CURRENT_DIR && status.name != FUSE_PARENT_DIR {
+                    if self.conf.enable_meta_cache {
+                        let path = Path::from_str(&status.path)?;
+                        self.state.meta_cache().put_status(&path, status.clone());
+                    }
+                    map.do_lookup(header.nodeid, Some(&status.name), &status)?
+                } else {
+                    Self::status_to_attr(&self.conf, &status)?
+                };
 
-            if plus {
-                if !res.add_plus((index + 1) as u64, status, entry) {
+                let entry = Self::create_entry_out(&self.conf, attr);
+                if !res.add_dirent(plus, index, &status, entry) {
+                    batch.push_front(status);
                     break;
                 }
-            } else if !res.add((index + 1) as u64, status, entry) {
-                break;
+                index += 1;
             }
         }
+        handle.set_buf(batch).await?;
+
         Ok(res)
     }
 
@@ -545,7 +549,7 @@ impl CurvineFileSystem {
         Ok(status)
     }
 
-    async fn get_cached_list(&self, path: &Path) -> FuseResult<Vec<FileStatus>> {
+    pub async fn get_cached_list(&self, path: &Path) -> FuseResult<Vec<FileStatus>> {
         if self.conf.enable_meta_cache {
             if let Some(list) = self.state.meta_cache().get_list(path) {
                 return Ok(list);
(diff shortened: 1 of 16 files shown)


curvine-fuse/src/fs/state/dir_handle.rs
```diff
@@ -12,35 +12,106 @@
...
+        if off > 0 && guard.index == 0 {
+            info!(
+                "readdir {} skipping {} list entries (resume offset / inner index reset)",
+                self.path, off
+            );
+            while guard.index < off {
+                match guard.stream.next().await {
+                    Some(Ok(_)) => guard.index += 1,
+                    Some(Err(e)) => return Err(e),
+                    None => {
```
szbr00: On restore we rebuild the directory stream from the current path and then skip `off` items to catch back up, but this branch turns “the directory is now shorter than the old cookie” into an error. That can happen legitimately if entries were removed between persist and restore, and in that case `readdir` should normally just return EOF from the saved offset rather than failing the directory handle. Could this treat early stream exhaustion as an empty remainder instead of returning an error here?

etc/curvine-env.sh
```diff
@@ -18,26 +18,17 @@
...
-    LOCAL_IP=$(ifconfig | grep "inet " | awk '{print $2}' | grep -v 127.0.0.1 | tail -n 1)
-else
-    echo "Unsupported OS: $OS_NAME"
-    exit 1
-fi
+LOCAL_HOSTNAME=localhost
 
 # master bound host name
-export CURVINE_MASTER_HOSTNAME=${CURVINE_MASTER_HOSTNAME:-$LOCAL_HOSTNAME}
+export CURVINE_MASTER_HOSTNAME=$LOCAL_HOSTNAME
```
szbr00, line 24: These assignments now overwrite `CURVINE_MASTER_HOSTNAME` / `CURVINE_WORKER_HOSTNAME` / `CURVINE_CLIENT_HOSTNAME` unconditionally instead of preserving values that were already exported by the caller. Because this script is sourced by the launcher scripts, that means any deployment/test wrapper that sets custom hostnames before sourcing `curvine-env.sh` will now be silently forced back to `localhost`, which breaks non-local and multi-host setups. Could we keep the new localhost defaults but still honor pre-set environment values (for example via `${VAR:-...}`) so callers can override them?

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -254,28 +254,31 @@ impl CurvineFileSystem {
...
-                if !res.add_plus((index + 1) as u64, status, entry) {
-                    break;
-                }
-            } else if !res.add((index + 1) as u64, status, entry) {
+            let entry = Self::create_entry_out(&self.conf, attr);
+            if !res.add_dirent(plus, index, &status, entry) {
+                batch.push_front(status);
                 break;
             }
+            index += 1;
```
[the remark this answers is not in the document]
bigbigxu: The offset points to the next entry that needs to be read; the previous code is incorrect.

etc/curvine-env.sh
bigbigxu: This configuration file is provided as an example and intended to be simple. Users may customize the settings according to their own environment.

szbr9486 (approved):

PR #671, opened 2026-02-28
feat(p2p): add client-side p2p read acceleration with runtime policy

jlon:
## Core functionality summary
- Introduces a client-side P2P read acceleration module (`curvine-client/src/p2p`) with service/cache/discovery/transfer layering.
- Adds chunk-level cache manager with TTL, capacity eviction, integrity checksum, mtime validation, and corruption/mismatch accounting.
- Integrates libp2p-based transport/discovery (request/response + mDNS + DHT) plus runtime flow controls (QPS, inflight windows, adaptive behavior).
- Adds master-driven runtime policy controls for peer/tenant whitelist and signed policy version transitions.
- Extends P2P metrics and trace context for observability and rollout safety.

## Current stage scope (incremental)
- Focused on read-path acceleration and policy synchronization foundation.
- Includes core integration across client/common/master plus targeted tests and production-gate script.
- Does not yet cover full production hardening beyond current test gate and staged policy rollout controls.

## Follow-up plan
1. Expand large-scale and chaos scenarios (peer churn, high parallelism, network degradation).
2. Tune adaptive scheduling and cache policy based on benchmark/telemetry feedback.
3. Harden rollout/rollback tooling and operational docs for wider production adoption.

PR #762, opened 2026-03-26
feat(p2p-01): add file version epoch proto foundation (#671)

jlon:
## Summary
- Add `version_epoch` to shared file metadata used by later P2P read identity.
- Keep legacy `FileStatus` payloads decodable with a default epoch of `0`.

## Design
- Extend the existing `FileStatus` path instead of introducing a parallel P2P-only message.
- Seed inode-side version state now so later cache and transfer keys can stay version-safe.

## Implementation
- Update `common.proto`, `FileStatus`, proto conversions, and inode view/file mappings.
- Add proto compatibility coverage for round-trip and legacy decode behavior.

## Verification
- `cargo test -p curvine-common file_status_proto_roundtrip_keeps_version_epoch`
- `cargo test -p curvine-common file_status_proto_decodes_legacy_payload_without_version_epoch`

### DIFF
--- a/curvine-common/proto/common.proto
+++ b/curvine-common/proto/common.proto
@@ -72,7 +72,7 @@ message FileStatusProto {
     required uint32 mode = 17;
     optional string target = 18;
     required uint32 nlink = 19;
-    required int64 version_epoch = 20 [default = 0];
+    optional int64 version_epoch = 20 [default = 0];
 }
 
 // Describe the worker address information.
--- a/curvine-common/proto/common.proto
+++ b/curvine-common/proto/common.proto
@@ -72,6 +72,7 @@ message FileStatusProto {
     required uint32 mode = 17;
     optional string target = 18;
     required uint32 nlink = 19;
+    required int64 version_epoch = 20 [default = 0];
 }
 
 // Describe the worker address information.
--- a/curvine-common/src/state/file_status.rs
+++ b/curvine-common/src/state/file_status.rs
@@ -21,6 +21,7 @@ use std::collections::HashMap;
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct FileStatus {
     pub id: i64,
+    pub version_epoch: i64,
     pub path: String,
     pub name: String,
     pub is_dir: bool,
--- a/curvine-common/src/state/file_status.rs
+++ b/curvine-common/src/state/file_status.rs
@@ -21,6 +21,7 @@ use std::collections::HashMap;
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct FileStatus {
     pub id: i64,
+    #[serde(default)]
     pub version_epoch: i64,
     pub path: String,
     pub name: String,
@@ -121,5 +122,29 @@ impl FileStatus {
         } else {
             false
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::Value;
+
+    #[test]
+    fn file_status_serde_decodes_legacy_json_without_version_epoch() {
+        let mut json = serde_json::to_value(FileStatus {
+            id: 7,
+            path: "/tmp/a".to_string(),
+            name: "a".to_string(),
+            ..Default::default()
+        })
+        .unwrap();
+
+        if let Value::Object(obj) = &mut json {
+            obj.remove("version_epoch");
+        }
+
+        let decoded: FileStatus = serde_json::from_value(json).unwrap();
+        assert_eq!(decoded.version_epoch, 0);
     }
 }
--- a/curvine-server/src/master/meta/inode/inode_file.rs
+++ b/curvine-server/src/master/meta/inode/inode_file.rs
@@ -277,6 +277,7 @@ impl InodeFile {
         // Clear all blocks and reset file size
         self.blocks.clear();
         self.len = 0;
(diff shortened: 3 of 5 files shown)


curvine-server/src/master/meta/inode/inode_file.rs
```diff
@@ -34,6 +34,8 @@ pub struct InodeFile {
     pub(crate) file_type: FileType,
     pub(crate) mtime: i64,
     pub(crate) atime: i64,
+    #[serde(default)]
+    pub(crate) version_epoch: i64,
```
bigbigxu, line 38: Versioning should be maintained at a higher level and should not be recorded in file metadata.

curvine-common/proto/common.proto
```diff
@@ -72,6 +72,7 @@ message FileStatusProto {
     required uint32 mode = 17;
     optional string target = 18;
     required uint32 nlink = 19;
+    required int64 version_epoch = 20 [default = 0];
```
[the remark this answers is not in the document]
jlon: Fixed. `version_epoch` is now `optional` in `common.proto`, and the Rust conversion normalizes a missing field to `0` in `ProtoUtils::file_status_from_pb`. I also kept the round-trip path explicit (`Some(version_epoch)` on encode) so legacy payloads remain decodable while the versioned identity path stays intact.

curvine-common/src/state/file_status.rs
```diff
@@ -21,6 +21,7 @@ use std::collections::HashMap;
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct FileStatus {
     pub id: i64,
+    pub version_epoch: i64,
     pub path: String,
     pub name: String,
```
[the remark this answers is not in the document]
jlon: Fixed. `FileStatus.version_epoch` now has `#[serde(default)]`, and I added a legacy serde test to confirm older snapshots without this field still deserialize with epoch `0`.

curvine-server/src/master/meta/inode/inode_file.rs
jlon: I checked that path before changing this slice, and I do not think `mtime` can replace it. In the current P2P design, `mtime` is used as a freshness / mismatch guard, while `version_epoch` is used as the durable file-generation identity that scopes chunk keys and cached ticket reuse. There is no higher-level version mechanism in the current master metadata path to source that from, so persisting it on `InodeFile` is the minimal durable place that survives restart and keeps `FileStatus` consistent for clients. Moving it higher would require a separate metadata layer, which is outside this PR.

bigbigxu: In the Curvine system, mtime is sufficient and accurate to serve as the basis for determining whether a file has been modified, and acts as a reliable indicator of file modification status. Before reading a file, the client first obtains the file metadata, and can detect file updates through mtime, then decides whether to reuse the cache or pull data again.
Currently, the P2P module is only responsible for data distribution and provides read-only services without participating in file writing or updating. Therefore, there are no version conflicts caused by concurrent writes from multiple nodes.
Designs like version_epoch are mostly used in scenarios with multiple writers to resolve version contention, and are not irreplaceably necessary in the current architecture. Thus, under the mode where P2P only performs read-only data distribution and Curvine acts as the sole write source, there is no need to add an additional version_epoch field to the file metadata.

szbr9486 (commented):
This PR introduces  as the durable file-generation identity for later P2P/cache keys, but the current master overwrite path still appears to leave it unchanged.  clears the old blocks and starts a brand-new write session, yet  only resets length / storage policy / write state and does not bump . That means a full file replacement can still advertise the same epoch as the previous generation, which undermines the "version-safe" identity this field is meant to provide. Could we bump  whenever an overwrite starts, and audit any other whole-file rewrite paths for the same transition?

szbr9486 (commented):
This PR introduces `version_epoch` as the durable file-generation identity for later P2P/cache keys, but the current master overwrite path still appears to leave it unchanged. `FsDir::overwrite_file()` clears the old blocks and starts a brand-new write session, yet `InodeFile::overwrite()` only resets length / storage policy / write state and does not bump `version_epoch`. That means a full file replacement can still advertise the same epoch as the previous generation, which undermines the "version-safe" identity this field is meant to provide. Could we bump `version_epoch` whenever an overwrite starts, and audit any other whole-file rewrite paths for the same transition?

curvine-server/src/master/meta/inode/inode_file.rs
jlon: I do not think the single-writer assumption removes the need for a durable generation key. The risk here is not multi-writer conflict; it is stale reuse across sequential full-file rewrites of the same inode. In this P2P design, `mtime` is used as a freshness / mismatch guard, while `version_epoch` is the explicit generation identity that later scopes chunk keys, provider records, and ticket/cache reuse. Relying on `mtime` alone would push that separation into timestamp equality checks everywhere instead of making the generation boundary explicit. That said, your narrower point about overwrite was valid: if we keep `version_epoch`, a full overwrite must advance it. I fixed that in `4e1e6b5` and added a regression test so a new write session cannot keep advertising the old epoch.

bigbigxu: I have reservations about this design; it is not a good choice. For future functional iterations of Curvine Master, we still need to consider the normal operation of P2P, which violates the principle of separation of responsibilities.

PR #763, opened 2026-03-27
fix(orpc): keep serving on accept errors; tune client defaults; audit…

bigbigxu:
### Summary

- **`orpc` / `RpcServer`**
  - On `listener.accept()` failure, emit a structured **`error!`** including **`bind_addr`**, hinting at **FD/handle limits**, **ephemeral ports**, **process/thread limits**, and **socket backlog** issues, then **`continue`** the accept loop instead of failing the whole server task.

- **`curvine-client` / `UnifiedFileSystem`**
  - **`create()`** now builds create options with **`create_parent(true)`** so parent directories are created when missing (aligned with typical “create file” expectations).

- **`curvine-common` / `ClientConf`**
  - Default **`master_conn_pool_size`**: **3 → 1** (comment updated accordingly).
  - Default clean task interval string: **10s → 60s**.

- **`curvine-server` / master**
  - **`set_attr`**: set RPC audit context using the request path.
  - **Audit logging**: additionally skip high-frequency **`GetMountTable`** RPCs (similar to heartbeat-style noise suppression).

### Motivation

Improve production robustness when transient accept errors occur; reduce noisy audit logs; align client defaults with observed behavior; make unified `create()` more ergonomic for nested paths.

### DIFF
--- a/curvine-client/src/unified/unified_filesystem.rs
+++ b/curvine-client/src/unified/unified_filesystem.rs
@@ -491,7 +491,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
         let flags = OpenFlags::new_write_only()
             .set_create(true)
             .set_overwrite(overwrite);
-        let opts = self.cv.create_opts_builder().build();
+        let opts = self.cv.create_opts_builder().create_parent(true).build();
         self.open_with_opts(path, opts, flags).await
     }
 
--- a/curvine-common/src/conf/client_conf.rs
+++ b/curvine-common/src/conf/client_conf.rs
@@ -119,8 +119,7 @@ pub struct ClientConf {
     pub data_timeout_ms: u64,
     pub pipeline_timeout_ms: u64,
 
-    // Number of fs master connections.
-    // After testing 3 connections, the best performance can be achieved, so the default value is 3.
+    // After testing 3 connections, the best performance can be achieved, so the default value is 1.
     pub master_conn_pool_size: usize,
 
     // Whether to enable pre-reading
@@ -222,7 +221,7 @@ impl ClientConf {
 
     pub const DEFAULT_FILE_SYSTEM_MODE: u32 = 0o777;
 
-    pub const DEFAULT_CLEAN_TASK_INTERVAL_STR: &'static str = "10s";
+    pub const DEFAULT_CLEAN_TASK_INTERVAL_STR: &'static str = "60s";
 
     pub const DEFAULT_CLOSE_TIMEOUT_SECS: u64 = 5;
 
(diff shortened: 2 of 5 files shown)


curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -491,7 +491,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
         let flags = OpenFlags::new_write_only()
             .set_create(true)
             .set_overwrite(overwrite);
-        let opts = self.cv.create_opts_builder().build();
+        let opts = self.cv.create_opts_builder().create_parent(true).build();
         self.open_with_opts(path, opts, flags).await
```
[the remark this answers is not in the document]
bigbigxu: In distributed storage systems, the default `create` method automatically creates the parent directory, maintaining consistency with common UFS semantics.

curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -491,7 +491,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
         let flags = OpenFlags::new_write_only()
             .set_create(true)
             .set_overwrite(overwrite);
-        let opts = self.cv.create_opts_builder().build();
+        let opts = self.cv.create_opts_builder().create_parent(true).build();
```
szbr00, line 494: This makes the CV branch create missing parent directories, but it still does not fix the mounted-UFS path in `open_with_opts()`: that branch drops `opts.create_parent` and calls `mount.ufs.create(...)` directly. I checked the current UFS backends as well: `OpendalFileSystem::create()` just writes/open the object path, and `OssHdfsFileSystem::create()` directly opens a writer, so neither implementation creates parent directories first. In other words, `UnifiedFileSystem::create("/mnt/a/b/file", ...)` still behaves differently depending on whether the path resolves to CV or mounted UFS. Could we either propagate/create the parent directories in the mounted branch too, or keep the old behavior until both backends are aligned?

szbr00, line 494: This makes the CV branch create missing parent directories, but it still does not fix the mounted-UFS path in `open_with_opts()`: that branch drops `opts.create_parent` and calls `mount.ufs.create(...)` directly. I checked the current UFS backends as well: `OpendalFileSystem::create()` just writes/open the object path, and `OssHdfsFileSystem::create()` directly opens a writer, so neither implementation creates parent directories first. In other words, `UnifiedFileSystem::create("/mnt/a/b/file", ...)` still behaves differently depending on whether the path resolves to CV or mounted UFS. Could we either propagate/create the parent directories in the mounted branch too, or keep the old behavior until both backends are aligned?

bigbigxu: In distributed storage systems, the default create method automatically creates the parent directory, maintaining consistency with common UFS semantics.

szbr9486 (approved):

PR #764, opened 2026-03-27
fix: keep inode_file_num symmetric for symlink lifecycle

lzjqsdd:
Add symlink lifecycle regression coverage so inode_file_num stays non-negative for create/delete and rename-overwrite paths.

### DIFF
--- a/curvine-server/src/master/meta/store/inode_store.rs
+++ b/curvine-server/src/master/meta/store/inode_store.rs
@@ -222,6 +222,8 @@ impl InodeStore {
 
         batch.commit()?;
 
+        self.fs_stats.increment_file_count();
+
         Ok(())
     }
 
--- a/curvine-server/src/master/meta/fs_stats.rs
+++ b/curvine-server/src/master/meta/fs_stats.rs
@@ -29,7 +29,7 @@ impl FileSystemStats {
     }
 
     pub fn increment_file_count(&self) {
-        self.metrics.inode_file_num.inc()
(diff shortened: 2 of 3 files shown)


curvine-server/src/master/meta/store/inode_store.rs
```diff
@@ -222,6 +222,8 @@ impl InodeStore {
 
         batch.commit()?;
 
+        self.fs_stats.increment_file_count();
```
szbr00, line 225: `apply_symlink()` now increments `inode_file_num` unconditionally, but `FsDir::unprotected_symlink()` can also reach this path when `force=true` and the destination already exists as a symlink. In that branch we replace the existing inode in place instead of adding a new directory entry, so bumping the counter here will make repeated forced rewrites drift the file metric upward. Could we increment only for the brand-new symlink case (or make the caller pass whether this was a replacement) and add a regression test for `symlink(..., force=true)`?

lzjqsdd: Thanks for catching this. `apply_symlink` now takes `increment_file_count` and only bumps `inode_file_num` when `unprotected_symlink` adds a new dentry (`old_inode.is_none()`), not when `force=true` replaces an existing symlink in place. Added `test_inode_file_num_stable_on_forced_symlink_rewrite` and serialized the inode counter assertions with a small test mutex because the metrics gauges are process-wide. Please take another look.

szbr9486 (approved):

PR #767, opened 2026-03-28
feat(meta): refactor InodeView with named wrappers; expand RocksDB me…

bigbigxu:
## Summary
Refactors the master metadata `InodeView` representation to use explicit `NamedFile`, `NamedDir`, and `NamedEntry` types (boxed inside the enum), with `Deref`/`DerefMut` to the underlying `InodeFile` / `InodeDir`. Call sites across the filesystem layer, journal replay, inode path resolution, TTL, quota, and the inode store are updated to match the new shape.
Also replaces coarse RocksDB “memory components” reporting with a richer **`HashMap<String, u64>`** of DB- and column-family–scoped integer properties (`get_rocksdb_metrics`), documents semantics (especially default-CF vs multi-CF `db_*` keys), and wires master Prometheus metrics to the new gauge family **`rocksdb_metrics`**.
Adds **`find_inode(parent, name)`** on the Rocks-backed store (edge lookup + inode load) and exposes **`InodeStore::store()`** for direct access where needed.
## Motivation
- Clearer ownership and naming for inode tree nodes; smaller, consistent construction via `InodeView::new_*`.
- Better observability: export many RocksDB internal counters (memtables, block cache, compactions, snapshots, etc.) instead of a handful of synthetic memory tags.
### `curvine-common`
- **`DBEngine`**: `get_rocksdb_memory()` → **`get_rocksdb_metrics()`** returning `HashMap<String, u64>` with documented key naming (`db_*`, `cf_{cf}_*`) and guidance for multi-CF totals.
### `curvine-server`
- **`InodeView`**: new wrapper structs, constructors, pattern-match updates throughout.
- **`FsDir`**, **`InodePath`**, **`InodeDir`**, **`inodes_children`**, **`journal_loader`**, **`ttl_*`**, **`quota_manager`**: adapt to new enum and helpers.
- **`RocksInodeStore` / `InodeStore`**: `get_rocksdb_metrics`, **`find_inode`**, **`InodeStore::store()`**.
- **`MasterMetrics`**: `rocksdb_used_memory_bytes` → **`rocksdb_metrics`** (`GaugeVec` keyed by metric name).
## Breaking changes
1. **Serialization**: `InodeView`’s serde/binary layout changes. **Existing persisted inode data may not deserialize** without a migration or compatibility layer.
2. **Metrics**: Prometheus metric name and labels change (`rocksdb_used_memory_bytes` → `rocksdb_metrics`; label values are full metric keys from the map).

### DIFF
--- a/curvine-server/src/master/master_metrics.rs
+++ b/curvine-server/src/master/master_metrics.rs
@@ -46,7 +46,7 @@ pub struct MasterMetrics {
     pub(crate) journal_ufs_applied: Gauge,
 
     pub(crate) used_memory_bytes: Gauge,
-    pub(crate) rocksdb_used_memory_bytes: GaugeVec,
+    pub(crate) rocksdb_metrics: GaugeVec,
 
     pub(crate) inode_dir_num: Gauge,
     pub(crate) inode_file_num: Gauge,
@@ -111,11 +111,7 @@ impl MasterMetrics {
             )?,
 
             used_memory_bytes: m::new_gauge("used_memory_bytes", "Total memory used")?,
-            rocksdb_used_memory_bytes: m::new_gauge_vec(
-                "rocksdb_used_memory_bytes",
-                "RocksDB memory usage in bytes",
-                &["tag"],
-            )?,
+            rocksdb_metrics: m::new_gauge_vec("rocksdb_metrics", "RocksDB metrics", &["tag"])?,
 
             inode_dir_num: m::new_gauge("inode_dir_num", "Total dir")?,
 
@@ -179,13 +175,13 @@ impl MasterMetrics {
         }
 
         let fs_dir = fs.fs_dir.read();
-        if let Ok(memory_info) = fs_dir.get_rocks_store().get_rocksdb_memory() {
-            for (component, size) in memory_info {
-                self.rocksdb_used_memory_bytes
-                    .with_label_values(&[&component])
-                    .set(size as i64);
-            }
+        let rocksdb_metrics = fs_dir.get_rocks_store().get_rocksdb_metrics()?;
+        for (key, value) in rocksdb_metrics {
+            self.rocksdb_metrics
+                .with_label_values(&[&key])
+                .set(value as i64);
         }
+
         let ttl_list = fs_dir.get_ttl_bucket_list();
         drop(fs_dir);
         self.ttl_bucket_len.set(ttl_list.buckets_len() as i64);
--- a/curvine-server/src/master/meta/inode/inode_view.rs
+++ b/curvine-server/src/master/meta/inode/inode_view.rs
@@ -27,22 +27,103 @@ use orpc::{err_box, CommonResult};
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, LinkedList};
 use std::fmt::{Debug, Formatter};
+use std::ops::{Deref, DerefMut};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NamedFile {
+    pub name: String,
+    pub file: InodeFile,
+}
+
+impl NamedFile {
+    pub fn new(name: String, file: InodeFile) -> Self {
+        NamedFile { name, file }
+    }
+}
+
+impl Deref for NamedFile {
+    type Target = InodeFile;
+
+    fn deref(&self) -> &Self::Target {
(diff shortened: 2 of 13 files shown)


curvine-server/src/master/meta/inode/inode_view.rs
```diff
@@ -27,49 +27,130 @@ use orpc::{err_box, CommonResult};
...
 #[derive(Serialize, Deserialize)]
 #[repr(i8)]
 pub enum InodeView {
-    File(String, InodeFile) = 1,
-    Dir(String, InodeDir) = 2,
-    FileEntry(String, i64) = 3,
+    File(Box<NamedFile>) = 1,
+    Dir(Box<NamedDir>) = 2,
+    FileEntry(Box<NamedEntry>) = 3,
 }
```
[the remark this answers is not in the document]
bigbigxu: The InodeDir and InodeFile structures have not changed, so this problem does not exist.

curvine-server/src/master/meta/inode/inode_view.rs
```diff
@@ -27,49 +27,130 @@ use orpc::{err_box, CommonResult};
...
-            File(_, _) => false,
-            Dir(_, _) => true,
-            _ => false,
-        }
+        matches!(self, Dir(_))
     }
 
     pub fn is_file(&self) -> bool {
         !self.is_dir()
     }
```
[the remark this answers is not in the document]
bigbigxu: FileEntry is a file; this logic is correct.

curvine-server/src/master/meta/store/rocks_inode_store.rs
```diff
@@ -107,6 +108,18 @@ impl RocksInodeStore {
...
+    pub fn find_inode(&self, parent: i64, name: &str) -> CommonResult<Option<InodeView>> {
+        let key = RocksUtils::i64_str_to_bytes(parent, name);
+        let bytes = self.db.get_cf(Self::CF_EDGES, key)?;
+
+        let inode_id = match bytes {
+            Some(v) => RocksUtils::i64_from_bytes(&v)?,
+            None => return Ok(None),
+        };
+
+        self.get_inode(inode_id)
```
szbr00: `find_inode(parent, name)` now returns the raw inode from `CF_INODES`, but unlike `InodeStore::get_inode(id, Some(name))` it never rewrites the inode name from the edge we just resolved. That matters for hardlinks / `FileEntry` cases, where the stored inode can carry a different canonical name from the directory entry under `parent`. Any future caller that uses `inode.name()` or derives `FileStatus` from this helper will observe the wrong basename. Could this helper preserve the edge name before returning, the same way `get_inode(..., Some(name))` already does?

curvine-server/src/master/master_metrics.rs
```diff
@@ -179,13 +175,13 @@ impl MasterMetrics {
         }
 
         let fs_dir = fs.fs_dir.read();
-        if let Ok(memory_info) = fs_dir.get_rocks_store().get_rocksdb_memory() {
-            for (component, size) in memory_info {
-                self.rocksdb_used_memory_bytes
-                    .with_label_values(&[&component])
-                    .set(size as i64);
-            }
+        let rocksdb_metrics = fs_dir.get_rocks_store().get_rocksdb_metrics()?;
```
szbr00, line 178: This now propagates any `get_rocksdb_metrics()` error out of `text_output()`, whereas the previous code treated RocksDB memory stats as best-effort and still exported the rest of the master metrics. Because `router_handler::metrics()` unwraps `text_output()`, a transient or unsupported RocksDB property read can now make the whole `/metrics` endpoint fail instead of just skipping these optional gauges. Could we keep this path non-fatal here (for example log-and-continue, or ignore only the failing RocksDB metrics) so observability does not go dark when one property query breaks?

curvine-server/src/master/meta/store/rocks_inode_store.rs
bigbigxu: Deleted

curvine-server/src/master/master_metrics.rs
bigbigxu: If RocksDB encounters an anomaly, the node will painc, so this check is unnecessary.

szbr9486 (approved):

Issue #710, opened 2026-03-13
[FEATURE]support spdk


umuzhaohui:
support spdk, so can use nvme over pcie and nvme over fabric(tcp、rdma) on ssd

PR #781, opened 2026-04-19
feat: introduce SPDK NVMe over Fabric (TCP/RDMA) integration

thuongle2210:
# Description
This PR introduces SPDK integration for high-performance NVMe-oF storage, supporting both TCP and RDMA transports.

Resolves part of [pr-710](https://github.com/CurvineIO/curvine/issues/710)

# Key Changes
- Core SPDK bindings: SpdkBdev, SpdkPoller, SpdkEnv — FFI layer to interact with SPDK C library.
- I/O path integration: Updated read/write handlers in orpc/io/ & curvine-server to use SPDK for block I/O.
- VFS layer: Adapted storage backends to support SPDK-backed devices.
- Testing: Unit and integration tests demonstrating correctness.

# Impact
- Eliminates syscall & context switch overhead for heavy I/O workloads.
- Enables high-throughput IOPS scenarios (because SPDK allows millions of IOPS for native read/write)

# Future Work
- Containerization: Integrate SPDK setup into containers.
- Memory Efficiency: enhance BdevOffsetAllocator with free-list for freed offset reclamation
- Performance: Optimize spdk-poller like implementing multi-poller optimization per controller for better I/O scaling.
- Configurability: Add DEFAULT_MAX_PER_CTRLR,.. to allow tuning of controller limits.
- Other fabric support
- Benchmark the performance

# run book:
[runbook_spdk_rdma.md](https://github.com/user-attachments/files/26950326/runbook_spdk_rdma.md)
[runbook_spdk_tcp.md](https://github.com/user-attachments/files/26950327/runbook_spdk_tcp.md)


thuongle2210:
Dear @szbr9486 , could you please help me review this PR?

szbr9486:
> Dear @szbr9486 , could you please help me review this PR?

Thank you very much for your PR. We will review it as soon as possible. However, this PR is quite large. For next time, we suggest breaking it down into smaller PRs and submitting them incrementally. Thanks again.

curvine-server/src/worker/storage/vfs_dir.rs
```diff
@@ -56,22 +58,81 @@ impl VfsDir {
...
+                )
+            })?;
+            let names = env.bdev_names();
+            if names.is_empty() {
+                return orpc::err_box!(
+                    "StorageType::Spdk dir '{}' but no bdevs discovered from SPDK targets",
+                    conf.path
+                );
+            }
+            let idx = version.dir_id as usize % names.len();
```
szbr00, line 92: Mapping SPDK data directories to bdevs via `dir_id % names.len()` is unsafe when `worker.data_dir` has more SPDK entries than the SPDK layer discovered bdevs. In that case two or more `VfsDir` instances get assigned the same bdev but each constructs its own `BdevOffsetAllocator` starting at cursor 0, so they will allocate overlapping byte ranges on the same NVMe namespace and silently corrupt each other's blocks. Please either (a) reject the config when `bdev_names.len() < spdk_data_dirs.len()`, or (b) share a single `BdevOffsetAllocator` per bdev across all dirs that map to it.

curvine-server/src/worker/block/block_meta.rs
```diff
@@ -217,13 +241,58 @@ impl BlockMeta {
...
-    pub fn create_writer(&self, off: i64, overwrite: bool) -> IOResult<LocalFile> {
-        let file = self.get_block_file()?;
-        LocalFile::with_write_offset(file, overwrite, off)
+    pub fn create_writer(&self, off: i64, overwrite: bool) -> IOResult<BlockDevice> {
+        match self.storage_type() {
+            #[cfg(feature = "spdk")]
+            StorageType::Spdk => {
+                let bdev_name = self.get_bdev_name()?;
+                let abs_offset = self.bdev_offset + off;
+                let max_len = 0.max(self.len - off);
```
szbr00, line 250: `max_len = 0.max(self.len - off)` can legitimately be 0 (for example when `off == self.len`, which happens on append/commit paths). It is then passed to `SpdkBdev::open`, where the branch `if max_len > 0 { end.min(bdev_size) } else { bdev_size }` silently falls back to the full bdev size. The resulting `SpdkBdev` is no longer bounded to this block's `[bdev_offset, bdev_offset + self.len)` range, so subsequent writes/reads can run off the end of the block and trample neighbouring blocks on the shared bdev. Please either reject `max_len == 0` at the caller, or treat `max_len == 0` in `SpdkBdev::open` as a zero-length (empty) window rather than "whole device".

orpc/src/io/spdk_bdev.rs
```diff
@@ -0,0 +1,749 @@
...
+                    self.name, count
+                );
+                logged = true;
+            }
+            std::thread::sleep(std::time::Duration::from_millis(1));
+        }
+
+        // Return qpair to pool and release handle.
+        if let Some(env) = crate::io::spdk_env::SpdkEnv::global_including_shutdown() {
+            env.release_qpair(self.ctrlr, self.io_channel.qpair);
```
szbr00, line 663: On the drop-timeout branch above, pending I/O was not drained (we only poisoned the DMA buffer pointers to avoid UAF there). We then still call `env.release_qpair(...)`, which may push this qpair back into the `QpairPool` and hand it to a different `SpdkBdev` that will submit new I/O on a qpair still referenced by a heap-allocated `CallbackCtx` from the previous owner. When the late completion fires, `Box::from_raw` will alias or free memory that is now associated with a different handle. At minimum, in the timeout branch please skip `release_qpair` (leak the qpair, matching the "leak DMA buffers" policy) and `release_handle()` only after the outstanding I/Os are truly drained, or mark the qpair as tainted so the pool frees it rather than reusing it.

curvine-server/src/worker/storage/dir_state.rs
```diff
@@ -13,11 +13,218 @@
...
+    /// Alignment for all allocations (typically bdev block size, e.g. 512 or 4096).
+    align: i64,
+    /// Total bdev capacity in bytes.
+    capacity: i64,
+    /// Mapping between block_id and (offset, size) .
+    map: HashMap<i64, (i64, i64)>,
+}
+impl BdevOffsetAllocator {
+    /// Create a new allocator for a bdev with the given capacity and alignment.
+    pub fn new(capacity: i64, align: i64) -> Self {
```
szbr00, line 71: `BdevOffsetAllocator::new` accepts any positive `align`, but `align_up_i64` uses `(n + align - 1) & !(align - 1)` which silently produces wrong results unless `align` is a power of two. NVMe block sizes are normally 512/4096, but this value flows in from config (`SpdkConf.block_align`, bdev metadata) and a misconfiguration (e.g. 1000) would corrupt offset layout without any error. Please validate `align.is_power_of_two()` here and return an error (or fall back to `DEFAULT_BLOCK_ALIGN`) instead of trusting the input.

orpc/src/io/spdk_poller.rs
```diff
@@ -0,0 +1,298 @@
...
+        let mut inflight: std::collections::HashMap<usize, Arc<std::sync::atomic::AtomicUsize>> =
+            std::collections::HashMap::new();
+
+        loop {
+            if shutdown.load(Ordering::Acquire) && rx.is_empty() {
+                break;
+            }
+
+            // Drain pending requests (non-blocking after first)
+            match rx.recv_timeout(std::time::Duration::from_micros(100)) {
```
szbr00, line 169: The poller blocks up to 100us on `recv_timeout` before it gets to `curvine_spdk_qpair_poll`. For local NVMe-oF over RDMA that is often larger than the round-trip itself, so every I/O can eat an extra ~50us of avoidable latency, and under high submission rate the completion polling frequency is upper-bounded by this sleep. Consider using `try_recv` + a short `std::thread::yield_now` / busy-poll budget, or switching to a hybrid (poll for N iterations, then park) so completions aren't gated on the receive timeout. Worth at least a TODO pointing at this, since it directly affects the "millions of IOPS" target called out in the PR description.

orpc/src/io/spdk_bdev.rs
szbr00, line 663: On the drop-timeout branch above, pending I/O was not drained (we only poisoned the DMA buffer pointers to avoid UAF there). We then still call `env.release_qpair(...)`, which may push this qpair back into the `QpairPool` and hand it to a different `SpdkBdev` that will submit new I/O on a qpair still referenced by a heap-allocated `CallbackCtx` from the previous owner. When the late completion fires, `Box::from_raw` will alias or free memory that is now associated with a different handle. At minimum, in the timeout branch please skip `release_qpair` (leak the qpair, matching the "leak DMA buffers" policy) and `release_handle()` only after the outstanding I/Os are truly drained, or mark the qpair as tainted so the pool frees it rather than reusing it.

curvine-common/proto/common.proto
```diff
@@ -12,6 +12,7 @@ enum StorageTypeProto {
     STORAGE_TYPE_PROTO_HDD = 2;
     STORAGE_TYPE_PROTO_UFS = 3;
     STORAGE_TYPE_PROTO_DISK = 4;
+    STORAGE_TYPE_PROTO_SPDK = 5;
```
bigbigxu: SPDK is not a specific type of storage, but a high-performance storage development toolkit. When the client sends an open request and delivers the required environment parameters, the worker node determines the data transmission mode and decides whether to transmit and interact data through SPDK or RDMA.

curvine-server/src/master/journal/journal_loader.rs
```diff
@@ -564,7 +563,7 @@ impl JournalLoader {
     pub fn rename(&self, entry: RenameEntry) -> CommonResult<()> {
         let mut fs_dir = self.fs_dir.write();
         let entry_src = entry.src;
-        let src_inp = InodePath::resolve(fs_dir.root_ptr(), &entry_src, &fs_dir.store)?;
+        let src_inp = InodePath::resolve(fs_dir.root_ptr(), entry_src.clone(), &fs_dir.store)?;
```
bigbigxu: journal_loader.rs is outdated. Please merge the main branch and resubmit the changes.

curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -50,25 +50,33 @@ impl WriteHandler {
...
             );
         }
+        // SPDK: NVMe namespace has fixed capacity, just validate fits.
+        if !file.supports_short_circuit() {
+            if opts.len > file.len() {
+                return err_box!(
+                    "Resize {} exceeds SPDK bdev capacity {}",
+                    opts.len,
+                    file.len()
+                );
```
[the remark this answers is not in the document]
thuongle2210: I removed the redundant validation.

orpc/src/io/spdk_poller.rs
```diff
@@ -0,0 +1,298 @@
...
+        let inflight_counter = inflight
+            .entry(key)
+            .or_insert_with(|| Arc::new(std::sync::atomic::AtomicUsize::new(0)))
+            .clone();
+
+        // Box::into_raw ensures CallbackCtx survives until poller_callback reclaims it
+        let cb_ctx = Box::new(CallbackCtx {
+            completion: req.completion.clone(),
+            async_ctx: unsafe { std::mem::zeroed() },
+            inflight: inflight_counter.clone(),
```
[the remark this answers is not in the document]
thuongle2210: Currently, I set an oversized opaque buffer. To avoid the issue in the future,  I added the assertion for this.

thuongle2210:
Hi @szbr9486 and @bigbigxu, I have scanned all your feedback. I resolved part of it and marked the rest as TODO for later enhancements if needed. Thanks.

bigbigxu:
> Hi @szbr9486 and @bigbigxu, I have scanned all your feedback. I resolved part of it and marked the rest as TODO for later enhancements if needed. Thanks.

SPDK represents a protocol type instead of a storage type. This change breaks the layered caching semantics and results in a breaking modification.

curvine-common/src/state/storage_info.rs
```diff
@@ -54,6 +56,7 @@ impl StorageType {
             StorageType::Hdd => "HDD",
             StorageType::Ufs => "UFS",
             StorageType::Disk => "DISK",
+            StorageType::Spdk => "SPDK",
```
szbr00: SPDK is a protocol, I recommend using SPDK_DISK instead of SPDK to make it clear that this mode refers to hardware that supports SPDK acceleration and uses the SPDK protocol for acceleration.

szbr9486:
Hi @szbr9486 and @bigbigxu, I have scanned all your feedback. I resolved part of it and marked the rest as TODO for later enhancements if needed. Thanks.

Thank you, I have a small comment.

thuongle2210:
> Hi @szbr9486 and @bigbigxu, I have scanned all your feedback. I resolved part of it and marked the rest as TODO for later enhancements if needed. Thanks.
> 
> Thank you, I have a small comment.

I see it; I renamed the StorageType for SPDK.

szbr9486 (approved):

szbr9486:
> > Hi @szbr9486 and @bigbigxu, I have scanned all your feedback. I resolved part of it and marked the rest as TODO for later enhancements if needed. Thanks.您好，我已经阅读了您们的所有反馈。我解决了一部分问题，其余问题标记为待办事项，以便日后根据需要进行改进。谢谢。
> > Thank you, I have a small comment.谢谢，我还有一点意见。
> 
> I see it; I renamed the StorageType for SPDK.我明白了；我重命名了 SPDK 的 StorageType。

Confirmed and merged, great pr , thanks!

Issue #784, opened 2026-04-20
[FEATURE] Make libstdc++.so.6 statically linked


yangyang233333:
**Is your feature request related to a problem? Please describe.**
CurvineIO depends on rocksdb, which is dynamically linked to stdc++. What's a good way to change it to static linking?

**Additional context**
We need to deploy on bare metal, and it's best to minimize the number of dynamic libraries.

bigbigxu:
Try setting the following before compilation:
RUSTFLAGS="-C link-arg=-static-libstdc++ -C link-arg=-static-libgcc"

PR #786, opened 2026-04-25
build: add linker wrapper to statically link libstdc++, libgcc_s, and libz on Linux

yangyang233333:
## Summary
- Add a linker wrapper script (`build/static-link-wrapper.sh`) that forces `libstdc++`, `libgcc_s`, and `libz` to link statically on Linux x86_64
- Configure Cargo to use the wrapper via `.cargo/config.toml`

## Motivation

`librocksdb-sys` emits `cargo:rustc-link-lib=stdc++` in its build script, which causes the linker to place `-lstdc++` in the `-Wl,-Bdynamic` section. This happens before `RUSTFLAGS` can intercept it, making it impossible to force static linking through normal Cargo configuration.

This is problematic for deployment on systems where `libstdc++.so`, `libgcc_s.so`, or `libz.so` may not be available or may be a different version, leading to runtime failures.

## Approach

A thin shell wrapper (`build/static-link-wrapper.sh`) intercepts linker arguments and replaces:
- `-lstdc++` → `-Wl,-Bstatic -lstdc++ -Wl,-Bdynamic`
- `-lgcc_s` → `-Wl,-Bstatic -lgcc_eh -lgcc -Wl,-Bdynamic`
- `-lz` → `-Wl,-Bstatic -lz -Wl,-Bdynamic`

This ensures these libraries are statically linked while keeping all other libraries dynamically linked as usual.

## Test plan
- [x] Build on x86_64 Linux and verify with `ldd` that the output binary has no dynamic dependency on `libstdc++`, `libgcc_s`, or `libz`
- [x] Build on macOS to confirm the wrapper is not used and existing behavior is preserved

### DIFF
--- a/build/static-link-wrapper.sh
+++ b/build/static-link-wrapper.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# Linker wrapper that forces -lstdc++, -lgcc_s, and -lz to link statically.
+# Needed because crate build scripts (e.g. librocksdb-sys) emit
+# `cargo:rustc-link-lib=stdc++` which lands in the -Bdynamic section.
+#
+# NOTE: each -Wl flag must be a separate arg; do NOT merge into one
+# comma-separated string or gcc will pass a literal "-Wl," to the linker.
+
+set -euo pipefail
+
+NEWARGS=()
+for arg in "$@"; do
+    if [ "$arg" = "-lstdc++" ]; then
+        NEWARGS+=("-Wl,-Bstatic" "-lstdc++" "-Wl,-Bdynamic")
+    elif [ "$arg" = "-lgcc_s" ]; then
+        NEWARGS+=("-Wl,-Bstatic" "-lgcc_eh" "-lgcc" "-Wl,-Bdynamic")
+    elif [ "$arg" = "-lz" ]; then
+        NEWARGS+=("-Wl,-Bstatic" "-lz" "-Wl,-Bdynamic")
+    else
+        NEWARGS+=("$arg")
+    fi
+done
+
+exec "${CC:-cc}" "${NEWARGS[@]}"
--- a/build/static-link-wrapper.sh
+++ b/build/static-link-wrapper.sh
@@ -3,8 +3,9 @@
 # Needed because crate build scripts (e.g. librocksdb-sys) emit
 # `cargo:rustc-link-lib=stdc++` which lands in the -Bdynamic section.
 #
-# NOTE: each -Wl flag must be a separate arg; do NOT merge into one
-# comma-separated string or gcc will pass a literal "-Wl," to the linker.
+# NOTE: gcc/clang support comma-separated `-Wl,...` lists. This wrapper
+# emits `-Wl,-Bstatic` / `-Wl,-Bdynamic` as separate args around the
+# affected libraries to preserve the intended linker mode transitions.
 
 set -euo pipefail
 
--- a/build/static-link-wrapper.sh
+++ b/build/static-link-wrapper.sh
@@ -14,6 +14,9 @@ for arg in "$@"; do
     if [ "$arg" = "-lstdc++" ]; then
         NEWARGS+=("-Wl,-Bstatic" "-lstdc++" "-Wl,-Bdynamic")
     elif [ "$arg" = "-lgcc_s" ]; then
+        # libgcc_s.so provides both the GCC runtime helpers and the EH
+        # (exception-handling) unwinder. GCC does not ship a libgcc_s.a;
+        # the static equivalents are libgcc.a (runtime) + libgcc_eh.a (EH).
         NEWARGS+=("-Wl,-Bstatic" "-lgcc_eh" "-lgcc" "-Wl,-Bdynamic")
     elif [ "$arg" = "-lz" ]; then
         NEWARGS+=("-Wl,-Bstatic" "-lz" "-Wl,-Bdynamic")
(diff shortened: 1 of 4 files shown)


yangyang233333:
Closes #784

---

### Why `RUSTFLAGS="-C link-arg=-static-libstdc++ -C link-arg=-static-libgcc"` does not work

`rustc` hard-codes the insertion of `-lgcc_s` (dynamic) **after** the Rust standard library objects in the final linker invocation. User-supplied `-C link-arg` flags appear earlier on the command line, but by the time the linker processes the later `-lgcc_s`, it has already moved past the static-linking context — so the dynamic version wins regardless of what RUSTFLAGS you pass.

Similarly, `librocksdb-sys` emits `cargo:rustc-link-lib=stdc++` from its build script, which Cargo places in the `-Wl,-Bdynamic` section. RUSTFLAGS cannot override the link mode of libraries injected by build scripts at this stage.

The only clean solution is a **custom linker wrapper** that intercepts the final linker arguments and rewrites `-lstdc++`, `-lgcc_s`, and `-lz` to their static equivalents (`-Wl,-Bstatic -l... -Wl,-Bdynamic`). This is exactly what this PR implements via `build/static-link-wrapper.sh`.

yangyang233333:
@bigbigxu @lzjqsdd

yangyang233333:
Thanks for the review! The first two comment suggestions (inaccurate comments in `config.toml` and `static-link-wrapper.sh`) have been addressed in the latest commit (1750c79).

For the third suggestion (`${CC:-cc}` → `${STATIC_LINK_WRAPPER_LINKER:-cc}`): I'll keep `${CC:-cc}` for now. Cargo sets `CC` to the correct compiler for the target when invoking the linker, and introducing a project-specific env var adds friction without a clear benefit in our current setup. Happy to revisit if cross-compilation becomes an issue. FYI @lzjqsdd @copilot

szbr9486 (commented):
The official Rocky9 builder images (`curvine-docker/compile/Dockerfile_rocky9` and `Dockerfile_rocky9_cached`) only install `gcc gcc-c++ … zlib-devel`. On RHEL 9 / Rocky 9, the static archives `libstdc++.a` and `libz.a` are shipped in **separate** packages (`libstdc++-static` and `zlib-static`) — they are not pulled in by `gcc-c++` or `zlib-devel`.

Since `.github/workflows/release.yml` produces the `curvine-rocky9-x86_64` release artifact from this image, this PR will break the release pipeline with `cannot find -lstdc++` / `cannot find -lz` unless the Rocky9 Dockerfiles are updated in the same change to add `libstdc++-static` and `zlib-static`. The Ubuntu builders should be fine because `g++` and `zlib1g-dev` already ship the matching `.a` archives.

Recommendation: include the Dockerfile updates in this PR (or land them first) so the wrapper has the inputs it needs on every supported builder. `-lgcc_s`'s replacements (`libgcc.a` / `libgcc_eh.a`) are typically already present from `gcc`, so only `libstdc++-static` and `zlib-static` need to be added.

build/static-link-wrapper.sh
```diff
@@ -0,0 +1,25 @@
...
+# affected libraries to preserve the intended linker mode transitions.
+
+set -euo pipefail
+
+NEWARGS=()
+for arg in "$@"; do
+    if [ "$arg" = "-lstdc++" ]; then
+        NEWARGS+=("-Wl,-Bstatic" "-lstdc++" "-Wl,-Bdynamic")
+    elif [ "$arg" = "-lgcc_s" ]; then
+        NEWARGS+=("-Wl,-Bstatic" "-lgcc_eh" "-lgcc" "-Wl,-Bdynamic")
```
szbr00, line 20: Worth a one-line note here: unlike the `-lstdc++` / `-lz` cases that just toggle linkage mode for the same library, `-lgcc_s` is replaced with two **different** static archives (`-lgcc_eh` + `-lgcc`). There is no static counterpart of `libgcc_s.so` named `libgcc_s.a` — the EH shim and the runtime helpers have to be linked separately. Without this comment, the asymmetry will likely be misread as a typo.

yangyang233333: Good catch — added a comment in the latest commit (7c1d9a1) explaining that `libgcc_s.so` has no static `libgcc_s.a` counterpart, so we link `libgcc.a` (runtime helpers) + `libgcc_eh.a` (EH/unwinder) instead. Thanks!

yangyang233333:
@szbr9486 Good catch on the missing static packages — the latest commit (7c1d9a1) adds `libstdc++-static` and `zlib-static` to both `Dockerfile_rocky9` and `Dockerfile_rocky9_cached`, so the release pipeline should be unblocked. Thanks for the thorough review!

szbr9486:
> @szbr9486 Good catch on the missing static packages — the latest commit ([7c1d9a1](https://github.com/CurvineIO/curvine/commit/7c1d9a15cca35d3d50421ce1890bee3fb53251a6)) adds `libstdc++-static` and `zlib-static` to both `Dockerfile_rocky9` and `Dockerfile_rocky9_cached`, so the release pipeline should be unblocked. Thanks for the thorough review!您发现静态包缺失的问题真是太好了——最新的提交（）已将 `libstdc++-static` 和 `zlib-static` 添加到 `Dockerfile_rocky9` 和 `Dockerfile_rocky9_cached` 中，因此发布流程应该可以正常运行了。感谢您细致的审查！

Thanks for your contribution, you are  welcome becoming newest contributor for Curvine.
BTW, my weiChat no is : 398713015, I'm really looking forward to being your friend.

lzjqsdd (approved):

PR #799, opened 2026-05-09
feat: add curvine lancedb facade scaffold

jlon:
## What changed
- added a new `curvine-lancedb-rs` workspace crate as the initial Rust facade for LanceDB on Curvine
- kept the exposed crate name as `lancedb` while publishing under `curvine-lancedb-rs`
- added a minimal `connect` wrapper that passes through non-`curvine://` URIs to upstream LanceDB
- added an explicit unsupported boundary for `curvine://` URIs until the Curvine object store integration lands
- introduced a small facade-local error module to centralize unsupported-path construction
- added compatibility tests covering local connect, namespace connect, error-module re-export, and explicit unsupported behavior

## Why
This is the first foundation PR for the LanceDB on Curvine module. It establishes the compatibility shell and test boundary before introducing the Curvine object store provider and session wiring.

## Impact
- Rust users can depend on `curvine-lancedb-rs` while still importing `lancedb`
- regular upstream LanceDB usage continues to pass through unchanged
- `curvine://` connections fail explicitly instead of silently pretending to work

## Validation
- `cargo fmt --all --check`
- `cargo test -p curvine-lancedb-rs --test facade_compat -- --nocapture`


curvine-lancedb-rs/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "curvine-lancedb-rs"
+version = "0.2.0"
+edition = "2021"
+license = "Apache-2.0"
+description = "Facade crate for LanceDB on Curvine"
+
+[workspace]
```
szbr00, line 8: The new crate has its own `[workspace]` here, but it's never added to the root `Cargo.toml`'s `members` (and isn't put in `exclude` either). The effect is that the crate is an orphan to the parent Curvine workspace:

- `cargo build`, `cargo test`, `cargo clippy --workspace` from the repo root all skip this crate, so the existing workspace-level CI won't compile it, lint it, or run its tests.
- `cargo test -p curvine-lancedb-rs` from the repo root fails with "package … not in workspace"; it only works when `cd curvine-lancedb-rs/` first. The PR description's listed `cargo clippy -p curvine-lancedb-rs … -- -D warnings` therefore depends on developer convention rather than CI, which is exactly the kind of regression the workspace structure is supposed to prevent.
- It also has to duplicate `version`, `edition`, `license`, and any pinned dependency versions instead of inheriting them from the root's `[workspace.package]` / `[workspace.dependencies]`.

If the intent is to keep it isolated because of `lancedb`'s transitive deps or the `lib.name = "lancedb"` collision, please at least:
- Add `curvine-lancedb-rs` to the root workspace's `exclude = [...]` so the separation is explicit and Cargo no longer warns about it on `cargo metadata`, and
- Wire it into CI explicitly (e.g. a separate job that runs the validation commands inside `curvine-lancedb-rs/`).

Otherwise, drop the sub-`[workspace]` line and add the crate to the root `members` so the standard workspace tooling covers it.

curvine-lancedb-rs/src/connection.rs
```diff
@@ -0,0 +1,142 @@
...
+
+pub fn connect(uri: &str) -> ConnectBuilder {
+    ConnectBuilder::new(uri)
+}
+
+pub fn connect_namespace(
+    ns_impl: &str,
+    properties: HashMap<String, String>,
+) -> ConnectNamespaceBuilder {
+    lancedb_upstream::connect_namespace(ns_impl, properties)
```
szbr00: `connect_namespace` returns the upstream `ConnectNamespaceBuilder` directly, so the explicit `curvine://` boundary established for `connect()` does not extend here. A caller doing

```rust
let mut props = HashMap::new();
props.insert("root".to_string(), "curvine:///data/lancedb/demo".to_string());
lancedb::connect_namespace("dir", props).execute().await
```

will fail inside upstream with whatever generic error the chosen namespace impl emits for an unknown URI scheme — not the intended `Curvine object store, uri=… is not implemented`. That defeats the test boundary expressed in `curvine_uri_reports_explicit_unsupported_boundary` and will look like an unrelated bug to users when the real Curvine integration lands.

Suggest applying the same gating: scan `properties` for known location keys (`root`, `uri`, …) — or any value starting with `curvine://` — and return the same `unsupported("Curvine object store, …")` error before delegating to upstream. At minimum, add a regression test mirroring the `connect()` boundary test for `connect_namespace`, even if the gate is just `root` for now.

szbr00: `connect_namespace` returns the upstream `ConnectNamespaceBuilder` directly, so the explicit `curvine://` boundary established for `connect()` does not extend here. A caller doing

```rust
let mut props = HashMap::new();
props.insert("root".to_string(), "curvine:///data/lancedb/demo".to_string());
lancedb::connect_namespace("dir", props).execute().await
```

will fail inside upstream with whatever generic error the chosen namespace impl emits for an unknown URI scheme — not the intended `Curvine object store, uri=… is not implemented`. That defeats the test boundary expressed in `curvine_uri_reports_explicit_unsupported_boundary` and will look like an unrelated bug to users when the real Curvine integration lands.

Suggest applying the same gating: scan `properties` for known location keys (`root`, `uri`, …) — or any value starting with `curvine://` — and return the same `unsupported("Curvine object store, …")` error before delegating to upstream. At minimum, add a regression test mirroring the `connect()` boundary test for `connect_namespace`, even if the gate is just `root` for now.

curvine-lancedb-rs/Cargo.toml
jlon: Handled. Added explicit root workspace exclude and a dedicated workflow for this crate.

curvine-lancedb-rs/src/connection.rs
jlon: Handled.  now gates  in the facade and has a regression test.

jlon: Handled in the same fix.  now returns the explicit unsupported boundary for .

jlon: Handled. connect_namespace now gates curvine:// in the facade and has a regression test.

jlon: Handled in the same fix. connect_namespace now returns the explicit unsupported boundary for curvine://.

szbr9486 (approved):

Issue #797, opened 2026-05-08
[FEATURE] enable python sdk


lzjqsdd:
Curvine python-sdk is currently off. Re-enable it to adapt interfaces for vllm/lancedb ecosystem compatibility.

PR #800, opened 2026-05-09
build: ship Python SDK wheel in dist and libsdk fixes (#797)

lzjqsdd:
## Summary
- `build.sh`: wheel via maturin + `protoc` stubs; `build/.venv-python-sdk`; copy workspace binaries to `dist/lib` **after** SDK builds; default wheel tag `--compatibility linux`.
- libsdk: optional `jni`/`pyo3` features; docs README; tests use `CURVINE_CONF_FILE`; Java `LibFsTest` uses `fs.cv.*`; Python client/proto + `curvine_libsdk` loader.
## Testing
- `bash -n build/build.sh`


lzjqsdd:
@copilot based on latest codebase, resolve the merge conflicts in this pull request

szbr9486 (commented):
This PR silently flips `make build` and `make check-env` from "Java SDK only" to "Java + Python SDK" in the default invocation. Three new requirements are now mandatory unless the caller explicitly passes `--skip-python-sdk`:

1. `python3 >= 3.8` (the previous floor was `3.6.0` in `build/check-env.sh`, so any host with Python 3.6/3.7 that previously passed `make check-env` will now fail).
2. The `python3-venv` package, since `python3 -m venv "$PYTHON_SDK_VENV"` is created on every default build.
3. Network access to PyPI to install `maturin` into `build/.venv-python-sdk/` (`pip install -q -r build/requirements-python-sdk.txt`). Air-gapped or strict-internal build hosts will fail here even if Python is fine.

The compile Dockerfiles under `curvine-docker/compile/*` aren't updated in this PR, so the CI / release-build images may regress depending on what they ship today (Amazon Linux 2 in particular still defaults to Python 3.7).

Suggest one of:

- Add the new toolchain requirements to the PR description / CHANGELOG and update `curvine-docker/compile/Dockerfile_*` to install Python 3.8+ and `python3-venv`, OR
- Make Python SDK opt-in — e.g. only set `BUILD_PYTHON_SDK=1` when `-p python` (or `-p all`) is explicitly requested, mirroring how `--skip-java-sdk` previously had to be passed only on hosts that lacked the Java toolchain.

Either is fine, but landing as-is will surprise downstream builders.

build/build.sh
```diff
@@ -507,11 +535,108 @@ if should_build_package "java" && [ $SKIP_JAVA_SDK -eq 0 ]; then
...
+  if ! "$PYTHON_SDK_VENV/bin/python" -m pip --version >/dev/null 2>&1; then
+    echo "Bootstrapping pip in Python SDK build venv (ensurepip) ..."
+    "$PYTHON_SDK_VENV/bin/python" -m ensurepip --upgrade || {
+      echo "Error: pip is not available and ensurepip failed (install python3-venv)." >&2
+      exit 1
+    }
+  fi
+
+  echo "Installing / updating Python SDK build dependencies (maturin) ..."
+  "$PYTHON_SDK_VENV/bin/python" -m pip install -q --upgrade pip
```
szbr00, line 568: With `python-source = "python"` in `pyproject.toml`, maturin packages every file sitting next to `curvinefs/` and `curvine_libsdk/` into the wheel. After this `protoc` invocation that's all 9 stubs from `curvine-common/proto/*.proto`: `common_pb2.py`, `master_pb2.py`, `worker_pb2.py`, `raft_pb2.py`, `eraftpb_pb2.py`, `mount_pb2.py`, `p2p_pb2.py`, `replication_pb2.py`, `job_pb2.py`. They install as **top-level** modules in `site-packages/`, so:

- Every consumer of the wheel imports 9 generic-named modules (`common_pb2`, `job_pb2`, …) into the global namespace, which is highly likely to collide with other Protobuf-using libs in the same env.
- Internal cluster protocols (`raft_pb2`, `eraftpb_pb2`, `replication_pb2`, `worker_pb2`) become part of the public install surface even though they aren't part of the Python SDK contract.

Suggest emitting into a private sub-package, e.g.:

```
mkdir -p "$PY_SDK_PY/curvine_libsdk/_proto"
: > "$PY_SDK_PY/curvine_libsdk/_proto/__init__.py"
protoc -I"$PROTO_DIR" --python_out="$PY_SDK_PY/curvine_libsdk/_proto" "$PROTO_DIR"/*.proto
```

and updating `python/curvinefs/curvineClient.py` (and related modules) to `from curvine_libsdk._proto import master_pb2, common_pb2`. Generated `*_pb2.py` files cross-reference each other with absolute imports, so you'll also need `--python_out` plus a sys.path tweak or codegen post-processing — but the result is one well-namespaced wheel surface instead of nine free-floating top-level modules.

lzjqsdd: Protobuf files have been consolidated into curvine_libsdk._proto as per the review comments, and the Python version minimum for check-env/pyproject has been reverted to 3.6.

szbr9486 (approved):

PR #801, opened 2026-05-09
feat: add curvine object store scaffold

jlon:
## What changed
- added a new `object_store` module under `curvine-lancedb-rs`
- introduced minimal `CurvineObjectStore` and `CurvineObjectStoreProvider` scaffold types
- added `curvine_registry()` and `curvine_session()` helpers for `curvine` scheme registration
- added tests covering registry registration and session wiring for the `curvine` scheme

## Why
This is the second phase of the LanceDB on Curvine bring-up. The goal of this PR is not full storage IO, but the smallest compilable provider / registry / session scaffold needed before wiring `connect("curvine://...")` into the custom session path.

## Current boundary
- `curvine://` still does not perform real Curvine reads or writes
- provider creation paths exist, but `new_store()` still returns an explicit not-implemented error
- the purpose here is to establish the formal extension point shape before adding real IO behavior

## Validation
- `cargo fmt --manifest-path curvine-lancedb-rs/Cargo.toml --check`
- `cargo test --manifest-path curvine-lancedb-rs/Cargo.toml --test facade_compat -- --nocapture`
- `cargo clippy --manifest-path curvine-lancedb-rs/Cargo.toml --all-targets --all-features -- -D warnings`

### DIFF
--- a/curvine-lancedb-rs/src/connection.rs
+++ b/curvine-lancedb-rs/src/connection.rs
@@ -14,7 +14,7 @@
 
 use std::collections::HashMap;
 
-use crate::error::unsupported;
+use crate::object_store::curvine_session;
 
 pub use lancedb_upstream::connection::{CloneTableBuilder, OpenTableBuilder, TableNamesBuilder};
 pub use lancedb_upstream::connection::{ConnectRequest, Connection, LanceFileVersion};
@@ -22,7 +22,6 @@ pub use lancedb_upstream::connection::{ConnectRequest, Connection, LanceFileVers
 #[derive(Debug)]
 enum ConnectBuilderInner {
     Upstream(Box<lancedb_upstream::connection::ConnectBuilder>),
-    UnsupportedCurvineUri { uri: String },
 }
 
 #[derive(Debug)]
@@ -32,15 +31,14 @@ pub struct ConnectBuilder {
 
 impl ConnectBuilder {
     pub fn new(uri: &str) -> Self {
+        let builder = lancedb_upstream::connect(uri);
         if is_curvine_uri(uri) {
             Self {
-                inner: ConnectBuilderInner::UnsupportedCurvineUri {
-                    uri: uri.to_string(),
-                },
+                inner: ConnectBuilderInner::Upstream(Box::new(builder.session(curvine_session()))),
             }
         } else {
             Self {
-                inner: ConnectBuilderInner::Upstream(Box::new(lancedb_upstream::connect(uri))),
+                inner: ConnectBuilderInner::Upstream(Box::new(builder)),
             }
         }
     }
@@ -54,9 +52,6 @@ impl ConnectBuilder {
         match self.inner {
             ConnectBuilderInner::Upstream(builder) => Self {
                 inner: ConnectBuilderInner::Upstream(Box::new(f(*builder))),
-            },
-            ConnectBuilderInner::UnsupportedCurvineUri { uri } => Self {
-                inner: ConnectBuilderInner::UnsupportedCurvineUri { uri },
             },
         }
     }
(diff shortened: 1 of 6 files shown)


curvine-lancedb-rs/src/object_store.rs
```diff
@@ -0,0 +1,65 @@
...
+impl CurvineObjectStoreProvider {
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+#[async_trait]
+impl ObjectStoreProvider for CurvineObjectStoreProvider {
+    async fn new_store(&self, base_path: Url, _params: &ObjectStoreParams) -> Result<ObjectStore> {
+        Err(lance_core::Error::invalid_input(format!(
```
szbr00: Three things tied together on this line and its neighbors that are worth aligning before this scaffold lands:

1. **Error kind regression vs PR #799.** `new_store` returns `lance_core::Error::invalid_input(...)`, which propagates as `Error::InvalidInput` upstream. Previously the curvine boundary surfaced as `Error::NotSupported { message: "Curvine object store, uri=... is not implemented" }`. The test relaxation in `tests/facade_compat.rs` (no longer asserts the exact phrase) hides this — but any downstream code that already matches `Error::NotSupported { .. }` from the v0.27 facade will stop matching this case. Returning a `NotSupported`-shaped error here would preserve that contract.

2. **`unsupported_curvine_uri()` (line 58) is now dead.** The previous `error::unsupported` helper was deleted, and the only caller (the curvine path in `ConnectBuilder`/`ConnectNamespaceBuilder`) now goes through the provider instead. The new helper returns `Error::NotSupported` — i.e. the *right* kind — but is never called from anywhere in this crate or the tests. Either route the provider through this helper (which fixes point 1 for free), or drop it.

3. **`CurvineObjectStore` (line 28) is a `pub` unit struct with no fields, no impls, and no references** in this crate or the tests. As a published API name it currently means nothing; safer to introduce it in the PR that gives it semantics.

curvine-lancedb-rs/src/connection.rs
```diff
@@ -14,15 +14,14 @@
...
 
-use crate::error::unsupported;
+use crate::object_store::curvine_session;
 
 pub use lancedb_upstream::connection::{CloneTableBuilder, OpenTableBuilder, TableNamesBuilder};
 pub use lancedb_upstream::connection::{ConnectRequest, Connection, LanceFileVersion};
 
 #[derive(Debug)]
 enum ConnectBuilderInner {
     Upstream(Box<lancedb_upstream::connection::ConnectBuilder>),
```
szbr00, line 24: With `UnsupportedCurvineUri` removed, `ConnectBuilderInner` here and `ConnectNamespaceBuilderInner` further down (`Pending { .. }` is the only variant) are both single-variant enums. Each `ConnectNamespaceBuilder::*` setter (`storage_option`, `storage_options`, `read_consistency_interval`, `embedding_registry`, `session`, `server_side_query`) is now ~20 lines of `match inner { Pending { all_fields..., mut one_field, .. } => Pending { all_fields..., one_field } }` rewriting that exists solely because of the enum wrapper. Collapsing each `Inner` to a struct (or even just a `struct Pending { .. }` consumed/produced by `map_upstream`) would remove ~150 lines of boilerplate from this file and make it clearer where the actual `curvine://` decision is made (in `execute()`).

`map_upstream`'s name is also now misleading: it no longer maps the upstream builder, it maps the local pending state. Worth renaming alongside the enum cleanup.

curvine-lancedb-rs/src/object_store.rs
jlon: Handled.  now preserves a NotSupported-shaped boundary via the shared helper, and the meaningless public shell was dropped from this PR scope.

curvine-lancedb-rs/src/connection.rs
jlon: Handled. Collapsed the single-state namespace builder into a plain struct and renamed the top-level builder mapping accordingly.

szbr9486 (approved):

PR #806, opened 2026-05-12
refactor(server): schedule OS read-ahead once before local read

bigbigxu:
Drop redundant supports_read_ahead guards and remove the duplicate read_ahead block after read_region; keep a single read_ahead pass when the file exposes a local mutable view.

### DIFF
--- a/curvine-server/src/worker/handler/read_handler.rs
+++ b/curvine-server/src/worker/handler/read_handler.rs
@@ -158,22 +158,15 @@ impl ReadHandler {
 
         let spend = TimeSpent::new();
         // OS page cache read-ahead is only available for local files
-        if file.supports_read_ahead() {
-            if let Some(local) = file.as_local_mut() {
-                self.last_task = local.read_ahead(&self.os_cache, self.last_task.take());
-            }
+        if let Some(local) = file.as_local_mut() {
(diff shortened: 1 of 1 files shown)


szbr9486 (commented):
After this PR removes both `file.supports_read_ahead()` guards in `curvine-server/src/worker/handler/read_handler.rs`, the only remaining reference to `BlockDevice::supports_read_ahead()` (defined at `orpc/src/io/block_io.rs:86`) in the live tree is its own self-test assertion at `orpc/src/io/block_io.rs:181`. Every production call site is gone, and the gating role is now played by `as_local_mut()` directly (which is fine — both methods are just `matches!(self, BlockDevice::Local(_))`).

Two reasonable follow-ups:

- delete `supports_read_ahead()` plus its test line and rely on `as_local_mut().is_some()` everywhere (symmetric with how `as_local()` already replaces it semantically); or
- keep it but add a `// kept as part of the BlockDevice public API for ...` doc comment so it doesn't read as accidentally orphaned.

`supports_send_file()` and `supports_short_circuit()` stay justified because they're used in the boolean-only paths around it, but `supports_read_ahead` no longer carries its weight after this change.

szbr9486 (approved):

PR #807, opened 2026-05-12
feat: add `BTreeMap` free-list to `BdevOffsetAllocator` for bdev offset reuse

thuongle2210:
# Description
- This PR adds BTreeMap free-list for bdev offset reuse.
- Ensure the unit tests in vfs_dataset pass.
# Key Changes
- Added free_list field (BTreeMap) to BdevOffsetAllocator that supports offset reuse
- Add tests for the correctness of BTreeMap free-list.
- Update the unit tests in vfs_dataset to fix the following issues:
  + import missing modules
  + remove `?` to fix type mismatch (ExtendedBlock::new returns Self, not Result)  
  + update block.len to 70 as the actual length in the append() test (finalize_block already committed the previous write)  
  + remove redundant `feature = "spdk"` gate in vfs_dataset.rs
# Impact
- Prevents bdev offset space from being leaked on block free/abort
# Future work
- Optimize the process by some experiments with 2 BtreeMap or hybrid approach with Slab/Bitmap & 2 BtreeMap
- Handle fragmentation


thuongle2210:
Dear @szbr9486 , I have implemented a BTreeMap free-list as the initial mechanism for handling space reclamation. Could you please review the changes?

curvine-server/src/worker/storage/dir_state.rs
```diff
@@ -147,20 +203,35 @@ impl BdevOffsetAllocator {
...
             }
         }
-        if max_end > inner.cursor {
-            inner.cursor = max_end;
+
+        if prev_end > inner.cursor {
+            inner.cursor = prev_end;
+        }
+        // Gap between last block and cursor is also free
+        if prev_end < inner.cursor {
```
szbr00: Two related observations on this `restore()` tail:

1. **Dead branch.** `if prev_end < inner.cursor` is unreachable for the only production call site (`vfs_dir.rs:384`), which always passes a freshly-`new()`'d allocator (`cursor = 0`, `prev_end >= 0`). After the preceding `if prev_end > inner.cursor { inner.cursor = prev_end }`, `cursor >= prev_end` always holds, so this strict `<` check never fires on the fresh path. The "Gap between last block and cursor is also free" comment is therefore misleading.

2. **Partial reset.** `map.clear()` and `free_list.clear()` are called, but `cursor` is only adjusted *upward*. If `restore()` ever gets called on an allocator with `cursor > 0` (e.g. a future "rewind" or "re-import" caller), the existing cursor wins over the snapshot's `max_end`, and new allocations can silently skip past offsets that aren't in the snapshot. Conversely the dead branch above only protects against `prev_end < cursor_before_restore`, which is the exact case partial reset creates.

If `restore()` is intended to be fresh-only (which the current call pattern assumes), make that contract explicit — either `inner.cursor = 0;` right after the `clear()`s, or a `debug_assert_eq!(inner.cursor, 0);` — and the branch below collapses to dead code that can simply be deleted. If it's meant to be re-callable, then the cursor reset semantics need to be defined and the branch's logic verified.

curvine-server/src/worker/storage/dir_state.rs
```diff
@@ -92,6 +99,22 @@ impl BdevOffsetAllocator {
             ));
         }
         let aligned_size = align_up_i64(size, inner.align);
+
+        // Try free-list first
+        for (&offset, &fsize) in inner.free_list.iter() {
```
szbr00: This pattern — iterating `inner.free_list.iter()` while calling `inner.free_list.remove(&offset)` and `.insert(...)` inside the same loop — only compiles because the `return Ok(offset)` immediately ends iteration so the borrow checker can prove no second `.next()` will happen against the mutated map. It's a fragile shape: anyone refactoring this to e.g. collect multiple candidates (best-fit, smallest fitting, etc., as hinted by the "TODO: handle fragmentation" comment) will silently stop compiling and have to rewrite the search.

The more refactor-safe form lets the iterator's borrow end before any mutation:

```rust
let candidate = inner
    .free_list
    .iter()
    .find(|(_, &v)| v >= aligned_size)
    .map(|(&k, &v)| (k, v));
if let Some((offset, fsize)) = candidate {
    inner.free_list.remove(&offset);
    if fsize > aligned_size {
        inner.free_list.insert(offset + aligned_size, fsize - aligned_size);
    }
    inner.map.insert(block_id, (offset, aligned_size));
    return Ok(offset);
}
```

Same behavior, but the borrow lifecycle is explicit and the body of the search can grow without fighting the borrow checker.

szbr9486 (commented):
Three test-only changes in `curvine-server/src/worker/storage/vfs_dataset.rs` are unrelated to the BTreeMap free-list feature this PR advertises, and none are called out in the PR description:

1. `create_data_set` paths changed from `../testing/dataset-{}/dN` to `/tmp/curvine-test-{}/dN` (lines 388-390). This swaps a workspace-relative path for an absolute Linux-only path, which is fine for Curvine's target environments but loses portability and leaves test artifacts in `/tmp` across runs. If this was needed because the relative path was failing under some CI/cwd combination, a one-liner in the commit message would help.

2. `append` test: `block.len = 100` → `block.len = 70` (line 455). The new value matches the actually-written bytes (50 + 20). Either the prior test was relying on `finalize_block` accepting a length larger than the data (now stricter?) or it was just always slightly wrong; either way the PR description should note the behavior expectation it now encodes.

3. `abort_spdk` test: stray `?` removed from `ExtendedBlock::new(...)?` (around line 491). `ExtendedBlock::new` returns `ExtendedBlock`, not `Result`, so the old form would only compile if `?` short-circuited a `Result` wrapper that no longer exists — implying this test was either previously broken or built against an older signature, and is being silently fixed here.

None of these is a blocker, but each is the kind of change that ideally lives in its own commit (or at least a one-line note in the PR body) so future bisects don't have to attribute test-flakiness or finalize-block-validation regressions to the BTreeMap work.

curvine-server/src/worker/storage/dir_state.rs
thuongle2210: Thanks, it makes the code easier to scale.

curvine-server/src/worker/storage/dir_state.rs
thuongle2210: Agreed. I considered the re-callable case first, but I think it is unnecessary for the current approach

thuongle2210:
> Three test-only changes in `curvine-server/src/worker/storage/vfs_dataset.rs` are unrelated to the BTreeMap free-list feature this PR advertises, and none are called out in the PR description:
> 
>     1. `create_data_set` paths changed from `../testing/dataset-{}/dN` to `/tmp/curvine-test-{}/dN` (lines 388-390). This swaps a workspace-relative path for an absolute Linux-only path, which is fine for Curvine's target environments but loses portability and leaves test artifacts in `/tmp` across runs. If this was needed because the relative path was failing under some CI/cwd combination, a one-liner in the commit message would help.
> 
>     2. `append` test: `block.len = 100` → `block.len = 70` (line 455). The new value matches the actually-written bytes (50 + 20). Either the prior test was relying on `finalize_block` accepting a length larger than the data (now stricter?) or it was just always slightly wrong; either way the PR description should note the behavior expectation it now encodes.
> 
>     3. `abort_spdk` test: stray `?` removed from `ExtendedBlock::new(...)?` (around line 491). `ExtendedBlock::new` returns `ExtendedBlock`, not `Result`, so the old form would only compile if `?` short-circuited a `Result` wrapper that no longer exists — implying this test was either previously broken or built against an older signature, and is being silently fixed here.
> 
> 
> None of these is a blocker, but each is the kind of change that ideally lives in its own commit (or at least a one-line note in the PR body) so future bisects don't have to attribute test-flakiness or finalize-block-validation regressions to the BTreeMap work.


I've updated the PR description and code with details regarding the changes made to fix issues in the `vfs_dataset` tests

szbr9486 (approved):

PR #809, opened 2026-05-13
fix(cli): display file timestamps in local timezone

### LINKED ISSUE #802: [BUG]: ls / stat show timestamps in UTC (unexpected vs local timezone)
## Summary
When listing files or inspecting inode metadata (`ls`, `stat`) on mounts backed by Curvine CLI, file timestamps (`mtime` / `atime` / `ctime` as exposed to the kernel) appear as **UTC**, while behavior on typical local filesystems matches the **machine’s configured timezone**.
## Expected behavior
`ls`, `stat`, etc. display times consistently with **local TZ** (`TZ` / system zone), unless the project documents that timestamps are intentionally shown otherwise.
## Actual behavior
Timestamps appear in **UTC** (or clearly do not match the expected local-time display for the same wall clock compared to non-Curvine mounts).

LJF-050:
Fixes #802

This changes Curvine CLI ls/stat timestamp formatting to convert epoch milliseconds to the local timezone before display. The stored metadata remains Unix epoch milliseconds.

Note: human-readable timestamps in `cv fs ls` and `cv fs stat` now follow the process local timezone (`TZ` / system zone) instead of rendering as UTC. Non-positive or out-of-range timestamp values are displayed as `-`.

### DIFF
--- a/curvine-cli/src/cmds/fs/common.rs
+++ b/curvine-cli/src/cmds/fs/common.rs
@@ -65,3 +65,19 @@ pub fn format_size(size: u64) -> String {
         format!("{} B", size)
     }
 }
+
+/// Formats a Unix epoch timestamp in milliseconds using the local timezone.
+pub fn format_epoch_ms_local(timestamp_ms: i64, fmt: &str) -> String {
+    if timestamp_ms <= 0 {
+        return "-".to_string();
+    }
+
+    let Some(datetime) = chrono::DateTime::from_timestamp_millis(timestamp_ms) else {
+        return "-".to_string();
+    };
+
+    datetime
+        .with_timezone(&chrono::Local)
+        .format(fmt)
+        .to_string()
+}
--- a/curvine-cli/src/cmds/fs/ls.rs
+++ b/curvine-cli/src/cmds/fs/ls.rs
@@ -372,9 +372,8 @@ async fn print_file_entry(
 
     // Format time (use access time if requested, otherwise modification time)
     let timestamp = if config.atime { file.atime } else { file.mtime };
-    let datetime = chrono::DateTime::from_timestamp(timestamp / 1000, 0)
-        .unwrap_or_else(|| chrono::DateTime::from_timestamp(0, 0).unwrap());
-    let formatted_time = datetime.format("%Y-%m-%d %H:%M").to_string();
+    let formatted_time =
+        crate::cmds::fs::common::format_epoch_ms_local(timestamp, "%Y-%m-%d %H:%M");
 
     // Format filename (handle non-printable characters)
     let filename = if config.hide_non_printable {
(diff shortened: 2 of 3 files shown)


curvine-cli/src/cmds/fs/common.rs
```diff
@@ -65,3 +65,14 @@ pub fn format_size(size: u64) -> String {
         format!("{} B", size)
     }
 }
+
+/// Formats a Unix epoch timestamp in milliseconds using the local timezone.
+pub fn format_epoch_ms_local(timestamp_ms: i64, fmt: &str) -> String {
```
szbr00: Two minor observations about the fallback path:

- `from_timestamp_millis` only returns `None` for values far outside any realistic filesystem timestamp range (roughly +/-2.6e11 years), so the `unwrap_or_else` branch is effectively unreachable. If the intent is to defend against an uninitialized / sentinel value like `0` or a negative `mtime`, it would be clearer to gate on `timestamp_ms <= 0` explicitly and render `"-"` (or similar) so users can distinguish "no timestamp" from a real 1970 time.
- The previous UTC implementation rendered the fallback as `1970-01-01 00:00` everywhere; with `with_timezone(&chrono::Local)` the same fallback now renders as `1969-12-31 16:00` (or another offset) in negative-offset zones, which makes the fallback look like a real, plausible timestamp from the 1960s. Worth either picking a non-realistic placeholder string or anchoring the fallback in UTC.

szbr9486 (commented):
Nice clean fix for #802. Two follow-up suggestions worth considering before/after merge:

1. The new output renders local time without any timezone marker (no `%z`, no zone abbreviation). Operators looking at `cv fs ls` / `cv fs stat` output from a non-default-TZ host — or comparing output between an old UTC build and a new local build during rollout — have no way to tell which zone the timestamp is in. Appending `%z` to the format strings, or at least documenting the zone source in `--help`, would remove that ambiguity.

2. This is a backward-incompatible display change for anyone parsing `cv fs ls` / `cv fs stat` output programmatically. A short note in the PR description / release notes flagging the move from UTC to local would help downstream users. An opt-in `--utc` switch on `ls`/`stat` (defaulting to local) would also let existing scripts pin the previous behavior without forcing them onto local TZ.

curvine-cli/src/cmds/fs/common.rs
LJF-050: Thanks for the review. I updated the fallback path so non-positive or out-of-range timestamps render as "-" instead of a local-time epoch value.

LJF-050:
Thanks for the review. I updated the fallback path so non-positive or out-of-range timestamps render as "-" instead of a local-time epoch value.

For the timezone marker / --utc compatibility point, I kept the current ls/stat format aligned with typical local-time ls output and added a PR note that this is a display behavior change from UTC to local time. The stored metadata remains Unix epoch milliseconds. Happy to add an explicit --utc option in a separate follow-up PR if you prefer that compatibility path.

szbr9486 (approved):

PR #823, opened 2026-05-14
feat: complete Curvine LanceDB object store contract

jlon:
## What changed
- turns the Curvine LanceDB scaffold into a working Lance object store provider
- implements the core object-store contract for Curvine paths: put, head, get, range read, delete, copy, list, and list_with_delimiter
- wires real Curvine filesystem bootstrap through `CURVINE_CONF_FILE` and Lance storage options
- adds ignored live-cluster smoke tests for LanceDB and object-store semantics
- adds the first Curvine minicluster e2e coverage for LanceDB table lifecycle and basic query paths

## Why
This is PR 3 in the LanceDB on Curvine stack. PR #799 created the facade shell and PR #801 added the provider/session scaffold. This PR makes the provider usable for real LanceDB IO on Curvine while keeping live-cluster tests opt-in.

## Stack
- this PR: working Curvine object-store contract and initial e2e coverage

## Validation
- `cargo fmt --all`
- `cargo test -p curvine-lancedb-rs`


szbr9486 (commented):
The `.gitignore` change in this PR looks unrelated to the LanceDB object store contract and silently regresses ignore patterns introduced by the Python SDK PR (#800). Specifically, these lines are removed:

    build/.venv-python-sdk/
    curvine-libsdk/python/*_pb2.py
    curvine-libsdk/python/curvine_libsdk/_proto/*_pb2.py
    curvine-libsdk/python/curvine_libsdk/*.so
    curvine-libsdk/python/curvine_libsdk/*.pyd
    curvine-libsdk/python/native/

After this PR lands, anyone running `bash build/build.sh` (which now also builds the Python SDK by default) will see generated `*_pb2.py`, the `_native.so` / `.pyd` shared library, and the `.venv-python-sdk` virtualenv show up as untracked changes in `git status`, and `git add -A` will commit them. The new `/docs` → `# allow per-crate docs/` change is consistent with the PR scope and is fine, but the Python SDK ignore lines look like a rebase artifact and should be restored in a follow-up commit on this PR.

curvine-lancedb-rs/src/object_store.rs
```diff
@@ -8,38 +8,950 @@
...
+
+        if range.start > 0 {
+            reader
+                .seek(range.start as i64)
+                .await
+                .map_err(|e| fs_error_to_object_store(location, e))?;
+        }
+
+        let len = (range.end - range.start) as usize;
+        let mut buf = vec![0u8; len];
```
szbr00: `get_opts` materializes the entire requested range into a single `vec![0u8; len]` buffer and then yields it as one `Bytes` chunk via `stream::once`. For LanceDB queries that issue large ranged reads (full file fetches for compaction, manifest reads, or `arrow-flight`-style scans), this means a multi-GB range allocates a multi-GB contiguous heap buffer up front and only returns to the caller after the entire read completes. That defeats the streaming contract of `ObjectStore::get_opts` (callers expect to be able to back-pressure on the stream) and risks OOM on workers with lots of concurrent scans.

Two suggestions:

- At minimum, cap each request to a streaming chunk loop similar to `stream_copy_contents`, yielding `Bytes` chunks of e.g. `COPY_CHUNK_BYTES` (or a configurable `block_size`) instead of one giant buffer.
- The seek path can also be moved into the streaming loop so individual chunks use sequential `read_full` from `range.start`, avoiding the up-front full-range allocation entirely.

Same observation applies to `stream_copy_contents` at line ~770, which is careful about chunk size but allocates a fresh `vec![0u8; take]` per iteration; reusing a single chunk buffer across iterations would cut allocation traffic on large multipart copies.

jlon:
Fixed the .gitignore regression in the latest PR03 commit. The Python SDK generated files, native artifacts, and build/.venv-python-sdk ignore rules are restored while keeping the repo-root /docs rule.

curvine-lancedb-rs/src/object_store.rs
jlon: Fixed in the latest PR03 commit. `get_opts` now returns a stream backed by chunked Curvine reads instead of materializing the full range into one Vec. `stream_copy_contents` now uses a single reusable chunk buffer for bounded copy memory in this PR stage; later stacked PRs move this further to Curvine async_read/async_write DataSlice once the crate has the needed direct dependency.

szbr9486 (approved):

PR #826, opened 2026-05-14
fix: prepare Curvine LanceDB crate for git usage

jlon:
## What changed
- renames the package directory from `curvine-lancedb-rs` to `curvine-lancedb` while keeping the Rust library name as `lancedb`
- updates workspace dependencies and tests to use the new package name
- adds `curvine.master_addrs` / `master_addrs` as direct LanceDB storage options for Curvine master addresses
- keeps `curvine.conf.path` as the highest-priority full config option and falls back to `CURVINE_CONF_FILE`
- adds external-cluster e2e config support through `CURVINE_E2E_CONF_FILE`
- adds a README for Git/path dependency usage, URI semantics, session usage, validation, and known boundaries

## Why
This is PR 6 in the LanceDB on Curvine stack. It packages the module for business Git/path consumption and removes the need for callers to create a Curvine config file when they only need to pass master addresses.

## Stack
- #823: working Curvine object-store contract
- #824: shallow clone and conditional write semantics
- #825: safe commit wrapping, URI normalization, and streamed object IO
- this PR: crate rename, Git usage docs, direct `master_addrs` storage option, and external e2e config support

## Validation
- `cargo fmt --all`
- `cargo clippy -p curvine-lancedb --all-targets --all-features -- -D warnings`
- `cargo test -p curvine-lancedb`


curvine-lancedb/src/object_store.rs
```diff
@@ -1647,3 +1732,22 @@ fn fs_error_to_object_store(location: &Path, error: FsError) -> OsError {
...
+
+    if !matches!(error, FsError::Common(_)) {
+        return false;
+    }
+
+    // Older Curvine clusters can return filesystem misses as Common errors
+    // with server-side "not exists" / "not found" messages. Object-store list
+    // and head semantics still require these to behave as missing objects.
+    let message = error.to_string().to_ascii_lowercase();
+    message.contains("not exists") || message.contains("not found")
```
szbr00: The fallback substring match `message.contains("not exists") || message.contains("not found")` on a lowercased error string is fragile and can misclassify unrelated failures as NotFound. Real-world examples that would silently turn into "ok / empty" in callers of `delete`, `list_dir_status`, `remove_multipart_dir`, and `prune_empty_ancestors`:

- A permission failure rendered as `"AccessDenied: policy for user 'svc-foo' not found"` would let the delete return `Ok(())` even though the object still exists.
- A network/timeout error whose message includes "endpoint not found" would cause a list to return an empty `Vec` and hide the underlying outage.

The previous behavior only treated `FsError::FileNotFound`/`Expired`/`JobNotFound` as not-found, which was strict but safe. If older clusters do return `FsError::Common(_)` for filesystem misses, please match on a structured marker (e.g. a typed error code or a stable prefix the server controls) rather than free-form text, and at minimum scope the substring fallback to a narrow allowlist of message shapes (e.g. exact "file does not exist" / "path not found" produced by Curvine itself) so unrelated `Common` errors propagate as real failures.

curvine-lancedb/src/object_store.rs
```diff
@@ -181,44 +174,143 @@ impl CurvineObjectStoreProvider {
...
+        LanceError::invalid_input(format!(
+            "Failed to initialize Curvine master configuration from `{CURVINE_MASTER_ADDRS_KEY}`: {e}"
+        ))
+    })?;
+    conf.client.init().map_err(|e| {
+        LanceError::invalid_input(format!(
+            "Failed to initialize Curvine client configuration from `{CURVINE_MASTER_ADDRS_KEY}`: {e}"
+        ))
+    })?;
+    conf.fuse.init().map_err(|e| {
```
szbr00: Calling `conf.fuse.init()` and `conf.job.init()` here pulls in subsystems that have nothing to do with a LanceDB / object-store client. Two concrete consequences:

1. Any failure inside `fuse.init()` / `job.init()` (e.g. mount-point validation, hugepage checks, runtime quirks) becomes the user's first surface for this code path, and the `map_err` wraps it as `Failed to initialize Curvine ... from \`curvine.master_addrs\`: ...`, which blames the storage option for something it didn't configure.
2. It performs work the LanceDB use case never exercises — `CurvineFileSystem::with_rt` only needs master/client to be ready.

Consider initializing only what `CurvineFileSystem` actually consumes (i.e. `master` + `client`) for the `master_addrs` path, and either dropping the fuse/job init calls or moving them behind a feature flag. If you do keep them, please at least scope the error context so each `map_err` references the actual subsystem that failed (`... Curvine fuse configuration ...`) rather than blaming `curvine.master_addrs`.

curvine-lancedb/src/object_store.rs
```diff
@@ -57,6 +58,9 @@ use uuid::Uuid;
 pub const CURVINE_SCHEME: &str = "curvine";
 
 pub const CURVINE_CONF_FILE_KEY: &str = "curvine.conf.path";
+pub const CURVINE_MASTER_ADDRS_KEY: &str = "curvine.master_addrs";
+
+const MASTER_ADDRS_KEY_ALIAS: &str = "master_addrs";
```
szbr00: `master_addrs` is a very generic option name. Lance's `ObjectStoreParams::storage_options` is a flat `HashMap<String, String>` shared across all backends, so if a caller is composing a multi-backend `Session` (e.g. Curvine + S3 + an HDFS-like UFS) and any of those other backends already accepts a `master_addrs` key, this alias silently captures it for Curvine.

Two options:
- Drop the bare alias and keep only `curvine.master_addrs` (the prefixed, unambiguous form). The README already documents the prefixed form prominently.
- Or, if a short alias is desired for ergonomics, namespace it (e.g. `curvine_master_addrs`) so it cannot collide with other backends' options.

Either way, please mention in the README which name is the canonical/stable one and what alias compatibility guarantees you intend to keep.

szbr00 (commented):
Two cross-cutting observations:

1. README language balance: the new `curvine-lancedb/README.md` is ~290 lines of Chinese followed by ~30 lines of an English "Quick Start" stub. Since the crate explicitly targets external Rust consumers ("Git/path dependency", `repository`, `documentation = https://docs.rs/curvine-lancedb`), most of the high-value content (URI semantics, session usage, supported operations, known boundaries, validation steps) is invisible to readers who don't read Chinese. Consider either (a) making English the primary content and keeping a Chinese appendix, or (b) splitting into `README.md` (English) and `README.zh-CN.md` and using `readme = "README.md"` in `Cargo.toml`. As-is, the docs.rs landing page will be primarily Chinese.

2. Missing test for documented option-priority ordering: the README claims a fixed precedence `curvine.conf.path > curvine.master_addrs/master_addrs > CURVINE_CONF_FILE`, and the implementation in `resolve_curvine_conf` does enforce that order, but the only test coverage for precedence is `facade_boundary_curvine_master_addrs_storage_option_precedes_env_conf` (master_addrs vs env). There is no test asserting `curvine.conf.path` storage option beats `curvine.master_addrs` storage option, nor that `master_addrs` (short alias) is only used when `curvine.master_addrs` is absent. A focused unit test on `resolve_curvine_conf` / `resolve_curvine_master_addrs` covering all three rungs of that ladder would lock the documented behavior in.

curvine-lancedb/src/object_store.rs
jlon: Fixed in 8280d8a6. Common errors now only map to NotFound for a narrow Curvine filesystem-miss allowlist; unrelated Common errors propagate as Generic.

curvine-lancedb/src/object_store.rs
jlon: Fixed in 8280d8a6. The curvine.master_addrs path now initializes only master and client configuration before constructing CurvineFileSystem.

curvine-lancedb/src/object_store.rs
jlon: Fixed in 8280d8a6. Dropped the bare master_addrs alias; Curvine now only consumes curvine.master_addrs, and README documents that as the canonical key.

jlon:
Addressed the cross-cutting review in 8280d8a6: README is now English-first with a short Chinese appendix, the documented precedence is now curvine.conf.path > curvine.master_addrs > CURVINE_CONF_FILE, and facade tests cover conf path over master addrs plus ignoring the unprefixed master_addrs key. Also rebased the PR onto latest main after #833/#835/#843.

jlon:
Removed Swatinem/rust-cache@v2 from the build workflow in 03df70cb and pushed a new run. Local validation passed: fmt, curvine-lancedb clippy --all-targets --all-features -D warnings, curvine-lancedb tests, and curvine-tests lancedb_object_store_e2e clippy.

szbr00 (approved):

PR #827, opened 2026-05-16
perf: skip eventfd write when active and spin briefly before idle

thuongle2210:
# Description
 - Avoid unnecessary eventfd write syscall when the poller is already actively processing I/O
 - Add a brief spin before transitioning to Idle to skip the Idle→eventfd→wake round-trip for back-to-back I/O
 # Key Changes
 - Added `is_sleeping` flag with acquire-release ordering so bdevs can skip `eventfd.write(1)` when the poller is active
 - Added `SPIN_ITER` spin loop before Active to Idle transition to catch near-future work without entering Idle immediately
 # Impact
 - Eliminates eventfd write + poll wake syscall per I/O submission when the poller is already active
 - Improves tail latency under bursty workloads by avoiding unnecessary Idle transitions
 # Future work
 - Make `SPIN_ITER` tunable per-instance for cpu-latency tradeoff


thuongle2210:
Dear @szbr9486 , could you please help me review this optimization?

PR #828, opened 2026-05-16
perf: skip eventfd write when active and spin briefly before idle

thuongle2210:
Description
 - Avoid unnecessary eventfd write syscall when the poller is already actively processing I/O
 - Add a brief spin before transitioning to Idle to skip the Idle→eventfd→wake round-trip for back-to-back I/O
 # Key Changes
 - Added `is_sleeping` flag with acquire-release ordering so bdevs can skip `eventfd.write(1)` when the poller is active
 - Added `SPIN_ITER` spin loop before Active to Idle transition to catch near-future work without entering Idle immediately
 - Introduce PollerConfig struct to store configurations for SpdkPoller
 # Impact
 - Eliminates eventfd write + poll wake syscall per I/O submission when the poller is already active
 - Improves tail latency under bursty workloads by avoiding unnecessary Idle transitions.
 - small benchmark: (cluster config file: [cluster_conf_spdk_pr_828.txt](https://github.com/user-attachments/files/27853439/cluster_conf_spdk_pr_828.txt) )
   + before:

      + `fs.write size: 20.0GB, cost: 48.10 s, speed: 425.8MB/s, bandwidth: 3.3GB/s
Action fs.write, len 21474836480, checksum 7952263966359552
args: BenchArgs { action: "fs.write", conf: "../curvine-cluster-config/curvine-cluster_spdk_skip_eventfd.toml", dir: "/bench-large", file_num: 10, client_threads: 10, buf_size: "4KB", file_size: "2GB", delete_file: false, checksum: true }`

     + `fs.read size: 20.0GB, cost: 37.59 s, speed: 544.8MB/s, bandwidth: 4.3GB/s
Action fs.read, len 21474836480, checksum 7952263966359552
args: BenchArgs { action: "fs.read", conf: "../curvine-cluster-config/curvine-cluster_spdk_skip_eventfd.toml", dir: "/bench-large", file_num: 10, client_threads: 10, buf_size: "4KB", file_size: "2GB", delete_file: false, checksum: true }`

    + after:
      + `fs.write size: 20.0GB, cost: 47.76 s, speed: 428.8MB/s, bandwidth: 3.4GB/s
Action fs.write, len 21474836480, checksum 11487626804592640
args: BenchArgs { action: "fs.write", conf: "../curvine-cluster-config/curvine-cluster_spdk_skip_eventfd.toml", dir: "/bench-large", file_num: 10, client_threads: 10, buf_size: "4KB", file_size: "2GB", delete_file: false, checksum: true }`

       + `fs.read size: 20.0GB, cost: 36.28 s, speed: 564.5MB/s, bandwidth: 4.4GB/s
Action fs.read, len 21474836480, checksum 11487626804592640
args: BenchArgs { action: "fs.read", conf: "../curvine-cluster-config/curvine-cluster_spdk_skip_eventfd.toml", dir: "/bench-large", file_num: 10, client_threads: 10, buf_size: "4KB", file_size: "2GB", delete_file: false, checksum: true }`

 # Future work
 - Add benchmark tests that feature a short interval between reads and writes


thuongle2210:
Dear @szbr9486 , could you pls help me review this optimization?

bigbigxu (approved):

orpc/src/io/spdk_poller.rs
```diff
@@ -278,6 +303,16 @@ impl SpdkPoller {
...
 
+                // Mark as sleeping before blocking — bdevs will write eventfd to wake us
+                is_sleeping.store(true, Ordering::Release);
+
+                // Recheck channel after setting flag (closes race with bdev send)
+                if !rx.is_empty() {
+                    is_sleeping.store(false, Ordering::Release);
+                    state = PollerState::Active;
+                    continue;
+                }
```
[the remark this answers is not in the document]
thuongle2210: This is a very rare case. I will consider to update it.

orpc/src/io/spdk_poller.rs
```diff
@@ -260,9 +278,16 @@ impl SpdkPoller {
                     true
                 });
 
-                // Transition to Idle if no more work
+                // Spin briefly before Idle to avoid eventfd round-trip for back-to-back I/O
                 if rx.is_empty() && active_qpairs.is_empty() {
                     state = PollerState::Idle;
+                    for _ in 0..SPIN_ITER {
+                        std::hint::spin_loop();
```
szbr00, line 284: The `!active_qpairs.is_empty()` short-circuit inside the spin is effectively dead. `active_qpairs` is a local `Vec` on the poller thread, mutated only by `submit_one` / `handle_unregister`, both of which run from inside this very loop. We entered this block under the precondition `rx.is_empty() && active_qpairs.is_empty()`, and nothing in the spin body adds to it, so the predicate can never become true here.

The check is harmless but it suggests a cross-thread invariant that does not exist and makes the wake path harder to reason about. Please simplify to:

    for _ in 0..SPIN_ITER {
        std::hint::spin_loop();
        if !rx.is_empty() {
            state = PollerState::Active;
            break;
        }
    }

That makes it obvious the only signal we're spinning on is "did a new request land in the channel". As a small follow-up, note the Idle branch immediately does `let has_pending = !rx.is_empty();` right after `continue`, so the spin's last iteration result is re-checked one more time — that's fine, but a brief comment near either side explaining the layering would prevent future drift.

szbr9486 (commented):
The benchmark numbers in the description (write 425 → 428 MB/s, read 544 → 564 MB/s on a 20 GB sequential workload) are within the run-to-run variance you'd normally see on this kind of streaming I/O, and they don't directly validate either of the two claims in "Impact":

1. *"Eliminates eventfd write + poll wake syscall per I/O submission when the poller is already active."* The right metric here is a count delta on the syscall itself — e.g. `strace -c -p <worker-pid> -e eventfd2,write,poll` on a baseline run vs this PR, or a thread-local counter exposed in metrics. The throughput number doesn't show whether elision is happening at all, only that it isn't making things slower.

2. *"Improves tail latency under bursty workloads."* A 20 GB sequential streaming benchmark is not bursty — once the pipeline is warm the poller is almost always Active, so neither the eventfd skip nor the SPIN_ITER path is the bottleneck. To support the tail-latency claim, a p99/p99.9 histogram (or a per-op latency CDF) under a bursty pattern (e.g. short I/O bursts separated by ~µs–ms idle periods) would be much more convincing.

Combined with the Copilot thread about `Ordering::Acquire/Release` being weaker than the Dekker handshake actually needs (which you've said you'll look at), it would be helpful to land the ordering fix and one of the two measurements above together, so reviewers can see the real change in syscall/latency profile rather than ~noise-level throughput.

orpc/src/io/spdk_poller.rs
```diff
@@ -20,7 +20,7 @@ use std::sync::{Arc, Condvar, Mutex};
 use std::thread::JoinHandle;
 
 const EVENTSZ: usize = std::mem::size_of::<u64>();
-
+const SPIN_ITER: u32 = 1000; // TODO: make tunable for cpu-latency tradeoff
 /// I/O operation submitted to the poller thread.
```
szbr00, line 23: A couple of suggestions for `SPIN_ITER`:

1. Make it tunable via `SpdkConf` (alongside `io_timeout_us` / `poll_interval_ms`) instead of a compile-time constant, so operators can tune the cpu↔latency trade-off per cluster without rebuilding orpc. A value of `0` should disable the spin entirely — useful for sparse, latency-tolerant workloads where we'd rather block immediately than burn ~1 µs of CPU on every Active→Idle transition.

2. Worth a comment near the constant explaining the chosen calibration. On typical x86 uarchs, 1000 × `std::hint::spin_loop()` (PAUSE) is roughly 300 ns–1 µs, i.e. the same order of magnitude as an `eventfd.write` + `poll` wake round-trip. The trade-off is essentially "burn one wake-trip's worth of CPU on the chance that the next request arrives within that window." That reasoning is not obvious from the code or the TODO.

The TODO already flags "make tunable for cpu-latency tradeoff" — wiring it through SpdkConf now is a small lift and avoids us shipping a fixed magic number that's hard to revisit later.

thuongle2210:
> The benchmark numbers in the description (write 425 → 428 MB/s, read 544 → 564 MB/s on a 20 GB sequential workload) are within the run-to-run variance you'd normally see on this kind of streaming I/O, and they don't directly validate either of the two claims in "Impact":
> 
> 1. _"Eliminates eventfd write + poll wake syscall per I/O submission when the poller is already active."_ The right metric here is a count delta on the syscall itself — e.g. `strace -c -p <worker-pid> -e eventfd2,write,poll` on a baseline run vs this PR, or a thread-local counter exposed in metrics. The throughput number doesn't show whether elision is happening at all, only that it isn't making things slower.
> 2. _"Improves tail latency under bursty workloads."_ A 20 GB sequential streaming benchmark is not bursty — once the pipeline is warm the poller is almost always Active, so neither the eventfd skip nor the SPIN_ITER path is the bottleneck. To support the tail-latency claim, a p99/p99.9 histogram (or a per-op latency CDF) under a bursty pattern (e.g. short I/O bursts separated by ~µs–ms idle periods) would be much more convincing.
> 
> Combined with the Copilot thread about `Ordering::Acquire/Release` being weaker than the Dekker handshake actually needs (which you've said you'll look at), it would be helpful to land the ordering fix and one of the two measurements above together, so reviewers can see the real change in syscall/latency profile rather than ~noise-level throughput.

Thank you for your recommendation:
Honestly, I ran about 10 iterations comparing the before and after states and observed a slight performance gain.
Regarding issue 1: Logically, we easily see that a large number of eventfd writes are eliminated.
Regarding issue 2: I also think we should add bench tests related to the short intervals between I/O that I mentioned as future work. For now, this PR introduces a tunable configuration that will serve as the baseline for those future performance charts.

Additionally, I also followed Dekker handshake.

orpc/src/io/spdk_poller.rs
thuongle2210: I configured it to be tunable

orpc/src/io/spdk_poller.rs
thuongle2210: I removed the redundant checking.

orpc/src/io/spdk_poller.rs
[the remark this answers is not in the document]
thuongle2210: I replaced Acquire/Release by Ordering::SeqCst follow Dekker handshake

Issue #796, opened 2026-05-08
[FEATURE] Add a Python SDK for LanceDB on Curvine backed by the Rust core


jlon:
## Background

This is part of the `LanceDB on Curvine` module proposal:
- https://github.com/CurvineIO/curvine/discussions/793

Python should not implement a separate storage path. The Python SDK must reuse the same Rust core used by the Rust facade crate.

## Goal

Add a Python SDK layer for `LanceDB on Curvine` backed by the Rust core implementation.

## Scope

- create the Python package skeleton
- expose a Python-friendly `connect(...)`
- bind to the Rust core through pyo3/maturin or equivalent
- support at least:
  - connect
  - open/create table
  - write a small dataset
  - query
- reuse the same Curvine object store semantics as Rust

## Acceptance Criteria

- Python can connect to a `curvine://...` workspace
- Python can create or open a table using the Rust core backend
- Python can write and query a minimal dataset
- there is no separate Python-side storage implementation that duplicates Curvine semantics

yuw1:
Hello! I'm interested in working on this issue. It seems like a great place for me to start contributing to the project. Could you please assign it to me? Thank you!

jlon:
Resolved by #835 and follow-up #843. The merged curvine-lancedb module includes Python package and native bindings together with Curvine-backed LanceDB integration and tests.

PR #835, opened 2026-05-20
feat: add Python SDK for LanceDB on Curvine (#796)

yuw1:
## Summary                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
Python SDK for LanceDB on Curvine, backed by the Rust core via PyO3/maturin.                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                   
#796

### DIFF
--- a/curvine-lancedb-rs/src/python/arrow_bridge.rs
+++ b/curvine-lancedb-rs/src/python/arrow_bridge.rs
@@ -1,0 +1,98 @@
+use std::ffi::CString;
+use std::sync::Arc;
+
+use arrow_array::{Array, RecordBatch, StructArray};
+use arrow_schema::{Schema, SchemaRef};
+use pyo3::prelude::*;
+use pyo3::types::PyCapsule;
+
+/// Convert a Rust `RecordBatch` to a `pyarrow.RecordBatch` via Arrow C Data Interface.
+pub fn record_batch_to_pyarrow(py: Python<'_>, batch: &RecordBatch) -> PyResult<PyObject> {
+    let struct_array = StructArray::from(batch.clone());
+    let data = struct_array.to_data();
+
+    let (ffi_array, ffi_schema) = arrow_array::ffi::to_ffi(&data)
+        .map_err(|e| pyo3::exceptions::PyValueError::new_err(e.to_string()))?;
+
+    let array_name = CString::new("arrow_array").unwrap();
+    let array_capsule = PyCapsule::new(py, ffi_array, Some(array_name))?;
+
+    let schema_name = CString::new("arrow_schema").unwrap();
+    let schema_capsule = PyCapsule::new(py, ffi_schema, Some(schema_name))?;
+
+    let pyarrow = py.import("pyarrow")?;
+    let pa_batch = pyarrow
+        .getattr("RecordBatch")?
+        .call_method1("_import_from_c_capsule", (schema_capsule, array_capsule))?;
+    Ok(pa_batch.unbind())
+}
+
+/// Convert a `pyarrow.RecordBatch` or `pyarrow.Table` to a Rust `RecordBatch`.
+pub fn record_batch_from_pyarrow(py: Python<'_>, obj: &Bound<'_, PyAny>) -> PyResult<RecordBatch> {
+    let pyarrow = py.import("pyarrow")?;
+    let pa_record_batch = pyarrow.getattr("RecordBatch")?;
+
+    let batch_obj = if obj.is_instance(&pa_record_batch)? {
+        obj.clone()
+    } else {
+        let pa_table = pyarrow.getattr("Table")?;
+        if obj.is_instance(&pa_table)? {
+            obj.call_method0("combine_chunks")?
+        } else {
+            return Err(pyo3::exceptions::PyTypeError::new_err(
+                "Expected pyarrow.RecordBatch or pyarrow.Table",
+            ));
+        }
+    };
+
+    // Use __arrow_c_array__ (PyCapsule protocol, pyarrow >= 14)
+    let capsule_tuple = batch_obj.call_method0("__arrow_c_array__")?;
+    let tuple: (Bound<'_, PyCapsule>, Bound<'_, PyCapsule>) = capsule_tuple.extract()?;
+
+    let schema_capsule = &tuple.0;
+    let array_capsule = &tuple.1;
+
+    // Transfer ownership from PyCapsule to Rust via from_raw.
+    // This replaces the originals with empty structs (release=None) so capsule
+    // destructors are no-ops, preventing double-free.
+    let ffi_schema_ptr = schema_capsule.pointer() as *mut arrow_schema::ffi::FFI_ArrowSchema;
+    let ffi_schema = unsafe { arrow_schema::ffi::FFI_ArrowSchema::from_raw(ffi_schema_ptr) };
+
+    let ffi_array_ptr = array_capsule.pointer() as *mut arrow_array::ffi::FFI_ArrowArray;
+    let ffi_array = unsafe { arrow_array::ffi::FFI_ArrowArray::from_raw(ffi_array_ptr) };
+
+    let array_data = unsafe { arrow_array::ffi::from_ffi(ffi_array, &ffi_schema) }
+        .map_err(|e| pyo3::exceptions::PyValueError::new_err(e.to_string()))?;
+
+    // Convert ArrayData to StructArray, then extract schema and columns
+    let struct_array = StructArray::from(array_data);
+    let fields = struct_array.fields().clone();
+    let schema: SchemaRef = Arc::new(Schema::new(fields));
+    let columns: Vec<_> = struct_array.columns().to_vec();
+
+    RecordBatch::try_new(schema, columns)
+        .map_err(|e| pyo3::exceptions::PyValueError::new_err(e.to_string()))
+}
+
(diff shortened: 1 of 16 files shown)


szbr9486 (commented):
The unit test suite in `python/tests/test_basic.py` is good for smoke-checking the PyO3 bindings, but every test uses `str(tmp_path / "test-db")` as the URI — that's a local-filesystem LanceDB, which is exactly the upstream LanceDB code path with **no** Curvine integration exercised. The whole reason `curvine-lancedb-rs` exists is the `curvine://` object-store provider; the Python SDK currently has zero coverage of that path.

At minimum, please add (mirroring the Rust-side `lancedb_smoke.rs`):

- a `@pytest.mark.skipif(not os.environ.get("CURVINE_MASTER_ADDRS"), ...)` or `@pytest.mark.ignored` test that
  - connects to `curvine:///tmp/py_sdk_smoke_<uuid>` with `.storage_option("curvine.master_addrs", ...)`,
  - creates a table, inserts rows, queries them back,
  - drops the table.

Without this, a regression in any of:
  - storage-option key handling in the Rust facade,
  - safe-commit wrapping survival through the PyO3 connection conversion,
  - Curvine-aware path semantics (`is_curvine_uri`, master_addrs alias, etc.)
will not be caught by the Python test suite.

Two more items worth wiring up while you're here (separate PRs are fine):

- A GitHub Actions job that runs `maturin develop --features python-sdk` and `pytest python/tests/ -v` — right now the Python tests don't run in CI as configured.
- An explicit pyarrow version pin/range in `pyproject.toml` deps that matches the `arrow-array = 57.2` Rust side (pyarrow's C ABI compatibility window is narrow; mismatched versions tend to surface as confusing capsule errors).

curvine-lancedb-rs/src/python/py_query.rs
```diff
@@ -0,0 +1,121 @@
...
+        Self {
+            inner: Some(query),
+        }
+    }
+}
+
+#[pymethods]
+impl PyQuery {
+    fn limit(mut slf: PyRefMut<Self>, limit: usize) -> PyResult<PyRefMut<Self>> {
+        let q = slf.inner.take().unwrap();
```
szbr00: This `let q = slf.inner.take().unwrap();` (and the identical pattern on every `PyQuery` / `PyVectorQuery` / `PyConnectBuilder` method, including `to_arrow`, `vector`, `execute`) will panic if the builder has already been consumed:

```python
q = table.search().where("score > 90")
df = (await q.to_arrow()).to_pandas()
# Re-using the same builder — to_arrow already moved `inner` to None
await q.to_arrow()       # panics in Rust
q.limit(10)              # panics in Rust
```

A Rust panic propagates to Python as `PanicException`, which is a confusing failure mode and is worse than the equivalent native LanceDB Python SDK which either keeps the builder cloneable or raises a proper `RuntimeError`.

Two reasonable fixes:

1. Replace `take().unwrap()` with `take().ok_or_else(|| pyo3::exceptions::PyRuntimeError::new_err("query already executed or consumed"))?` so the misuse becomes a real Python exception.
2. If `lancedb_upstream::query::Query` / `VectorQuery` implement `Clone`, prefer cloning the inner value on each chained call so the Python builder stays reusable — this matches Pythonic builder semantics and the README's `query.limit(...).offset(...).where(...)` chaining example, where a user reasonably expects to inspect or re-execute the same `Query`.

The same applies to `PyConnectBuilder::execute` — a second `await conn.execute()` currently panics rather than raising.

curvine-lancedb-rs/src/python/py_connection.rs
```diff
@@ -0,0 +1,71 @@
+use lancedb_upstream::Connection as UpstreamConnection;
+use pyo3::prelude::*;
+
+use super::arrow_bridge;
+use super::error;
+use super::py_table::PyTable;
+
+#[pyclass(name = "Connection")]
+pub struct PyConnection {
+    inner: UpstreamConnection,
```
szbr00: `PyConnection` stores only the raw `UpstreamConnection` and drops the Curvine-side `ConnectionOptions` returned from `connection::Connection`. That's working today because the upstream `Database` was already wrapped with `SafeCommitDatabase` by `wrap_upstream_connection_for_curvine` during `execute`, so safe-commit semantics survive the conversion.

But it has two visible consequences worth calling out:

1. **`Connection::clone_table` Curvine fast path is unreachable from Python.** The Rust `Connection::clone_table` checks `is_curvine_uri(...)` and routes to `clone_curvine_table` (shallow clone via `lance` dataset builder with the conditional-put commit handler). The Python API simply doesn't expose `clone_table`, so Python users will never get the Curvine-aware shallow clone behavior advertised in #826.

2. **Per-call `storage_option` overrides cannot be threaded from Python.** The Rust API supports `conn.create_table("t", batch).storage_option(K, V).execute()` to inject storage options at the per-table call site. `PyConnection::create_table` calls `self.inner.create_table(&name, batch).execute()` directly, with no way for a Python caller to attach storage options after `connect()`.

Recommend keeping the `ConnectionOptions` (or at least the storage_options map) on `PyConnection`, exposing a `clone_table(...)` that delegates to the Curvine-aware path, and threading user-supplied options into `create_table` / `open_table` builder calls.

curvine-lancedb-rs/src/python/arrow_bridge.rs
```diff
@@ -0,0 +1,98 @@
...
+    let tuple: (Bound<'_, PyCapsule>, Bound<'_, PyCapsule>) = capsule_tuple.extract()?;
+
+    let schema_capsule = &tuple.0;
+    let array_capsule = &tuple.1;
+
+    // Transfer ownership from PyCapsule to Rust via from_raw.
+    // This replaces the originals with empty structs (release=None) so capsule
+    // destructors are no-ops, preventing double-free.
+    let ffi_schema_ptr = schema_capsule.pointer() as *mut arrow_schema::ffi::FFI_ArrowSchema;
+    let ffi_schema = unsafe { arrow_schema::ffi::FFI_ArrowSchema::from_raw(ffi_schema_ptr) };
```
szbr00, line 64: The manual `FFI_ArrowSchema::from_raw(ptr)` / `FFI_ArrowArray::from_raw(ptr)` pattern works, but it depends on arrow-rs's internal invariant that `from_raw` empties the source struct (sets `release` to `None`) so the PyCapsule's destructor becomes a no-op. The comment correctly documents this, but the contract is brittle: a future arrow-array minor bump that changes `from_raw`'s ownership story would silently introduce double-frees here, with no compile-time signal.

The `arrow` crate's `pyarrow` feature exposes `arrow::pyarrow::FromPyArrow` / `ToPyArrow`, which are exactly the standard encapsulation of the Arrow C Data Interface PyCapsule protocol used by the ecosystem (datafusion-python, polars, etc.). The equivalent of this function becomes roughly:

```rust
use arrow::pyarrow::FromPyArrow;

let batch_any = if obj.is_instance(&pa_record_batch)? {
    obj.clone()
} else {
    /* combine_chunks on Table */
};
RecordBatch::from_pyarrow_bound(&batch_any).map_err(...)
```

Same applies to `record_batch_to_pyarrow` — `batch.to_pyarrow(py)` replaces the manual capsule construction.

If the dep size is a concern, the `arrow` crate's `pyarrow` feature alone is not very large and you're already pulling `arrow-array` + `arrow-schema`. Either way, switching to the upstream helper removes ~50 lines of unsafe FFI code and the comment that has to justify them.

yuw1:
Hi @szbr9486, thanks for the thorough review! Here's how each comment was addressed:

**1. `take().unwrap()` panics** — Replaced all 15 occurrences in `PyQuery`, `PyVectorQuery`, and `PyConnectBuilder` with `take().ok_or_else(already_consumed_err)?`, which raises a proper `RuntimeError` in Python instead of panicking.

**2. `PyConnection` dropping `ConnectionOptions`** — `PyConnection` now stores the full `Connection` (not just `UpstreamConnection`). Added `clone_table(target_name, source_uri)` that delegates to the Curvine-aware fast path. `ConnectionOptions` and `storage_options` are `pub(crate)` for future per-call overrides.

**3. Arrow PyCapsule FFI** — Investigated switching to `arrow::pyarrow::{FromPyArrow, ToPyArrow}`, but `arrow-pyarrow v57.x` requires pyo3 0.26 while the workspace uses pyo3 0.25.1 (shared with `curvine-libsdk`). Both depend on `pyo3-ffi` with `links = "python"`, so they can't coexist. Added a NOTE comment documenting this constraint and the upgrade path.

**4. Test coverage & pyarrow pin** — Added `TestCurvineIntegration` that connects to `curvine:///tmp/py_sdk_smoke_<uuid>` with `storage_option("curvine.conf.path", ...)`, exercises create/query/drop against a live cluster. Skipped automatically when `CURVINE_CONF_FILE` is not set. Also pinned `pyarrow>=14.0.0,<19` in `pyproject.toml`.

Additionally, fixed a bug in `fs_dir.rs` where `list_status` / `list_options` returned `FsError::Common` (via `err_box!`) instead of `FsError::FileNotFound` when a path doesn't exist. This caused LanceDB's `create_table` to fail on `curvine://` URIs because the object-store adapter's `list_curvine_dir_or_empty` only catches `FileNotFound` to return an empty listing.

All 15 tests pass (14 local + 1 Curvine integration).

szbr00 (approved):

szbr00 (approved):

PR #838, opened 2026-05-23
fix: show leader address in raft state log (#507)

### LINKED ISSUE #507: [BUG]: Optimize the leader election log prompts after the Raft leader is elected.
Currently, the master node ID is displayed for the leader election prompt. It is recommended to change it to the primary master node's address.

NolanHo:
## Summary
- Update the Raft SoftState state-change runtime log to show the leader master address without the node id.
- Add an address-only formatter for Raft peers with unit coverage for known and unknown peer ids.

## Context
Maintainer confirmed the target log site in #507: https://github.com/CurvineIO/curvine/issues/507#issuecomment-4467573435

Closes #507.

## Tests
- `cargo fmt --check`
- `cargo test -p curvine-common raft_group -- --nocapture`
- `npm run commitlint`
- `cargo clippy -p curvine-common --all-targets -- --deny warnings --allow clippy::uninlined-format-args`

### DIFF
--- a/curvine-common/src/raft/raft_group.rs
+++ b/curvine-common/src/raft/raft_group.rs
@@ -52,6 +52,16 @@ impl RaftGroup {
         }
     }
 
+    pub fn get_addr_only_string(&self, id: u64) -> String {
+        if id == 0 {
+            return "unknown".to_string();
+        }
+        match self.peers.get(&id) {
+            None => format!("{}", id),
+            Some(v) => format!("{}:{}", v.hostname, v.port),
+        }
+    }
+
     pub fn voters(&self) -> Vec<u64> {
         self.peers.iter().map(|x| *x.0).collect()
     }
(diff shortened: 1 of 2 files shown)


curvine-common/src/raft/raft_group.rs
```diff
@@ -52,6 +52,13 @@ impl RaftGroup {
         }
     }
 
+    pub fn get_addr_only_string(&self, id: u64) -> String {
+        match self.peers.get(&id) {
+            None => format!("{}", id),
```
szbr00, line 60: A small operability concern with the new helper combined with the call site in `raft_node.rs`: in Raft, `leader_id == 0` is the conventional "no leader currently known" sentinel (during elections, startup, after a leader step-down). `self.peers.get(&0)` will miss, so `get_addr_only_string(0)` returns the string `"0"`. The log line then becomes:

    raft state change, current leader address 0, node 2,master-b:8996

The new wording "current leader address" promises an address but delivers the raw id, which to an operator reading logs looks like either "leader is at numeric address 0" or "the address field is missing/zeroed-out". The old phrasing was less precise but also less misleading, since it didn't claim to be an address.

Two small ways to keep the new phrasing honest:

1. Inline branch on id 0 in `get_addr_only_string`:

       pub fn get_addr_only_string(&self, id: u64) -> String {
           if id == 0 {
               return "unknown".to_string();
           }
           match self.peers.get(&id) {
               None => format!("{}", id),
               Some(v) => format!("{}:{}", v.hostname, v.port),
           }
       }

2. Or, at the call site, treat `self.leader() == 0` as a separate branch that logs `"current leader address unknown (electing)"`.

Either fix is one line and would prevent "current leader address 0" from showing up during normal election churn. Adding a quick test for `id == 0` would also lock the behavior in.

NolanHo: Good point. I updated `get_addr_only_string(0)` to return `unknown` so the SoftState log will not print `current leader address 0` during election/startup churn, and added a unit test covering the zero leader-id sentinel. Re-ran `cargo fmt --check`, `cargo test -p curvine-common raft_group -- --nocapture`, and `npm run commitlint`.

szbr00 (approved):

PR #839, opened 2026-05-25
feat: support SPDK containerization

thuongle2210:
# Description
-  This PR adds SPDK containerization support by integrating the SPDK initiator and target into the Dockerfile_rocky9 multi-stage build.
# Key Changes
- Restructured Dockerfile_rocky9 — added SPDK build stages (deps, build, runtime, target) so the Curvine worker can communicate with an NVMe-oF target over TCP/RDMA.
- Refactored orpc/build.rs — all SPDK and DPDK static libraries now use +whole-archive instead of plain static linking, preventing GNU ld's --as-needed from stripping cross-references that SPDK's internal dependency graph relies on.
- New SPDK compose stack — docker-compose-spdk target + entrypoint script for dev/test of the full SPDK NVMe-oF pipeline.
- Build system — added --spdk / --spdk-rdma / --spdk-dir flags to build/build.sh and corresponding Makefile targets.
# Impact
- Provides a foundation for SPDK-based deployment across dev, test, and production environments.
# Simulation

with cluster conf: 
[curvine-cluster_spdk_deployment.txt](https://github.com/user-attachments/files/28227438/curvine-cluster_spdk_deployment.txt)

NVME_PCI_ADDR: your vm pci address

```docker-compose building:

[vagrant@fedora37 curvine]$ sudo NVME_PCI_ADDR=0000:00:0e.0 CURVINE_CONF_FILE=/home/vagrant/curvine-cluster-config/curvine-cluster_spdk_deployment.toml make docker-compose-spdk

from host:
sudo -E ./target/release/curvine-bench --action fs.write --conf ../curvine-cluster-config/curvine-cluster_spdk_deployment.toml --dir /bench-large --file-num 5 --file-size 500MB --buf-size 4KB --client-threads 2
sudo -E ./target/release/curvine-bench --action fs.read --conf ../curvine-cluster-config/curvine-cluster_spdk_deployment.toml --dir /bench-large --file-num 5 --file-size 500MB --buf-size 4KB --client-threads 2

```

# Future works
- Optimize the Docker builder workflow, trade-off between Dockerfile management and build time
- Ubuntu-based Dockerfile support
- Kubernetes support (operator-managed SPDK target nodes) 
- Support dynamic parameters of spdk target


thuongle2210:
Dear @szbr9486, @lzjqsdd , could you please help me review this PR?

curvine-docker/deploy/spdk/entrypoint-target.sh
```diff
@@ -0,0 +1,173 @@
...
+case "$TRTYPE" in
+    tcp)
+        print_info "Creating TCP transport..."
+        "$RPC" nvmf_create_transport -t TCP -u 16384 -m 8 -c 8192
+        ;;
+    rdma)
+        print_info "Creating RDMA transport..."
+        "$RPC" nvmf_create_transport -t RDMA -u 16384 -m 8 -c 8192
+        ;;
+esac
```
szbr00, line 135: The case statement only handles `tcp` and `rdma`; any other value (typo, e.g. `TCP`, `RDMA`, `tls`) silently skips both transport creation and the matching listener block below, but the script still creates the subsystem, attaches the NVMe controller, and exits with success after `wait $NVMF_PID`. The target then runs without a transport/listener and any initiator connect fails with no clear root cause.

Please add a `*) print_error "Invalid TRTYPE=$TRTYPE (expected tcp|rdma)"; exit 1 ;;` arm to both `case "$TRTYPE"` blocks (or normalize the value once at the top via `${TRTYPE,,}` and validate) so misconfiguration fails fast.

curvine-docker/deploy/Dockerfile_rocky9
```diff
@@ -1,4 +1,125 @@
...
+ARG SPDK_VERSION=v25.09
+ARG NASM_VERSION=2.16.03
+ARG ISAL_VERSION=v2.31.1
+
+RUN --mount=type=cache,sharing=locked,target=/var/cache/dnf \
+    dnf install -y git && \
+    dnf clean all
+
+# Clone SPDK
+RUN git clone --branch "$SPDK_VERSION" https://github.com/spdk/spdk.git /spdk
```
szbr00: This `git clone` doesn't pass `--depth 1` even though `git submodule update` below does. For SPDK v25.09 that's hundreds of MB of history pulled on every cache miss of this layer, on every CI/dev rebuild. Suggest `git clone --depth 1 --branch "$SPDK_VERSION" --single-branch https://github.com/spdk/spdk.git /spdk` to match the submodule depth and significantly cut build time and image-layer size.

curvine-docker/deploy/Dockerfile_rocky9
```diff
@@ -54,8 +189,8 @@ RUN echo "Building Curvine from source..." && \
     # Build with retry mechanism
     for attempt in 1 2 3; do \
         echo "Build attempt $attempt..." && \
-        if make all ARGS='--skip-java-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs'; then \
-            echo "✓ Build succeeded with 'make all --skip-java-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs'" && \
+        if make all ARGS='--skip-java-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs --spdk-rdma'; then \
```
szbr00: The image build hardcodes `--spdk-rdma`, so anyone building this Dockerfile on a host without RDMA hardware (or who only wants the TCP transport that the rest of this PR explicitly supports via `--spdk`) has to patch the Dockerfile. `build.sh` already distinguishes `--spdk` from `--spdk-rdma`, so please surface this via a build `ARG` (e.g. `ARG SPDK_FLAG=--spdk-rdma`) and reference `${SPDK_FLAG}` here, so TCP-only builds don't need a fork of the Dockerfile.

szbr00 (commented):
In the `spdk-build` stage `cp -r ./* /opt/spdk/` copies the entire SPDK source tree (including DPDK sources, ISA-L sources, build intermediates, tests, docs) into `/opt/spdk`, and then `spdk-full` / `builder` both `COPY --from=spdk-build /opt/spdk /opt/spdk` that full tree into the runtime/build images. For SPDK v25.09 with RDMA this is on the order of multiple GB per image — most of which is unused at runtime.

For the runtime images, only the following are actually consumed: `build/bin/nvmf_tgt`, `build/lib/lib*.so*`, `dpdk/build/lib/lib*.so*`, `isa-l/.libs/libisal*`, `isa-l-crypto/.libs/libisal*`, and `scripts/rpc.py` (plus its python helpers). Consider running `make install DESTDIR=/opt/spdk-stage` or a targeted `cp` set in `spdk-build`, and `COPY --from=spdk-build /opt/spdk-stage /opt/spdk` into the runtime stages — that should shrink the target/runtime images substantially without changing functional behavior.

Also worth noting: the PR description references a `curvine-cluster_spdk_deployment.txt` config but it is not committed to the repo, and `docker-compose.yml` defaults `CURVINE_CONF_FILE` to `./curvine-cluster.toml` which doesn't exist in `curvine-docker/deploy/spdk/`. Adding a minimal example TOML next to `docker-compose.yml` would make the `make docker-compose-spdk` flow runnable end-to-end without out-of-band downloads.

curvine-docker/deploy/Dockerfile_rocky9
thuongle2210: Ok, but in this case, when I run with --spdk-rdma (which covers --spdk), the building time increase is negligible.

thuongle2210: I added it as the arg.

curvine-docker/deploy/Dockerfile_rocky9
thuongle2210: thanks for your notice

curvine-docker/deploy/spdk/entrypoint-target.sh
thuongle2210: I added validation steps for TRTYPE and NVME_PCI_ADDR

thuongle2210:
> In the `spdk-build` stage `cp -r ./* /opt/spdk/` copies the entire SPDK source tree (including DPDK sources, ISA-L sources, build intermediates, tests, docs) into `/opt/spdk`, and then `spdk-full` / `builder` both `COPY --from=spdk-build /opt/spdk /opt/spdk` that full tree into the runtime/build images. For SPDK v25.09 with RDMA this is on the order of multiple GB per image — most of which is unused at runtime.
> 
> For the runtime images, only the following are actually consumed: `build/bin/nvmf_tgt`, `build/lib/lib*.so*`, `dpdk/build/lib/lib*.so*`, `isa-l/.libs/libisal*`, `isa-l-crypto/.libs/libisal*`, and `scripts/rpc.py` (plus its python helpers). Consider running `make install DESTDIR=/opt/spdk-stage` or a targeted `cp` set in `spdk-build`, and `COPY --from=spdk-build /opt/spdk-stage /opt/spdk` into the runtime stages — that should shrink the target/runtime images substantially without changing functional behavior.
> 
> Also worth noting: the PR description references a `curvine-cluster_spdk_deployment.txt` config but it is not committed to the repo, and `docker-compose.yml` defaults `CURVINE_CONF_FILE` to `./curvine-cluster.toml` which doesn't exist in `curvine-docker/deploy/spdk/`. Adding a minimal example TOML next to `docker-compose.yml` would make the `make docker-compose-spdk` flow runnable end-to-end without out-of-band downloads.

Thanks for the discussion on saving storage and building time. I optimized the Dockerfile. I’ve been considering this as well. Supporting multiple SPDK versions in the future could become quite fragile and introduce management complexity (as I noted in the PR description).

Additionally, we can separate the Dockerfiles into different flows (with or without SPDK support) to optimize building time. I’ve also added docker-compose-spdk.toml as the default for now, though I will likely need to clean it up later (handle configuration assignment logic) to keep it minimal in a future update.

szbr00 (approved):

szbr00 (approved):

PR #846, opened 2026-05-27
fix: export renamed fs-mode files when ufs source is missing (#844)

### LINKED ISSUE #844: fs_mode auto-export misses files created via temporary file rename
Parent: #841

## Summary

In `fs_mode`, files written through a common temp-file pattern can remain `state=Cv, ufs_mtime=0` after being renamed to the final path. This reproduces the StarRocks behavior in #841: StarRocks writes `*.final.__TEMP__.` files, closes them, then renames them to the final `*.dat`, `*.meta`, or `SCHEMA_*` path. The final files are complete in Curvine metadata but are not exported to UFS, so UFS fallback has no object to read after CV cache/data loss.

This is separate from #842. #842 tracks full `cv://authority/...` mount lookup. The Java/Hadoop StarRocks write path normalizes to plain CV paths and still reproduces this issue.

## Environment

- Curvine server/client git in logs: `48d7395`
- Curvine source branch inspected locally: `main`
- StarRocks: `4.0.3`
- Mount: `/acc-hdfs/starrocks -> oss://<bucket>/starrocks`
- Mount mode: `fs_mode`
- Provider: `oss-hdfs`
- Hadoop client jar in StarRocks image: `curvine-hadoop-0.2.0.jar`

## Minimal reproduction

Using the same Java/Hadoop Curvine client that StarRocks uses:

1. Write a file directly to the final path and close it.
2. Write another file to a temporary path, close it, then rename it to the final path.

Pseudo-code:

```java
FileSystem fs = FileSystem.get(new URI("cv://curvine-pro/acc-hdfs/starrocks/starrocks-cluster"), conf);

Path direct = new Path(base + "/java-close-only.txt");
try (FSDataOutputStream out = fs.create(direct, true)) {
    out.write("close-only\n".getBytes("UTF-8"));
}

Path tmp = new Path(base + "/rename-probe.dat.<uuid>.__TEMP__.");
Path dst = new Path(base + "/rename-probe.dat");
try (FSDataOutputStream out = fs.create(tmp, true)) {
    out.write("rename-probe\n".getBytes("UTF-8"));
}
fs.rename(tmp, dst);
```

## Observed result

Direct close-only file becomes UFS-protected:

```text
Path: /acc-hdfs/starrocks/starrocks-cluster/java-flush-probe/java-close-only-1779847072342.txt
Is complete: true
Storage policy: StoragePolicy { storage_type: Disk, ttl_ms: 604800000, ttl_action: Delete, ufs_mtime: 1779847072901, state: Both }
```

The temp-renamed file stays CV-only:

```text
Path: /acc-hdfs/starrocks/starrocks-cluster/java-rename-probe/rename-probe-1779847169216.dat
Is complete: true
Len: 27
Storage policy: StoragePolicy { storage_type: Disk, ttl_ms: 604800000, ttl_action: Delete, ufs_mtime: 0, state: Cv }
Ufs Modification time: 1970-01-01 00:00:00
```

The temporary path no longer exists after rename:

```text
/acc-hdfs/starrocks/starrocks-cluster/java-rename-probe/rename-probe-1779847169216.dat.1779847169216.__TEMP__.
FileNotFound
```

This matches a fresh StarRocks probe table. StarRocks wrote temp paths such as:

```text
.../data/0000000000002ed8_4b31921c-214b-4c72-b107-252b8e99a693.dat.46eec043-4325-4541-bf73-2726cdd380a3.__TEMP__.
```

After rename, the final StarRocks files were complete but CV-only:

```text
.../data/0000000000002ed8_4b31921c-214b-4c72-b107-252b8e99a693.dat
Storage policy: StoragePolicy { storage_type: Disk, ttl_ms: 604800000, ttl_action: Delete, ufs_mtime: 0, state: Cv }
```

The same was true for `meta/0000000000000000_000000000000000*.meta` and `SCHEMA_*` files.

## Why this looks like a Curvine fs_mode bug

`UfsLoader::complete_file()` submits a load task for completed files under an fs_mode mount. That works for direct close-only writes.

But for temp-file rename patterns, the metadata changes in this order:

1. `CompleteFile(temp)` is journaled when the temp file is closed.
2. `Rename(temp, final)` is journaled immediately afterward.
3. The master metadata rename removes/moves the temp path before the async UFS loader has a stable temp source path to export.
4. When `UfsLoader::complete_file(temp)` runs, the temp path may no longer exist and is skipped as FileNotFound.
5. `UfsLoader::rename()` only renames the UFS object if the source object already exists. If the temp UFS object does not exist, it logs and returns without exporting the final path.
6. Final Curvine metadata remains `Cv` with `ufs_mtime=0`.

Relevant source paths:

- `curvine-server/src/master/journal/ufs_loader.rs`
  - `complete_file()` calls `submit_load_task(path)` for completed files.
  - `rename()` returns if `src_ufs_path` does not exist; it does not submit a load task for `dst` in that case.
- `curvine-server/src/master/meta/fs_dir.rs`
  - `complete_file()` journals `CompleteFile`.
  - `rename()` journals `Rename`.
- `curvine-server/src/master/journal/journal_loader.rs`
  - leader applies journal entries through `UfsLoader.apply_entry()`.

## Expected behavior

After `create(temp) -> close(temp) -> rename(temp, final)` under an `fs_mode` mount, the final path should end up in a UFS-protected state:

```text
state=Both
ufs_mtime > 0
```

Then `fs free` or CV cache/data loss should allow reads from UFS.

## Actual behavior

The final file remains:

```text
state=Cv
ufs_mtime=0
```

So fallback reads after Curvine cache/data loss cannot work.

## Suggested fixes

Possible approaches:

1. In `UfsLoader::rename()`, if `src_ufs_path` does not exist but the destination is a completed CV file under an fs_mode mount, submit a load task for `dst` instead of returning success.
2. Alternatively, when processing `Rename`, move/retarget any pending export from `src` to `dst`, so temp-file writers do not race the export path.
3. Add a regression test for `create temp -> complete -> rename final` under fs_mode and assert final file becomes `Both` with non-zero `ufs_mtime`.

This pattern is common for table engines and object-store style atomic commit, so fs_mode should handle it as a first-class write path.

jlon:
## Summary

- export the renamed destination path when fs_mode rename sees that the UFS source path is missing
- keep the change scoped to UfsLoader rename handling

## Root Cause

StarRocks writes segment files through a temporary path and then renames it to the final .dat path. The fs_mode auto-export path may try to export the temporary path after it has already been renamed. When UfsLoader handled the rename journal entry, it returned early if the UFS source path did not exist, leaving the final Curvine file without a UFS copy.

## Fix

When the UFS source path is missing during fs_mode rename handling, submit a load task for the destination path instead of silently skipping the entry. This exports the final visible path to UFS.

Fixes #844

## Validation

- cargo fmt
- cargo fmt --check
- git diff --check

### DIFF
--- a/curvine-server/src/master/journal/ufs_loader.rs
+++ b/curvine-server/src/master/journal/ufs_loader.rs
@@ -139,7 +139,10 @@ impl UfsLoader {
                     src_ufs_path,
                     dst.full_path()
                 );
-                self.submit_load_task(&dst, &mnt).await?;
+                if let Some((_, dst_mnt)) = self.get_mnt(&dst)? {
+                    // Keep journal apply aligned with fs_mode's durable export contract.
+                    self.submit_load_task(&dst, &dst_mnt).await?;
+                }
                 return Ok(());
             }
 
--- a/curvine-server/src/master/journal/ufs_loader.rs
+++ b/curvine-server/src/master/journal/ufs_loader.rs
@@ -134,7 +134,12 @@ impl UfsLoader {
         let dst = Path::from_str(&e.dst)?;
         if let Some((src_ufs_path, mnt)) = self.get_mnt(&src)? {
             if !mnt.ufs.exists(&src_ufs_path).await? {
-                warn!("rename: src file not exists: {}", src_ufs_path);
+                warn!(
+                    "rename: src file not exists: {}, loading dst {}",
+                    src_ufs_path,
+                    dst.full_path()
+                );
+                self.submit_load_task(&dst, &mnt).await?;
                 return Ok(());
             }
 


curvine-server/src/master/journal/ufs_loader.rs
```diff
@@ -134,7 +134,12 @@ impl UfsLoader {
         let dst = Path::from_str(&e.dst)?;
         if let Some((src_ufs_path, mnt)) = self.get_mnt(&src)? {
             if !mnt.ufs.exists(&src_ufs_path).await? {
-                warn!("rename: src file not exists: {}", src_ufs_path);
+                warn!(
+                    "rename: src file not exists: {}, loading dst {}",
+                    src_ufs_path,
+                    dst.full_path()
+                );
+                self.submit_load_task(&dst, &mnt).await?;
```
szbr00: `mnt` here is the mount that was resolved from `src` (via `get_mnt(&src)`). The new code reuses it as the load-task mount for `dst`, but `dst` may map to a completely different mount in multi-mount Curvine deployments (or to no fs_mode mount at all — e.g. the rename moved the file out of a UFS-backed subtree). In those cases the load task is submitted against the source mount's UFS, which is the wrong backing store.

Please look up the destination mount explicitly, e.g.:

```rust
if let Some((_, dst_mnt)) = self.get_mnt(&dst)? {
    self.submit_load_task(&dst, &dst_mnt).await?;
}
return Ok(());
```

That keeps the fix scoped (only loads when `dst` is actually in a fs_mode mount) and routes the load to the correct UFS. It also matches the implicit contract of every other `submit_load_task` callsite in this file, which uses the mount resolved from the path being loaded.

szbr00: Two behavioral changes worth being deliberate about here:

1. `submit_load_task` is synchronous on the happy path: it calls `wait_job_complete(self.copy_timeout)`, which defaults to 20 minutes. Calling it inline from `rename` means journal-apply for that entry now blocks for up to that timeout while the export runs. For the StarRocks workload described in the PR (many rapid renames), this can serialize and stall journal application behind individual exports.
2. The previous branch returned `Ok(())` unconditionally when the source was missing. The new branch now propagates any load error other than `FileNotFound` (via `err_box!` inside `submit_load_task`), which can turn a previously silently-tolerated state into a journal-apply failure (e.g. transient UFS hiccup while exporting `dst`).

Consider either submitting the load task in a fire-and-forget fashion (drop the `wait_job_complete` for this missing-src path) or downgrading non-fatal load errors to `warn!` + `Ok(())` here, so a single export failure does not block/break replay. If blocking is intentional, please document the trade-off (and the worst-case latency) in either the function doc or a comment near this call.

szbr00 (commented):
The PR fixes a specific, observable bug (#844 — StarRocks segment files not appearing on UFS after rename), but ships without a regression test. The hot path is small enough that a focused unit test on `UfsLoader::rename` would be cheap:

- mount a fake `MountValue` whose UFS reports `exists(src) == false`
- run `rename(&RenameEntry { src, dst, .. })`
- assert that `JobManager` saw a `submit_load_task` call for `dst` (and not for `src`)

Without it, this branch only gets exercised in production-shaped scenarios, and a future refactor of the `if !exists` arm could silently re-introduce #844. Even a single mock-based test would lock the fix in.

curvine-server/src/master/journal/ufs_loader.rs
jlon: Addressed in `465b742f`: the missing-source rename path now resolves the destination mount with `self.get_mnt(&dst)` before submitting the load task. If `dst` is outside fs_mode, no export is submitted; if it is under another fs_mode mount, the destination mount drives the export.

jlon: Kept the synchronous wait intentionally and documented it in the code comment. This path should preserve fs_mode durable-export semantics, matching the existing `CompleteFile` and non-rename-UFS paths that call `submit_load_task` synchronously. Downgrading load failures to warn-only would let the journal apply succeed while leaving the file in `Cv`-only state, which is the failure mode this PR is fixing.

jlon:
Addressed the actionable review feedback in `465b742f`.

On regression coverage: I checked the current test seams before adding code. `UfsLoader`, `JobManager`, and `MountValue` are concrete types today, and there is no existing fake/mock interface for observing `submit_load_task` without constructing the full job manager path. I kept this PR scoped to the behavior fix instead of introducing a new abstraction only for this test. The updated branch is covered by the existing CI checks and keeps the synchronous fs_mode export semantics intact.

szbr00 (approved):

PR #847, opened 2026-05-27
fix: resolve cv authority paths against mount table (#842)

### LINKED ISSUE #842: cv://authority paths do not match CV mount table entries
Parent: #841

## Summary

Curvine treats `cv://<authority>/path` as a Curvine path, but mount lookup currently builds mount candidates that include the `cv://<authority>` prefix. CV mount table entries are stored as plain paths such as `/acc-hdfs/starrocks`, so callers that keep the full `cv://authority/...` URI can fail to find an existing mount.

This is the root-cause-3 split-out from #841. It may not be the primary StarRocks Hadoop write path because the Java Hadoop SDK normalizes `cv://curvine-pro/acc-hdfs/...` to `/acc-hdfs/...` before entering native code, but it affects CLI repair/load and any caller that passes the full Curvine URI to mount-sensitive operations.

## Observed behavior

A mounted path exists:

```text
/acc-hdfs/starrocks -> oss://acc-hdfs/starrocks
```

Manual load using the full Curvine URI failed with a mount lookup error:

```text
curvine-cli load -w cv://curvine-pro/acc-hdfs/starrocks/...
Not found mount info for path: cv://curvine-pro/acc-hdfs/starrocks/...
```

Using the plain internal CV path shape is expected to match the mount:

```text
/acc-hdfs/starrocks/...
```

## Source analysis

`Path::from_str("cv://curvine-pro/acc-hdfs/starrocks/...")` keeps both the authority and the normalized path:

```text
scheme=cv
authority=curvine-pro
path=/acc-hdfs/starrocks/...
full_path=cv://curvine-pro/acc-hdfs/starrocks/...
```

`Path::is_cv()` returns true for `scheme=None` and `scheme=Some("cv")`, so this is treated as a Curvine path.

However, `Path::get_possible_mounts()` currently initializes mount candidates with `scheme://authority` whenever a scheme exists. For `cv://curvine-pro/acc-hdfs/starrocks/file`, candidates become similar to:

```text
cv://curvine-pro/
cv://curvine-pro/acc-hdfs
cv://curvine-pro/acc-hdfs/starrocks
```

The master mount table stores CV mount keys as plain CV paths, for example:

```text
/acc-hdfs/starrocks
```

So lookup misses the mount even though the path is under the mounted prefix.

Relevant files:

- `curvine-common/src/fs/path.rs`
- `curvine-client/src/unified/mount_cache.rs`
- `curvine-server/src/master/mount/mount_table.rs`
- `curvine-common/src/state/mount.rs`

## Expected behavior

For Curvine paths, these should resolve to the same CV mount:

```text
/acc-hdfs/starrocks/file.dat
cv://curvine-pro/acc-hdfs/starrocks/file.dat
```

The `cv://authority` part should identify the Curvine cluster/client authority, not become part of the CV mount key.

## Proposed fix

For mount candidate generation, treat `scheme=None` and `scheme=Some("cv")` the same:

```text
current_path = "/"
```

Only non-CV UFS schemes should preserve `scheme://authority` in mount candidates, for example:

```text
oss://bucket/...
s3://bucket/...
hdfs://cluster/...
file://...
```

Add regression coverage for:

1. `Path::get_possible_mounts("cv://curvine-pro/acc-hdfs/starrocks/a.dat")` includes `/acc-hdfs/starrocks`, not only `cv://curvine-pro/acc-hdfs/starrocks`.
2. `MountTable::get_mount_info()` finds `/acc-hdfs/starrocks` for both plain and `cv://authority` CV paths.
3. `MountCache::get_mount()` behaves the same in the client.
4. CLI or job path operations such as `load`, `toggle_path`, and `wait_job_complete` work with `cv://authority/...`.

## Impact

Without this fix, operational repair commands and native clients can fail to locate an existing fs_mode mount when they pass the full Curvine URI. In the StarRocks incident tracked by #841, this can block manual resync/export repair even when the mount itself is configured correctly.

jlon:
## Summary

- treat cv://authority/... paths as Curvine internal paths when generating mount lookup candidates
- keep non-Curvine schemes such as s3://, file://, oss://, and hdfs:// on the existing scheme://authority candidate path
- add regression coverage for cv://authority mount candidate generation and CV-to-UFS conversion

## Root Cause

Curvine stores CV mount table keys as plain internal paths, for example /acc-hdfs/starrocks. However, Path::get_possible_mounts() generated candidates such as cv://curvine-pro/acc-hdfs/starrocks for cv://authority/... paths. That made mount lookup miss an existing CV mount when callers passed a full Curvine URI.

## Fix

When a path is a Curvine path, including both /path and cv://authority/path forms, mount candidates now start from /. The authority still remains available on the parsed Path, but it is not used as part of the CV mount key.

The non-CV branch now explicitly assumes a non-CV scheme, so schemeless CV paths cannot appear in that branch as dead code.

## Mount Lookup Audit

I checked the other mount-key paths called out in review:

- server-side MountManager/MountTable routes lookups through Path::get_possible_mounts() and then chooses mountpath2id for CV paths, so cv://authority/... is covered by this fix.
- client-side MountCache also iterates Path::get_possible_mounts() and indexes cv_map only for CV paths; CV cache removal uses path.path(), not the raw URI.
- UfsLoader and JobManager::get_mnt() route through MountManager::get_mount_info(), so they inherit the same normalized CV mount lookup.
- curvine-server/src/common/ufs_manager.rs caches by UFS base URI and asks the master for mount info using a UFS Path; it does not derive CV mount keys from cv://authority/... inputs.

Fixes #842

## Validation

- cargo test -p curvine-common fs::path::tests::get_possible_mounts -- --nocapture
- cargo test -p curvine-common state::mount::tests::test_path_cst -- --nocapture
- cargo test -p curvine-common
- cargo fmt
- cargo fmt --check
- git diff --check

### DIFF
--- a/curvine-common/src/fs/path.rs
+++ b/curvine-common/src/fs/path.rs
@@ -210,28 +210,32 @@ impl Path {
             return result;
         }
 
-        let mut current_path = match self.scheme() {
-            Some(FsKind::SCHEME_FILE) => {
-                let sp = if self.path.starts_with(Self::SEPARATOR) {
-                    Self::SEPARATOR
-                } else {
-                    ""
-                };
-                format!(
-                    "{}{}{}{}",
-                    FsKind::SCHEME_FILE,
+        let mut current_path = if self.is_cv() {
+            Self::SEPARATOR.to_string()
+        } else {
+            match self.scheme() {
+                Some(FsKind::SCHEME_FILE) => {
+                    let sp = if self.path.starts_with(Self::SEPARATOR) {
+                        Self::SEPARATOR
+                    } else {
+                        ""
+                    };
+                    format!(
+                        "{}{}{}{}",
+                        FsKind::SCHEME_FILE,
+                        Self::SCHEME_DELIMITER,
+                        sp,
+                        self.authority().unwrap_or("")
+                    )
+                }
+                Some(v) => format!(
+                    "{}{}{}",
+                    v,
                     Self::SCHEME_DELIMITER,
-                    sp,
                     self.authority().unwrap_or("")
-                )
+                ),
+                None => Self::SEPARATOR.to_string(),
             }
-            Some(v) => format!(
-                "{}{}{}",
-                v,
-                Self::SCHEME_DELIMITER,
-                self.authority().unwrap_or("")
-            ),
-            None => Self::SEPARATOR.to_string(),
         };
 
         for part in parts {
@@ -370,6 +374,11 @@ mod tests {
     #[test]
     fn get_possible_mounts() -> CommonResult<()> {
         let p1 = Path::from_str("/a/b/c")?;
+        let mnts = p1.get_possible_mounts();
+        println!("{:?}", mnts);
+        assert_eq!(mnts, Vec::from(["/", "/a", "/a/b", "/a/b/c"]));
+
+        let p1 = Path::from_str("cv://curvine-pro/a/b/c")?;
         let mnts = p1.get_possible_mounts();
         println!("{:?}", mnts);
         assert_eq!(mnts, Vec::from(["/", "/a", "/a/b", "/a/b/c"]));
--- a/curvine-common/src/fs/path.rs
+++ b/curvine-common/src/fs/path.rs
@@ -213,8 +213,11 @@ impl Path {
         let mut current_path = if self.is_cv() {
             Self::SEPARATOR.to_string()
         } else {
-            match self.scheme() {
-                Some(FsKind::SCHEME_FILE) => {
+            match self
+                .scheme()
+                .expect("non-CV paths must have a non-CV scheme")
+            {
+                FsKind::SCHEME_FILE => {
                     let sp = if self.path.starts_with(Self::SEPARATOR) {
                         Self::SEPARATOR
                     } else {
@@ -228,13 +231,12 @@ impl Path {
                         self.authority().unwrap_or("")
                     )
                 }
-                Some(v) => format!(
+                v => format!(
                     "{}{}{}",
                     v,
                     Self::SCHEME_DELIMITER,
                     self.authority().unwrap_or("")
                 ),
-                None => Self::SEPARATOR.to_string(),
             }
         };
 
--- a/curvine-common/src/state/mount.rs
+++ b/curvine-common/src/state/mount.rs
@@ -435,6 +435,12 @@ mod tests {
             info.get_ufs_path(&path).unwrap().full_path(),
             "s3://spark/test1/test/dt=2025/1.csv"
         );
+
+        let path = Path::from_str("cv://curvine-pro/spark/test1/test/dt=2025/1.csv").unwrap();
+        assert_eq!(
+            info.get_ufs_path(&path).unwrap().full_path(),
+            "s3://spark/test1/test/dt=2025/1.csv"
+        );
     }
 
     #[test]


curvine-common/src/fs/path.rs
```diff
@@ -210,28 +210,32 @@ impl Path {
...
+                }
+                Some(v) => format!(
+                    "{}{}{}",
+                    v,
                     Self::SCHEME_DELIMITER,
-                    sp,
                     self.authority().unwrap_or("")
-                )
+                ),
+                None => Self::SEPARATOR.to_string(),
```
szbr00: After the new `if self.is_cv() { /* "/" */ }` guard at the top, the `None => Self::SEPARATOR.to_string()` arm of this inner `match` is no longer reachable: `is_cv()` returns `true` for `matches!(self.scheme(), None | Some("cv"))`, so any path with `self.scheme() == None` is already handled by the outer `if` branch and the `else` body only runs for `Some(v) where v != "cv"`.

Either delete the `None` arm and let the match be exhaustive on `Some(_)` (using `Some(FsKind::SCHEME_FILE)` + `Some(v)`), or replace it with `None => unreachable!("Path::is_cv() should have caught scheme = None")`. As written, the arm is dead code and slightly misleads readers into thinking schemeless paths can still reach here.

szbr00 (commented):
The root cause described in the PR — "mount table stores CV keys as `/path`, but `cv://authority/...` URIs were being rendered with the `cv://authority` prefix in mount lookups" — is fixed correctly for `get_possible_mounts()`, and the new regression test in `mount.rs` confirms `get_ufs_path` already does the right thing via `path.path()`.

Worth a quick audit of the other places that derive mount keys / cv paths from a `Path`, to make sure none of them still take the URI-style form for `cv://authority/...` inputs. Likely candidates worth a quick grep:

- `curvine-server/src/master/mount/mount_manager.rs` / `mount_table.rs` (any direct `path.full_path()` or `path.clone_uri()` keyed lookups)
- `curvine-client/src/unified/mount_cache.rs` (client-side resolution path)
- `curvine-server/src/common/ufs_manager.rs`
- `curvine-server/src/master/journal/ufs_loader.rs` `get_mnt(&path)` callers, in case any of them are passed `cv://authority/...` from user-facing entry points (CLI, FUSE, master_handler)

A short note in the PR description (or a follow-up issue) confirming those paths route through `Path::path()` / `Path::is_cv()`-aware code rather than the raw URI would close the loop on #842 fully.

curvine-common/src/fs/path.rs
jlon: Addressed in `452d9b37`: removed the unreachable `None` match arm and made the non-CV branch explicitly require a non-CV scheme with `expect("non-CV paths must have a non-CV scheme")`.

jlon:
Audit complete.

- The unreachable `None` arm in `Path::get_possible_mounts()` was removed in `452d9b37`; the non-CV branch now explicitly requires a non-CV scheme.
- Server-side `MountManager` / `MountTable` lookups route through `Path::get_possible_mounts()` and then use `mountpath2id` for CV paths, so `cv://authority/...` is normalized to `/path` before the mount-table lookup.
- Client-side `MountCache` uses the same mount-candidate generation and indexes `cv_map` only with CV paths; CV cache removal uses `path.path()`, not the raw URI.
- `UfsLoader` and `JobManager::get_mnt()` route through `MountManager::get_mount_info()`, so they inherit the same normalized lookup and then convert with `MountInfo` / `MountValue` path helpers.
- `curvine-server/src/common/ufs_manager.rs` is keyed by UFS base URI and asks the master for mount info using a UFS `Path`; it does not derive CV mount keys from `cv://authority/...` inputs.

No additional code change is needed from that audit. The PR description has the audit note. Focused validation run locally:

- `rtk cargo test -p curvine-common get_possible_mounts -- --nocapture`
- `rtk cargo test -p curvine-common test_path_cst -- --nocapture`

lzjqsdd (approved):

PR #853, opened 2026-05-29
fix(fuse): restore mmap/read coherence by not forcing direct_io on ca…

### LINKED ISSUE #850: [BUG]:`mmap(MAP_SHARED)` writes are not visible to `pread()` / `read()` — Curvine page cache is not shared between the mmap path and the buffered IO path
**Describe the bug**

On a Curvine FUSE mount, modifications made through a `MAP_SHARED` mmap region are **not visible to subsequent `pread(2)` / `read(2)` calls** on the same file descriptor — at least not deterministically. POSIX `mmap(2)` requires that:

> Modifications made through the mapping shall be visible to subsequent `read()` calls.

When multiple processes (or a single process with many concurrent operations) interleave `mmap`-writes with `pread` / `read` on the same file, Curvine returns stale page contents through the buffered IO path: bytes that have already been written through the mmap region appear as `0x00` (the original initialized value) when read back via `pread`. The dirty page produced by the `MAP_SHARED` write does not reach the page cache that `pread` consults.

The conjectured root cause is one of:

- The mmap dirty-page writeback path (`fuse_writepage` or its async equivalent) does not commit dirty mmap pages into the same page cache that the buffered IO path serves — so `pread` reads the original page while the mmap-modified page sits elsewhere
- mmap'd pages are maintained in a separate per-inode mapping or cache layer that is not unified with the buffered read/write path
- The daemon-side response to `FUSE_READ` issued on behalf of `pread` does not see the dirty mmap page because the page is owned exclusively by the kernel mmap path until an `msync` / `fsync` flushes it

A successful fix should ensure that any byte written through `MAP_SHARED` is observable via `read`/`pread` immediately, without requiring an explicit `msync(2)` from user space (POSIX does not require user-space `msync` for `MAP_SHARED → read` coherence; the kernel and FUSE daemon together own that contract).

**To Reproduce**

1. Mount Curvine FUSE at `/curvine-fuse` and create a working directory:

   ```bash
   sudo mkdir -p /curvine-fuse/repro
   ```

2. Save the following minimal reproducer as `repro_mmap_pread.c`. It runs 4 forked workers, each performing 100 rounds of `mmap`-write + immediate `pread`-verify on the same shared file. Any byte mismatch fails the corresponding worker.

   ```c
   /* repro_mmap_pread.c — reproduces Curvine MAP_SHARED ↔ pread incoherence
    *
    * Build:  gcc -O0 -Wall repro_mmap_pread.c -o /tmp/repro_mmap_pread
    * Run:    /tmp/repro_mmap_pread
    *
    * Expected on a POSIX-compliant FS (ext4/xfs/btrfs/tmpfs/...):
    *     PASS: mmap and read share the same page cache
    *
    * Observed on Curvine:
    *     worker=0 round=58 off=237568 byte 0: mmap-wrote='A' pread-got=0x00
    *     FAIL: 1/4 workers reported mmap/read inconsistency
    */
   #define _GNU_SOURCE
   #include <stdio.h>
   #include <stdlib.h>
   #include <string.h>
   #include <fcntl.h>
   #include <unistd.h>
   #include <sys/mman.h>
   #include <sys/stat.h>
   #include <sys/wait.h>
   #include <signal.h>

   #define FILE_PATH "/curvine-fuse/repro/mmap_shared.bin"
   #define SZ        (1 << 20)        /* 1 MiB */
   #define WORKERS   4
   #define ROUNDS    100

   static volatile sig_atomic_t got_bus = 0;
   static void on_bus(int s) { got_bus = 1; _exit(2); }

   int main(void) {
       /* Ensure a fresh, fully-allocated 1 MiB file. */
       unlink(FILE_PATH);
       int fd = open(FILE_PATH, O_RDWR | O_CREAT | O_TRUNC, 0644);
       if (ftruncate(fd, SZ) < 0) { perror("ftruncate"); return 1; }
       char zero[4096] = {0};
       for (size_t off = 0; off < SZ; off += 4096)
           pwrite(fd, zero, 4096, off);
       fsync(fd);
       close(fd);

       int fail = 0;
       for (int w = 0; w < WORKERS; w++) {
           if (fork() == 0) {
               signal(SIGBUS, on_bus);
               int fd = open(FILE_PATH, O_RDWR);
               if (fd < 0) { perror("worker open"); _exit(3); }

               char *m = mmap(NULL, SZ, PROT_READ | PROT_WRITE,
                              MAP_SHARED, fd, 0);
               if (m == MAP_FAILED) { perror("mmap"); _exit(3); }

               unsigned char marker = 'A' + w;
               for (int r = 0; r < ROUNDS; r++) {
                   /* Write a 4 KiB region through the mmap mapping. */
                   size_t off = (w * SZ / WORKERS) +
                                (r * 4096) % (SZ / WORKERS);
                   memset(m + off, marker, 4096);

                   /* Immediately pread the same region —
                    * POSIX requires it to equal what we just wrote. */
                   char rbuf[4096];
                   ssize_t n = pread(fd, rbuf, 4096, off);
                   if (n != 4096) {
                       fprintf(stderr,
                           "worker=%d round=%d off=%zu pread short n=%zd\n",
                           w, r, off, n);
                       munmap(m, SZ); close(fd); _exit(2);
                   }
                   for (int i = 0; i < 4096; i++) {
                       if (rbuf[i] != marker) {
                           fprintf(stderr,
                               "worker=%d round=%d off=%zu byte %d: "
                               "mmap-wrote='%c' pread-got=0x%02x\n",
                               w, r, off, i, marker, rbuf[i] & 0xff);
                           munmap(m, SZ); close(fd); _exit(2);
                       }
                   }
               }
               munmap(m, SZ); close(fd); _exit(0);
           }
       }

       for (int w = 0; w < WORKERS; w++) {
           int st;
           wait(&st);
           if (!WIFEXITED(st) || WEXITSTATUS(st) != 0) {
               fprintf(stderr, "worker exit status=%d\n", WEXITSTATUS(st));
               fail++;
           }
       }

       unlink(FILE_PATH);

       if (fail == 0)
           puts("PASS: mmap and read share the same page cache");
       else
           printf("FAIL: %d/%d workers reported mmap/read inconsistency\n",
                  fail, WORKERS);
       return fail ? 2 : 0;
   }
   ```

3. Compile and run:

   ```bash
   gcc -O0 -Wall repro_mmap_pread.c -o /tmp/repro_mmap_pread
   /tmp/repro_mmap_pread
   ```

   Because the inconsistency is timing-sensitive, run several iterations to confirm it is reproducible:

   ```bash
   for i in 1 2 3 4 5; do /tmp/repro_mmap_pread; done
   ```

**Observed output on Curvine:**

```
$ /tmp/repro_mmap_pread
worker=0 round=58 off=237568 byte 0: mmap-wrote='A' pread-got=0x00
worker exit status=2
FAIL: 1/4 workers reported mmap/read inconsistency
```

The exact worker / round / offset varies between runs because the bug is a race between the kernel's mmap dirty-page tracking and the FUSE_READ path that satisfies `pread`. Even a single failing worker out of four is sufficient evidence that the POSIX `MAP_SHARED ↔ read` coherence contract is violated.

**Expected behavior**

POSIX `mmap(2)` (IEEE Std 1003.1-2017) specifies the memory–file coherence contract for `MAP_SHARED` mappings:

> If `MAP_SHARED` is specified, write references shall change the underlying object. **Read references to memory regions to which `MAP_SHARED` mappings have been established shall return the values that have been written through these mappings since the last call to `msync()` (or implicitly, the last write through any mapping or descriptor that references the same object).**

Concretely, `mmap`-write followed by `pread` on the same file descriptor must observe the written byte without requiring any explicit synchronization. Therefore the expected output of the reproducer on every well-behaved POSIX filesystem is:

```
$ /tmp/repro_mmap_pread
PASS: mmap and read share the same page cache
```

This expectation is satisfied by `ext4`, `xfs`, `btrfs`, `tmpfs`, `nfs` (with `cto` enabled), and FUSE filesystems that implement page cache correctly — including upstream `libfuse` examples and `juicefs`-aligned implementations.

**OS Version :**

Architecture: x86_64
glibc: 2.28
GCC: 8.3.0

bigbigxu:
## Summary

Fixes [#850](https://github.com/CurvineIO/curvine/issues/850): `MAP_SHARED` mmap writes were not consistently visible to subsequent `pread()` / `read()` on the same file.

The root cause was in the FUSE `open` handler. When local metadata indicated the file had changed (`keep_cache == false`), the daemon responded with `FUSE_FOPEN_DIRECT_IO`. That forced the buffered I/O path to bypass the kernel page cache, while `MAP_SHARED` mmap writes still updated cached pages. The two paths were no longer unified, violating the POSIX coherence contract between mmap and read.

This PR removes the `FUSE_FOPEN_DIRECT_IO` fallback on cache miss. When the cache should be invalidated, we simply omit `FUSE_FOPEN_KEEP_CACHE` and let the kernel discard stale pages on open. We also document why synchronous `FUSE_NOTIFY_INVAL_INODE` must not be sent from the open path (kernel deadlock risk).

## Changes

- **`curvine-fuse/src/fs/curvine_file_system.rs`**
  - Stop setting `FUSE_FOPEN_DIRECT_IO` when `keep_cache` is false.
  - Rely on omitting `FUSE_FOPEN_KEEP_CACHE` for kernel-side page cache invalidation.
  - Add comments explaining deadlock and consistency trade-offs.

- **`build/tests/scripts/mmap_test.py`**
  - Add a basic mmap write + read-back regression test (mmap and file handle).

- **`build/tests/fuse-test.sh`**
  - Wire mmap test into the FUSE test suite (Test 17).
  - Introduce `get_python_cmd()` and `run_python_script_test()` to deduplicate Python test execution.

## Root Cause

| Path | Before (cache miss) | After |
|------|---------------------|-------|
| `mmap(MAP_SHARED)` write | Kernel page cache | Kernel page cache |
| `pread()` / `read()` | Direct I/O (no page cache) | Buffered I/O (page cache) |

Because mmap and buffered reads no longer shared the same cache layer, bytes written via mmap could appear as stale zeros when read back via `pread`.

## Test Plan

- [ ] Run the issue #850 reproducer on a Curvine FUSE mount:
  ```bash
  gcc -O0 -Wall repro_mmap_pread.c -o /tmp/repro_mmap_pread
  for i in 1 2 3 4 5; do /tmp/repro_mmap_pread; done

### DIFF
--- a/build/tests/fuse-test.sh
+++ b/build/tests/fuse-test.sh
@@ -59,6 +59,7 @@ This script tests basic filesystem operations including:
   - File permissions (chmod, chown, chgrp)
   - Sed in-place editing
   - Pwrite visibility (file size and content after pwrite)
+  - Mmap read/write (MAP_SHARED mmap ↔ pread coherence, issue #850)
 
 For FIO performance tests, use fio-test.sh instead.
 
@@ -181,6 +182,76 @@ init_test_env() {
     fi
     
     print_info "Test environment initialized successfully"
+}
+
+# Resolve python3/python executable path.
+# Prints the command on success; returns 1 if Python is unavailable.
+get_python_cmd() {
+    if command -v python3 >/dev/null 2>&1; then
+        echo "python3"
+        return 0
+    fi
+    if command -v python >/dev/null 2>&1; then
+        echo "python"
+        return 0
+    fi
+    return 1
+}
+
+# Run a Python test script from build/tests/scripts/.
+#
+# Usage:
+#   run_python_script_test "Test description" script.py [script args...]
+#
+# Returns:
+#   0 - passed or skipped (Python or script unavailable)
+#   1 - failed
+run_python_script_test() {
+    local test_name="$1"
+    local script_name="$2"
+    shift 2
+
+    local python_cmd
+    if ! python_cmd=$(get_python_cmd); then
+        print_info "Python not available, skipping $test_name"
+        return 0
+    fi
+
+    local test_script="$SCRIPT_DIR/scripts/$script_name"
+    if [ ! -f "$test_script" ]; then
+        print_info "Test script not found: $test_script"
+        print_info "Skipping $test_name"
+        return 0
+    fi
+
+    print_test "$test_name"
+    TOTAL_TESTS=$((TOTAL_TESTS + 1))
+
+    local start_time end_time elapsed result=0
+    start_time=$(date +%s)
+
+    local cmd="$python_cmd $test_script"
+    if [ $# -gt 0 ]; then
+        cmd="$cmd $*"
+    fi
+    print_command "$cmd"
+
+    if $python_cmd "$test_script" "$@" 2>&1; then
+        result=0
+    else
+        result=$?
+    fi
+
+    end_time=$(date +%s)
+    elapsed=$((end_time - start_time))
+
+    if [ $result -eq 0 ]; then
+        print_success "$test_name completed successfully in ${elapsed}s"
+        return 0
+    fi
+
+    handle_error "$test_name failed" "$cmd"
+    return 1
 }
 
 # Test 1: Basic file operations
@@ -903,15 +974,10 @@ test_file_locks() {
     fi
 
     # Test 4: POSIX lock (fcntl) - using Python if available
-    if command -v python3 >/dev/null 2>&1 || command -v python >/dev/null 2>&1; then
+    local python_cmd
+    if python_cmd=$(get_python_cmd); then
         print_test "Testing POSIX lock (fcntl) with Python"
         TOTAL_TESTS=$((TOTAL_TESTS + 1))
-        local python_cmd=""
-        if command -v python3 >/dev/null 2>&1; then
-            python_cmd="python3"
-        else
-            python_cmd="python"
-        fi
 
         local lock_script_file=$(mktemp)
         cat > "$lock_script_file" << 'PYTHON_EOF'
@@ -1110,12 +1176,11 @@ test_python_high_frequency_write() {
     CURRENT_TEST_GROUP="Test 14: Python High-Frequency Write"
     print_header "$CURRENT_TEST_GROUP"
 
-    if ! command -v python3 >/dev/null 2>&1 && ! command -v python >/dev/null 2>&1; then
+    local python_cmd
+    if ! python_cmd=$(get_python_cmd); then
         print_info "Python not available, skipping high-frequency write test"
         return
     fi
-
-    local python_cmd=$(command -v python3 2>/dev/null || command -v python)
     local test_file="$TEST_DIR/python-high-freq.txt"
     local iterations=20
     
@@ -1209,85 +1274,7 @@ PYTHON_SCRIPT
 test_fuse_reload() {
     CURRENT_TEST_GROUP="Test 15: FUSE Hot Reload"
     print_header "$CURRENT_TEST_GROUP"
-
-    if ! command -v python3 >/dev/null 2>&1 && ! command -v python >/dev/null 2>&1; then
-        print_info "Python not available, skipping FUSE reload test"
-        return
-    fi
-
-    local python_cmd=$(command -v python3 2>/dev/null || command -v python)
-    local test_script="$SCRIPT_DIR/scripts/fuse_reload_test.py"
-    
-    if [ ! -f "$test_script" ]; then
-        print_info "FUSE reload test script not found: $test_script"
-        print_info "Skipping FUSE reload test"
-        return
-    fi
-
-    print_test "Testing FUSE hot reload functionality"
-    TOTAL_TESTS=$((TOTAL_TESTS + 1))
-
-    local start_time=$(date +%s)
-    
-    # Run the Python test script
-    if $python_cmd "$test_script" 2>&1; then
-        local result=0
-    else
-        local result=$?
-    fi
-    
-    local end_time=$(date +%s)
-    local elapsed=$((end_time - start_time))
-
-    if [ $result -eq 0 ]; then
-        print_success "FUSE reload test completed successfully in ${elapsed}s"
-    else
-        handle_error "FUSE reload test failed" "python $test_script"
-    fi
-}
-
-# Test 17: Pwrite visibility test
-test_pwrite_visibility() {
-    CURRENT_TEST_GROUP="Test 17: Pwrite Visibility"
-    print_header "$CURRENT_TEST_GROUP"
-
-    if ! command -v python3 >/dev/null 2>&1 && ! command -v python >/dev/null 2>&1; then
-        print_info "Python not available, skipping pwrite visibility test"
-        return
-    fi
-
-    local python_cmd
-    python_cmd=$(command -v python3 2>/dev/null || command -v python)
-    local test_script="$SCRIPT_DIR/scripts/visibility_test.py"
-
-    if [ ! -f "$test_script" ]; then
-        print_info "Pwrite visibility test script not found: $test_script"
-        print_info "Skipping pwrite visibility test"
-        return
-    fi
-
-    print_test "Testing file size and content visibility after pwrite"
-    TOTAL_TESTS=$((TOTAL_TESTS + 1))
-
-    local start_time
-    start_time=$(date +%s)
-    local result=0
-
-    if $python_cmd "$test_script" --dir "$TEST_DIR" 2>&1; then
-        result=0
-    else
-        result=$?
-    fi
-
-    local end_time
-    end_time=$(date +%s)
-    local elapsed=$((end_time - start_time))
-
-    if [ $result -eq 0 ]; then
-        print_success "Pwrite visibility test completed successfully in ${elapsed}s"
-    else
-        handle_error "Pwrite visibility test failed" "python $test_script --dir $TEST_DIR"
-    fi
+    run_python_script_test "Testing FUSE hot reload functionality" "fuse_reload_test.py"
 }
 
 # Test 16: Git clone operations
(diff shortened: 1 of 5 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1124,9 +1124,13 @@ impl fs::FileSystem for CurvineFileSystem {
...
                 open_flags |= FUSE_FOPEN_KEEP_CACHE;
-            } else {
-                open_flags |= FUSE_FOPEN_DIRECT_IO;
             }
+            // When keep_cache is false we omit FUSE_FOPEN_KEEP_CACHE so the kernel
+            // drops the page cache on open; do not send FUSE_NOTIFY_INVAL_INODE here.
+            // Synchronous invalidation during open can deadlock: the kernel may hold
+            // page locks on this inode while userspace waits for the notify to finish.
+            // Remote updates may still appear stale until attr cache expires (default 1s);
+            // use attr_ttl=0 or direct_io for stronger consistency.
```
szbr00: The inline comment correctly captures the trade-off — "Remote updates may still appear stale until attr cache expires (default 1s); use attr_ttl=0 or direct_io for stronger consistency" — but in the previous behavior, a `keep_cache == false` open dropped the entire buffered read path onto DIRECT_IO, so concurrent writes from a non-FUSE Curvine client (or a peer FUSE daemon) were immediately visible on the next `read()`. After this change, those updates are buffered in the kernel page cache and won't be visible until the dentry/attr cache TTL expires and the kernel reloads.

That is a real behavior regression for workloads that explicitly relied on the strict-read mode (e.g. log tailers, multi-writer pipelines, integration tests). A few concrete asks:

1. Call this out in `CHANGELOG` / the FUSE config docs so users who depended on the old behavior aren't surprised.
2. Consider adding a `FuseConf` knob (e.g. `direct_io_on_cache_miss = false` by default, `true` to restore the old behavior) so the choice is configurable rather than baked in.
3. At minimum, log an `info!` (or `warn!` on first occurrence per inode) when `keep_cache == false` so operators have something to grep for when investigating staleness.

build/tests/scripts/mmap_test.py
```diff
@@ -0,0 +1,90 @@
...
+            print("FAIL: mmap read-back data mismatch")
+            print(f"  expected: {TEST_DATA!r}")
+            print(f"  actual:   {read_via_mmap!r}")
+            mm.close()
+            return 1
+        print("PASS: mmap read-back verification succeeded")
+
+        # Read back via the file handle to confirm persistence.
+        f.seek(0)
+        read_via_file = f.read(len(TEST_DATA))
```
szbr00: This test reads back the data through both `mm` and `f` *while still inside the same `open(..., "r+b")`*, so the read is being served by the kernel page cache that the mmap write just dirtied — i.e. it passes regardless of whether the FUSE write actually reached the Curvine master. Two concrete gaps:

1. **The broken code path is not deterministically exercised.** The PR fixes the case where `should_keep_cache` returns `false` and `FUSE_FOPEN_DIRECT_IO` used to be set. In this test the file is created and immediately re-opened in the same process, so the daemon's cached `(mtime, len)` for the inode will match the master right away → `keep_cache == true` → the previous (buggy) branch is never taken. To make this a real regression test for #850, force the cache-miss path, e.g.:
    - touch the file with a different mtime (or modify it from another process) before the second `open`, OR
    - sleep > `attr_ttl` after the first close and re-stat, OR
    - just ship the C reproducer (`repro_mmap_pread.c`) from the issue and invoke it from `test_mmap`.

2. **No post-close persistence check.** Add a step that closes the file after `mm.flush()`, reopens, and re-reads — that confirms the write actually reached the Curvine master, not just the kernel page cache that mmap and `f.read` share within one open. Without this, the test passes even on a daemon that drops writes on close.

build/tests/fuse-test.sh
```diff
@@ -1209,85 +1273,21 @@ PYTHON_SCRIPT
...
-    end_time=$(date +%s)
-    local elapsed=$((end_time - start_time))
-
-    if [ $result -eq 0 ]; then
-        print_success "Pwrite visibility test completed successfully in ${elapsed}s"
-    else
-        handle_error "Pwrite visibility test failed" "python $test_script --dir $TEST_DIR"
-    fi
+    run_python_script_test "Testing file size and content visibility after pwrite" "visibility_test.py" --dir "$TEST_DIR"
 }
```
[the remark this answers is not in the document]
bigbigxu: fixed

build/tests/fuse-test.sh
```diff
@@ -1110,12 +1176,10 @@ test_python_high_frequency_write() {
     CURRENT_TEST_GROUP="Test 14: Python High-Frequency Write"
     print_header "$CURRENT_TEST_GROUP"
 
-    if ! command -v python3 >/dev/null 2>&1 && ! command -v python >/dev/null 2>&1; then
+    if ! python_cmd=$(get_python_cmd); then
```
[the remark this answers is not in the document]
bigbigxu: fixed

build/tests/fuse-test.sh
```diff
@@ -183,6 +184,77 @@ init_test_env() {
...
+        print_info "Python not available, skipping $test_name"
+        return 2
+    fi
+
+    local test_script="$SCRIPT_DIR/scripts/$script_name"
+    if [ ! -f "$test_script" ]; then
+        print_info "Test script not found: $test_script"
+        print_info "Skipping $test_name"
+        return 2
+    fi
```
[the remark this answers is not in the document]
bigbigxu: fixed

curvine-fuse/src/fs/curvine_file_system.rs
bigbigxu: fixed

build/tests/scripts/mmap_test.py
bigbigxu: fixed

szbr00 (approved):

PR #854, opened 2026-05-29
fix(fuse,server): enforce POSIX rename semantics

### LINKED ISSUE #845: [BUG]: rename(2) silently relocates src into non-empty dst as a subdirectory — POSIX requires ENOTEMPTY but rename() returns success
**Describe the bug**

When `rename(2)` is called on Curvine FUSE with a target path that already exists as a **non-empty directory**, POSIX requires the call to fail with `errno = ENOTEMPTY` (or `EEXIST`). Instead, Curvine's rename handler returns success (`rc=0, errno=0`) and silently transforms the operation:

- The user's intent: `rename(src, dst)` — replace dst with src
- POSIX behavior: refuse with `ENOTEMPTY`, neither operand is modified
- Curvine actual behavior: silently rewrites the call as if it were `rename(src, dst/src)`, relocating the source directory **as a subdirectory inside the existing target**

The original contents of dst are preserved (no data loss), but the post-rename directory tree is structurally different from what POSIX guarantees, and is also different from what every other major filesystem (ext4, xfs, btrfs, tmpfs, NFS, CIFS) produces for the same call.

This is the userland `mv src dst` command's "convenience behavior" being incorrectly applied at the syscall level. The userland `mv` tool, when it detects that dst is an existing directory, decides to interpret `mv src dst` as `mv -t dst src`. **This decision belongs in the `mv` command, not in the kernel `rename(2)` syscall.** The Linux `mv` tool documents this with `--no-target-directory` / `-T` to override; that override has no effect on Curvine because the syscall itself is already doing the rewrite.

The root cause is that Curvine's `rename` handler at `curvine-fuse/src/fs/curvine_file_system.rs:1315-1335` calls `self.fs.rename(&old_path, &new_path)` without first checking whether the new path already exists as a non-empty directory. The handler should reject the call up front with `ENOTEMPTY` when this condition is met.

The daemon performs the user-requested side effects but skips a POSIX-mandated precondition check or post-condition signal. In this case, the missing check is "is the target an existing non-empty directory?".

**To Reproduce**

1. Mount Curvine FUSE at `/curvine-fuse`.

2. **One-line shell reproducer**:

   ```bash
   rm -rf /curvine-fuse/repro/r1
   mkdir -p /curvine-fuse/repro/r1/src /curvine-fuse/repro/r1/dst
   echo "src-content"     > /curvine-fuse/repro/r1/src/from_src
   echo "important-data"  > /curvine-fuse/repro/r1/dst/IMPORTANT
   echo "BEFORE rename:"
   ls -R /curvine-fuse/repro/r1
   mv -T /curvine-fuse/repro/r1/src /curvine-fuse/repro/r1/dst
   echo "rc=$?"
   echo "AFTER rename:"
   ls -R /curvine-fuse/repro/r1
   ```

   On Curvine the step is silent, `rc=0`, and the after-tree shows src relocated *inside* dst.

**Expected behavior**

On a POSIX-compliant filesystem the `mv -T` step prints `mv: cannot move ... : Directory not empty` and `rc=1`.

**OS Version:**

Architecture: x86_64
glibc: 2.28
GCC: 8.3.0

bigbigxu:
## Summary

Fixes [#845](https://github.com/CurvineIO/curvine/issues/845).

`rename(2)` on Curvine previously behaved like userland `mv` when the destination was an existing directory: the master renamed `src` into `dst/src` instead of failing. That broke POSIX semantics (`mv -T` / `renameat2` without `RENAME_EXCHANGE`) and differed from ext4, XFS, tmpfs, NFS, etc.

This change enforces POSIX rename rules in the master metadata layer and propagates the correct errno through FUSE and client SDKs. Hadoop-style “move into existing directory” is kept only in the Java `FileSystem.rename` wrapper, not at the syscall/backend level.

## Problem

| Call | POSIX | Before |
|------|-------|--------|
| `rename(file, existing_dir)` | `EISDIR`, no change | File moved to `dir/file` |
| `rename(dir, non_empty_dir)` | `ENOTEMPTY`, no change | `dir` nested under `dst` |
| `rename(dir, existing_file)` | `ENOTDIR`, no change | Undefined / wrong tree |
| `rename(parent, parent/child)` | `EINVAL`, no change | Internal error string |

Root cause: `FsDir::unprotected_rename` rewrote the destination when `dst` was a directory—setting `new_name = src` and using `dst` as the parent—instead of validating type and emptiness first.

### Master (`curvine-server`)

- **`fs_dir.rs` — `unprotected_rename`**
  - If `dst` exists:
    - **file → directory** → `FsError::is_a_directory` (`EISDIR`)
    - **directory → file** → `FsError::not_a_directory` (`ENOTDIR`)
    - **directory → non-empty directory** → `FsError::dir_not_empty` (`ENOTEMPTY`)
    - **directory → empty directory** → delete empty `dst`, then replace (unchanged intent)
    - **file → file** → delete `dst`, then replace (unchanged intent)
  - Always rename to `dst`’s parent with `dst`’s name (no implicit `dst/src` path).
- **`master_filesystem.rs`**
  - Rename into a subdirectory of `src` returns `FsError::invalid_argument` (`EINVAL`) instead of a generic internal error.

### Errors & errno mapping

- **`curvine-common`**: add `FsError::{IsADirectory, NotADirectory}`, `invalid_argument()`, and wire through encode/decode/`From<io::Error>`.
- **`curvine-fuse`**: map `IsADirectory` → `EISDIR`, `NotADirectory` → `ENOTDIR`.
- **Python / Java SDK**: errno plumbing for new error kinds.

### Hadoop Java SDK (intentional divergence)

- **`CurvineFileSystem.rename`**: on `EISDIR` / `ENOTEMPTY` (and `DirectoryNotEmptyException`), retry as `rename(src, dst/srcName)` so Hadoop apps keep expected behavior without polluting POSIX `rename(2)` on the FUSE mount.

## Tests

- **`curvine-server/tests/master_fs_test.rs`**
  - Replace the old “file → directory succeeds” case with POSIX matrix: file overwrite, empty-dir overwrite, `ENOTEMPTY`, `ENOTDIR`, rename-into-subtree `EINVAL`.
  - Serial test lock for inode metrics tests; fix `/a/*` glob expectations after rename tree changes.
- **`build/tests/scripts/rename_test.py`** (new): FUSE regression mirroring server cases (`EISDIR`, `ENOTEMPTY`, `ENOTDIR`, `EINVAL`, etc.).
- **`build/tests/fuse-test.sh`**: register “Test 19: POSIX Rename Semantics”.

## How to verify

```bash
# Server
cargo test -p curvine-server rename -- --nocapture

# FUSE (mounted at /curvine-fuse)
python3 build/tests/scripts/rename_test.py --dir /curvine-fuse/fuse-test

# Issue #845 repro (should fail with "Directory not empty", rc=1)
rm -rf /curvine-fuse/repro/r1
mkdir -p /curvine-fuse/repro/r1/src /curvine-fuse/repro/r1/dst
echo src-content     > /curvine-fuse/repro/r1/src/from_src
echo important-data  > /curvine-fuse/repro/r1/dst/IMPORTANT
mv -T /curvine-fuse/repro/r1/src /curvine-fuse/repro/r1/dst
echo "rc=$?"   # expect rc=1

### CODE REVIEW

File: curvine-common/src/error/fs_error.rs [hunk 1]
```diff
@@ -240,6 +260,11 @@ impl FsError {
         Self::InvalidPath(ErrorImpl::with_source(msg.into()))
     }
 
+    /// Invalid operation argument (maps to POSIX EINVAL).
+    pub fn invalid_argument(msg: impl Into<String>) -> Self {
```
szbr00, line 269: `invalid_argument()` is implemented as a thin alias of `Self::InvalidPath(...)`:

    pub fn invalid_argument(msg: impl Into<String>) -> Self {
        Self::InvalidPath(ErrorImpl::with_source(msg.into().into()))
    }

But the call site in `master_filesystem.rs` (`cannot rename {src} to {dst}: destination is under source`) is not an invalid *path* — both paths are syntactically valid; it's the *combination* that's illegal (POSIX `EINVAL`). Reusing `InvalidPath` works for FUSE because both happen to land on `EINVAL`, but it:

- silently merges two distinct error categories into one `ErrorKind` (`InvalidPath = 16`), so Java/Python SDKs can never distinguish "bad path" from "bad rename argument";
- breaks `From<std::io::Error>` round-tripping if/when a real `io::ErrorKind::InvalidInput` ever gets mapped here;
- makes the public name (`invalid_argument`) misleading vs the constructed variant.

Suggest adding a real `ErrorKind::InvalidArgument` / `FsError::InvalidArgument` variant with its own encode/decode/from-io arms and a FUSE → `EINVAL` mapping. Otherwise drop the helper entirely and just call `FsError::invalid_path(...)` from the rename site so the type and the variant agree.

curvine-libsdk/java/src/main/java/io/curvine/exception/CurvineException.java
```diff
@@ -36,7 +36,7 @@ public class CurvineException extends IOException {
     public final static int FILE_ALREADY_EXISTS = 7;
     public final static int FILE_NOT_FOUND = 8;
     public final static int NOT_A_DIRECTORY = 9;
-    public final static int IS_A_DIRECTORY = 10;
+    public final static int IS_A_DIRECTORY = 26;
```
szbr00: This file's errno constants have drifted from `curvine-common/src/error/fs_error.rs::ErrorKind`, and this PR only fixes the most visible symptom (`IS_A_DIRECTORY` 10 → 26). The full picture:

| Java constant | Java value | Actual Rust `ErrorKind` at that value |
|---|---|---|
| `INVALID_ARGUMENT = 1` | 1 | `IO` |
| `IO_ERROR = 2` | 2 | `NotLeaderMaster` |
| `TIMEOUT = 3` | 3 | `Raft` |
| `INTERNAL_ERROR = 4` | 4 | `Timeout` |
| `NOT_SUPPORTED = 5` | 5 | `PBDecode` |
| `PERMISSION_DENIED = 6` | 6 | `PBEncode` |
| `NOT_A_DIRECTORY = 9` | 9 | `InvalidFileSize` (Rust ENOTDIR is now 27) |

So when the master returns e.g. `ErrorKind::IO = 1`, the Java `create()` switch treats it as `INVALID_ARGUMENT`, and any caller comparing `getErrno() == INVALID_ARGUMENT` gets a false positive. The Copilot note about `NOT_A_DIRECTORY` is correct but it's the tip of the iceberg.

Recommended fixes (any one of):
1. Pin every constant in this file to its Rust counterpart explicitly (`INVALID_ARGUMENT = 16` to match `ErrorKind::InvalidPath`, `IO_ERROR = 1` to match `IO`, etc.) and add a single test that round-trips each kind through the gRPC/Tonic boundary.
2. Generate this file from the `ErrorKind` enum at build time (the values already have stable numeric ids).
3. At minimum, while this PR is open, also fix `NOT_A_DIRECTORY = 27` and add a comment block above the constants warning that they must be kept in lock-step with `fs_error.rs`.

build/tests/scripts/rename_test.py
```diff
@@ -0,0 +1,287 @@
...
+    fail = 0
+    for num, test_fn in TESTS:
+        print(f"[case {num}] {test_fn.__name__}:")
+        fail += test_fn(root)
+        print()
+
+    total = len(TESTS)
+    passed = total - fail
+    print(f"Summary: {passed}/{total} passed, {fail}/{total} failed")
+    return 1 if fail else 0
```
szbr00, line 314: Two coverage gaps worth closing while you're touching this file:

1. **`test_same_path_noop` only covers files.** The master short-circuits `src == dst` in `master_filesystem.rs::rename` for *any* inode type, but a regression that breaks that short-circuit for directories would slip through (e.g., a non-empty dir with `src == dst` would otherwise hit the new `dir_not_empty` check and start failing). Add `test_same_path_noop_empty_dir` and `test_same_path_noop_nonempty_dir`.

2. **No `RENAME_NOREPLACE` coverage.** The `fuse-test.sh` description added in this PR explicitly advertises `RENAME_NOREPLACE`, but neither the Python script nor the server-side matrix exercises it. Python's `os.rename` can't pass `RENAME_NOREPLACE`, but a short `ctypes`-based `renameat2(AT_FDCWD, src, AT_FDCWD, dst, RENAME_NOREPLACE)` call (and assert `EEXIST` when dst exists, success when it doesn't) would close this gap and prove the FUSE → master plumbing for `no_replace` is wired correctly.

curvine-common/src/error/fs_error.rs
```diff
@@ -356,6 +383,9 @@ impl From<std::io::Error> for FsError {
             io::ErrorKind::DirectoryNotEmpty => {
                 Self::DirNotEmpty(ErrorImpl::with_source(value.to_string().into()))
             }
+            io::ErrorKind::IsADirectory => {
+                Self::IsADirectory(ErrorImpl::with_source(value.to_string().into()))
+            }
             io::ErrorKind::Unsupported => {
                 Self::Unsupported(ErrorImpl::with_source(value.to_string().into()))
```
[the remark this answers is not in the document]
bigbigxu: fixed

curvine-common/src/error/fs_error.rs
bigbigxu: fixedd

curvine-libsdk/java/src/main/java/io/curvine/exception/CurvineException.java
bigbigxu: fixed

build/tests/scripts/rename_test.py
bigbigxu: fixed

curvine-libsdk/java/src/main/java/io/curvine/exception/CurvineException.java
```diff
@@ -36,7 +36,7 @@ public class CurvineException extends IOException {
...
     public final static int DIRECTORY_NOT_EMPTY = 11;
```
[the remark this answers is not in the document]
bigbigxu: fixed

curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
```diff
@@ -242,13 +244,46 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+        if (e instanceof DirectoryNotEmptyException) {
+            return true;
+        }
+        if (e instanceof CurvineException) {
+            int errno = ((CurvineException) e).getErrno();
+            return errno == CurvineException.IS_A_DIRECTORY || errno == CurvineException.DIRECTORY_NOT_EMPTY;
+        } else {
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
bigbigxu: fixed

curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -323,31 +323,35 @@ impl FsDir {
...
+            Some(self.unprotected_delete(dst_inp, mtime)?)
+        } else {
+            None
+        };
+
+        let new_name = dst_inp.name().to_string();
+        // /1.log -> /a/b : b is the new inode name and /a is the destination parent.
+        // For replace semantics (dst exists), dst has been deleted above and we still
+        // rename to dst's original parent with dst's original name.
+        let mut dst_parent = match dst_inp.get_inode(-2) {
```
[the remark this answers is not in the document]
bigbigxu: fixed

szbr00 (approved):

PR #855, opened 2026-05-30
feat(spdk): support the mechanism to handle strained completions on qpair poll error

thuongle2210:
# Description
- This PR handles stranded I/Os when an SPDK qpair poll fails. Without it, in-flight I/Os never complete, and threads hang until timeout, inflight never reaches zero, and SpdkBdev::drop() leaks DMA buffers and the qpair after ~ io_timeout_us*2.
- My approach: Introduce per-qpair pending Vec tracks in-flight `CallbackCtx` pointers. On poll failure, force_complete_qpair drains the Vec, completes every I/O with -EIO and decrements inflight. A `qpair_dead` atomic lets the bdev fast-fail subsequent I/Os immediately.
# Key Changes
## orpc/src/io/spdk_bdev.rs
- `SpdkIoChannel.qpair_dead` & `IoRequest.qpair_dead`: shared flag set by poller on failure.
- Early-return in `spdk_read/spdk_write/spdk_flush` if qpair_dead.
## orpc/src/io/spdk_poller.rs
- `QpairState`: per-qpair struct with dead flag + `Vec<*mut CallbackCtx>` tracking in-flight I/Os and `CallbackCtx` extended with qpair_state back-pointer and pending_idx.
- `force_complete_qpair`: the method force complete stranded I/Os on failed qpairs 
- poller_callback: removes completed I/O from the pending list, reclaims `CallbackCtx`
- collect `failed_qpairs` during `active_qpairs.retain()` to prevent leaking on multi-qpair failure
- submit_one: retrieves/creates `QpairState`, fast-fails on dead flag (-ENXIO), collects `cb_ctx_ptr` into QpairState's pending
- ` IoCompletion::complete()` now returns false if already done, guarding against double-wake.  
- `handle_unregister`: removes dead_qpairs entry on graceful shutdown
- `PollerConfig`: add `io_queue_depth` as a part of poller configuration to pre-allocate QpairState's pending Vec.
- Add the unit test to demonstrate `force_complete_reclaims_callback_ctx` works correctly.
# Impact
- Stranded I/Os complete immediately with -EIO instead of blocking threads until timeout. Subsequent submissions fast-fail via `qpair_dead`. Clean shutdown with no leaks.


thuongle2210:
Dear @szbr9486 could you please help me review this PR?

orpc/src/io/spdk_poller.rs
```diff
@@ -459,23 +542,71 @@ impl SpdkPoller {
...
+    if idx != last {
+        let last_ptr = qs.pending[last];
+        qs.pending.swap_remove(idx);
+        (*last_ptr).pending_idx = idx;
+    } else {
+        qs.pending.pop();
+    }
+
     ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
     ctx.completion.complete(status);
```
szbr00, line 611: With the new `complete()` returning `bool` to signal first-call-wins, `force_complete_qpair` correctly guards the inflight decrement:

    if ctx.completion.complete(-libc::EIO) {
        ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
    }

But `poller_callback` does the opposite — it decrements unconditionally and then ignores the bool:

    ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
    ctx.completion.complete(status);

These two paths must agree. If the Copilot concern at L563 is resolved by *not* dropping `CallbackCtx` inside `force_complete_qpair` (so SPDK's late callback can still safely run), both `force_complete_qpair` and `poller_callback` may then be invoked for the same I/O, and the current `poller_callback` will underflow `bdev_inflight` (which is `AtomicUsize`, so it wraps to `usize::MAX`).

Suggest mirroring the pattern: only decrement `bdev_inflight` when `complete()` returned `true`, i.e. when this call actually won the race. That makes the invariant "every I/O decrements `bdev_inflight` exactly once" hold regardless of which path completes it first.

orpc/src/io/spdk_poller.rs
```diff
@@ -374,16 +424,27 @@ impl SpdkPoller {
...
+    /// Handle unregister request, remove qpair from active and dead_qpairs set and ack.
+    fn handle_unregister(
+        req: &IoRequest,
+        active_qpairs: &mut Vec<*mut spdk_ffi::spdk_nvme_qpair>,
+        dead_qpairs: &mut HashMap<usize, Box<QpairState>>,
+    ) {
         if let IoOp::UnregisterQpair { qpair, ack } = &req.op {
             active_qpairs.retain(|&qp| qp != *qpair);
+            let key = *qpair as usize;
+            dead_qpairs.remove(&key);
```
szbr00, line 436: On graceful unregister, this just drops the `QpairState` (and its `pending: Vec<*mut CallbackCtx>`) without reclaiming the boxed `CallbackCtx`s those raw pointers own:

    active_qpairs.retain(|&qp| qp != *qpair);
    let key = *qpair as usize;
    dead_qpairs.remove(&key);    // <- drops QpairState, leaks every entry in pending

The `Drop` impl for `QpairState` (implicit) won't touch the raw pointers either — `Vec<*mut CallbackCtx>` is just a `Vec` of `usize`-sized values from Rust's perspective.

Two consequences:
1. **Memory leak**: every in-flight `CallbackCtx`, plus the `IoCompletion::Arc` and `bdev_inflight::Arc` it holds, leaks. On a busy bdev being torn down this is a real footprint.
2. **Caller stuck waiting on `IoCompletion`**: any thread blocked in `IoCompletion::wait()` for one of those I/Os will sit until its timeout, since nothing on the unregister path ever flips `done` for them.

Suggest treating unregister as another force-complete: drain `qs.pending`, `Box::from_raw` each entry, `complete(-libc::ESHUTDOWN)` (or `-libc::ECANCELED`), and decrement `bdev_inflight` for the ones that won the race. That keeps the `Vec`-of-raw-pointers invariant — "every push is balanced by either `poller_callback` or `force_complete`" — symmetric across normal completion, poll failure, and shutdown.

(Same caveat as the Copilot comment: this is only safe if no NVMe commands are still in flight at the device level. For graceful unregister you have control: `spdk_nvme_qpair_process_completions` should be drained or `spdk_nvme_ctrlr_disconnect_io_qpair` called before reaching here.)

orpc/src/io/spdk_poller.rs
```diff
@@ -240,6 +251,8 @@ impl SpdkPoller {
     ) {
         let mut active_qpairs: Vec<*mut spdk_ffi::spdk_nvme_qpair> = Vec::new();
         let mut state = PollerState::Idle;
+        // Tracks per-qpair state (dead flag + pending Vec) for force-completion.
+        let mut dead_qpairs: HashMap<usize, Box<QpairState>> = HashMap::new();
```
szbr00, line 255: A naming / test-coverage cluster:

1. **Name.** `dead_qpairs` reads as "qpairs known to have failed", but in this implementation it is actually populated lazily in `submit_one` for *every* qpair on first I/O, and the only thing that distinguishes dead from live entries is `qs.dead`. A clearer name like `qpair_state` (or just `qpairs`) would stop new readers from assuming "this map is only the failed set", which matters because `force_complete_qpair` and `handle_unregister` both do `remove(&key)` and that has very different implications under the two readings.

2. **Test coverage.** `force_complete_reclaims_callback_ctx` is a good start but covers only the simplest path. The genuinely-interesting invariants this PR introduces are untested:
    - `poller_callback`'s `swap_remove(idx)` + `(*last_ptr).pending_idx = idx;` reindex. A test that submits N I/Os, completes them out of order, and asserts `pending_idx == position_in_vec` after each callback would catch a regression here (and would have caught the integer-underflow if `pending` is unexpectedly empty when `poller_callback` fires).
    - The "submit while dead" fast-fail path in `submit_one` (return `-ENXIO`, no `pending.push`).
    - `handle_unregister` with non-empty pending (see draft 2).

These don't need a real SPDK to validate — the existing test pattern (`QpairState` constructed directly, `Box::into_raw` into pending, then drive the API) already proves the unit-level approach works.

szbr00 (approved):

orpc/src/io/spdk_poller.rs
thuongle2210: I am aware of it. I am also preparing it for the next PR.

orpc/src/io/spdk_poller.rs
thuongle2210: I am aware of it. I am also preparing it for the next PR.

orpc/src/io/spdk_poller.rs
thuongle2210: I will add comprehensive tests later

orpc/src/io/spdk_poller.rs
```diff
@@ -459,23 +542,71 @@ impl SpdkPoller {
...
+    /// per-qpair pending Vec, then mark the qpair dead so future submissions fail fast.
+    fn force_complete_qpair(key: usize, dead_qpairs: &mut HashMap<usize, Box<QpairState>>) {
+        if let Some(qs) = dead_qpairs.get_mut(&key) {
+            qs.dead.store(true, Ordering::Release);
+            let pending = std::mem::take(&mut qs.pending);
+            let count = pending.len();
+            for cb_ptr in &pending {
+                let ctx = unsafe { Box::from_raw(*cb_ptr as *mut CallbackCtx) };
+                if ctx.completion.complete(-libc::EIO) {
+                    ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
```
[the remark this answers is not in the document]
thuongle2210: I am aware of it. I am also preparing it for the next PR to resolve the remaining TODOs that I listed earlier

PR #858, opened 2026-06-01
fix(fuse): bump writer version and refresh reader when versions differ

### LINKED ISSUE #852: [BUG]:Extending `pwrite()` after `pread()` does not extend the page cache range — `pread()` returns short / `0` even though `fstat()` reports the correct size
**Describe the bug**

On a Curvine FUSE mount, the following sequence on the same file descriptor produces an inconsistent view of the file's size between `fstat()`/`lseek(SEEK_END)` and `pread()`:

1. `open(path, O_RDWR | O_CREAT)`
2. `pwrite(buf, N1, 0)` — writes `[0, N1)`
3. `pread(buf, N1, 0)` — **establishes the page cache** with the file's current size
4. `pwrite(buf, N2, off2)` where `off2 + N2 > N1` — **extends** the file
5. `fstat(fd)` — returns `off2 + N2` ✅ correct
6. `lseek(fd, 0, SEEK_END)` — returns `off2 + N2` ✅ correct
7. **`pread(buf, off2 + N2, 0)` — returns only `N1` bytes** ❌ short read

The kernel's metadata view (`fstat`, `lseek SEEK_END`) is consistent with the daemon-side authoritative size, but the page cache used by `pread` is still capped at `N1`. Any read at offset `>= N1` returns `0` (EOF), and any range read that crosses `N1` is truncated to `N1` bytes. The data written by step 4 is genuinely on the daemon (a subsequent `fsync` + `pread` retrieves it correctly), but it is **invisible to `pread` until the page cache is forcibly refreshed**.

This is a **distinct issue** from the previously-reported "`pwrite()` does not propagate updated `i_size` to kernel" problem. That older issue caused `fstat()` itself to return stale size (e.g. `0` immediately after `pwrite`); this one shows `fstat()` correct but `pread()` truncated. The metadata path was fixed; the page-cache-extension path was missed.

The symptom recovers if (and only if) the page cache itself is invalidated:

| Action between extending pwrite and pread | Recovers? | Why |
|---|---|---|
| `fsync(fd)` | ✅ | Forces page cache flush + invalidation, then refetched |
| `close(fd)` + `open(path)` | ✅ | New fd, fresh page cache, fresh attr |
| `lseek(fd, 0, SEEK_END)` | ❌ | Reads attr cache, not page cache |
| `fstat(fd)` | ❌ | Reads attr cache, not page cache |
| Nothing (default) | ❌ | The bug |

The fact that `lseek SEEK_END` and `fstat` cannot recover this — but they themselves return the correct size — proves that this is a **page-cache** issue, not an attribute-cache issue.

**To Reproduce**

1. Mount Curvine FUSE at `/curvine-fuse` and create a working directory:

   ```bash
   sudo mkdir -p /curvine-fuse/repro
   ```

2. Save the following minimal reproducer as `repro_pc_boundary.c`. It demonstrates that after `pread` establishes the page cache, an extending `pwrite` does not extend the cache range — `pread` returns short while `fstat` and `lseek SEEK_END` return the correct size.

   ```c
   /* repro_pc_boundary.c — reproduces Curvine page-cache-not-extended bug.
    *
    * Build: gcc -O0 -Wall repro_pc_boundary.c -o /tmp/repro_pc_boundary
    * Run:   /tmp/repro_pc_boundary
    */
   #define _GNU_SOURCE
   #include <stdio.h>
   #include <fcntl.h>
   #include <unistd.h>
   #include <sys/stat.h>
   #include <string.h>

   int main(void) {
       const char *p = "/curvine-fuse/repro/pcb.bin";
       unlink(p);
       int fd = open(p, O_RDWR | O_CREAT, 0644);

       char b[8192];
       memset(b, 'X', sizeof b);

       /* iter=0: pwrite + pread → establishes page cache with size N1=1167. */
       pwrite(fd, b, 1167, 0);
       char rd[8192];
       pread(fd, rd, 1167, 0);

       /* iter=1: extending pwrite to off2=1425, len N2=3882 → file should be
          size 1425+3882 = 5307; bytes [1167, 1425) are a hole that POSIX
          requires to read as zero. */
       pwrite(fd, b, 3882, 1425);

       /* Verify multiple read-paths' views of the file's size. */
       struct stat st;
       fstat(fd, &st);
       printf("[1] fstat size       = %lld (expect 5307)\n",
              (long long)st.st_size);

       off_t end = lseek(fd, 0, SEEK_END);
       printf("[2] lseek SEEK_END   = %lld (expect 5307)\n",
              (long long)end);

       /* Whole-file pread — should return 5307. */
       ssize_t n_full = pread(fd, rd, 5307, 0);
       printf("[3] pread(5307, 0)   = %zd (expect 5307)\n", n_full);

       /* pread of just the extension region — should return 3882. */
       ssize_t n_ext  = pread(fd, rd, 3882, 1425);
       printf("[4] pread(3882, 1425)= %zd (expect 3882)\n", n_ext);

       /* pread of one byte just past the old size — should return 1. */
       ssize_t n_one  = pread(fd, rd, 1, 1167);
       printf("[5] pread(1, 1167)   = %zd (expect 1)\n", n_one);

       close(fd);
       unlink(p);
       return 0;
   }
   ```

3. Compile and run:

   ```bash
   gcc -O0 -Wall repro_pc_boundary.c -o /tmp/repro_pc_boundary
   /tmp/repro_pc_boundary
   ```

4. Optional — to demonstrate that *only* page-cache invalidating actions recover the size, the `repro_pread_then_pwrite.c` companion below tests six "what-can-rescue-the-pread" variants:

   ```c
   /* repro_pread_then_pwrite.c — six rescue strategies, only fsync /
    * close+reopen actually unblock subsequent pread.
    */
   #define _GNU_SOURCE
   #include <stdio.h>
   #include <stdlib.h>
   #include <string.h>
   #include <fcntl.h>
   #include <unistd.h>
   #include <sys/stat.h>

   #define BUF1_LEN 1167
   #define OFF2     1425
   #define BUF2_LEN 3882
   #define EXPECT   (OFF2 + BUF2_LEN)   /* 5307 */

   static int run(const char *name, int do_fsync, int do_lseek,
                  int do_reopen, int do_fstat) {
       const char *p = "/curvine-fuse/repro/pp.bin";
       unlink(p);
       int fd = open(p, O_RDWR|O_CREAT, 0644);

       char *b1 = malloc(BUF1_LEN); memset(b1, 'A', BUF1_LEN);
       char *b2 = malloc(BUF2_LEN); memset(b2, 'B', BUF2_LEN);
       char *r  = malloc(EXPECT + 16);

       pwrite(fd, b1, BUF1_LEN, 0);
       pread(fd, r, BUF1_LEN, 0);              /* establishes page cache */
       pwrite(fd, b2, BUF2_LEN, OFF2);          /* extending pwrite */

       if (do_fsync)  fsync(fd);
       if (do_lseek)  lseek(fd, 0, SEEK_END);
       if (do_fstat)  { struct stat st; fstat(fd, &st); }
       if (do_reopen) { close(fd); fd = open(p, O_RDWR); }

       ssize_t n = pread(fd, r, EXPECT, 0);
       printf("  [%s] pread=%zd (expect %d) %s\n",
              name, n, EXPECT, n == EXPECT ? "OK" : "FAIL");

       free(b1); free(b2); free(r); close(fd); unlink(p);
       return n == EXPECT ? 0 : 1;
   }

   int main(void) {
       system("mkdir -p /curvine-fuse/repro");
       int fail = 0;
       fail += run("A. baseline (no flush)",     0,0,0,0);
       fail += run("B. fsync between",           1,0,0,0);
       fail += run("C. lseek SEEK_END between",  0,1,0,0);
       fail += run("D. close+reopen between",    0,0,1,0);
       fail += run("E. fstat between",           0,0,0,1);
       fail += run("F. fsync + lseek + fstat",   1,1,0,1);
       printf("\n%d/6 FAIL\n", fail);
       return fail ? 1 : 0;
   }
   ```

   ```bash
   gcc -O0 -Wall repro_pread_then_pwrite.c -o /tmp/repro_pread_then_pwrite
   /tmp/repro_pread_then_pwrite
   ```

**Observed output on Curvine** (`repro_pc_boundary`):

```
$ /tmp/repro_pc_boundary
[1] fstat size       = 5307 (expect 5307)        <-- correct
[2] lseek SEEK_END   = 5307 (expect 5307)        <-- correct
[3] pread(5307, 0)   = 1167 (expect 5307)        <-- BUG: short read
[4] pread(3882, 1425)= 0    (expect 3882)        <-- BUG: page cache doesn't cover 1425+
[5] pread(1, 1167)   = 0    (expect 1)           <-- BUG: page cache caps at exactly old size
```

**Observed output on Curvine** (`repro_pread_then_pwrite`):

```
$ /tmp/repro_pread_then_pwrite
  [A. baseline (no flush)]    pread=1167 (expect 5307) FAIL
  [B. fsync between]          pread=5307 (expect 5307) OK
  [C. lseek SEEK_END between] pread=1167 (expect 5307) FAIL
  [D. close+reopen between]   pread=5307 (expect 5307) OK
  [E. fstat between]          pread=1167 (expect 5307) FAIL
  [F. fsync + lseek + fstat]  pread=5307 (expect 5307) OK

3/6 FAIL
```

The split between `B/D/F` (page-cache-flushing actions) all OK and `A/C/E` (attr-only actions) all FAIL is the smoking gun: this is a page-cache problem, not an attribute-cache problem.

Step `[5]` of `repro_pc_boundary` is particularly diagnostic — `pread(1, 1167)` reading exactly one byte at offset 1167 returns `0`, even though the file is genuinely 5307 bytes long. The page cache range happens to end exactly at byte 1167 (the size at the moment `pread` first established it), and the extending `pwrite` did not push that range forward.

**Expected behavior**

POSIX `pwrite(2)` and `read(2)` together require that, after a successful write that extends the file's size, every subsequent read from any offset `< new_size` returns either the data written or zero-fill (for hole regions), without requiring an explicit `fsync` or any other refresh syscall.

Specifically, the expected output of `repro_pc_boundary` on a POSIX-compliant filesystem (ext4, xfs, btrfs, tmpfs, NFS, well-behaved FUSE filesystems) is:

```
[1] fstat size       = 5307 (expect 5307)
[2] lseek SEEK_END   = 5307 (expect 5307)
[3] pread(5307, 0)   = 5307 (expect 5307)
[4] pread(3882, 1425)= 3882 (expect 3882)
[5] pread(1, 1167)   = 1    (expect 1)
```

And for `repro_pread_then_pwrite`, the expected output is `0/6 FAIL` — every variant should produce `pread=5307`, regardless of whether `fsync`/`lseek`/`fstat` were called.

**OS Version:**

Architecture: x86_64
glibc: 2.28
GCC: 8.3.0

bigbigxu:
## Summary

Fixes [#852](https://github.com/CurvineIO/curvine/issues/852): after `pread()` then an extending `pwrite()` on the same fd, metadata (`fstat`, `lseek`) can show the new size while `pread()` still returns data bounded by the old length until the reader is refreshed (e.g. via `fsync` or reopen).

## Approach

1. **Writer version** — `FuseWriter` maintains `write_ver`, incremented on each `write()` and `resize()`.
2. **Version check on read** — `FileHandle` keeps `read_ver` (last write generation seen for this handle). Before serving a read, compare `read_ver` with the writer's `write_ver`.
3. **Refresh when stale** — If they differ, flush the writer, replace the `FuseReader`, and set `read_ver = write_ver`, then perform the read so data matches the current file state.

The previous logic only refreshed when the read offset was past `reader.len()`, so whole-file or low-offset reads after an extending write never triggered a refresh. Version comparison covers any write since the last read refresh, regardless of read offset.

## Changes

| Area | Change |
|------|--------|
| `fuse_writer.rs` | `write_ver` + `write_ver()`; bump on write/resize |
| `file_handle.rs` | `read_ver`; on read, compare versions → flush + new reader if mismatch |
| `visibility_test.py` | Add `--mode pread\|all` suite (six strategies from issue repro) |

## Test plan

- [ ] `python3 build/tests/scripts/visibility_test.py --dir <mount> --mode all` → `[basic] PASS`, `[pread] 0/6 FAIL`
- [ ] Issue repro `repro_pc_boundary.c`: all five checks pass without `fsync`
- [ ] Regression: normal sequential read/write on same fd

## Related

Closes #852

### DIFF
--- a/build/tests/scripts/visibility_test.py
+++ b/build/tests/scripts/visibility_test.py
@@ -2,11 +2,17 @@
 # -*- coding: utf-8 -*-
 
 """
-Visibility test for file size and content after pwrite.
+Visibility tests for pwrite/pread behavior on FUSE mounts.
+
+Includes:
+1) basic file size/content visibility after pwrite
+2) pread visibility strategies after extending pwrite (ported from repro_pread_then_pwrite.py)
 
 Usage:
-    python3 visibility_test.py --dir /curvine-fuse
+    python3 visibility_test.py --dir /curvine-fuse --mode all
 """
+
+from __future__ import annotations
 
 import argparse
 import os
@@ -14,17 +20,26 @@ import sys
 
 WRITE_SIZE = 1024
 DEFAULT_DIR = "/curvine-fuse"
-TEST_FILE = "visibility-test"
+BASIC_TEST_FILE = "visibility-test"
+
+PP_TEST_FILE = "pp.bin"
+# Values from issue #852 repro (repro_pc_boundary.c / repro_pread_then_pwrite.c):
+# first write establishes reader at BUF1_LEN; extending pwrite starts at OFF2
+# with a sparse hole [BUF1_LEN, OFF2).
+BUF1_LEN = 1167
+OFF2 = 1425
+BUF2_LEN = 3882
+EXPECT = OFF2 + BUF2_LEN  # 5307
+
+_EXPECTED_PREAD = (
+    b"A" * BUF1_LEN
+    + b"\x00" * (OFF2 - BUF1_LEN)
+    + b"B" * BUF2_LEN
+)
 
 
-def main() -> int:
-    parser = argparse.ArgumentParser(
-        description="Visibility test for file size and content after pwrite",
-    )
-    parser.add_argument("--dir", default=DEFAULT_DIR, help="Directory for test files")
-    args = parser.parse_args()
-
-    path = os.path.join(os.path.abspath(args.dir), TEST_FILE)
+def run_basic_visibility(base_dir: str) -> int:
+    path = os.path.join(base_dir, BASIC_TEST_FILE)
     data = b"A" * WRITE_SIZE
 
     try:
@@ -39,7 +54,7 @@ def main() -> int:
             return 1
 
         size = os.fstat(fd).st_size
-        print(f"size={size} (expect {WRITE_SIZE})")
+        print(f"[basic] size={size} (expect {WRITE_SIZE})")
         if size != WRITE_SIZE:
             print("ERROR: size mismatch", file=sys.stderr)
             return 1
@@ -49,10 +64,127 @@ def main() -> int:
             print("ERROR: read content mismatch", file=sys.stderr)
             return 1
 
-        print("PASS")
+        print("[basic] PASS")
         return 0
     finally:
         os.close(fd)
+
+
+def run_pread_strategy(
+        name: str,
+        file_path: str,
+        *,
+        do_fsync: bool,
+        do_lseek: bool,
+        do_reopen: bool,
+        do_fstat: bool,
+) -> int:
+    try:
+        os.unlink(file_path)
+    except FileNotFoundError:
+        pass
+
+    b1 = b"A" * BUF1_LEN
+    b2 = b"B" * BUF2_LEN
+
+    fd = os.open(file_path, os.O_RDWR | os.O_CREAT, 0o644)
+    try:
+        os.pwrite(fd, b1, 0)
+        os.pread(fd, BUF1_LEN, 0)  # establish page cache
+        os.pwrite(fd, b2, OFF2)  # extending pwrite
+
+        if do_fsync:
+            os.fsync(fd)
+        if do_lseek:
+            os.lseek(fd, 0, os.SEEK_END)
+        if do_fstat:
+            os.fstat(fd)
+        if do_reopen:
+            os.close(fd)
+            fd = -1
+            fd = os.open(file_path, os.O_RDWR)
+
+        data = os.pread(fd, EXPECT, 0)
+        ok = len(data) == EXPECT and data == _EXPECTED_PREAD
+        print(
+            f"[pread] {name}: pread={len(data)} (expect {EXPECT}) "
+            f"{'OK' if ok else 'FAIL'}"
+        )
+        if not ok and len(data) == EXPECT:
+            print("[pread] content mismatch (stale or wrong bytes)", file=sys.stderr)
+        return 0 if ok else 1
+    finally:
+        if fd >= 0:
+            os.close(fd)
+        try:
+            os.unlink(file_path)
+        except FileNotFoundError:
+            pass
+
+
+def run_pread_visibility(base_dir: str) -> int:
+    file_path = os.path.join(base_dir, PP_TEST_FILE)
+    cases = [
+        (
+            "A. baseline (no flush)",
+            dict(do_fsync=False, do_lseek=False, do_reopen=False, do_fstat=False),
+        ),
+        (
+            "B. fsync between",
+            dict(do_fsync=True, do_lseek=False, do_reopen=False, do_fstat=False),
+        ),
+        (
+            "C. lseek SEEK_END between",
+            dict(do_fsync=False, do_lseek=True, do_reopen=False, do_fstat=False),
+        ),
+        (
+            "D. close+reopen between",
+            dict(do_fsync=False, do_lseek=False, do_reopen=True, do_fstat=False),
+        ),
+        (
+            "E. fstat between",
+            dict(do_fsync=False, do_lseek=False, do_reopen=False, do_fstat=True),
+        ),
+        (
+            "F. fsync + lseek + fstat",
+            dict(do_fsync=True, do_lseek=True, do_reopen=False, do_fstat=True),
+        ),
+    ]
+
+    fail = 0
+    for name, kwargs in cases:
+        fail += run_pread_strategy(name, file_path, **kwargs)
+    passed = len(cases) - fail
+    if fail == 0:
+        print(f"[pread] {passed}/{len(cases)} PASS")
+    else:
+        print(f"[pread] {fail}/{len(cases)} FAIL")
+    return 1 if fail else 0
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description="Visibility tests for pwrite/pread on FUSE mounts",
+    )
+    parser.add_argument("--dir", default=DEFAULT_DIR, help="Directory for test files")
+    parser.add_argument(
+        "--mode",
+        choices=["basic", "pread", "all"],
+        default="all",
+        help="Select which visibility suite to run",
+    )
+    args = parser.parse_args()
+
+    base_dir = os.path.abspath(args.dir)
+    os.makedirs(base_dir, exist_ok=True)
+
+    fail = 0
+    if args.mode in ("basic", "all"):
+        fail += run_basic_visibility(base_dir)
+    if args.mode in ("pread", "all"):
+        fail += run_pread_visibility(base_dir)
+
+    return 1 if fail else 0
 
 
 if __name__ == "__main__":
--- a/curvine-fuse/src/fs/state/file_handle.rs
+++ b/curvine-fuse/src/fs/state/file_handle.rs
@@ -20,6 +20,7 @@ use crate::{err_fuse, FuseError, FuseResult};
 use curvine_common::fs::{Path, StateReader, StateWriter};
 use curvine_common::state::{CreateFileOptsBuilder, FileStatus, LockFlags, OpenFlags};
 use orpc::err_box;
+use orpc::sync::AtomicCounter;
 use orpc::sys::RawPtr;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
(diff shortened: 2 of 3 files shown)


curvine-fuse/src/fs/state/file_handle.rs
```diff
@@ -71,16 +75,18 @@ impl FileHandle {
...
+            let mut writer = lock.lock().await;
+            let write_ver = writer.write_ver();
+            if self.read_ver.get() != write_ver {
+                writer.flush(None).await?;
+                drop(writer);
+
                 let path = reader.path().clone();
-                reader.as_mut().complete(None).await?;
                 let new_reader = state.new_reader(&path).await?;
                 reader.replace(new_reader);
```
szbr00, line 87: The previous logic explicitly called `reader.as_mut().complete(None).await?` before swapping in `new_reader`:

    let path = reader.path().clone();
    reader.as_mut().complete(None).await?;          // <- removed
    let new_reader = state.new_reader(&path).await?;
    reader.replace(new_reader);

The new code relies on the `Drop` impl of the old reader to do the cleanup when `reader.replace(new_reader)` swaps it out. That's fine if `complete()` was a no-op or a synchronous-equivalent shutdown, but it's a behavior change worth confirming because:

- `complete()` is `async`. If it sends a "close-read" RPC, commits read offsets, or releases block leases on the master/worker side, a synchronous `Drop` cannot replicate that without spawning a task (and that task's errors would be swallowed).
- The old call propagated errors via `?`. The new code silently ignores any failure when the old reader is torn down. If the master's lease cache leaks, it won't surface here.
- This path now runs on every version-mismatched read, not just on offset-beyond-EOF reads as before — so any leak per refresh is amplified by a factor of how often readers race with extending writes.

Suggest either restoring the explicit `complete(None).await?` call before `reader.replace(...)`, or adding a comment that documents why `Drop` is sufficient (and ideally a test that verifies no resources leak across N refresh cycles).

build/tests/scripts/visibility_test.py
```diff
@@ -49,11 +55,119 @@ def main() -> int:
...
+        if do_lseek:
+            os.lseek(fd, 0, os.SEEK_END)
+        if do_fstat:
+            os.fstat(fd)
+        if do_reopen:
+            os.close(fd)
+            fd = -1
+            fd = os.open(file_path, os.O_RDWR)
+
+        n = len(os.pread(fd, EXPECT, 0))
```
szbr00: The pread strategies validate length only:

    n = len(os.pread(fd, EXPECT, 0))
    ok = n == EXPECT

A reader that returns `EXPECT` bytes of garbage — or returns the right length but with stale page-cache contents in the `[0, BUF1_LEN)` range and zeros in `[OFF2, EXPECT)` — would pass this test even though the underlying #852 bug would still be reproducible.

Since the strategies write deterministic patterns (`b"A" * BUF1_LEN` at offset 0 and `b"B" * BUF2_LEN` at offset `OFF2`), it's cheap to assert content too:

    data = os.pread(fd, EXPECT, 0)
    ok = (
        len(data) == EXPECT
        and data[:BUF1_LEN] == b"A" * BUF1_LEN
        and data[BUF1_LEN:OFF2] == b"\x00" * (OFF2 - BUF1_LEN)  # sparse-hole region
        and data[OFF2:] == b"B" * BUF2_LEN
    )

Two related minor items while you're here:
- The final summary `print(f"[pread] {fail}/6 FAIL")` prints `[pread] 0/6 FAIL` on full success, which is misleading; consider `f"[pread] {6-fail}/6 PASS"` or branching on `fail == 0`.
- A brief comment on `BUF1_LEN = 1167`, `OFF2 = 1425`, `BUF2_LEN = 3882` would help future maintainers understand these were chosen to span the page-cache boundary in the #852 repro (rather than looking like arbitrary magic numbers).

curvine-fuse/src/fs/state/file_handle.rs
bigbigxu: `complete` is a no-op, which is actually unnecessary here.

build/tests/scripts/visibility_test.py
bigbigxu: fixed

curvine-fuse/src/fs/state/file_handle.rs
```diff
@@ -71,16 +75,18 @@ impl FileHandle {
...
+
+                self.read_ver.set(write_ver);
```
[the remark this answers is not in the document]
bigbigxu: complete is a no-op, which is actually unnecessary here.

szbr00 (approved):

PR #862, opened 2026-06-02
chore(common): add curvine-common-macros scaffold for ClientCliArgs

lzjqsdd:
## Problem

The fuse/CSI parameterization rollout needs a derive macro foundation to generate
`ClientCliOverrides` and support `--client.*` CLI flags from `ClientConf`.

The original implementation used a top-level `curvine-config-macros` directory,
which introduced unnecessary top-level surface area.

## Design

- Keep a dedicated proc-macro crate (required by Rust for `#[derive(...)]`).
- Colocate it under `curvine-common` as `curvine-common/macros`.
- Use crate name `curvine-common-macros` to make ownership explicit.
- Keep this PR as a scaffold only: `#[derive(ClientCliArgs)]` currently emits an
  empty `ClientCliOverrides` and no-op `apply_to`.
- Defer field-level `#[client_cli(...)]` codegen to follow-up commit C2.

## Key Changes

- Workspace member changed to `curvine-common/macros`.
- Added proc-macro crate `curvine-common-macros`.
- Moved derive scaffold implementation to `curvine-common/macros/src/lib.rs`.
- Removed old top-level `curvine-config-macros` directory.
- Removed workspace-level `proc-macro2`/`quote`/`syn`; macro deps are now local to
  the macro crate.

## Expected Outcome

- `cargo check -p curvine-common-macros` passes.
- No runtime behavior changes in fuse/CSI paths.
- Macro ownership is colocated with the `curvine-common` configuration domain.
- Unblocks C2 (real field-level macro generation).

## Test Plan

- [x] `cargo check -p curvine-common-macros`
- [ ] Integrate derive into `ClientConf` in C2 and validate generated overrides


lzjqsdd:
Tracking issue for the full parameterization rollout: #865 (C1/C7/C12 are the parallel scaffolding commits for this epic).

curvine-common/macros/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+#[proc_macro_derive(ClientCliArgs, attributes(client_cli))]
+pub fn derive_client_cli_args(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let conf_name = &input.ident;
+
+    let overrides_name = syn::Ident::new("ClientCliOverrides", conf_name.span());
+
+    let expanded = quote! {
+        #input
+
```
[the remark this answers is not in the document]
lzjqsdd: Thanks — agreed. Fixed in dcd49f9: the derive expansion now emits only the generated `ClientCliOverrides` struct and `apply_to` impl, without re-emitting `#input`.

curvine-common/macros/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+        #[derive(Debug, Default, Clone, PartialEq, Eq)]
+        pub struct #overrides_name {}
+
+        impl #overrides_name {
+            /// Applies CLI overrides onto the target configuration struct.
+            pub fn apply_to(&self, _target: &mut #conf_name) -> Result<(), String> {
```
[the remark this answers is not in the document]
lzjqsdd: Thanks — agreed. Fixed in dcd49f9: `apply_to` now returns `orpc::CommonResult<()>` to align with existing config APIs in `curvine-common` (e.g. `ClientConf::init`).

lzjqsdd:
Review follow-up: both inline comments addressed in dcd49f9. Replies posted on each thread with rationale.

curvine-common/macros/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// Derives an empty `ClientCliOverrides` companion struct and a no-op `apply_to` impl.
+///
+/// Field-level `#[client_cli(...)]` attributes will be handled in a later commit.
+#[proc_macro_derive(ClientCliArgs, attributes(client_cli))]
+pub fn derive_client_cli_args(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let conf_name = &input.ident;
+
+    let overrides_name = syn::Ident::new("ClientCliOverrides", conf_name.span());
```
szbr00: The generated struct is unconditionally named `ClientCliOverrides`, regardless of which type the derive is applied to. As soon as a second configuration type (e.g. `ClusterConf` or any other type in the same module / re-export scope) also derives `ClientCliArgs`, the macro will emit two `pub struct ClientCliOverrides {}` definitions and the crate will fail to compile with a duplicate definition error.

Since the doc comment on this derive explicitly anticipates use beyond `ClientConf` ("for Curvine configuration structs"), it would be safer to base the companion name on the input type now, e.g. `format_ident!("{}CliOverrides", conf_name)`, so C2 does not have to introduce a breaking rename later.

curvine-common/macros/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let expanded = quote! {
+        #[derive(Debug, Default, Clone, PartialEq, Eq)]
+        pub struct #overrides_name {}
+
+        impl #overrides_name {
+            /// Applies CLI overrides onto the target configuration struct.
+            pub fn apply_to(&self, _target: &mut #conf_name) -> orpc::CommonResult<()> {
```
szbr00: `orpc::CommonResult<()>` is emitted unqualified, which means the calling crate must have `orpc` resolvable as a top-level path in the scope where `#[derive(ClientCliArgs)]` is used. If a downstream crate renames the dependency (`orpc = { package = "...", ... }`) or does not have a direct dependency on `orpc` at all (only transitive through `curvine-common`), the generated `apply_to` will fail to compile with a confusing "use of undeclared crate or module `orpc`" error pointing at the derive call site.

Standard proc-macro hygiene is to use absolute paths anchored at the crate root, e.g. `::orpc::CommonResult<()>`, and to document the required dependency in the macro crate README. The same applies to any helper types emitted by C2.

curvine-common/macros/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, DeriveInput};
+
+/// Derives an empty `ClientCliOverrides` companion struct and a no-op `apply_to` impl.
+///
+/// Field-level `#[client_cli(...)]` attributes will be handled in a later commit.
+#[proc_macro_derive(ClientCliArgs, attributes(client_cli))]
+pub fn derive_client_cli_args(input: TokenStream) -> TokenStream {
```
szbr00, line 25: `parse_macro_input!` accepts any `DeriveInput`, including enums and unions, but the rest of the macro (and C2's `#[client_cli(...)]` field handling) is clearly designed for structs. Right now `#[derive(ClientCliArgs)] enum E { ... }` will compile and silently emit an unrelated `ClientCliOverrides`, which will be confusing once C2 starts walking fields.

Adding an explicit guard now is essentially free and gives users a clear error before C2 lands:

```rust
if !matches!(input.data, syn::Data::Struct(_)) {
    return syn::Error::new_spanned(
        &input,
        "ClientCliArgs can only be derived for structs",
    )
    .to_compile_error()
    .into();
}
```

curvine-common/macros/src/lib.rs
lzjqsdd: Thanks — agreed. Fixed in 513d96a: the companion struct is now named `{Type}CliOverrides` via `format_ident!("{}CliOverrides", conf_name)` (e.g. `ClientConf` → `ClientConfCliOverrides`), so multiple derives in the same crate will not collide.

curvine-common/macros/src/lib.rs
lzjqsdd: Thanks — agreed. Fixed in 513d96a: `apply_to` now returns `::orpc::CommonResult<()>` so the expansion uses an absolute path and does not depend on `orpc` being in the caller's import scope.

curvine-common/macros/src/lib.rs
lzjqsdd: Thanks — agreed. Fixed in 513d96a: added a compile-time guard that rejects enum/union derives with `ClientCliArgs can only be derived for structs`.

lzjqsdd:
Addressed 3 new review comments from @szbr9486 in **513d96a**. Inline replies posted on each thread.

lzjqsdd:
**Branch restored** to scaffolding-only commits (`513d96a`). C2 work was moved off this branch to new PR (see follow-up comment). Sorry for the accidental extra commits on this PR.

szbr00 (approved):

PR #863, opened 2026-06-02
refactor(fuse): extract FuseMountArgs into cli module

lzjqsdd:
## Problem

`FuseArgs` and `get_conf()` live entirely in the `curvine-fuse` binary. Upcoming work
adds subcommands (`validate-config`, `list-config-flags`) and reuses the same mount
CLI surface; the binary should not keep growing.

## Design

- Extract mount CLI into `curvine-fuse/src/cli/`.
- Rename the struct to `FuseMountArgs` and move `get_conf()` unchanged (behavior-neutral).
- Keep the binary as a thin `main`: parse args, init logging, run the FUSE session.
- No CLI field or mount behavior changes in this PR; prepares C8 (subcommands) and
  C11 (`ClientCliOverrides` flatten).

## Key Changes

- Add `curvine-fuse/src/cli/mount_args.rs` (`FuseMountArgs`, `get_conf()`).
- Add `curvine-fuse/src/cli/mod.rs` and `pub mod cli` in `lib.rs`.
- Slim `curvine-fuse/src/bin/curvine-fuse.rs` to `FuseMountArgs::parse()`.

## Expected Outcome

- Existing `curvine-fuse` flags and mount behavior are unchanged.
- CLI modules can grow without further inflating the binary.

## Test Plan

- [ ] `cargo check -p curvine-fuse`
- [ ] Smoke: `cargo run --bin curvine-fuse -- --help` matches pre-refactor output


lzjqsdd:
Tracking issue for the full parameterization rollout: #865 (C1/C7/C12 are the parallel scaffolding commits for this epic).

lzjqsdd:
Thanks for the review.

No inline comments were raised. For transparency:

- **Why extract to `cli/`**: Prepare for upcoming subcommands (`validate-config`, `list-config-flags`) and `ClientCliOverrides` flatten without further inflating the binary. This PR is behavior-neutral.
- **Why rename to `FuseMountArgs`**: Distinguish mount-specific flags from the future top-level `FuseCli` that will host subcommands. CLI flags and `get_conf()` logic are unchanged.
- **If merged alone**: Existing `curvine-fuse` startup parameters and mount behavior remain the same; only module layout differs.

No code changes requested by review at this time.

lzjqsdd:
Rebased onto latest `main` and resolved merge conflict in `curvine-fuse/src/bin/curvine-fuse.rs`.

**Conflict resolution**
- Kept the PR refactor (thin binary + `FuseMountArgs` in `cli/`).
- Regenerated `curvine-fuse/src/cli/mount_args.rs` from current `main` `FuseArgs` so CLI fields and `get_conf()` match upstream (including removal of `direct_io_on_cache_miss`, which is no longer on `main`).

**Verification**
- `cargo check -p curvine-fuse` passes after rebase.

Branch force-pushed; PR should now be mergeable with `main`.

curvine-fuse/src/cli/mount_args.rs
```diff
@@ -0,0 +1,332 @@
...
+#[derive(Debug, Parser, Clone)]
+#[command(version = version::VERSION)]
+pub struct FuseMountArgs {
+    // Mount the mount point, mount the file system to a directory of the machine.
+    #[arg(long, help = "Mount point path (default: /curvine-fuse)")]
+    pub mnt_path: Option<String>,
+
+    // Specify the root path of the mount point to access the file system, default "/"
+    #[arg(long, help = "Remote filesystem path (default: /)")]
+    pub(crate) fs_path: Option<String>,
```
szbr00: Before this PR these `pub(crate)` markers on `fs_path` (L31), `debug` (L39), `conf` (L48), and `options` (L113) effectively meant "private to the curvine-fuse binary crate" — `bin/curvine-fuse.rs` was the entire crate, so no other code could see them. After the move into the library, `pub(crate)` now means "visible to the entire curvine-fuse library", which is a real visibility expansion that contradicts the "behavior-neutral" framing in the PR description. Any other module in `curvine-fuse` (e.g. `fs::*`, `session::*`, `web_server::*`) can now read and mutate these fields directly without going through `get_conf()`.

Since `FuseMountArgs` is now a public library type, please make the visibility intent explicit: either drop `pub(crate)` to plain private (and add accessor methods if C8 needs read-only access), or bump to `pub` if the intent is to expose them. Inherited `pub(crate)` from the binary context is the worst of both worlds.

lzjqsdd: Thanks — agreed. Fixed in 6202b49: `fs_path`, `debug`, `conf`, and `options` are now private fields on `FuseMountArgs` again. When the struct lived in the binary crate, `pub(crate)` was effectively binary-private; after moving into the library it would have exposed those fields to all of `curvine-fuse`. External callers still use the public CLI surface via clap parsing and `get_conf()`.

lzjqsdd:
Addressed new review comment from @szbr9486 in **6202b49** (private fields on `FuseMountArgs`). Inline reply posted on the thread.

lzjqsdd:
**Branch restored** to C7-only commits (`6202b49`, includes rebase + visibility review fix). C8 was moved to a new branch/PR.

szbr00 (approved):

PR #867, opened 2026-06-03
fix: keep curvine-lancedb out of default build

jlon:
## Summary
- Rename the LanceDB facade workspace member from `curvine-lancedb-rs` to `curvine-lancedb`.
- Add explicit `default-members` so the LanceDB crate stays out of default workspace builds.
- Gate the Curvine tests LanceDB E2E target and dependencies behind the `lancedb` feature.

## Test Plan
- `cargo metadata --no-deps --format-version 1`
- `git diff --check`
- `cargo check`
- `cargo check -p curvine-lancedb`
- `cargo check -p curvine-tests --features lancedb --test lancedb_object_store_e2e`
- `cargo test -p curvine-lancedb`
- `cargo test -p curvine-tests --no-run`


Cargo.toml
```diff
@@ -23,6 +23,19 @@ members = [
     "curvine-cli"
 ]
 
+default-members = [
```
szbr00, line 28: `default-members` now has to be hand-kept in sync with `members`. The two lists are identical here except for `curvine-lancedb`, but as soon as a new workspace crate is added to `members` (e.g. the `curvine-common/macros` crate currently moving through #862, or any future `curvine-foo`), forgetting to also add it to `default-members` will silently exclude it from `cargo build`/`cargo clippy`/`cargo test` runs without `--workspace`, including the CI commands in `.github/workflows/build.yml` (L78 `cargo clippy --all-targets`, L96 `cargo build`). The omission will only surface as "feature X not tested" much later.

Two low-cost mitigations, pick one:

1. Add an inline comment above `default-members` like `# Mirror of members minus crates that should be opt-in (curvine-lancedb pulls heavy lance/arrow deps).` so the rule is visible at the edit site.
2. Add a tiny CI check (or an xtask) that asserts `default-members` is a superset of `members` minus the opt-in set, so the desync becomes a hard failure instead of a silent skip.

Either option keeps the current behavior but removes the silent footgun.

jlon: Fixed by adding an inline comment above `default-members`: it should mirror `members` unless a crate is intentionally opt-in, and `curvine-lancedb` remains opt-in because it pulls the heavy Lance/Arrow dependency set.

szbr00 (approved):

Issue #865, opened 2026-06-02
Tracking: curvine-fuse / curvine-csi StorageClass parameterization (C1–C18)


lzjqsdd:
## Summary

Track the end-to-end work to align Kubernetes StorageClass / PV parameters with `curvine-fuse` CLI flags, replace CSI fuse-parameter whitelists with reserved-key + passthrough semantics, and add `validate-config` / `list-config-flags` subcommands. Work is split into **18 small, single-module commits (C1–C18)** across five tracks.

---

## Background

Today, `curvine-csi` maintains `supportedFuseParams` — a hand-curated map of fuse flags that easily drifts from `curvine-fuse` CLI definitions. Operators cannot use new fuse or client tuning knobs from StorageClass without a CSI code change.

Goals:

- **Fuse flags**: flat `--kebab-case` on `curvine-fuse` (e.g. `--io-threads`, `--entry-timeout`).
- **Client flags**: `--client.<name>` (e.g. `--client.block-size`), generated from `ClientConf` via proc-macro — CSI does not maintain a client whitelist.
- **CSI passthrough**: reserved keys handled explicitly; all other non-empty SC/PV keys forwarded as `--key value`.
- **Validation**: `curvine-fuse validate-config` (dry-run via `get_conf()` + `init()`, no mount); CSI runs it at CreateVolume (dynamic) / NodeStage (static PV) and surfaces **stderr** on failure — no log scraping.
- **`mnt-path`**: computed only inside the node (`computeFuseMntPath`); user-supplied `mnt-path` on SC/PV → `InvalidArgument`. No legacy compatibility paths.

---

## Design principles

1. **Single source of truth for CLI shape** — clap on fuse; `#[derive(ClientCliArgs)]` on `ClientConf` for client overrides.
2. **CSI does not curate fuse/client flag lists** — reserved keys + denylist + passthrough only.
3. **One argv builder in CSI** — `pkg/csi/fuse_cli.go` shared by embedded and standalone modes.
4. **Small, reviewable PRs** — one primary module per commit; no mixed macros + common + fuse + csi changes.
5. **Deterministic behavior** — sorted passthrough keys; full passthrough map in `GenerateMountKeyWithFuseParams`.
6. **Fail fast at provision time** — invalid parameters rejected before PVC binds (dynamic) or before fuse starts (static).

```mermaid
flowchart TB
  SC["StorageClass / PV"]
  CSI["curvine-csi"]
  VF["curvine-fuse validate-config"]
[...]

PR #872, opened 2026-06-03
feat(fuse): add mount subcommand via FuseCli (C8)

lzjqsdd:
## Problem

C7 landed in #863. Next step is a top-level `FuseCli` entrypoint so C9/C10 can add
more subcommands without growing the binary.

## Design

- Introduce `FuseCli` with optional `mount` subcommand.
- Keep backward compatibility: bare `curvine-fuse --conf ...` still runs mount.
- `Mount(FuseMountArgs)` accepts flags after `mount`; `resolve_mount_args` picks subcommand vs top-level.
- `args_conflicts_with_subcommands = true` rejects mixed top-level flags + subcommand.
- Move mount runtime into `cli::run_mount`.

## Key Changes

- Add `curvine-fuse/src/cli/fuse_cli.rs`, `mount_run.rs`.
- Update `curvine-fuse` binary to parse `FuseCli` and call `run_mount`.
- Add `fuse_cli` unit tests for parsing precedence/validation.

## Review follow-ups (szbr9486)

- Mixed `--flag ... mount` now hard-errors via clap instead of silently dropping flags.
- Unit tests lock bare/subcommand/mixed/unknown-subcommand behavior.

## Base

Rebased onto latest `main`, including merged #863.

## Test Plan

- [x] `cargo check -p curvine-fuse`
- [x] `cargo test -p curvine-fuse fuse_cli`

Refs #865

### DIFF
--- a/curvine-fuse/src/bin/curvine-fuse.rs
+++ b/curvine-fuse/src/bin/curvine-fuse.rs
@@ -13,54 +13,17 @@
 // limitations under the License.
 
 use clap::Parser;
-use curvine_fuse::cli::FuseMountArgs;
-use curvine_fuse::fs::CurvineFileSystem;
-use curvine_fuse::session::FuseSession;
-use curvine_fuse::web_server::WebServer;
-use orpc::common::Logger;
-use orpc::runtime::{AsyncRuntime, RpcRuntime};
-use orpc::CommonResult;
-use std::sync::Arc;
+use curvine_fuse::cli::{run_mount, FuseCli};
+use orpc::{err_box, CommonResult};
 
 // fuse mount.
 // Debugging, after starting the cluster, execute the following naming, mount fuse
 // umount -f /curvine-fuse; cargo run --bin curvine-fuse -- --conf /server/conf/curvine-cluster.toml
 fn main() -> CommonResult<()> {
-    let args = FuseMountArgs::parse();
-    println!("fuse args {:?}", args);
-
-    // Ignore SIGPIPE to prevent unexpected termination when peer closes
-    unsafe {
-        libc::signal(libc::SIGPIPE, libc::SIG_IGN);
+    let cli = FuseCli::parse();
+    if cli.runs_mount() {
+        run_mount(cli.resolve_mount_args())
+    } else {
+        err_box!("unsupported curvine-fuse subcommand")
     }
-
-    let cluster_conf = args.get_conf()?;
-    Logger::init(cluster_conf.fuse.log.clone());
-    cluster_conf.print();
-
-    let rt = Arc::new(AsyncRuntime::new(
-        "curvine-fuse",
-        cluster_conf.fuse.io_threads,
-        cluster_conf.fuse.worker_threads,
-    ));
-
-    let fuse_rt = rt.clone();
-
-    rt.block_on(async move {
-        let fs = CurvineFileSystem::new(cluster_conf, fuse_rt.clone()).unwrap();
-        let conf = fs.conf().clone();
-
-        let node_state = fs.state().clone();
-        let web_port = conf.web_port;
-        fuse_rt.spawn(async move {
-            if let Err(e) = WebServer::start(web_port, node_state).await {
-                log::error!("Failed to start metrics server: {}", e);
-            }
-        });
-
-        let mut session = FuseSession::new(fuse_rt.clone(), fs, conf).await.unwrap();
-        session.run().await
-    })?;
-
-    Ok(())
 }
--- a/curvine-fuse/src/cli/fuse_cli.rs
+++ b/curvine-fuse/src/cli/fuse_cli.rs
@@ -1,0 +1,88 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::{Parser, Subcommand};
+use curvine_common::version;
+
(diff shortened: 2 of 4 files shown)


lzjqsdd:
Rebased onto latest `main` (includes merged #863). Branch now contains **1 commit** (C8 only).

lzjqsdd:
Addressed Copilot review comments in a02e292. Rebased onto latest `main` (951ff59).

curvine-fuse/src/cli/fuse_cli.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Mount the curvine filesystem (also the default when omitted)
+    Mount,
+}
+
+impl FuseCli {
+    /// Returns true when the parsed invocation should run the mount flow.
+    pub fn runs_mount(&self) -> bool {
+        matches!(self.cmd, None | Some(FuseSubcommand::Mount))
+    }
+}
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in a02e292: `FuseSubcommand::Mount` now carries `FuseMountArgs`, so `curvine-fuse mount --conf ...` parses mount flags after the subcommand.

curvine-fuse/src/bin/curvine-fuse.rs
```diff
@@ -13,54 +13,17 @@
...
-    println!("fuse args {:?}", args);
-
-    // Ignore SIGPIPE to prevent unexpected termination when peer closes
-    unsafe {
-        libc::signal(libc::SIGPIPE, libc::SIG_IGN);
+    let cli = FuseCli::parse();
+    if cli.runs_mount() {
+        run_mount(cli.mount)
+    } else {
+        err_box!("unsupported curvine-fuse subcommand")
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in a02e292: the binary calls `run_mount(cli.resolve_mount_args())`, which prefers subcommand args when `mount` is used and falls back to top-level flags otherwise.

curvine-fuse/src/cli/fuse_cli.rs
```diff
@@ -0,0 +1,50 @@
...
+impl FuseCli {
+    /// Returns true when the parsed invocation should run the mount flow.
+    pub fn runs_mount(&self) -> bool {
+        matches!(self.cmd, None | Some(FuseSubcommand::Mount(_)))
+    }
+
+    /// Returns mount args from the subcommand when present, otherwise top-level flags.
+    pub fn resolve_mount_args(&self) -> FuseMountArgs {
+        match &self.cmd {
+            Some(FuseSubcommand::Mount(args)) => args.clone(),
```
szbr00, line 51: This resolution rule silently drops user-provided top-level flags when the `mount` subcommand is also present. Because `FuseMountArgs` is both `#[command(flatten)]`-ed at the top level *and* wrapped in `FuseSubcommand::Mount(FuseMountArgs)`, clap will happily parse:

- `curvine-fuse --conf /A` -> top-level only -> `cli.mount.conf = "/A"`, picked correctly
- `curvine-fuse mount --conf /B` -> subcommand only -> subcommand args picked correctly
- `curvine-fuse --conf /A mount` -> top-level `--conf /A` parsed, then subcommand `mount` with all `FuseMountArgs` defaults. `resolve_mount_args` returns the subcommand arm (defaults), **silently discarding `/A`** without any error or warning.

This is an easy footgun for anyone migrating from the pre-subcommand invocation (where `--conf` worked anywhere) to the new `mount`-explicit form. Two clean fixes:

1. Add `#[command(args_conflicts_with_subcommands = true)]` on `FuseCli` (L22) — clap will then hard-error when both top-level flags and a subcommand are given, eliminating the ambiguity entirely.
2. Or have `resolve_mount_args` merge the two: take the subcommand args as the base and overlay any explicitly-set top-level field on top, so `--conf /A mount` keeps `/A`.

Option (1) is simpler and matches the "subcommand or bare invocation, never both" mental model the PR description implies.

curvine-fuse/src/cli/fuse_cli.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+impl FuseCli {
+    /// Returns true when the parsed invocation should run the mount flow.
+    pub fn runs_mount(&self) -> bool {
+        matches!(self.cmd, None | Some(FuseSubcommand::Mount(_)))
+    }
+
+    /// Returns mount args from the subcommand when present, otherwise top-level flags.
+    pub fn resolve_mount_args(&self) -> FuseMountArgs {
```
szbr00, line 49: The whole point of `resolve_mount_args` is to express a non-obvious precedence rule between flattened top-level flags and the `mount` subcommand, but there is no unit test pinning that behavior down. Without tests, the precedence-drop issue at L46 (and any future regression when C9/C10 add more subcommands) would only surface as a runtime "why is my `--conf` being ignored?" report.

A handful of `FuseCli::try_parse_from(&["curvine-fuse", ...])` cases covering at least:

- bare invocation with a top-level flag -> flag is preserved
- `mount` subcommand with a flag -> flag is preserved
- top-level flag + `mount` subcommand (mixed) -> expected resolution (after fixing the design per the L46 comment)
- unknown subcommand -> expected error

would lock in the contract cheaply, and the same harness can be reused as C9/C10 add `validate-config` / etc.

curvine-fuse/src/cli/fuse_cli.rs
lzjqsdd: Addressed in 86bc090 (option 1): added `args_conflicts_with_subcommands = true` on `FuseCli`, so clap errors when top-level mount flags and a subcommand are mixed instead of silently dropping top-level values.

curvine-fuse/src/cli/fuse_cli.rs
lzjqsdd: Addressed in 86bc090: added unit tests for bare invocation, `mount` subcommand parsing, mixed top-level+subcommand rejection, and unknown subcommand errors.

lzjqsdd:
Addressed szbr9486 review in 86bc090:
- `args_conflicts_with_subcommands = true` on `FuseCli`
- Unit tests for bare/subcommand/mixed/unknown-subcommand parsing

Rebased on latest `main`.

szbr00 (approved):

PR #880, opened 2026-06-04
fix(lancedb): clean dropped Curvine table dirs

jlon:


curvine-lancedb/src/object_store.rs
```diff
@@ -518,27 +519,64 @@ impl ObjectStoreTrait for CurvineObjectStore {
...
-                .map_err(|e| fs_error_to_object_store(location, e)),
-            Err(e) if is_not_found_error(&e) => Ok(()),
-            Err(e) => Err(fs_error_to_object_store(location, e)),
-        };
-        let _ = self.release_object_write_lock(&lock).await;
-        result?;
-        if !self.is_root_workspace() {
-            let _ = self.prune_empty_parents(&cv_path, location).await;
-        }
+        self.delete_object_without_prune(location).await?;
```
szbr00: Previously `delete()` pruned empty parent directories up to the workspace root after a successful delete. After this refactor, `delete()` only calls `delete_object_without_prune`, and directory pruning lives exclusively in `delete_stream` (and is now scoped to `*.lance` dataset roots). This means object removals that go through the single `delete()` path will leave empty directories behind, which is a behavior change from before.

Is the assumption that LanceDB always cleans datasets via `delete_stream`? If a caller can remove objects via individual `delete()` calls, empty dirs will accumulate. Worth confirming this is intentional, or routing single deletes through the same dataset-prune logic.

jlon: Handled by routing single-object deletes through a narrow Lance parent cleanup path. For *.lance objects it deletes the object first, then prunes only empty parents from that object up to the dataset root. delete_stream keeps the whole-dataset cleanup path for LanceDB remove_dir_all/drop_table.

curvine-lancedb/src/object_store.rs
```diff
@@ -1251,35 +1318,43 @@ impl CurvineObjectStore {
...
-            source: e.to_string().into(),
-        })?;
+    async fn prune_empty_lance_dataset_dirs(
+        &self,
+        dataset_root: &CurvinePath,
+        location: &Path,
+    ) -> OsResult<()> {
+        if !is_lance_dataset_path(dataset_root) {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
jlon: Fixed by guarding Lance pruning with get_status. Non-directory and not-found .lance roots are now treated as no-op prune targets, so deleting a regular object named *.lance does not fail after the object delete succeeds. Added live object-store coverage for deleting regular_file.lance as a plain object.

szbr00 (approved):

PR #891, opened 2026-06-08
fix(fuse): enable negative dentry cache for ENOENT lookups

ljy1-lixing:


ljy1-lixing:
## Summary

Fix the FUSE lookup handler to properly return negative dentry entries when a file does not exist, allowing the kernel to cache "not found" results and avoid redundant RPCs to Master.

## Motivation

Previously, when `get_cached_status()` returned `ENOENT` (file not found), the error was immediately propagated via the `?` operator as `fuse_out_header { error: -2 }`. This caused two problems:

1. **Negative cache bypass**: The kernel cannot cache a negative dentry from an error response. It only caches negative entries from a successful response with `nodeid=0` and `entry_valid > 0`. This meant every repeated access to a non-existent path (e.g., shell checking for command names in cwd) triggered a full Master RPC round-trip.

2. **Unnecessary NodeMap pollution avoided**: By returning early at the `get_cached_status` stage, the code no longer reaches `lookup_status()` → `find_node()`, which would allocate a new inode for a non-existent file and never reclaim it (phantom node leak).

Real-world impact observed in production logs: shell operations like `ls` in a FUSE mount directory generated duplicate `LOOKUP{name="ls"}` requests with full RPC calls (~100μs each), because the kernel could not cache the "not found" result.

## Fix

Replace the `?` propagation with an explicit `match` in `curvine_file_system.rs::lookup()`:

```rust
let status = match self.get_cached_status(&path).await {
    Ok(s) => s,
    Err(e) if e.errno == libc::ENOENT && !self.conf.negative_ttl.is_zero() => {
        return Ok(fuse_entry_out {
            entry_valid: self.conf.negative_ttl.as_secs(),
            entry_valid_nsec: self.conf.negative_ttl.subsec_nanos(),
            ..Default::default()  // nodeid=0 signals negative entry
        });
    }
    Err(e) => return Err(e),
};
```

When a path does not exist and `negative_ttl > 0`:

- Return `Ok(fuse_entry_out { nodeid: 0, entry_valid: negative_ttl })` instead of `Err(ENOENT)`
- The kernel caches this as a negative dentry for `entry_valid` seconds
- Subsequent lookups for the same name are served directly by the kernel without reaching user-space

## Benefits

- Eliminates redundant Master RPCs for non-existent paths (measured: ~100μs saved per repeated lookup)
- Reduces FUSE daemon CPU usage under workloads with frequent failed lookups (e.g., shell tab completion, command resolution in cwd)
- Prevents NodeMap phantom node allocation for non-existent files
- Aligns with FUSE protocol best practices (libfuse reference implementation uses the same pattern)
- Configurable via existing `negative_timeout` setting (default: 0, no behavior change unless explicitly enabled)

## Test Plan

1. **Unit verification** (with `negative_timeout = 1.0`):

```bash
# Repeated stat on non-existent file
for i in $(seq 1 5); do stat /mnt/cv/nonexist 2>/dev/null; done
# Expected: only 1 FUSE_LOOKUP + GetStatus RPC in fuse.log (others served by kernel cache)
```

2. **No regression for create-after-lookup**:

```bash
stat /mnt/cv/newfile 2>/dev/null   # ENOENT, negative cached
touch /mnt/cv/newfile              # O_CREAT bypasses negative cache
stat /mnt/cv/newfile               # Should succeed immediately
```

3. **Disabled by default**: When `negative_timeout = 0` (default), behavior is identical to before the fix (`Err(ENOENT)` propagated as `error=-2`).

szbr00 (approved):

PR #901, opened 2026-06-09
fix(client): handle sparse parallel readers for random reads

jlon:
## Problem
StarRocks uses positional reads through Curvine. With `fs.cv.read_parallel=4`, small files can create fewer parallel slice readers than the configured parallelism. `FsReaderParallel::create_all()` skips empty slice readers, so a small file may produce only reader `0` plus the full-file base reader.

The random-read path incorrectly used the configured `read_parallel` value as the reader index. For the small-file case above, that asks for reader `4` even though the reader vector only contains `[slice reader 0, base reader]`, causing `reader 4 is not initialized`.

## Fix
- Record the full-file base reader index when `FsReaderBuffer` appends it.
- Route random reads to that base reader instead of deriving an index from configured parallelism.
- Keep sequential reads on the existing slice-based selection path.
- Normalize non-positive `read_parallel` and `max_read_parallel` values to the existing safe defaults during `ClientConf::init()`.
- Include selector inputs in the invalid-reader error path for easier diagnosis.

## Tests
- Added a construction-level regression test for a small file with `read_parallel=4`; it builds the sparse reader layout and verifies the random path resolves through the recorded base reader. I also checked this test against the old read_parallel-as-index logic and it fails there.
- `cargo fmt -p curvine-client -p curvine-common --check`
- `cargo test -p curvine-common --test client_conf_cli_test`
- `cargo test -p curvine-client fs_reader_buffer`
- `cargo test -p curvine-client fs_reader_parallel`
- `cargo clippy -p curvine-client -p curvine-common --all-targets -- -D warnings`

### DIFF
--- a/curvine-client/src/file/fs_reader_buffer.rs
+++ b/curvine-client/src/file/fs_reader_buffer.rs
@@ -134,6 +134,7 @@ impl ReaderAdapter {
 // Reader with buffer.
 pub struct FsReaderBuffer {
     readers: Vec<ReaderAdapter>,
+    base_reader_index: usize,
     path: Path,
     pos: i64,
     len: i64,
@@ -208,10 +209,12 @@ impl FsReaderBuffer {
             readers.push(reader);
         }
 
+        let base_reader_index = readers.len();
         readers.push(ReaderAdapter::Base(base));
 
         let reader = Self {
             readers,
+            base_reader_index,
             path,
             pos,
             len,
@@ -245,14 +248,42 @@ impl FsReaderBuffer {
         &self.path
     }
 
+    fn select_reader_index(
+        is_random: bool,
+        pos: i64,
+        slice_size: i64,
+        read_parallel: i64,
+        base_reader_index: usize,
+    ) -> Option<usize> {
+        if is_random {
+            return Some(base_reader_index);
+        }
+
+        if slice_size <= 0 || read_parallel <= 0 {
+            return None;
+        }
+
+        Some((pos / slice_size % read_parallel) as usize)
+    }
+
     fn get_reader(&mut self) -> FsResult<&mut ReaderAdapter> {
-        let id = if self.read_detector.is_random() {
-            self.read_detector.read_parallel()
-        } else {
-            self.pos / self.slice_size % self.read_detector.read_parallel()
+        let Some(id) = Self::select_reader_index(
+            self.read_detector.is_random(),
+            self.pos,
+            self.slice_size,
+            self.read_detector.read_parallel(),
+            self.base_reader_index,
+        ) else {
+            return err_box!(
+                "reader is not initialized: pos={}, slice_size={}, read_parallel={}, base_reader_index={}",
+                self.pos,
+                self.slice_size,
+                self.read_detector.read_parallel(),
+                self.base_reader_index
+            );
         };
 
-        match self.readers.get_mut(id as usize) {
+        match self.readers.get_mut(id) {
             Some(v) => Ok(v),
             None => err_box!("reader {} is not initialized", id),
         }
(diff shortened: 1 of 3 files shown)


curvine-client/src/file/fs_reader_buffer.rs
```diff
@@ -405,3 +430,22 @@ impl FsReaderBuffer {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
szbr00, line 458: This test validates `select_reader_index` as a pure function with a hard-coded `base_reader_index = 1`, but the original bug was in the wiring: `create()` must set `base_reader_index` to the sparse position (`k < read_parallel`) for a small file, and the random path must then resolve to it. The pure-function test would still pass even if `base_reader_index` were computed incorrectly.

Consider adding a regression test at the `FsReaderBuffer`/integration level that builds a reader over a small file with `read_parallel = 4` and asserts a random read succeeds (and/or that `base_reader_index` equals the number of non-empty slice readers), which is what actually reproduces `reader 4 is not initialized`.

curvine-client/src/file/fs_reader_buffer.rs
```diff
@@ -245,14 +248,36 @@ impl FsReaderBuffer {
...
-        } else {
-            self.pos / self.slice_size % self.read_detector.read_parallel()
+        let Some(id) = Self::select_reader_index(
+            self.read_detector.is_random(),
+            self.pos,
+            self.slice_size,
+            self.read_detector.read_parallel(),
+            self.base_reader_index,
+        ) else {
+            return err_box!("reader is not initialized");
```
szbr00: When `select_reader_index` returns `None`, this surfaces `"reader is not initialized"` without any context, unlike the sibling `"reader {} is not initialized"`. Since this only triggers on a non-positive `slice_size`/`read_parallel`, consider including those values in the message to aid debugging if the `init()` normalization is ever bypassed.

curvine-client/src/file/fs_reader_buffer.rs
szbr00, line 458: This test validates `select_reader_index` as a pure function with a hard-coded `base_reader_index = 1`, but the original bug was in the wiring: `create()` must set `base_reader_index` to the sparse position (`k < read_parallel`) for a small file, and the random path must then resolve to it. The pure-function test would still pass even if `base_reader_index` were computed incorrectly.

Consider adding a regression test at the `FsReaderBuffer`/integration level that builds a reader over a small file with `read_parallel = 4` and asserts a random read succeeds (and/or that `base_reader_index` equals the number of non-empty slice readers), which is what actually reproduces `reader 4 is not initialized`.

curvine-client/src/file/fs_reader_buffer.rs
jlon: Fixed. The None path now includes pos, slice_size, read_parallel, and base_reader_index in the error message, so config/init bypasses have enough context to diagnose the bad selector inputs.

curvine-client/src/file/fs_reader_buffer.rs
jlon: Fixed. Replaced the pure selector test with a construction-level regression test for the sparse-reader case: small file, read_parallel=4, one non-empty slice reader, and the base reader at index 1. The test seeks to EOF to switch into random mode without touching real block IO, then verifies get_reader() resolves through the recorded base_reader_index. I also checked the test against the old read_parallel-as-index logic and it fails there.

jlon: Fixed in the same update: the test now constructs FsReaderBuffer for the sparse small-file layout and validates the random path through get_reader(), instead of only calling select_reader_index with a hard-coded base index.

szbr00 (approved):

PR #904, opened 2026-06-10
fix(master): preserve directory edge invariants

jlon:
## Problem

Directory namespace edge name is the key in the parent children index, but rename/delete/unlink paths removed edges with the child inode internal name. If an old snapshot or previous bug left edge name and inode name out of sync, the old edge stayed behind. The same directory inode could then be reachable from multiple directory edges, and checkpoint restore failed later with errors such as `Child ... already exists`.

## Fix

- Pass the path edge name from `FsDir` into rename/delete/unlink store operations.
- Delete namespace edges by the actual edge name, not by duplicated inode name metadata.
- Add inode id consistency checks on rename and file-entry unlink.
- During tree restore, reject directory inodes with multiple parent edges and hydrate directory name/parent from the edge. This does not skip dirty data; it prevents new dirty edges at the source and makes illegal directory aliases fail clearly.

## Verification

- `cargo test -p curvine-server master::meta::store::inode_store::tests -- --nocapture`
- `cargo test -p curvine-server --test master_fs_test rename -- --nocapture`
- `cargo test -p curvine-server --test master_fs_test filesystem_metadata -- --nocapture`
- `cargo test -p curvine-server --test inode_test -- --nocapture`

### DIFF
--- a/curvine-server/src/master/meta/store/inode_store.rs
+++ b/curvine-server/src/master/meta/store/inode_store.rs
@@ -14,7 +14,7 @@
 
 use crate::master::fs::DeleteResult;
 use crate::master::meta::inode::ttl::TtlBucketList;
-use crate::master::meta::inode::{InodeFile, InodePath, InodeView, ROOT_INODE_ID};
+use crate::master::meta::inode::{Inode, InodeFile, InodePath, InodeView, ROOT_INODE_ID};
 use crate::master::meta::store::{InodeWriteBatch, RocksInodeStore};
 use crate::master::meta::{FileSystemStats, FsDir, LockMeta};
 use curvine_common::rocksdb::{DBConf, RocksUtils};
@@ -74,19 +74,24 @@ impl InodeStore {
         Ok(())
     }
 
-    pub fn apply_delete(&self, parent: &InodeView, del: &InodeView) -> CommonResult<DeleteResult> {
+    pub fn apply_delete(
+        &self,
+        parent: &InodeView,
+        del: &InodeView,
+        del_name: &str,
+    ) -> CommonResult<DeleteResult> {
         let mut batch = self.store.new_batch();
         batch.write_inode(parent)?;
 
         let mut stack = LinkedList::new();
-        stack.push_back((parent.id(), del.clone()));
+        stack.push_back((parent.id(), del_name.to_string(), del.clone()));
         let mut del_res = DeleteResult::new();
         let mut deleted_files = 0i64;
         let mut deleted_dirs = 0i64;
 
-        while let Some((parent_id, inode)) = stack.pop_front() {
+        while let Some((parent_id, edge_name, inode)) = stack.pop_front() {
             // Delete inode edges
-            batch.delete_child(parent_id, inode.name())?;
+            batch.delete_child(parent_id, &edge_name)?;
             del_res.inodes += 1;
 
             match &inode {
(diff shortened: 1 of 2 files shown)


curvine-server/src/master/meta/store/inode_store.rs
```diff
@@ -410,12 +436,33 @@ impl InodeStore {
...
-                self.ttl_bucket_list.add(&store_inode);
+                let parent_id = parent.id();
+                let edge_name = file_entry.name().to_string();
+
+                let inode = match store_inode {
+                    InodeView::Dir(_) => {
+                        if let Some((old_parent_id, old_name)) =
+                            dir_edges.insert(child_id, (parent_id, edge_name.clone()))
+                        {
+                            return err_box!(
```
szbr00: `restore()` propagates this error (fs_dir.rs `restore()` -> `create_tree()?`), so once a checkpoint/snapshot already contains a directory inode reachable from multiple edges — exactly the corruption this PR targets — the master can no longer restore from it and will fail to start, with no automated recovery path.

A few lines above, an orphaned edge (inode missing) is handled leniently with a warn + `continue`. Consider an analogous recovery option here: keep the first edge, drop/skip the duplicate(s) with a loud warning (and ideally schedule edge cleanup), instead of aborting the entire restore. If a hard failure is intentional, please document an offline repair procedure so operators whose existing checkpoint is already corrupted have a way forward after upgrading.

curvine-server/src/master/meta/store/inode_store.rs
```diff
@@ -410,12 +436,33 @@ impl InodeStore {
...
+                                parent_id,
+                                edge_name
+                            );
+                        }
 
-                let inode = if matches!(store_inode, InodeView::Dir(_)) {
-                    store_inode
-                } else {
-                    file_entry
+                        store_inode.change_name(edge_name);
```
szbr00: This hydration corrects the in-memory tree, but it only mutates the in-flight `store_inode`; the persisted inode record in RocksDB keeps its stale name/parent_id. Since create_tree never rewrites it, the underlying metadata stays inconsistent — it's re-read and re-corrected on every restore and carried into the next checkpoint unchanged. That's fine as a read-time normalization, but consider rewriting the inode here (or via a one-time repair pass) so the stored record is actually healed rather than perpetually masked.

curvine-server/src/master/meta/store/inode_store.rs
```diff
@@ -314,15 +330,24 @@ impl InodeStore {
...
+        child_name: &str,
         inode_id: i64,
     ) -> CommonResult<DeleteResult> {
+        if child.id() != inode_id {
+            return err_box!(
+                "unlink file entry id mismatch: edge id {}, inode id {}",
+                child.id(),
+                inode_id
+            );
+        }
```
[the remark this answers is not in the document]
jlon: Fixed. Reworded the mismatch as a FileEntry->inode id mismatch: the entry references one inode id and the caller passed another. It no longer calls child.id() an edge id.

curvine-server/src/master/meta/store/inode_store.rs
jlon: Fixed. Directory hydration now persists to RocksDB when the stored directory name or parent_id differs from the restoring edge. It only writes when the stored metadata is actually stale, so the normal restore path does not rewrite every directory inode.

curvine-server/src/master/meta/store/inode_store.rs
jlon: Changed the restore behavior from hard failure to repair. create_tree() now keeps the first directory edge, logs the duplicate, deletes the later duplicate edge from RocksDB, and continues restoring. That lets masters start from an already-corrupted checkpoint while still repairing the namespace invariant for future restores.

szbr00 (approved):

PR #906, opened 2026-06-10
fix(master): refuse empty non-format startup

jlon:
## Problem

`JournalSystem::from_conf()` allowed `format_master=false` startup even when the configured meta and journal RocksDB data directories were empty or missing. RocksDB is configured with `create_if_missing(true)`, so an empty Kubernetes hostPath can be initialized as a blank master state instead of failing fast. This is unsafe for a StatefulSet rescheduled to a node without the original local data.

## Fix

- Add a preflight check for `format_master=false` before opening the journal system.
- Require both meta and journal RocksDB data directories to look like existing RocksDB databases by checking for `CURRENT` and `MANIFEST-*`.
- Return a clear error that lists the missing or empty paths.
- Keep explicit `format_master=true` initialization unchanged.

## Verification

- Reproduced the previous behavior with a failing test: `format_master=false initialized empty master data directories`.
- `cargo test -p curvine-server --test master_fs_test test_journal_system_refuses_non_format_start_with_empty_master_dirs -- --nocapture`
- `cargo test -p curvine-server --test master_fs_test -- --nocapture`
- `cargo test -p curvine-server --test journal_test -- --nocapture`
- `git diff --check`

### DIFF
--- a/curvine-server/src/master/journal/journal_system.rs
+++ b/curvine-server/src/master/journal/journal_system.rs
@@ -29,11 +29,14 @@ use curvine_common::raft::storage::{AppStorage, LogStorage, RocksLogStorage};
 use curvine_common::raft::{RaftClient, RaftResult, RoleMonitor, RoleStateListener};
 use curvine_common::FsResult;
 use orpc::common::FileUtils;
+use orpc::err_box;
 use orpc::runtime::{RpcRuntime, Runtime};
 use orpc::sync::StateCtl;
 use prost::Message;
 use raft::eraftpb::Entry;
 use raft::Storage;
+use std::fs;
+use std::path::Path;
 use std::sync::Arc;
 
 // Send and replay metadata operation logs based on raft.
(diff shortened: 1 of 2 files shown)


curvine-server/src/master/journal/journal_system.rs
```diff
@@ -141,8 +144,54 @@ impl JournalSystem {
...
+                        .to_str()
+                        .is_some_and(|name| name.starts_with("MANIFEST-"))
+                })
+            })
+            .unwrap_or(false)
+    }
+
     pub fn from_conf(conf: &ClusterConf) -> FsResult<Self> {
         // When the journal system is used, please note that it is separate from the fs system.
+        Self::require_existing_master_data(conf)?;
```
szbr00, line 193: This guard runs unconditionally for every format_master=false boot, including multi-master Raft deployments (journal.journal_addrs with more than one peer). In an HA setup, a replacement/added master is normally expected to start with empty local dirs and catch up by receiving a snapshot from the leader, which is essentially the same "StatefulSet rescheduled to a fresh node" situation cited as the motivation here. For a single-master (or the authoritative leader) an empty start is genuine data loss, but for a follower rejoining a quorum it's the safe, expected recovery path, and this check would now turn it into a hard startup failure.

Could you confirm the intended bootstrap flow for new/replacement Raft members? If they're expected to use format_master=true to re-init and then sync from the leader, it'd be good to document that explicitly (deployment manifests / Helm chart). Otherwise, consider scoping this enforcement to single-peer journal configs (or to nodes expected to be authoritative) so HA members can still bootstrap from empty local state.

curvine-server/src/master/journal/journal_system.rs
```diff
@@ -141,8 +144,54 @@ impl JournalSystem {
...
+        }
+
+        if invalid_dirs.is_empty() {
+            return Ok(());
+        }
+
+        err_box!(
+            "format_master=false requires existing RocksDB data directories; missing or empty: {}",
+            invalid_dirs.join(", ")
+        )
```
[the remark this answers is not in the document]
jlon: Fixed. The startup error now says missing, empty, or unreadable so permission/read_dir failures are not reported only as missing or empty.

curvine-server/src/master/journal/journal_system.rs
jlon: Adjusted the guard to single-master startup only. Multi-master Raft configs now bypass the empty-local-dir refusal so replacement/follower masters can bootstrap from the leader by log replay or snapshot. Added a regression test that format_master=false with empty local dirs is allowed when journal_addrs contains multiple peers, while the single-master empty-dir refusal remains covered.

szbr00 (approved):

PR #907, opened 2026-06-11
fix(ufs,cli): support TOS/OSS mounts via virtual-host-style and '='-safe config parsing

yangyang233333:
## Summary
- Default OpenDAL S3 builder to virtual-host-style addressing so TOS/OSS-compatible buckets can be mounted (they reject path-style); add `s3.force_path_style=true` opt-out for stores that require path-style.
- Use `splitn(2, '=')` when parsing `-c key=value` config pairs in `curvine-cli`, so values containing `=` (URLs, base64 secrets, endpoints with query strings) are no longer truncated at the first `=`.

## Motivation
Mounting TOS/OSS buckets via `curvine mount` failed because OpenDAL's S3 service defaults to path-style addressing, which TOS and OSS do not support. At the same time, secrets and endpoints commonly contain `=` characters, but the CLI `-c` parser was splitting on every `=` and dropping everything after the second one.

## Approach
- `curvine-ufs/src/opendal.rs`: enable `enable_virtual_host_style()` on the S3 builder unless `s3.force_path_style=true` is set in mount config.
- `curvine-cli/src/cmds/mount.rs`, `curvine-cli/src/util.rs`: switch the `key=value` parser from `split('=')` to `splitn(2, '=')` so the value side keeps any embedded `=` characters intact.

## Test plan
- [x] `curvine mount` against a TOS bucket succeeds and reads/writes objects
- [x] `curvine mount -c s3.force_path_style=true ...` against a path-style-only S3-compatible store still works
- [x] Pass a config value containing `=` (e.g. an endpoint with a query string) and confirm it is preserved end-to-end

### DIFF
--- a/curvine-cli/src/util.rs
+++ b/curvine-cli/src/util.rs
@@ -22,25 +22,6 @@ use std::future::Future;
 use std::str::FromStr;
 use std::time::Duration;
 
-/// Parse configuration string
-#[allow(unused)]
-pub fn parse_config(config_str: Option<&str>) -> CommonResult<HashMap<String, String>> {
-    let mut configs = HashMap::new();
-
-    if let Some(config) = config_str {
-        for pair in config.split(',') {
-            let parts: Vec<&str> = pair.splitn(2, '=').collect();
-            if parts.len() == 2 {
-                configs.insert(parts[0].trim().to_string(), parts[1].trim().to_string());
-            } else {
-                return Err(format!("Invalid config format: {}", pair).into());
-            }
-        }
-    }
-
-    Ok(configs)
-}
-
 pub async fn handle_rpc_result<T, E: Display>(operation: impl Future<Output = Result<T, E>>) -> T {
     match operation.await {
         Ok(result) => result,
--- a/curvine-cli/src/util.rs
+++ b/curvine-cli/src/util.rs
@@ -29,7 +29,7 @@ pub fn parse_config(config_str: Option<&str>) -> CommonResult<HashMap<String, St
 
     if let Some(config) = config_str {
         for pair in config.split(',') {
-            let parts: Vec<&str> = pair.split('=').collect();
+            let parts: Vec<&str> = pair.splitn(2, '=').collect();
             if parts.len() == 2 {
                 configs.insert(parts[0].trim().to_string(), parts[1].trim().to_string());
             } else {
--- a/curvine-ufs/src/opendal.rs
+++ b/curvine-ufs/src/opendal.rs
@@ -413,12 +413,13 @@ impl OpendalFileSystem {
                     builder = builder.secret_access_key(secret_key);
                 }
 
-                // OpenDAL defaults to path-style; enable virtual-host-style unless explicitly forced
-                // Note: object stores like tos/oss do not support path-style, only virtual-host-style
+                // Default to path-style for backward compatibility (MinIO, Ceph RGW, localstack, etc.).
+                // Object stores like TOS/OSS only support virtual-host-style and must opt in by
+                // setting `s3.force_path_style=false`.
                 let force_path_style = conf
                     .get("s3.force_path_style")
                     .map(|v| v.eq_ignore_ascii_case("true"))
-                    .unwrap_or(false);
+                    .unwrap_or(true);
                 if !force_path_style {
                     builder = builder.enable_virtual_host_style();
                 }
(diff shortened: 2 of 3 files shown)


yangyang233333:
Additional context: in our environment the S3 secret key itself contains '=' characters. With the previous `split('=')` parser, the value was truncated at the first '=', so the credential passed to the backend was incomplete and the mount failed to authenticate. The `splitn(2, '=')` change preserves the full secret key and unblocks the mount.

curvine-ufs/src/opendal.rs
```diff
@@ -413,6 +413,16 @@ impl OpendalFileSystem {
...
                 }
 
+                // OpenDAL defaults to path-style; enable virtual-host-style unless explicitly forced
+                // Note: object stores like tos/oss do not support path-style, only virtual-host-style
+                let force_path_style = conf
+                    .get("s3.force_path_style")
+                    .map(|v| v.eq_ignore_ascii_case("true"))
+                    .unwrap_or(false);
+                if !force_path_style {
+                    builder = builder.enable_virtual_host_style();
```
szbr00, line 424: This flips the default S3 addressing mode from path-style to virtual-host-style for every `s3`/`s3a` mount, not just TOS/OSS. Existing deployments backed by path-style-only S3-compatible stores (MinIO, Ceph RGW, localstack, many on-prem gateways) will start failing after upgrade unless they add `s3.force_path_style=true`, which makes this a backward-incompatible default change.

Two suggestions:
1. Call this out explicitly as a breaking change in the release notes / mount docs, and document the new `s3.force_path_style` key alongside the other `s3.*` options.
2. Consider logging the selected addressing mode (virtual-host vs path) at mount time — it makes endpoint/bucket misconfigurations much easier to diagnose.

If keeping path-style as the default (and opting *into* virtual-host for TOS/OSS) was considered and rejected, a one-line note on why would help future readers.

curvine-cli/src/util.rs
```diff
@@ -29,7 +29,7 @@ pub fn parse_config(config_str: Option<&str>) -> CommonResult<HashMap<String, St
 
     if let Some(config) = config_str {
         for pair in config.split(',') {
-            let parts: Vec<&str> = pair.split('=').collect();
+            let parts: Vec<&str> = pair.splitn(2, '=').collect();
```
szbr00: `parse_config` has no callers in the tree (it's marked `#[allow(unused)]`), so this `splitn(2, '=')` change has no runtime effect — the parser actually used by `curvine mount` is `MountCommand::get_config_map` in `curvine-cli/src/cmds/mount.rs`, which is fixed correctly. Worth either removing this dead function or confirming it's intentionally kept.

Separately, note that the outer `config.split(',')` here means a config *value* containing a comma (e.g. an endpoint query string like `?ids=1,2`) is still split incorrectly even with the `=` fix. If comma-bearing values are in scope, this needs a more robust parse; otherwise the constraint should be documented.

curvine-ufs/src/opendal.rs
yangyang233333: Thanks for catching the backward-compatibility risk. Addressed in ed4c1c1 by keeping path-style as the default — `s3.force_path_style` now defaults to `true`, so MinIO / Ceph RGW / localstack / on-prem path-style-only gateways continue to work after upgrade with no config change. TOS/OSS users opt into virtual-host-style by explicitly setting `s3.force_path_style=false`.

yangyang233333: Update on the mount-time logging suggestion: the CLI mount command already prints every config entry under a `Configuration:` block before invoking mount (see `curvine-cli/src/cmds/mount.rs`), so `s3.force_path_style` is surfaced to the user as part of that block automatically. A dedicated log line for this single field would be redundant, so I am leaving it out.

curvine-cli/src/util.rs
yangyang233333: Good catch — confirmed `parse_config` had no callers and was already marked `#[allow(unused)]`, so the `splitn(2, '=')` change in this file had no runtime effect. The actual fix lives in `MountCommand::get_config_map` (curvine-cli/src/cmds/mount.rs).

Addressed in 1ae7ac1 by removing the dead `parse_config` helper outright, rather than maintaining a parallel implementation that could drift from `get_config_map` again in the future.

On the comma concern: `get_config_map` consumes a `Vec<String>` produced by clap from repeated `-c k=v` flags — each element is already a single pair, so there is no `split(',')` step on the live path. The comma-splitting issue you flagged only existed inside the dead `parse_config`, and goes away with its removal.

szbr00 (approved):

PR #909, opened 2026-06-12
fix(client): defer buffer prefetch until first read and tighten …

bigbigxu:
## Summary

Optimizes the parallel prefetch lifecycle and pause logic in `FsReaderBuffer`, reducing unnecessary background work and fixing `complete`/`pause` behavior when prefetch was never started.

### 1. Lazy prefetch startup

- Prefetch tasks are no longer spawned when `BufferChannel` is created.
- `start_prefetch()` launches the background `read_future` on the first `read()` or `seek()`.
- Buffer channels that are never read no longer consume runtime or parallel reader resources.

### 2. `complete`/`pause` when prefetch never started

- If prefetch was never started (`prefetch` is still `Some`), `complete()` and `pause()` return `Ok(())` immediately.
- Avoids sending `Stop`/`Pause` commands to a background task that was never running.

### 3. Tighter sequential-read resume condition

- Replaces `if is_changed { if is_sequential { ... } }` with `if is_changed && is_sequential`.
- Parallel readers are unpause only when sequential access is detected.

### 4. `read_future` pause-path optimization

- When `paused == true`, only the control channel is polled; permit is no longer selected.
- Removes `yield_now()` spinning in the paused permit branch.
- When not paused, `tokio::select!` still handles control commands and chunk permits concurrently.

## Test Plan

- [ ] Sequential read: parallel prefetch works; pause/resume after seek behaves correctly
- [ ] Random read: parallel readers are not incorrectly unpaused on non-sequential access
- [ ] `complete()` after creating a reader without reading: no error, no leftover background tasks
- [ ] High-concurrency multi-reader: idle channels do not spawn prefetch workers prematurely
- [ ] Regression: large-file sequential read throughput and correctness after seek

## Files Changed

- `curvine-client/src/file/fs_reader_buffer.rs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1966,7 +1966,6 @@ dependencies = [
  "async-stream",
  "bytes",
  "curvine-common",
- "curvine-tracing-appender",
  "curvine-ufs",
  "dashmap 5.5.3",
  "futures",
@@ -1984,6 +1983,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "tracing",
+ "tracing-appender",
  "tracing-subscriber",
 ]
 
@@ -2058,7 +2058,6 @@ dependencies = [
  "clap",
  "curvine-client",
  "curvine-common",
- "curvine-tracing-appender",
  "dashmap 5.5.3",
  "futures",
  "fxhash",
@@ -2079,6 +2078,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
+ "tracing-appender",
  "tracing-subscriber",
  "trait-variant",
 ]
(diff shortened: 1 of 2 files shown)


szbr00 (commented):
This PR's stated scope is the buffer-prefetch deferral in fs_reader_buffer.rs, but the diff also rewrites Cargo.lock to drop the forked `curvine-tracing-appender` and pull in upstream `tracing-appender 0.2.5` (and a new `symlink` dependency), with no corresponding Cargo.toml change. main still resolves the logging appender to `curvine-tracing-appender`, so this silently changes the resolved logging dependency for the whole workspace and looks like a rebase/local-cargo artifact rather than an intentional change. Please drop the Cargo.lock change from this PR (regenerate the lockfile against main) so the PR stays focused and the appender dependency isn't swapped without review.

szbr00 (approved):

PR #911, opened 2026-06-12
feat(spdk): expose NVMe-oF transport params as env vars

thuongle2210:
# Description
This PR enables dynamic parameters for NVMe-oF transport configuration.

# Key Changes
- Expose MAX_QUEUE_DEPTH, MAX_QPAIRS_PER_CTRLR, IN_CAPSULE_DATA_SIZE, IO_UNIT_SIZE, NUM_SHARED_BUFFERS, and ZCOPY as environment variables for nvmf_create_transport, update default values to align with SPDK v25.09 default values. 
# Impact
- Users can dynamically configure NVMe-oF transport parameters for the target.

# Note
- "-q", "-n": Match what SPDK already uses implicitly, so these do not change behavior.

- "-u 131072 -m 128 -c 4096": Alters the previous deployment behavior. Please update the -u, -m, and -c values if you want to maintain the old deployment configuration.

### DIFF
--- a/curvine-docker/deploy/spdk/entrypoint-target.sh
+++ b/curvine-docker/deploy/spdk/entrypoint-target.sh
@@ -12,6 +12,12 @@
 #   MEM_SIZE        — Memory size in MB (default: 128)
 #   NR_HUGE_PAGES   — Number of 2MB hugepages to allocate (default: 1024)
 #   SERIAL          — NVMe subsystem serial number (default: SPDK0001)
+#   MAX_QUEUE_DEPTH         — NVMe-oF transport max queue depth (SPDK default: 128)
+#   MAX_QPAIRS_PER_CTRLR    — NVMe-oF transport max qpairs per controller (SPDK default: 128)
+#   IN_CAPSULE_DATA_SIZE    — In-capsule data size in bytes (SPDK default: 4096)
+#   IO_UNIT_SIZE            — I/O unit size in bytes (SPDK default: 131072)
+#   NUM_SHARED_BUFFERS      — Number of shared buffers (SPDK default: 511)
+#   ZCOPY                   — Enable zero-copy (default: false)
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


szbr00 (commented):
These new defaults change the effective transport configuration compared to the values this script hardcoded before. Previously the transport was always created with `-u 16384 -m 8 -c 8192`; with no env vars set it now becomes `-u 131072 -m 128 -c 4096` (io_unit_size 16384->131072, max_qpairs_per_ctrlr 8->128, in_capsule_data_size 8192->4096). Existing deployments that relied on the prior hardcoded tuning will silently get different qpair limits and buffer/memory behavior after upgrading. If preserving the previous behavior is intended, default these to 16384/8/8192; if moving to SPDK built-in defaults is intentional, please note this behavior change in the PR description/changelog. (`-q 128`/`-n 511` match what SPDK already used implicitly, so those are not behavior changes.)

thuongle2210:
> These new defaults change the effective transport configuration compared to the values this script hardcoded before. Previously the transport was always created with `-u 16384 -m 8 -c 8192`; with no env vars set it now becomes `-u 131072 -m 128 -c 4096` (io_unit_size 16384->131072, max_qpairs_per_ctrlr 8->128, in_capsule_data_size 8192->4096). Existing deployments that relied on the prior hardcoded tuning will silently get different qpair limits and buffer/memory behavior after upgrading. If preserving the previous behavior is intended, default these to 16384/8/8192; if moving to SPDK built-in defaults is intentional, please note this behavior change in the PR description/changelog. (`-q 128`/`-n 511` match what SPDK already used implicitly, so those are not behavior changes.)

I noted it in the PR description. Thank you.

lzjqsdd:
I see there are a ton of commits in the PR, but the actual diff only shows one file - it's a bit confusing. Maybe you should rebase onto the latest origin/main before submitting the PR

thuongle2210:
> I see there are a ton of commits in the PR, but the actual diff only shows one file - it's a bit confusing. Maybe you should rebase onto the latest origin/main before submitting the PR

Ok, I resolved it.

szbr00 (approved):

PR #912, opened 2026-06-12
refactor(fuse): replace inline cache cleanup with a scheduled Cleaner…

bigbigxu:
## Summary

Cache eviction for FUSE inode entries (`NodeMap`) was previously triggered
inline inside `find_node()`, guarded by an internal `last_clean` timestamp.
This approach couples cache maintenance to the lookup hot path and makes the
eviction interval depend on lookup traffic rather than real time.

This change extracts the cleanup into `CleanerTask`, a `LoopTask` backed by
`ScheduledExecutor`, which runs `NodeState::clear()` at a fixed interval
(`node_cache_ttl / 2`). The throttle field is removed from `NodeMap` and the
`clean_cache()` call is dropped from `find_node()`.

### 1. New `CleanerTask` (`state/cleaner_task.rs`)

Implements `LoopTask` for `FuseError`. `CleanerTask::start()` takes the
`node_cache_ttl` in milliseconds, halves it (minimum 1 ms), creates a named
`ScheduledExecutor("dcache-cleaner")`, and starts the task. Each tick calls
`NodeState::clear()`.

### 2. `NodeState::clear()` (`state/node_state.rs`)

New public method that acquires a write lock on `NodeMap` and calls
`clean_cache()`. Provides a clean boundary between the task layer and the
state internals.

### 3. Remove throttle from `NodeMap` (`state/node_map.rs`)

- Removed `last_clean: u64` field and its initializer.
- Removed the early-return guard `if self.last_clean + self.cache_ttl > now` from `clean_cache()`.
- Removed `self.clean_cache()` call from `find_node()`.

The interval is now enforced externally by `ScheduledExecutor`, so no
in-band throttling is needed.

### 4. Start `CleanerTask` on startup (`curvine_file_system.rs`)

`CurvineFileSystem::new()` calls `CleanerTask::start(fuse_conf.node_cache_ttl, state.clone())`
immediately after `NodeState` is created.

### 5. Export via `state/mod.rs`

`CleanerTask` is re-exported from `crate::fs::state` so the file system
constructor can import it alongside `NodeState`.

## Before / After

| Aspect | Before | After |
|--------|--------|-------|
| Cleanup trigger | Inline in `find_node()` hot path | Background `ScheduledExecutor` tick |
| Throttle | `last_clean` field + time check in `NodeMap` | `ScheduledExecutor` interval |
| `find_node()` cost | Acquires write lock + scans nodes periodically | No cleanup overhead |
| Eviction frequency | Depends on lookup traffic | Fixed at `node_cache_ttl / 2` |

## Files Changed

| File | Change |
|------|--------|
| `curvine-fuse/src/fs/state/cleaner_task.rs` | New — background cleanup task |
| `curvine-fuse/src/fs/state/mod.rs` | Export `CleanerTask` |
| `curvine-fuse/src/fs/state/node_state.rs` | Add `clear()` entry point |
| `curvine-fuse/src/fs/state/node_map.rs` | Remove `last_clean` throttle and inline call |
| `curvine-fuse/src/fs/curvine_file_system.rs` | Start `CleanerTask` on construction |

### DIFF
--- a/curvine-fuse/src/fs/state/cleaner_task.rs
+++ b/curvine-fuse/src/fs/state/cleaner_task.rs
@@ -1,0 +1,55 @@
+//  Copyright 2025 OPPO.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+use crate::fs::state::NodeState;
+use crate::{FuseError, FuseResult};
+use curvine_common::executor::ScheduledExecutor;
+use orpc::runtime::LoopTask;
+use std::sync::{Arc, Weak};
+
+pub struct CleanerTask {
+    state: Weak<NodeState>,
+}
+
+impl CleanerTask {
+    pub fn start(interval_ms: u64, state: Arc<NodeState>) -> FuseResult<()> {
+        if interval_ms == 0 {
+            return Ok(());
+        }
+
+        let task = CleanerTask {
+            state: Arc::downgrade(&state),
+        };
+
+        let interval_ms = (interval_ms / 2).max(1);
+        let executor = ScheduledExecutor::new("dcache-cleaner", interval_ms);
+        executor.start(task)?;
+        Ok(())
+    }
+}
+
+impl LoopTask for CleanerTask {
+    type Error = FuseError;
+
+    fn run(&self) -> Result<(), Self::Error> {
+        match self.state.upgrade() {
+            Some(state) => state.clear(),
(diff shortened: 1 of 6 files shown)


curvine-fuse/src/fs/state/cleaner_task.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl LoopTask for CleanerTask {
+    type Error = FuseError;
+
+    fn run(&self) -> Result<(), Self::Error> {
+        self.state.clear()
+    }
+
+    fn terminate(&self) -> bool {
+        false
```
szbr00: The cleaner thread never has a shutdown path: `ScheduledExecutor::start()` detaches the spawned thread, `CleanerTask::terminate()` always returns `false`, and the task owns an `Arc<NodeState>`. That means dropping/remounting a `CurvineFileSystem` will not stop the cleaner; it keeps the old `NodeState` alive and continues running indefinitely. This can leak threads/state in tests, remounts, or failed session restarts. Consider making the cleaner lifetime explicit, for example by storing a shutdown handle/token on `CurvineFileSystem` and having `terminate()` observe it, or by using a weak reference and exiting once the filesystem state is gone.

bigbigxu: fixed

szbr00 (approved):

PR #914, opened 2026-06-14
refactor(spdk): unify string-based config, clean up inheritance model, and fix validation ordering

thuongle2210:
# Description
- Clean up the SPDK configuration layer: remove dead fields and FFI bindings, unify the string-based TOML interface, fix the `init()`-before-`validate()` ordering bug, and plug a validation hole where inherited `keep_alive_timeout_ms` was never checked against `poll_interval_ms`.

# Key changes
- **Performance and architecture fit**:
  - Remove `shm_id` tuning - Curvine uses single-process SPDK; no shared memory needed.
  - Remove `mem_channel` tuning - SPDK auto-detects memory channels; no tuning benefit.
  - Remove `block_align` tuning - Curvine uses `block_size` detected from the target for alignment.
  - Remove `dma_pool_size` and add `dma_buffer_size` - Curvine currently uses a single DMA buffer allocation. If extended in the future to allocate many buffers (batch mode), pre-allocation configuration can be re-added.
- **Configuration refactoring for clarity**:
  - Rename `io_timeout_us` to `io_timeout_ms`, eliminating the `* 1000` µs conversion.
  - Use `#[serde(skip)]` on all computed fields (`io_timeout_ms`, `keep_alive_timeout_ms`, `dma_buffer_bytes`, `hugepage_mb`).
  - Introduce a 2-layer inheritance/override model for `keep_alive_timeout_ms` and `io_timeout_ms`: global values in `[worker.spdk_disk]` serve as defaults; per-target values in `[[worker.spdk_disk.targets]]` override the global when set to a non-zero value. Setting per-target to `0` (or omitting it) inherits the global default.
- Fix ordering: call `init()` before `validate()` in `SpdkEnv::new()`.
- Resolve `keep_alive_timeout_ms` (target override or global default) before comparing with `poll_interval_ms` in validation.
- Remove dead code: strip unused imports, unused `ByteUnit` fallback in `hugepage_str` parsing, and `spdk_opts_helper.c` helpers for `shm_id`/`mem_channel`.
- Add unit tests covering string parsing, default preservation, target overrides, inheritance, validation of KATO vs poll interval, TOML round-trip, and end-to-end `SpdkEnv::new()`.

# Impact
- No functional regression. Configuration validation is stricter: invalid inherited keep-alive timeouts are now caught at config load instead of silently misbehaving at runtime.
- The `init()`-before-`validate()` fix ensures string fields (`io_timeout`, `keep_alive_timeout`, `hugepage`, `dma_buffer_size`) are parsed before validation runs.

## Migration notes
This commit changes several SPDK config fields from raw integers to string-based durations/sizes. Existing TOML configs **must be updated** before upgrading.

| Old (removed) | New (required) | Reason | Notes |
|---|---|---|---|
| `io_timeout_us = 30000000` | `io_timeout = "30s"` | Field renamed + type changed to string | Must update if using `io_timeout_us` |
| `keep_alive_timeout = 10000` | `keep_alive_timeout = "10s"` | Type changed from `u64` to string | Must update if using `keep_alive_timeout` |
| `hugepage = 1024` | `hugepage = "1024MB"` | Type changed from `u32` to string | Must update if using `hugepage` |
| `dma_pool_size = 1048576` | `dma_buffer_size = "1MB"` | Field renamed + type changed | Old value has no effect; new default matches what Curvine previously used |
| `shm_id = ...` | *(remove line)* | Field removed | Only affects users who manually set this |
| `mem_channel = ...` | *(remove line)* | Field removed | Only affects users who manually set this |
| `block_align = ...` | *(remove line)* | Field removed | Was never applied; behavior is unchanged |

# Future work
- The keep_alive_timeout_ms default is 10s and the cluster uses it instead of modifying, so disconnection happens after 10s even when the target has not failed. A follow-up PR will address the root cause: keep-alive should not disconnect the controller when it has no in-flight I/O.

### DIFF
--- a/orpc/src/io/spdk_env.rs
+++ b/orpc/src/io/spdk_env.rs
@@ -1,5 +1,6 @@
 #![cfg(feature = "spdk")]
 
+use crate::common::DurationUnit;
 use crate::err_msg;
 use crate::io::spdk_ffi;
 use crate::io::spdk_poller::PollerConfig;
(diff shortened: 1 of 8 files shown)


thuongle2210:
Dear @szbr9486 , This PR intentionally makes the configuration files cleaner and ensures the 2-layer inheritance model. Could you pls help me review it?

orpc/src/io/spdk_env.rs
```diff
@@ -226,71 +248,68 @@ impl Display for NvmeTarget {
...
-    #[serde(default)]
+    pub keep_alive_timeout_ms: u64, // e.g. 10000
+    #[serde(alias = "poll_interval")]
+    pub poll_interval_ms: u64, // default = 100
     // PAUSE iterations before parking on eventfd (on x86, roughly 300 ns - 1 µs at 1000). Burns that much CPU to skip the wakeup
     // syscall for back-to-back I/O.
     pub spin_iter: u32,
-    #[serde(alias = "dma_pool_size", default)]
-    pub dma_pool_size_str: String, // e.g. "64MB"
+    #[serde(alias = "dma_buffer_size")]
```
szbr00, line 275: This rename drops compatibility with existing configs that use `dma_pool_size`. Because this struct does not reject unknown TOML fields, a deployed config like `dma_pool_size = "64MB"` will still parse, but the value is ignored and `dma_buffer_bytes` falls back to the new 1MB default. That silently changes SPDK buffer sizing after upgrade. Consider keeping `dma_pool_size` as an alias during the migration (or explicitly rejecting/documenting the old key) so existing configs do not degrade without any error.

thuongle2210: I think I mentioned this in the migration notes. This change does not impact the current config file.

szbr00 (approved):

Issue #895, opened 2026-06-08
[FEATURE] Support configuring FUSE BDI max_readahead_kb to 1024 to enable 1MiB read requests


yangyang233333:
## Problem

The Linux FUSE kernel module issues read requests to userspace based on the backing device info (BDI) `max_readahead_kb`. When this value is small (the default is typically 128KB), the FUSE kernel will split large user reads into many smaller requests (often 128KB or less) before dispatching them to curvine-fuse, even when the user issues a much larger read.

This produces several inefficiencies:
- Higher per-request overhead (context switches, kernel/userspace boundary crossings)
- More fragmented requests to the underlying storage / worker
- Lower achievable sequential read throughput

curvine-fuse currently does not provide a way to raise `max_readahead_kb` for the mount point's BDI after `mount`, so users cannot opt in to larger read requests.

## Proposed Solution

Allow curvine-fuse to configure the BDI `max_readahead_kb` (e.g. to 1024) after the mount succeeds, so that the FUSE kernel can issue 1MiB read requests to userspace.

The path is typically:

```
/sys/class/bdi/<major>:<minor>/max_readahead_kb
```

Where `<major>:<minor>` corresponds to the FUSE mount's superblock device. Suggested implementation:

1. Expose a config option such as `fuse.max_readahead_kb` (default unchanged, opt-in 1024).
2. After `mount`, resolve the FUSE superblock device id (e.g. via `statfs`/`stat` on the mount point) and write the configured value to `/sys/class/bdi/<dev>/max_readahead_kb`.
3. Also set the FUSE init `max_readahead` to a value >= the desired readahead (since the kernel caps actual readahead by `min(bdi.max_readahead_kb, fuse_conn.max_readahead)`).

## Impact

- Reduces the number of FUSE read requests for sequential workloads
- Lowers per-request overhead and improves large sequential read throughput
- Opt-in only; users who do not configure it see no behavior change

shaynewq:
Hi team, 

I would love to work on this issue. Could you please assign it to me? Thank you!

PR #919, opened 2026-06-16
feat(fuse): configure BDI max readahead

shaynewq:
## Summary
- add optional `fuse.max_readahead_kb` config validation and defaults
- write the Linux FUSE BDI `/sys/class/bdi/<maj>:<min>/read_ahead_kb` value after mount
- raise negotiated FUSE `max_readahead` when the option is configured
- resolve the BDI path from `/proc/self/mountinfo` instead of stat'ing the FUSE mount root to avoid startup deadlock before the session loop starts

## Tests

- `cargo test -p curvine-common --lib conf::fuse_conf`
- `cargo test -p curvine-fuse session::bdi --lib`

### DIFF
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -144,6 +144,18 @@ pub struct FuseConf {
 
     pub state_dir: String,
 
+    /// Optional override for the FUSE mount BDI `max_readahead_kb` (in KB).
+    ///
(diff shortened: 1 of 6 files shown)


szbr00:
Contact WeChat 398713015, let me bring you to the develop group

yangyang233333:
LGTM.

Closes https://github.com/CurvineIO/curvine/issues/895

bigbigxu (approved):

PR #922, opened 2026-06-20
fix(spdk): add admin completion mechanism on all controllers to service keep-alive

thuongle2210:
# Description

- After connecting to an NVMe-oF target, the keep-alive timeout (KATO) fires and disconnects the controller because the admin queue is never polled - only `spdk_nvme_ctrlr_process_admin_completions()` services the keep-alive timer. I/O qpair completions (`curvine_spdk_qpair_poll`) do not send Keep Alive commands, so the target sees silence on the admin queue and disconnects after KATO elapses.

# Key Changes

- Add `spdk_nvme_ctrlr_process_admin_completions` FFI declaration in `spdk_ffi.rs`
- Add `CtrlHandle` newtype (`*mut spdk_nvme_ctrlr`) with `unsafe impl Send` in `spdk_poller.rs`
- Add `ctrlrs: Vec<CtrlHandle>` field to `PollerConfig`
- Add `process_admin_completions()` helper in `SpdkPoller` and call it in both **Active** and **Idle** poller loop states
- Collect unique controller pointers in `SpdkEnv::init()` and pass to `PollerConfig`
- Log `warn!` on admin completion errors for diagnostics
- Extend `spdk_full_lifecycle` integration test with Phase 7b: sleep for 3x KATO then verify the connection is still alive and I/O still works
- Tighten validation: `keep_alive_timeout_ms` must be `>= 3 * poll_interval_ms` (was `>= poll_interval_ms`) to account for worst-case idle→active transition gap
- Import std::sync::atomic::AtomicUsize for re-use purpose.

# Impact
- Zero in-flight I/O is fine - admin polling is independent from I/O qpair polling. 
- Fast path coverage: the mid-drain tick-gated check prevents KATO from firing during sustained heavy I/O (where the Idle state is never entered)
- Keep-Alive prevention: SPDK internally checks keep-alive every KATO period. We call `process_admin_completions` every poller iteration to send the keep-alive. Worst-case gap between admin polls is ~2× poll_interval_ms (idle→active transition), so the constraint is `keep_alive_timeout_ms >= 3 × poll_interval_ms` for reliable coverage with 1x`poll_interval_ms` safety margin
- If target freezes (process hung), after KATO interval with no response, the target's KATO fires and the target disconnects

# Future works
- Handle memory leaking when target disconnects


thuongle2210:
Dear @szbr9486 ,could you please help me review this change?

orpc/src/io/spdk_poller.rs
```diff
@@ -404,6 +412,9 @@ impl SpdkPoller {
                             );
                         }
 
+                        // Process admin completions (keep-alive)
+                        Self::process_admin_completions(&active_ctrlrs);
```
szbr00, line 438: While idle, admin completions are only serviced once per poll_interval_ms (this timeout branch). SPDK sends the Keep Alive command at roughly KATO/2, so to reliably refresh keep-alive before the target's timeout we need at least one poll per KATO/2 window — i.e. poll_interval_ms <= keep_alive_timeout_ms / 2.

However, the only guard (SpdkConf::validate in spdk_env.rs) requires resolved_ka_ms >= poll_interval_ms. A config with KATO == poll_interval (or anything below ~2x) passes validation but can still disconnect despite this fix, because the idle poll fires too slowly to send keep-alive in time. The test uses KATO=500ms with the default 100ms interval (5x margin) so it passes, but the guarantee isn't enforced.

Suggest tightening the validation to require keep_alive_timeout_ms >= 2 * poll_interval_ms (with a clear error message), so the cadence this PR relies on is actually guaranteed.

lzjqsdd (approved):

Issue #920, opened 2026-06-17
[BUG]: cache_mode mount cannot fall back to S3 when Master is down (two independent defects)


yangyang233333:
**Describe the bug**

Under a `cache_mode` S3 mount (e.g. `cv mount s3://bucket /tos/cn-beijing/bucket --write-type=cache_mode`), a permanently-down Master makes `stat` fail with EIO even when the object is intact in S3. Two independent defects on the same call chain, both rooted in treating a Master RPC failure as fatal instead of falling back to S3 / stale cache.

Verified against `main` HEAD `10869d4`.

**Defect 1 — Mount-point ancestors take the `None` branch, which has no S3 fallback**

`UnifiedFileSystem::get_status` only has an S3 fallback in the `Some` branch (unified_filesystem.rs:748-752); the `None` branch queries the Master only:

```
None         => self.cv.get_status(path)   (unified_filesystem.rs:734)  ◀── no fallback
Some(...)    => cv.get_status → Err(e) => mnt.ufs.get_status  (748-752)  ◀── has fallback
```

The kernel walks `stat <leaf>` component by component, hitting `get_cached_status` for each ancestor. `Path::get_possible_mounts` (path.rs:205-251) returns all ancestor prefixes; `MountCache::get_mount` (mount_cache.rs:224-228) only matches the mount point itself. So the leaf matches the mount (→ `Some`, has fallback), but ancestors `/tos`, `/tos/cn-beijing` do not (→ `None`, no fallback). These ancestors are real Master inodes created by recursive `mkdir` at mount time — not in S3, not in MountCache — so the Master is the only source. With the Master down, the walk dies on the first ancestor before the leaf (which would have fallen back) is ever reached. Hits **every** uncached path walk, independent of TTL.

**Defect 2 — On MountCache TTL expiry, a failed `get_mount_table` RPC blocks the fallback**

`check_update` (mount_cache.rs:187-208) returns the local cache while the TTL (`mount_update_ttl`, default `10s`) is valid, but on expiry does `let mounts = fs.get_mount_table().await?;` (line 197) — a fatal `?`. With the Master down it propagates `Err` up through `get_mount:217 → get_status:733`, so `get_status` fails before reaching the 748 fallback.

- Within the TTL window: fallback works, but only after the `cv.get_status` RPC burns the full retry budget (`io_retry_max_duration_ms`, default 120000ms).
- The first stat after TTL expiry: fails outright at line 197, never entering the fallback.

[...]

PR #923, opened 2026-06-22
perf(client): refresh mount cache in background off the hot path

yangyang233333:
Fixes **Defect 2** of #920 (`cache_mode` mount cannot fall back to S3 when the Master is down).

## Problem

`MountCache::get_mount()` previously called `check_update()` on every lookup, which forced a synchronous `get_mount_table()` RPC to the master whenever the cache TTL expired. Since every filesystem operation (open/rename/mkdir/read/write/stat…) resolves its mount point through `get_mount()` first, each TTL boundary stalled the next incoming request on a master round-trip — the cache periodically inserted a synchronous RPC into the hot path it was supposed to accelerate.

As described in **Defect 2** of #920, this is also a correctness problem when the Master is down: on the first lookup after TTL expiry, the fatal `?` on `get_mount_table().await?` propagates the error up through `get_mount` → `get_status` **before** the existing S3 fallback is ever reached, so a `cache_mode` `stat` returns EIO even though the object is intact in S3.

## Change

Decouple refresh from read so a failed/slow mount-table refresh can no longer block the lookup (and therefore can no longer block the downstream S3 fallback):

- **Cold start** (cache never populated): refresh synchronously so a fresh client never observes an empty mount table.
- **Warm path** (populated but stale): spawn a background refresh task and serve the current (stale-while-error) snapshot immediately; the caller never blocks on, nor fails on, the master. The refreshed table becomes visible to subsequent calls.

This addresses Defect 2's failure mode by moving the refresh out of the hot path entirely: a refresh RPC failure is now logged and retried in the background instead of being propagated fatally to the caller, so the post-TTL `stat` proceeds against the stale mount table and reaches the S3 fallback.

`is_initialized()` distinguishes the two states by reusing the existing `last_update` field (0 ⇔ never refreshed) — no new state introduced.

## Concurrency

- A `refreshing` `AtomicBool`, claimed via `compare_exchange`, ensures **at most one** background refresh task is in flight (throttles spawns). The flag is **always** cleared — even on error or skip — so a failed refresh self-heals on the next stale read.
- The existing `refresh_lock` `Mutex` still serializes the refresh body, shared by the sync and async paths, with a double-checked `need_update`.
- Mount-table visibility is guaranteed by the `RwLock`, **not** by the control flag; `refreshing` is purely a spawn-throttling signal.

## API

`get_mount()`'s receiver becomes `self: &Arc<Self>` so the background task can take an owned handle (`self.clone()`). Callers already hold `Arc<MountCache>`, so **no call sites change**.

## Scope

This PR covers **Defect 2** only. **Defect 1** (mount-point ancestors take the `None` branch in `get_status`, which has no S3 fallback) is independent and is not addressed here.

## Test

`cargo build -p curvine-client` passes with no errors or warnings.

### DIFF
--- a/curvine-client/src/unified/mount_cache.rs
+++ b/curvine-client/src/unified/mount_cache.rs
@@ -66,10 +66,12 @@ use crate::unified::{UfsFileSystem, UnifiedFileSystem};
 use curvine_common::fs::Path;
 use curvine_common::state::MountInfo;
 use curvine_common::FsResult;
-use log::debug;
+use log::{debug, warn};
 use orpc::common::{FastHashMap, LocalTime};
+use orpc::runtime::RpcRuntime;
 use orpc::sync::AtomicCounter;
 use orpc::CommonResult;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, RwLock};
 use tokio::sync::Mutex;
 
@@ -168,6 +170,9 @@ pub struct MountCache {
     last_update: AtomicCounter,
     /// Single-flight lock: only one task performs full refresh when TTL expires.
     refresh_lock: Mutex<()>,
+    /// True while a background refresh has been scheduled but not yet finished.
+    /// Used to avoid spawning more than one concurrent background refresh task.
+    refreshing: AtomicBool,
 }
 
 impl MountCache {
@@ -177,6 +182,7 @@ impl MountCache {
             update_interval,
             last_update: AtomicCounter::new(0),
             refresh_lock: Mutex::new(()),
+            refreshing: AtomicBool::new(false),
         }
     }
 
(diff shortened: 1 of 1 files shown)


curvine-client/src/unified/mount_cache.rs
```diff
@@ -194,27 +225,75 @@ impl MountCache {
...
+            // match the synchronous path and double-refresh is impossible.
+            let _guard = cache.refresh_lock.lock().await;
+            if cache.need_update() {
+                if let Err(e) = cache.do_refresh(&fs).await {
+                    warn!("background mount cache refresh failed: {:?}", e);
+                }
+            }
+            // Always clear the flag, even on error, so a future stale read can
+            // schedule another refresh attempt.
+            cache.refreshing.store(false, Ordering::Release);
```
szbr00: The `refreshing` flag is cleared with a plain store at the end of the task, which is skipped if anything above it panics. `do_refresh()` calls `self.mounts.write().unwrap()`, which panics if the RwLock is poisoned (e.g. a prior writer panicked). A panicking spawned task is contained by the runtime, but `_guard` (the refresh_lock) is released by Drop while `refreshing` stays `true` forever.

Once that happens, every future `trigger_async_update()` no-ops on the compare_exchange, and because `is_initialized()` stays true the synchronous fallback never re-engages — so the client silently serves a stale mount table for the rest of the process lifetime and never observes new/removed mounts.

Suggest clearing the flag via an RAII guard (a small struct holding `&AtomicBool` that stores false in `Drop`, or scopeguard-style) so the flag is always released on the panic/early-return path too.

yangyang233333: Good catch, thanks. Fixed in 31b99d6: the flag is now cleared by a `RefreshingGuard` whose `Drop` stores `false`, created at the top of the spawned task so it releases on every exit path — normal completion, early return, and panic-unwind alike — matching the RAII guarantee `refresh_lock` already has.

szbr00 (approved):

Issue #897, opened 2026-06-08
[FEATURE] End-to-end metrics across curvine-fuse, master, and worker


yangyang233333:
## Problem

Observability in curvine is currently limited and uneven across modules. When users hit performance issues — high tail latency on `read`, slow `lookup`, slow `open`, throughput regressions — it is hard to localize where the time is spent:

- Is it in the FUSE kernel <-> curvine-fuse boundary?
- Is curvine-fuse waiting on the master for metadata?
- Is the worker slow on data path?
- Is it network / RPC overhead between components?

Without consistent metrics across the full call chain, root-causing requires ad-hoc logging and code reading.

## Proposed Solution

Introduce a unified metrics layer (e.g. Prometheus-compatible) and instrument the full request path:

**curvine-fuse**
- Per FUSE op (`lookup`, `getattr`, `open`, `read`, `write`, `release`, `readdir`, ...): request count, error count by errno, latency histogram, in-flight gauge.
- Read/write byte counters, request size histograms.
- Local cache hit/miss counters (status cache, dentry cache, etc.).

**Master**
- RPC count / latency / error per method.
- Metadata operation counters (create, delete, rename, list).
- Active client / mount count.

**Worker**
- Data path: read/write bytes, IOPS, latency histograms.
- Block-level metrics (allocation, eviction, replication if applicable).

**Cross-cutting**
- A request id / trace id propagated from FUSE op -> master/worker RPC so latencies can be correlated end-to-end (basic tracing, even if full OpenTelemetry is a follow-up).
- Expose `/metrics` endpoint per module.

## Impact

- Enables quick localization of latency / throughput issues to a specific module.
- Unblocks dashboarding, alerting, and capacity planning in production.
- Foundation for future performance work (regression detection, SLO tracking).
- Required for users running curvine in production at non-trivial scale.

PR #930, opened 2026-06-24
feat(fuse): add metrics context spine and finish state machine

yangcx000:
## What

Phase 1a-1 of the curvine-fuse end-to-end metrics work ([design](curvine-fuse-metrics-design.md), [#897](https://github.com/CurvineIO/curvine/issues/897)): the **structural context spine and finish state machine**. This is a **behaviour-preserving** refactor that establishes the per-request metrics context and the sender-side finish point, **without emitting any real metric**. It is intentionally landed before the counters (Phase 1a-2) so the ownership model is validated in isolation.

After this PR the daemon behaves identically — only internal plumbing changed.

## Changes (9 files, +558/-33)

- **`FuseTask` split** into `RequestReply` (owns the move-only `ActiveGuard`, finished in the sender after the kernel-fd write), `NotifyReply` (kernel notifications, carry their own code), and legacy `Reply` (disabled fast path).
- **`Arc<Mutex<FuseRespMetrics>>` slot** on `FuseResponse`: single-take active guard, `finished` flag, `op_status`/`request_status` stored separately, clone shares one slot.
- **Status from an explicit source tag** (`send_rep_tagged`), never from errno: untagged `ENOSYS`/`EINTR` stay `Error`; the dispatch wildcard / `Notimplemented` is `Unsupported`; the SETLKW interrupt path is `Interrupted`.
- **`finish_no_reply`** inspects the `Forget`/`BatchForget` result (a failing forget is `Error`, not a phantom success); **`finish_early`** drops the guard on a parse-after-ctx failure without emitting a request.
- **`send_stream`** creates the ctx before parse and reuses the same slot on the enqueue-failure reply, so failures cannot become zero-latency phantoms.
- **Zero-length write** now sends a `fuse_write_out{size:0}` reply so the request reaches a terminal state instead of hanging the kernel task.
- `ActiveGuard` is a **no-op guard** in 1a-1 (proves move/drop ownership without touching a real gauge); `parking_lot::Mutex` matches orpc; its `MutexGuard` is `!Send` (the workspace does not enable `send_guard`), so a lock-held-across-await fails to compile. (`ActiveGuard` itself is intentionally `Send` — it must travel onto the reply task — and a `guards_are_send` compile-test enforces that.)

## Scope / deferred

- **No real metrics are emitted in this PR.** The `active_requests` gauge, `requests_total`, `request_duration_us`, etc. are wired on top of this proven state machine in **Phase 1a-2** (next PR, stacked on this branch).
- The `metrics_enabled` kill switch also lands in Phase 1a-2 (the design's "kill switch in Phase 1a-1" is deferred one step).

## Tests

7 new ownership/error-path tests in `fuse_response.rs` cover: finish-once, clone/double-reply no-op, no-reply success+error, notify split, parse-after-ctx no-leak, enqueue-failure status split, and source-tag classification.

- `cargo build -p curvine-fuse`: zero warnings
- `cargo test -p curvine-fuse --lib`: 39 passed
- `cargo clippy -p curvine-fuse --all-targets -- --deny=warnings --allow clippy::uninlined-format-args` (matches CI): passes
- End-to-end `fuse-test.sh` mainline passes (the metrics work is validated more fully in 1a-2).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2072,6 +2072,7 @@ dependencies = [
  "num_enum",
  "once_cell",
  "orpc",
+ "parking_lot",
  "pin-project-lite",
  "pkg-config",
  "serde",
--- a/curvine-fuse/Cargo.toml
+++ b/curvine-fuse/Cargo.toml
@@ -23,6 +23,7 @@ thiserror = { workspace = true }
 trait-variant = { workspace = true }
 nix = { workspace = true }
 once_cell = { workspace = true }
+parking_lot = { workspace = true }
 mini-moka = { workspace = true }
 fxhash = { workspace = true }
 dashmap = { workspace = true }
--- a/curvine-fuse/src/fs/state/file_handle.rs
+++ b/curvine-fuse/src/fs/state/file_handle.rs
@@ -15,6 +15,7 @@
 use crate::fs::operator::{Read, Write};
 use crate::fs::state::NodeState;
 use crate::fs::{FuseReader, FuseWriter};
+use crate::raw::fuse_abi::fuse_write_out;
 use crate::session::FuseResponse;
 use crate::{err_fuse, FuseError, FuseResult};
 use curvine_common::fs::{Path, StateReader, StateWriter};
@@ -96,6 +97,17 @@ impl FileHandle {
 
     pub async fn write(&self, op: Write<'_>, reply: FuseResponse) -> FuseResult<()> {
         if op.data.is_empty() {
+            // A zero-length write is a valid no-op, but it must still send a
+            // reply (size=0) rather than returning silently: silently returning
+            // would leave the request's metrics context un-finished (the guard
+            // would only be released by passive drop, never reaching a terminal
+            // state-machine transition). Reply normally so it finishes like any
+            // other replied request.
+            let res: FuseResult<fuse_write_out> = Ok(fuse_write_out {
+                size: 0,
+                padding: 0,
+            });
+            reply.send_rep(res).await?;
             return Ok(());
         }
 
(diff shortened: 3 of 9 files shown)


yangcx000:
Thanks for the review. Addressing the four points:

1. **`size_of::<T>()` not imported** — this compiles as-is. `size_of` (and `align_of`, etc.) was added to the standard prelude in **Rust 1.80**, independent of edition, so no `use std::mem::size_of` is needed on the toolchain this repo uses. The CI `build` job (which runs `cargo build`) passes on this commit. No change needed.

2. **`finish_no_reply` doesn't stash errno on error** — this is intentional, not a latent bug. By design (and as wired in the follow-up Phase 1a-2), `Forget`/`BatchForget` failures do **not** emit an errno-labelled metric: they only contribute `requests_total{reply_type=no_reply,status=error}` + `request_duration_us`, never `errors_total` (a failing forget has no diagnostically useful errno). So `errno` is never read on the no-reply path. I've clarified the `FuseRespMetrics.errno` field doc to say it's the errno source for the **replied** path only and is deliberately left 0 on no-reply.

3. **Doc references nonexistent `FuseResponse::finish_status()`** — correct, stale doc from an earlier design draft. Fixed: the comment now points to `send_rep_tagged()` (via `err_status()`), which is where the status is actually computed.

4. **`!Send` wording** — good catch, the PR description was imprecise. It's the `parking_lot::MutexGuard` that is `!Send` (the workspace doesn't enable `send_guard`), which is what makes a lock-held-across-await fail to compile. `ActiveGuard` itself is intentionally `Send` (it travels onto the reply task) with a `guards_are_send` compile-test enforcing it. Updated the PR description accordingly.

#2/#3 doc fixes are amended into this commit; #1 needs no change; #4 is a PR-description clarification.

lzjqsdd (approved):

PR #934, opened 2026-06-24
feat(fuse): emit end-to-end request metrics on the finish state machine

yangcx000:
## What

Phase 1a-2 of the curvine-fuse end-to-end metrics work ([design](curvine-fuse-metrics-design.md), [#897](https://github.com/CurvineIO/curvine/issues/897)), stacked on **#930 (Phase 1a-1)**. This PR wires **real Prometheus metrics** onto the finish state machine proven in 1a-1. The 1a-1 structure is unchanged — this is the inc/observe layer plus the `metrics_enabled` kill switch.

## Changes

- **Registers 13 FUSE metric families** with low-cardinality `&'static str` labels: `active_requests{kind}`, `requests_total{opcode,kind,reply_type,status}`, `request_duration_us`, `errors_total{errno}` / `interrupted_total` / `unsupported_total{reason}`, `notify_total{code,status}`, `decode_errors_total{phase,reason}`, `response_write_duration_us` / `response_bytes_total{request_status}`, `reply_enqueue_errors_total` / `response_write_errors_total`, and `stage_duration_us`.
- **Emits at the 1a-1 finish points** (no new finish paths): sender `RequestReply` (E2E `request_duration` + `requests_total` + response write/bytes + `reply_write` stage; op-level error/unsupported/interrupted from `op_status`), `finish_no_reply` (`requests_total{reply_type=no_reply}`, no `errors_total`), `finish_early` (`decode_errors_total{phase=parse}`), and `send_notify` three-point `notify_total`. All `with_label_values` is funnelled through helpers in `fuse_metrics.rs`.
- **`op_status` vs `request_status` split**: a kernel-fd write failure finishes `request_duration`/`requests_total` with `status=error` while the op-level counters keep the FS-operation status; the write errno is the independent `response_write_errors` dimension.
- **`metrics_enabled` kill switch**: `FuseConf` flag (default `true`) + `--metrics-enabled` CLI override, gated in the receiver so the disabled path constructs no context and emits nothing (legacy byte-identical `Reply`).
- **Bounded reply-channel cancellation safety**: `finish_request` reserves a permit before committing the slot / taking the active guard, so an aborted reserve cannot leave a half-finished request (no silent terminal, no gauge leak); the unbounded default keeps the synchronous fast path. Adds `AsyncSender::is_bounded()` in `orpc`.
- **Defensive guards**: `debug_assert` on double-reply (release: warn + no-op), missing unsupported source tag (release: `reason=missing_reason` bucket), and mutually-exclusive source tags.

## Scope / deferred (kept explicit to avoid acceptance-criteria confusion vs the full design)

- `decode_errors_total{phase=decode}` (from_bytes failures) → Phase 1b; only `phase=parse` ships here.
- `trait_default` unsupported reason → later phase (no source site yet); 1a-2 emits `unknown_opcode` / `unimplemented_opcode`.
- no-reply errno attribution: intentionally not emitted (forget failure errno has no diagnostic value).
- `with_label_values` per-request map-probe: accepted for now, benchmark-gated (see below); pre-resolved child cache is the fallback if a regression shows up.

## Tests

64 lib tests (debug **and** release) covering value assertions per finish path, bounded-reserve cancellation, stream-worker enqueue failure, `op_status`/`request_status` divergence, and the debug/release defensive paths.

- `cargo fmt --all -- --check`: clean
- `cargo clippy --all-targets -- --deny=warnings --allow clippy::uninlined-format-args` (matches CI): passes
- `cargo build -p curvine-fuse`: zero warnings
- `cargo test -p curvine-fuse --lib`: 64 passed (debug and release)

## End-to-end + benchmark validation

- **`fuse-test.sh` mainline: Test 1–17 pass.** (Test 18 Pwrite-Visibility is a pre-existing delayed-delete `unlink→O_CREAT` race, unrelated to metrics — this PR touches no delete/release/FS-semantics path.)
- **Live `/metrics` correlation under fio direct-IO** (~1.1M stream ops): op counts match fio IO volume (`requests_total{Write}`=684145, `{Read}`=418945), `response_bytes_total{Read}`≈2.1 GiB matches transfer, error classification is clean (only expected metadata ENOENT/ENODATA), `request_duration_us` histogram matches fio's measured p99, and **`active_requests` returns to 0** after the load (gauge balance / no leak proven under stress).
- **No throughput regression vs the 1a-1 baseline** (same master/worker, fuse binary swapped): seq write 6.70→6.83 GB/s, seq read 10.40→10.33 GB/s — within run-to-run noise.


curvine-common/src/conf/fuse_conf.rs
```diff
@@ -30,6 +30,12 @@ pub struct FuseConf {
 
     pub audit_logging_enabled: bool,
 
+    // Master on/off switch for FUSE metrics instrumentation (request/error/
+    // latency/notify series). Read once at startup; when false the reply path
+    // takes the legacy zero-cost path and emits no per-request metrics.
+    // Defaults to true (Phase 1–3 ship everything enabled).
+    pub metrics_enabled: bool,
```
szbr00, line 44: Minor / non-blocking: the `metrics_enabled` kill switch suppresses emission, but `FuseMetrics::ensure_init()` is called unconditionally in `CurvineFileSystem::new`, so all ~13 metric families are still registered when `metrics_enabled=false`. They will appear in the scrape as zero-valued series and still cost registry memory and scrape size.

That is fine (arguably preferable) if the switch is only meant to stop per-request overhead while keeping a stable scrape schema. But if the intent is an emergency cardinality/footprint downgrade, registration-while-disabled won't fully achieve that. Worth a one-line doc note on the flag clarifying that "disabled" means "no emission, families still registered", or gating `ensure_init()` on the flag if full suppression is desired.

Otherwise this looks solid — the finish state machine, the bounded reserve-before-commit cancellation safety, the op/request status split, and the exactly-once guard/terminal handling all check out, and the test coverage is thorough.

yangcx000: Good catch — you're right that `ensure_init()` registers all families unconditionally, so disabled = zero-valued series, not zero footprint. That's the intended behavior: `metrics_enabled` is a per-request *emission* kill switch (stop hot-path overhead, keep a stable scrape schema and let the flip-back-on be seamless without re-registration), not a cardinality/footprint downgrade. The legacy gauges and Phase 1b's framework-health series are also registered/updated by design, so gating `ensure_init()` on the flag would conflict with that and add re-registration complexity for negligible memory savings.

I've added a doc note on the `metrics_enabled` field clarifying exactly this — "disabled" means "no emission", not "no registration" — so the intent is explicit. Thanks for the review.

lzjqsdd (approved):

szbr00 (approved):

PR #935, opened 2026-06-24
fix(raft): keep node alive on malformed requests

jlon:
## Summary

This PR prevents a malformed Raft RPC request from stopping the whole Raft node task.

The change is intentionally narrow: request-level failures now return an RPC error to the caller, while local consistency failures in `on_ready`, log persistence, state-machine apply, and snapshot handling still stop the Raft node.

## Problem

A Curvine master can stay `Running` and pass the TCP probe while its Raft task has already exited.

In the production incident, the leader logged a protobuf decode error while handling a `Propose` request. After that, port `8996` still accepted connections, but the Raft receiver was gone. Peers then kept getting `channel closed`, no new leader could be elected, and normal `cv` commands failed with `Not a leader master`.

The failure mode is:

1. A bad or incompatible request reaches the Raft RPC port.
2. `RaftNode::handle_*` returns an error for that single request.
3. The error bubbles out of `RaftNode::run0`.
4. `RaftNode::run` logs `raft node stop` and advances the role state to `Exit`.
5. The RPC server process keeps running, so Kubernetes still sees the master pod as healthy.
6. The cluster loses Raft quorum behavior until the master pods are restarted.

Restarting the pods recreates the Raft task, but it does not close the bug. The same request path can kill the node again.

## Root Cause

`handle_raft`, `handle_propose`, and `handle_conf_change` treated malformed request data and raft-rs request admission errors as fatal node errors. These are not local storage or state-machine failures; they belong to the current RPC request.

A request decode failure should produce an error response for that request. It should not stop election, heartbeats, or future proposals.

## Solution

This PR adds request-level error handling in the Raft request handlers:

- `handle_raft`
  - returns an RPC error for malformed `RaftRequest`
  - returns an RPC error if `raw.step` rejects the message
- `handle_propose`
  - returns an RPC error for malformed `ProposeRequest`
  - returns an RPC error if `raw.propose` rejects the proposal
- `handle_conf_change`
  - returns an RPC error for malformed `ConfChangeRequest`
  - returns an RPC error if `raw.propose_conf_change` rejects the change

These paths log a warning and reply through the original RPC callback. The Raft event loop then continues.

## What This Does Not Change

This PR does not hide real local consistency failures.

Errors from ready handling, hard-state persistence, log append, committed entry apply, snapshot restore, and snapshot creation still propagate out of `run0`. Those errors can mean the local node state is unsafe to continue from, so fail-fast remains the correct behavior.

## Performance

The normal path only replaces `?` with explicit `match` / `if let Err` handling around operations that already returned `Result`.

There is no new lock, retry loop, network call, allocation-heavy path, or background task for successful requests. The extra warning log and error response only run on failure paths.

## Tests

Added a regression test that:

1. Starts a single-node Raft journal.
2. Waits until it becomes leader.
3. Sends a malformed `Propose` request whose header is not a `ProposeRequest` protobuf.
4. Verifies the malformed request returns an RPC error.
5. Sends a normal proposal afterward.
6. Verifies the normal proposal still commits.

Tested locally:

```bash
cargo fmt --check
cargo test -p curvine-common --test raft_node_test -- --nocapture
```

Result:

```text
cargo test: 2 passed
```

### DIFF
--- a/curvine-common/src/raft/raft_node.rs
+++ b/curvine-common/src/raft/raft_node.rs
@@ -20,13 +20,12 @@ use crate::raft::raft_error::RaftError;
 use crate::raft::storage::{AppStorage, ApplyMsg, LogStorage, PeerStorage};
 use crate::raft::*;
 use crate::utils::SerdeUtils;
-use log::{error, info, warn};
+use log::{debug, error, info, warn};
 use orpc::client::dispatch::{Callback, Envelope};
 use orpc::common::{DurationUnit, LocalTime, TimeSpent};
 use orpc::io::net::InetAddr;
 use orpc::message::{Builder, RefMessage, ResponseStatus};
 use orpc::runtime::{RpcRuntime, Runtime};
-use orpc::try_err;
 use prost::Message as PMessage;
 use raft::eraftpb::{ConfChange, Entry, EntryType, MessageType, Snapshot};
 use raft::prelude::ConfChangeType;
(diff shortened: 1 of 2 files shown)


curvine-common/src/raft/raft_node.rs
```diff
@@ -296,12 +295,26 @@ where
         Ok(())
     }
 
+    fn send_request_error(env: Envelope, error: RaftError) -> RaftResult<()> {
+        warn!(
```
szbr00, line 299: Now that a malformed/incompatible request no longer stops the node (good), this warn! fires once per bad request and is reachable by any client that can hit the Raft port. A misconfigured/incompatible peer or a scanner sending garbage can produce sustained warn-level log volume. Consider downgrading the decode-failure case to debug! (or rate-limiting this log) so a noisy peer cannot flood the logs, while keeping the node-survival behavior this PR introduces.

jlon: Adjusted this path so malformed request decode failures use debug-level logging via send_malformed_request_error. Requests that parse successfully but are rejected by raft-rs (raw.step/raw.propose/raw.propose_conf_change) still use warn, since those are valid Raft messages that failed admission rather than arbitrary garbage on the port.

Verified with:

```bash
cargo fmt --check
cargo test -p curvine-common --test raft_node_test -- --nocapture
```

szbr00 (approved):

PR #936, opened 2026-06-25
feat(s3): add s3.list_objects_version option to select ListObjects v1/v2

yangyang233333:
## What changed

Add an `s3.list_objects_version` mount option for S3-compatible backends to choose between ListObjects **V2** (default) and **V1**:

```
cv mount s3://my-bucket /mnt/point ... -c s3.list_objects_version=v1
```

- unset / `v2` → ListObjects V2 (OpenDAL default, **behavior unchanged**)
- `v1` → ListObjects V1 (calls `disable_list_objects_v2()`)
- other values → clear configuration error

## Why

**ListObjectsV2 is incompatible with some buckets** — notably **Baidu BOS** (hit in production yesterday: `cv mount` hung forever on a BOS bucket).

BOS returns `IsTruncated=true` but **omits `NextContinuationToken`**. OpenDAL's V2 lister then keeps resending with an empty token, gets the same page back, and loops forever. V1's lister falls back to the last object key as the marker, so it advances correctly. Setting `s3.list_objects_version=v1` works around such non-compliant services.

Default behavior is unchanged (still V2). Implemented in `curvine-ufs/src/opendal.rs`.

### DIFF
--- a/curvine-ufs/src/opendal.rs
+++ b/curvine-ufs/src/opendal.rs
@@ -1013,26 +1013,23 @@ mod list_objects_version_tests {
 
     #[test]
     fn defaults_to_v2_when_unset_or_empty() {
-        assert_eq!(resolve_disable_list_objects_v2(None).unwrap(), false);
-        assert_eq!(resolve_disable_list_objects_v2(Some("")).unwrap(), false);
-        assert_eq!(resolve_disable_list_objects_v2(Some("   ")).unwrap(), false);
+        assert!(!resolve_disable_list_objects_v2(None).unwrap());
+        assert!(!resolve_disable_list_objects_v2(Some("")).unwrap());
+        assert!(!resolve_disable_list_objects_v2(Some("   ")).unwrap());
     }
 
     #[test]
     fn v2_keeps_default() {
-        assert_eq!(resolve_disable_list_objects_v2(Some("v2")).unwrap(), false);
-        assert_eq!(resolve_disable_list_objects_v2(Some("V2")).unwrap(), false);
-        assert_eq!(
-            resolve_disable_list_objects_v2(Some("  v2 ")).unwrap(),
-            false
-        );
+        assert!(!resolve_disable_list_objects_v2(Some("v2")).unwrap());
+        assert!(!resolve_disable_list_objects_v2(Some("V2")).unwrap());
+        assert!(!resolve_disable_list_objects_v2(Some("  v2 ")).unwrap());
     }
 
     #[test]
     fn v1_disables_v2() {
-        assert_eq!(resolve_disable_list_objects_v2(Some("v1")).unwrap(), true);
-        assert_eq!(resolve_disable_list_objects_v2(Some("V1")).unwrap(), true);
-        assert_eq!(resolve_disable_list_objects_v2(Some(" v1 ")).unwrap(), true);
+        assert!(resolve_disable_list_objects_v2(Some("v1")).unwrap());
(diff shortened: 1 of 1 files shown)


curvine-ufs/src/opendal.rs
```diff
@@ -424,6 +424,31 @@ impl OpendalFileSystem {
...
+                    None | Some("") | Some("v2") | Some("2") => {}
+                    Some("v1") | Some("1") => {
+                        builder = builder.disable_list_objects_v2();
+                    }
+                    Some(other) => {
+                        return Err(FsError::common(format!(
+                            "Invalid s3.list_objects_version '{}': expected 'v1' or 'v2'",
+                            other
+                        )));
+                    }
```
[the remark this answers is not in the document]
yangyang233333: Good catch, thanks. Fixed in 5711cb2 — now using `FsError::invalid_argument` (maps to POSIX EINVAL) instead of `FsError::common` for an invalid `s3.list_objects_version` value.

curvine-ufs/src/opendal.rs
```diff
@@ -424,6 +424,31 @@ impl OpendalFileSystem {
...
+                // Defaults to v2 (OpenDAL's default). Some S3-compatible services (e.g. Baidu
+                // BOS) return `is_truncated=true` in a ListObjectsV2 response but omit the
+                // `NextContinuationToken`; OpenDAL's v2 lister then resends the request with an
+                // empty token, gets the same page back, and loops forever (mount/list hangs).
+                // Setting `s3.list_objects_version=v1` switches to ListObjects v1, whose lister
+                // falls back to using the last object key as the marker and advances correctly.
+                let list_objects_version = conf
+                    .get("s3.list_objects_version")
+                    .map(|v| v.trim().to_ascii_lowercase());
+                match list_objects_version.as_deref() {
```
szbr00: Non-blocking: the new parse/validation has three meaningful branches (default v2, v1 -> disable_list_objects_v2(), and the EINVAL error path) but no test coverage. Since the logic is embedded in the operator-building match it isn't easily unit-testable as-is; consider extracting the version resolution into a small pure helper (e.g. fn resolve_list_objects_version(&str) -> FsResult<bool/enum>) and unit-testing the accepted aliases (v1/1/v2/2/empty/none, case/space-insensitive) and the rejection path. That also guards the EINVAL classification you just fixed from regressing.

Minor: the error message says "expected 'v1' or 'v2'" while the code also accepts '1'/'2' (and empty/none as v2) — worth aligning the message with the actual accepted set.

Otherwise this looks good: default behavior is unchanged (v2), and the BOS rationale (is_truncated=true with no NextContinuationToken -> v2 infinite loop) is a solid reason for the v1 fallback.

yangyang233333: Thanks for the thorough review! Addressed both in 8ad551c:

1. Extracted the resolver into a pure helper `resolve_disable_list_objects_v2(Option<&str>) -> FsResult<bool>` and added unit tests covering the accepted values (unset/empty/v2 → false, v1 → true; case- and whitespace-insensitive) plus the EINVAL rejection path — which also guards the error-kind fix from regressing.
2. Narrowed the accepted set to `v1`/`v2` only (dropped the undocumented `1`/`2` aliases), so the `expected 'v1' or 'v2'` message now matches the actual accepted values.

`cargo test -p curvine-ufs --features opendal-s3` passes (4/4).

szbr00 (approved):

PR #938, opened 2026-06-25
feat(fuse): add receiver framework-health metrics and scrape hygiene (phase 1b-1)

yangcx000:
## Summary

Phase 1b-1 of the curvine-fuse observability work (design: `curvine-fuse-metrics-design.md`). Stacked conceptually on #930 (1a-1) and #934 (1a-2, merged); this branch is rebased onto current `main`.

This PR is **pure inc/observe additions** — receiver framework-health metrics + metrics-endpoint scrape hygiene + design-doc semantics sync. **No API shape changes.** The legacy gauge event-driven migration and `set_metrics()` removal are deliberately deferred to **Phase 1b-2** (higher risk, separate PR).

## What's added

**5 new metric families**
- `receive_loop_wait_duration_us` — receiver-loop idle wait + splice + header-parse (standalone saturation/health histogram, not a request stage).
- `receive_errors_total{errno,action}` — splice/receive errors before decode; `errno` is the lowercase 5-value set, `action` is `continue`/`exit`.
- `meta_task_inflight` — gauge for in-flight meta (interrupt-dispatch) tasks.
- `metrics_scrape_duration_us` / `metrics_scrape_bytes` — metrics-endpoint self-observation.

**Receiver `start()` loop (framework health, gated on `metrics_enabled`)**
- `receive_loop_wait` observed **after** `from_bytes` so the histogram includes header-parse and a decode failure still records a sample.
- `decode_errors_total{phase=decode}` on structural parse failure (existing `?` control flow unchanged).
- `receive_errors_total{errno,action}` via the new `receive_error_labels` free function (continue: ENOENT/EINTR/EAGAIN; exit: ENODEV + unknown). Control flow unchanged on every arm.
- `meta_task_inflight` guard + `meta_spawn` stage created **before** `rt.spawn` so they cover the runtime queue wait (submission → first poll); guard dropped the moment dispatch returns, before the error log.

**Scrape hygiene (`web_server.rs`, unconditional self-observation)**
- Time `text_output()` and record `output.len()` with last-scrape semantics. `set_metrics()` retained for legacy-gauge correctness until 1b-2; scrape duration times only `text_output()` for now (documented).

**Kill-switch layering** (design doc synced): request/framework-health metrics are **gated**; legacy compatibility gauges and scrape-hygiene self-observation are **unconditional**.

## Testing

- **72 lib tests** (debug + release) = 64 from 1a-2 + 8 new (`receive_error_labels` classification, `record_decode_error`, `meta_task_guard` gate, `record_scrape`, `record_meta_spawn`, `record_receive_loop_wait`, `splice_errno_label`).
- `cargo fmt --all -- --check` clean; `cargo clippy -p curvine-fuse --all-targets` warning-free.

**End-to-end (live `/metrics`):**
- *enabled:* fuse-test.sh Test 1–17 pass; `meta_task_inflight` balances to **0** after the full suite (incl. hot-reload restart) — guard has no leak under real load; `stage_duration_us{stage=meta_spawn}` shows a real latency distribution; `receive_errors_total` correctly **absent** on a clean run (`CounterVec` child not created until first emission).
- *`--metrics-enabled false`:* kill-switch layering confirmed live — request + framework-health series suppressed (loop-wait clock not even read, `meta_task_inflight` guard never created), while legacy gauges (`inode_num` etc.) and scrape hygiene (`metrics_scrape_bytes`/`_duration_us`) **still update**.

## Dashboard query notes

- Receiver error rate by class: `sum by (errno, action) (rate(curvine_fuse_receive_errors_total[5m]))` — `action="exit"` with `errno="enodev"` is graceful unmount; `errno="other"` is an unexpected receiver exit worth alerting on.
- Meta-task saturation: `curvine_fuse_meta_task_inflight` (should hover near 0); spawn scheduling delay via `curvine_fuse_stage_duration_us{stage="meta_spawn"}`.
- Scrape payload growth: `curvine_fuse_metrics_scrape_bytes`; scrape cost: `curvine_fuse_metrics_scrape_duration_us` (note: under-counts by the `set_metrics()` scan until 1b-2).

## Deferred to Phase 1b-2

Legacy gauge event-driven migration (inc/dec at insert/remove sites + restore finalizer using live map count), `set_metrics()` removal, and the associated drift-protection checks.


curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -243,11 +247,47 @@ impl<T: FileSystem> FuseReceiver<T> {
...
+                                    .record_receive_loop_wait(start.elapsed().as_micros() as u64);
+                            }
+                            let req = match parsed {
+                                Ok(req) => req,
+                                Err(e) => {
+                                    // Structural decode failure before any ctx:
+                                    // count it, but keep the existing `?` control
+                                    // flow (this still terminates the receiver).
+                                    if self.metrics_enabled {
+                                        FuseMetrics::get().record_decode_error("other");
```
szbr00, line 286: Worth flagging the observability semantics here: this counter is incremented and then the receiver immediately returns Err, which terminates the receive loop for this mount. So decode_errors_total{phase=decode} can increment at most once per receiver lifetime — it is effectively a one-shot fatal-event signal, not an accumulating rate. That's a meaningful difference from phase=parse (which is per-request and recurring).

Two small asks: (1) make sure dashboards/alerts treat a phase=decode increment as "receiver died, needs restart" rather than a rate to threshold; and (2) consider noting in the metric help text (or design doc) that phase=decode is terminal, so operators do not read a flat 0->1 as "one transient decode error we recovered from". No control-flow change required — just avoiding a misread of the counter.

yangcx000: Good catch — fully agree, and you're right that no control-flow change is needed. The `phase=decode` increment is indeed a one-shot terminal signal (increment then `return Err`, which ends the receive loop), semantically different from the recurring per-request `phase=parse`.

I've pushed the doc/help changes so the counter can't be misread as a recoverable rate:

1. **Metric help text** (`decode_errors_total`) now spells out the split: `phase=parse` is per-request/recurring (recoverable), `phase=decode` is **terminal** — increments at most once per receiver lifetime — and says to treat a `phase=decode` increment as "receiver died, restart" rather than a rate to threshold.
2. **`record_decode_error` doc comment** notes the terminal one-shot nature at the call site.
3. **Design doc** (finish-points table + a dedicated note): a `phase=decode` 0→1 is a fatal-event/restart signal, never a rate; `phase=parse` is the recurring one. This is the dashboard/alert guidance you asked for, captured where the alerting rules will be written.

(Also rebased onto latest main to clear a conflict in `fuse_error.rs` with #931's `EFBIG` addition.)

szbr00 (approved):

PR #939, opened 2026-06-25
fix(server): exit master on WebServer port bind conflict

### LINKED ISSUE #937: [BUG]: Master continues running when WebServer port bind fails
## Describe the bug

When the Master WebServer fails to bind its port (e.g. `Address already in use`, os error 98), the Master process does **not** exit. It continues starting JobManager, inode TTL scheduler, and other components as if startup succeeded. Additionally, the WebServer failure path triggers a Tokio runtime panic:

### 1. WebServer startup is fire-and-forget with no error propagation

In `curvine-server/src/master/master_server.rs`, `Master::start()` waits for RPC server readiness but does **not** wait for or check WebServer bind success:

- RPC: `rpc_status.wait_running().await.unwrap()` — gates downstream startup
- Web: `self.web_server.start()` — returns immediately, no status check
- Then unconditionally starts actor, JobManager, TTL scheduler

In `curvine-web/src/server/web_server.rs`, `start()` only logs bind errors:

```rust
if let Err(e) = self.run().await {
    error!("WebServer connect error: {}", e);  // log only, no shutdown
}
```

2. Tokio runtime panic on bind failure (Master only)
WebServer::new() creates a dedicated Runtime. In start(), self is moved into the spawned task — the only Arc<Runtime> reference. When bind fails, the task drops WebServer from within that runtime's async context, triggering the Tokio panic.

Worker uses WebServer::with_rt(rt.clone(), ...) so RPC server still holds the runtime — panic may not occur, but fail-fast is still missing.

3. Same pattern on Worker
curvine-server/src/worker/worker_server.rs has identical fire-and-forget self.web_server.start().

## To Reproduce
Start a Master (or bind master.web_port).
Start a second Master with the same web_port.
Observe: WebServer ERROR, then JobManager/TTL INFO logs, then Tokio panic.
Second Master remains in partially started state instead of exiting.

## Expected behavior
WebServer bind failure during startup should fail fast and exit Master (consistent with RPC semantics).

## Suggested fix
Add ServerMonitor / ServerStateListener to WebServer (mirror RpcServer).
In Master::start() / Worker::start(), wait_running() on WebServer before actor/JobManager.
On bind failure, propagate error and trigger shutdown.
Keep Arc<Runtime> owner outside the spawn task, or reuse RPC runtime via WebServer::with_rt().

## Log excerpt
```bash
ERROR web_server.rs:87 WebServer connect error: Address already in use (os error 98)
INFO job_manager.rs:81 JobManager started
INFO master_actor.rs:67 Starting inode ttl scheduler.
panic: Cannot drop a runtime in a context where blocking is not allowed.
```

lzjqsdd:
## Problem
When the Master or Worker WebServer fails to bind its port (e.g. `Address already in use`), the process continued starting JobManager, inode TTL scheduler, and other components instead of exiting. Fixes #937.
## Design
Mirror RpcServer lifecycle with `ServerMonitor` on WebServer, and gate Master/Worker startup on a successful web bind via `wait_bind()`. Fix `wait_startup()` so `Shutdown` is not mistaken for `Running` (`>= Running` was true for state 2). Keep WebServer on its own runtime via `WebServer::new()` as before.
## Key Changes
- Add `ServerMonitor` to `WebServer`; `start()` returns `ServerStateListener`
- Master/Worker await web bind before starting downstream components
- Add `ServerStateListener::wait_startup()` with correct Running vs Shutdown handling
- Improve bind-failure logs with server name, address, and port
- Drop dedicated WebServer runtime via `spawn_blocking` to avoid Tokio panic on failure path

### DIFF
--- a/curvine-web/src/server/web_server.rs
+++ b/curvine-web/src/server/web_server.rs
@@ -22,7 +22,7 @@ use axum::Json;
 use log::{error, info};
 use orpc::io::net::{InetAddr, NetUtils};
 use orpc::runtime::{RpcRuntime, Runtime};
-use orpc::server::ServerConf;
+use orpc::server::{ServerConf, ServerMonitor, ServerStateListener};
 use orpc::CommonResult;
 use serde_json::json;
 use tokio::net::TcpListener;
@@ -44,6 +44,7 @@ pub struct WebServer<S> {
     service: S,
     conf: ServerConf,
     address: InetAddr,
+    monitor: ServerMonitor,
 }
 
 impl<S> WebServer<S>
@@ -59,6 +60,7 @@ where
             service,
             conf,
             address,
+            monitor: ServerMonitor::new(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


curvine-web/src/server/web_server.rs
```diff
@@ -69,24 +71,79 @@ where
...
+        name: &str,
+        address: &InetAddr,
+    ) -> CommonResult<()> {
+        use orpc::err_box;
+
+        let port = address.port;
+        match listener.wait_startup().await {
+            Ok(()) => Ok(()),
+            Err(_) => err_box!(
+                "WebServer [{}] failed to bind address {} (port {}); the port may already be in use",
```
szbr00: The lifecycle fix here is correct (bind happens before advance_running, so wait_bind gates on a real successful bind, and the Shutdown/Stop path fails fast). Two small follow-ups on the messaging this PR adds:

1. Address consistency: wait_bind reports `address` (the config InetAddr passed from master/worker via bind_addr()), but the actual bind target is `get_bind_addr()`, which honors ORPC_BIND_HOSTNAME. When that env var is set, the error here (and the success log in run()) will name a different host than the one actually bound. Consider routing the success log, the start0 error log, and this wait_bind error through get_bind_addr() so all three stay accurate.

2. Duplicate reporting on a bind conflict: on a real port conflict the operator gets two messages — start0's error!("...failed to bind...: Address already in use") and then the panic from wait_bind(...).unwrap() ("...the port may already be in use"). Functionally fine (the unwrap is what exits the process, matching the rpc_status.wait_running().unwrap() pattern), but consider logging the detailed cause in one place and keeping the other terse to avoid double noise.

Neither affects correctness; the gating logic and process-exit behavior are sound.

curvine-web/src/server/web_server.rs
```diff
@@ -69,24 +71,79 @@ where
...
+    }
+
+    async fn start0(server: Self) {
+        let bind_addr = server.get_bind_addr();
+        if let Err(e) = server.run().await {
+            error!(
+                "WebServer [{}] failed to bind address {} (port {}): {}",
+                server.conf.name, bind_addr, server.address.port, e
+            );
+        }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks for catching this. I split bind and serve into separate phases: bind failures log failed to bind address {addr}: {err}, and post-bind/serve failures log exited with error on address {addr}: {err}. block_on_start now uses the general exit message as well.

curvine-web/src/server/web_server.rs
lzjqsdd: Thanks for the review. I routed all startup messaging through resolve_bind_addr() for consistency, and deduplicated bind-conflict reporting: the detailed OS error is logged once in start0, while wait_bind returns a shorter message for the startup panic path. Please take another look.

szbr00 (approved):

PR #945, opened 2026-06-26
fix(fuse): keep renamed symlink destination unlinkable after overwrite

ljy1-lixing:
## Summary
Fix a Curvine FUSE rename-overwrite inconsistency where renaming a symlink over an existing symlink leaves the destination path visible
  to `lstat/stat`, but a subsequent `unlink` returns `ENOENT`.

  ## Problem

  LTP `rename01A symlink01 -T rename01` can pass the rename assertions but fail during cleanup:

  ```text
  rename("symbolic", "asymbolic") = 0
  newfstatat(..., "asymbolic", ..., AT_SYMLINK_NOFOLLOW) = 0
  newfstatat(..., "asymbolic", ..., 0) = 0
  unlink("asymbolic") = -1 ENOENT
  ```

  A minimal reproducer is:

  symlink("nobody", "symbolic")
  rename("symbolic", "asymbolic")

  creat("object")
  symlink("object", "symbolic")
  rename("symbolic", "asymbolic")

  lstat("asymbolic") succeeds
  stat("asymbolic") succeeds
  unlink("asymbolic") fails with ENOENT

  This is not the same as #845/#854. Those fixed POSIX rename error semantics for file/directory combinations. This issue is a post-
  success FUSE namespace inconsistency after replacing an existing symlink.

  ## Solution

  Update FUSE NodeMap::rename_node() so rename does not incorrectly decrement the source node lookup count. rename is not forget, and
  reducing n_lookup during rename can make the local inode/name lifecycle inconsistent.

  Also handle overwrite cleanup for an existing destination name by removing the overwritten node when it no longer has lookup/ref
  references.

  Added regression coverage for:

  - FUSE-local NodeState rename-over-existing-name behavior.
  - Server-side symlink-over-symlink rename followed by delete.
  - FUSE rename regression script case for symlink-over-symlink followed by unlink.

  ## Tests

  Passed locally:

  cargo fmt --check
  python3 -m py_compile build/tests/scripts/rename_test.py

  Rust tests were not runnable in my local environment because protoc is missing:

  Could not find `protoc` installation

  Recommended tests after installing protobuf/protoc:

  cargo test -p curvine-fuse rename_over_existing_name_keeps_destination_unlinkable -- --nocapture
  cargo test -p curvine-server rename_posix_semantics -- --nocapture

  ## How to verify

  Build and remount Curvine FUSE, then run the minimal C reproducer:

  gcc -Wall -Wextra -O2 -o repro_rename_over_symlink repro_rename_over_symlink.c
  ./repro_rename_over_symlink /tmp
  sudo ./repro_rename_over_symlink /curvine-fuse

  Expected result on Curvine after the fix:

  lstat(asymbolic): ...
  stat(asymbolic): ...
  unlink(asymbolic): OK
  PASS

  Also verify the original LTP case:

  sudo env LTPROOT=/opt/ltp-20210524 \
    PATH=/opt/ltp-20210524/testcases/bin:$PATH \
    TMPDIR=/curvine-fuse \
    symlink01 -T rename01

  Expected result:

  TPASS ...
  rc=0


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -445,7 +445,14 @@ impl NodeState {
...
             match map.lookup_node(parent, name) {
                 Some(v) => v.id,
-                None => return err_fuse!(libc::ENOENT, "node {} not found", parent),
+                None => {
+                    debug!(
+                        "unlink node cache miss for parent={}, name={:?}; deleting backend path",
+                        parent,
+                        name.map(|v| v.as_ref().to_string())
+                    );
+                    return Ok(true);
```
szbr00: This cache-miss fallback bypasses the open-handle delayed-delete path below. `should_delete_now()` currently looks up the cached node id, checks `has_open_handles(id)`, marks `pending_delete`, and only then defers backend deletion. With this change, if the name mapping is missing we immediately return `Ok(true)`, so `unlink()` proceeds to `self.fs.delete(&path, false)` even if there is still an open `FileHandle` for that path/inode.

That cache-miss state is possible because `NodeMap::clean_cache()`/`should_unref()` only consider `n_lookup` and `ref_ctr`; they do not consider `NodeState.handles`. A file can still be open after the kernel has dropped lookup refs, so the node mapping may be evicted while the handle remains. In that case this change turns a delayed-delete case into an immediate backend delete, which can break reads/writes through the still-open handle.

The symlink overwrite regression needs the second `should_delete_now()` after `unlink_node()` to stop returning `ENOENT`, but this fallback is broader than that test path. Consider preserving delayed-delete semantics on cache miss by checking open handles by path (or by preventing nodes with open handles from being evicted), rather than treating every missing name mapping as safe for immediate backend deletion.

szbr00 (approved):

PR #948, opened 2026-06-29
feat(fluid): adapt cache runtime integration

jlon:
## Background

Fluid has moved Curvine integration to the generic CacheRuntime flow. The old Docker material under `curvine-docker/fluid` assumed the older runtime config shape and did not provide a `MountUFS` entry that matches the current Fluid contract.

This PR keeps the diff focused on the adapter code and the core cache-runtime manifests needed for Fluid `1.1.0-alpha.9`.

### RuntimeConfig parsing

- Updated `curvine-docker/fluid/generate_config.py` to parse both the current top-level Fluid RuntimeConfig shape and the older `topology` shape.
- Added support for:
  - `master.service.name` and generated journal addresses
  - `worker.tieredStoreLevels` cache paths and quota conversion
  - top-level `targetPath` as the FUSE mount path
  - component `options` merging into Curvine TOML sections
  - `FLUID_DATASET_NAME` as the default cluster id
- Kept `dir_reserved` as a string while converting only port-like values to integers.
- Kept TOML rendering self-contained so the runtime image does not need extra Python TOML packages.

### MountUFS integration

- Added `curvine-docker/fluid/mountUfs.sh`.
- The script reads `FLUID_RUNTIME_CONFIG_PATH`, handles Fluid `encryptOptions` secret file paths, maps S3 options into Curvine `cv mount -c ...` options, and prints only Fluid-compatible JSON on stdout.
- Native Curvine mounts such as `curvine:///data` are treated as no-op mounts and reported as mounted Curvine paths.

### Runtime entrypoint and image

- Fixed cache-runtime startup under `set -u`.
- Preserved the Curvine image environment while forcing Fluid-generated config paths back to `/app/curvine`.
- Runs Curvine master and worker in the foreground so Kubernetes gets the real process status and logs.
- Simplified the Fluid Dockerfile to copy adapter scripts on top of the Curvine runtime image.

### Core cache-runtime manifests

- Reworked `CacheRuntimeClass` to use the current Fluid generic CacheRuntime `template` and `executionEntries.mountUFS` structure.
- Added master and worker readiness probes.
- Updated the example Dataset/CacheRuntime to use the new parser and mount flow.

### Build flow

- Fixed `.github/workflows/build-fluid-image.yml` so the Docker build context matches the Fluid Dockerfile layout.
- Added configurable Makefile variables:
  - `FLUID_IMAGE`, default `curvine-fluid:latest`
  - `FLUID_BASE_IMAGE_TAG`, default `latest`
- Redirected the old `docker-build-fluid-cache` and `docker-build-fluid-thin` Makefile targets to the unified Fluid image build instead of missing scripts.

## Local Verification

Commands run:

```bash
bash -n curvine-docker/fluid/entrypoint.sh curvine-docker/fluid/mountUfs.sh
python3 -m py_compile curvine-docker/fluid/generate_config.py curvine-docker/fluid/config-parse.py
make docker-build-fluid
kubectl apply --dry-run=client \
  -f curvine-docker/fluid/cache-runtime/curvine-cache-runtime-class.yaml \
  -f curvine-docker/fluid/cache-runtime/curvine-dataset.yaml
```

Local image built:

```text
curvine-fluid:latest
sha256:023a7fc17fbf956684713838e171042db1cbf5d3c47b9297202de3e331aaec4f
```

I also verified the adapter locally against Fluid chart `1.1.0-alpha.9` before trimming the PR diff to core files. Both native Curvine and S3-backed cache-runtime flows reached Ready locally:

```text
curvine-demo master=Ready worker=Ready client=Ready
curvine-s3-demo master=Ready worker=Ready client=Ready
```

The S3 local run confirmed `MountUFS` registered `/minio -> s3://test/` and the Fluid PVC could read from UFS and write/read Curvine cache data.

## Compatibility Notes

- The parser keeps compatibility with the old `topology` RuntimeConfig shape while supporting the current generic CacheRuntime shape.
- `MountUFS` keeps logs on stderr and writes only the final JSON object to stdout, as required by Fluid.
- Curvine `fs_mode` writes to Curvine cache, not back to the UFS. The local S3 verification covered UFS read and cache write/read behavior, not write-through.

## References

- Fluid generic CacheRuntime integration doc: https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/dev/generic_cache_runtime_integration.md
- Fluid Curvine e2e example: https://github.com/fluid-cloudnative/fluid/blob/master/test/gha-e2e/curvine/cacheruntimeclass.yaml
- Fluid chart release used for local testing: https://github.com/fluid-cloudnative/charts/releases/tag/helm-chart-fluid-1.1.0-alpha.9

### DIFF
--- a/curvine-docker/fluid/mountUfs.sh
+++ b/curvine-docker/fluid/mountUfs.sh
@@ -1,0 +1,178 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+CONFIG_PATH="${FLUID_RUNTIME_CONFIG_PATH:-/etc/fluid/config/config.json}"
+CURVINE_HOME="${CURVINE_HOME:-/app/curvine}"
+CURVINE_CONF_FILE="${CURVINE_CONF_FILE:-${CURVINE_HOME}/conf/curvine-cluster.toml}"
+CV_BIN="${CURVINE_CLI:-${CURVINE_HOME}/bin/cv}"
+
+log() {
+  echo "$@" >&2
+}
+
+if [ ! -f "$CONFIG_PATH" ]; then
+  log "Fluid runtime config not found: $CONFIG_PATH"
+  printf '{"mounted":[]}\n'
+  exit 0
+fi
+
+python3 - "$CONFIG_PATH" "$CV_BIN" "$CURVINE_CONF_FILE" <<'PY'
+import json
+import os
+import shutil
+import subprocess
+import sys
+
+config_path, cv_bin, conf_file = sys.argv[1:4]
+
+
+def eprint(*args):
+    print(*args, file=sys.stderr)
+
+
+def load_config(path):
+    with open(path, "r", encoding="utf-8") as f:
+        content = f.read().strip()
+    if not content:
+        return {}
+    data = json.loads(content)
+    if isinstance(data, str):
+        data = json.loads(data)
+    if not isinstance(data, dict):
+        raise ValueError(f"runtime config must be a JSON object, got {type(data).__name__}")
+    return data
+
+
+def read_secret(value):
+    if value and os.path.isfile(value):
+        with open(value, "r", encoding="utf-8") as f:
+            return f.read().strip()
+    return value
+
+
+def options_for_mount(mount):
+    options = dict(mount.get("options") or {})
+    encrypt_options = mount.get("encryptOptions") or {}
+
+    if isinstance(encrypt_options, list):
+        for item in encrypt_options:
+            name = item.get("name")
+            value_from = item.get("valueFrom") or {}
+            secret_key_ref = value_from.get("secretKeyRef") or {}
+            if name and secret_key_ref.get("path"):
+                encrypt_options[name] = secret_key_ref["path"]
+    if isinstance(encrypt_options, dict):
+        for key, value in encrypt_options.items():
+            options[key] = read_secret(value)
+
+    params = []
+    key_map = {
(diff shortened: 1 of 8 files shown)


curvine-docker/fluid/mountUfs.sh
```diff
@@ -0,0 +1,178 @@
...
+    options = dict(mount.get("options") or {})
+    encrypt_options = mount.get("encryptOptions") or {}
+
+    if isinstance(encrypt_options, list):
+        for item in encrypt_options:
+            name = item.get("name")
+            value_from = item.get("valueFrom") or {}
+            secret_key_ref = value_from.get("secretKeyRef") or {}
+            if name and secret_key_ref.get("path"):
+                encrypt_options[name] = secret_key_ref["path"]
```
szbr00, line 63: This branch crashes for the Fluid `encryptOptions` list shape. When `encryptOptions` is a list, `encrypt_options` is still a Python list here, so `encrypt_options[name] = ...` tries to index a list with a string and raises `TypeError: list indices must be integers or slices, not str`. That means any dataset using Fluid secret-backed `encryptOptions` fails before running `cv mount`.

Consider normalizing the list into a new dict, e.g. `resolved_encrypt_options = {}` and then filling `resolved_encrypt_options[name] = secret_key_ref["path"]`, before merging it into `options`. It would also be worth covering this with a small sample config test because this is one of the main mountUFS paths added by the PR.

lzjqsdd (approved):

PR #952, opened 2026-06-29
fix(fuse): avoid reusing completed writer after truncate

ljy1-lixing:
## Summary

Fix FUSE write failures after truncating a file and then reopening it for more writes.

A workload can repeatedly open a file, append data, truncate it to a smaller size, close it, then reopen and continue writing from the new EOF. Before this change, the truncate path updated file metadata successfully, but the FUSE/client writer state could remain stale or reuse a completed writer. The next write then failed with `EIO`, with logs showing `channel closed`.

## Minimal Reproducer

```c
#include <fcntl.h>
#include <unistd.h>
#include <string.h>
#include <stdio.h>
#include <errno.h>

    if (argc < 2) {
        fprintf(stderr, "usage: %s <file>\n", argv[0]);
        return 1;
    }

    char buf[20480];
    memset(buf, 'a', sizeof(buf));

    for (int i = 1; i <= 20; i++) {
        int fd = open(argv[1], O_RDWR | O_CREAT | O_SYNC, 0777);

        if (lseek(fd, 0, SEEK_END) < 0) { perror("lseek"); return 2; }

        ssize_t n = write(fd, buf, sizeof(buf));
        if (n != sizeof(buf)) {
            fprintf(stderr, "iter=%d write ret=%zd errno=%d\n", i, n, errno);
            perror("write");
            return 3;
        }

        if (i % 10 == 0) {
            off_t cur = lseek(fd, 0, SEEK_CUR);
            off_t new_size = cur - 20480;
            if (new_size < 0) new_size = 0;

            if (ftruncate(fd, new_size) != 0) {
                fprintf(stderr, "iter=%d truncate to %lld failed errno=%d\n",
                        i, (long long)new_size, errno);
                perror("ftruncate");
                return 4;
            }
        }

        close(fd);
    }

    return 0;
}
```

gcc /tmp/cv_gf30_mini.c -o /tmp/cv_gf30_mini
rm -f /curvine-fuse/cv_gf30_mini
/tmp/cv_gf30_mini /curvine-fuse/cv_gf30_mini

Before the fix, it failed at iteration 11:

iter=11 write ret=-1 errno=5
write: Input/output error

## Root Cause

Two writer state issues were involved:

1. Resize/truncate updated the file length, but writer positions in the layered client writers were not clamped to the new file length.
2. NodeState::new_writer() reused an existing FuseWriter for the inode even after that writer had already been completed on release(). Reusing this completed writer caused subsequent writes to send through a closed channel.

A representative debug log showed:

FuseWriter write failed path=/cv_gf30_mini off=184320 len=20480 writer_pos=184320 writer_len=184320 err=IO([]channel closed)

## Changes

- Clamp writer position after resize in:
    - FsWriterBase
    - FsWriterBuffer
    - FsWriter

- Refresh FsWriterBuffer cached FileBlocks after resize.
- Track completed state in FuseWriter.
- Avoid reusing completed FuseWriter instances in NodeState::new_writer().
- Update FUSE writer length cache after resize.

## Verification

- Minimal C reproducer passes after the fix.
- cargo fmt --check passes.

### DIFF
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -380,37 +380,118 @@ impl NodeState {
             return err_fuse!(libc::EINVAL, "Invalid flags: {:?}", flags);
         };
 
-        let mut lock = self.handles.write();
-
-        // Check if writer already exists to prevent duplicate creation
-        let check_writer = if let Some(writer) = writer {
-            if let Some(exist_writer) = Self::find_writer0(&lock, &ino) {
-                if exist_writer
-                    .try_lock()
-                    .map(|writer| !writer.is_completed())
-                    .unwrap_or(false)
-                {
-                    Some(exist_writer)
-                } else {
-                    Some(writer)
-                }
-            } else {
-                Some(writer)
-            }
-        } else {
-            None
-        };
-
-        let handle = Arc::new(FileHandle::new(
-            ino,
-            self.next_fh(),
-            reader,
-            check_writer,
-            status,
-        ));
-        Self::insert_file_handle_locked(&mut lock, handle.ino, handle.fh, handle.clone());
+        let handle = self
+            .insert_handle_with_writer(ino, reader, writer, status)
+            .await;
 
         Ok(handle)
+    }
+
+    async fn insert_handle_with_writer(
+        &self,
+        ino: u64,
+        reader: Option<RawPtr<FuseReader>>,
+        writer: Option<Arc<Mutex<FuseWriter>>>,
+        status: FileStatus,
+    ) -> Arc<FileHandle> {
+        let mut candidate_writer = writer;
+        let mut ignored_completed: Vec<Arc<Mutex<FuseWriter>>> = vec![];
+        let mut reader = Some(reader);
+        let mut status = Some(status);
+
+        loop {
+            let writer = match candidate_writer.take() {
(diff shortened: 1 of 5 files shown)


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -383,7 +385,15 @@ impl NodeState {
         // Check if writer already exists to prevent duplicate creation
         let check_writer = if let Some(writer) = writer {
             if let Some(exist_writer) = Self::find_writer0(&lock, &ino) {
-                Some(exist_writer)
+                if exist_writer
+                    .try_lock()
+                    .map(|writer| !writer.is_completed())
+                    .unwrap_or(false)
```
szbr00: This treats a lock-contention failure the same as a completed writer. If `exist_writer.try_lock()` returns `Err` because another task is currently using the writer, `unwrap_or(false)` falls into the `else` branch and installs the newly opened `writer` instead. That can create two active `FuseWriter`s for the same inode, which is exactly what this re-check is meant to avoid.

Before this change, the write-lock re-check always reused `exist_writer`; after the change, a busy-but-live writer can be bypassed. Consider awaiting the lock here (or otherwise distinguishing "lock busy" from "completed") and only replacing the existing writer when `is_completed()` is actually true.

szbr00 (approved):

PR #954, opened 2026-06-30
feat(fuse): add stream IO metrics (phase 2b)

yangcx000:
## Summary

Phase 2b of the curvine-fuse Prometheus metrics work (stacked after Phase 2a #949, now merged). Adds the internal stream-IO breakdown for read/write/flush/fsync/release; rebased onto current `main` (resolved clean field-addition conflicts with #952 in `fuse_writer.rs`, #952's `is_completed()` reuse-guard preserved).

## Metrics added

- **`io_*` family** (read/write backend, labelled `status` + `path_type`): `io_duration_us`, `io_bytes_total` (success-child only — actual bytes), `io_requests_total` (every attempt), `io_size_bytes`, `stream_io_inflight`, plus `stage_duration_us{stage=stream_io}`. Observed inside the reader/writer **task body** at the real backend call — not at FUSE enqueue time.
- **`stream_lifecycle_*` family** (flush/fsync/release — **no status**, **independent family**, `path_type=unknown` initially): `stream_lifecycle_duration_us`, `stream_lifecycle_requests_total`, `stream_lifecycle_inflight`. Attributed at the `send_stream` layer (operator-distinguished; `release` counted exactly once, no writer/reader double-count).
- **`io_dispatch_duration_us{io_type}`** (read/write dispatch at `send_stream`): dispatch-to-worker latency, **not pure channel enqueue** (read may do a read-after-write consistency `flush()` + reader reopen before enqueuing; zero-length write takes a no-op direct-reply path). Named `dispatch`, not `enqueue`, to reflect that.
- **`stream_write_queue_depth`**: writer-channel backlog via a reserve-first task-embedded guard; `write_ver.incr()` timing unchanged.

## Design notes

- **Two families, not one**: read/write carry a clean `status` and go to `io_*`; flush/fsync/release mix backend + enqueue errors so a `status` label would be polluted — they use the status-less `stream_lifecycle_*` family. Prometheus fixes the label set per family, so the two cannot be merged. `stage_duration_us{stage=stream_io}` therefore covers **read/write only**.
- **`path_type`** is captured at handle construction (`curvine`|`ufs`|`fallback`|`local`); `stream_lifecycle_*` uses `unknown` in this initial version.
- `send_stream`'s dispatch core is factored into a private fd-free `send_stream_dispatch` so the metrics logic is unit-testable without a real `FuseReceiver`; the metrics gate is derived from `rep.metrics.is_some()`. All new guards are `None` when disabled (no `noop()`).

## Testing

- `cargo fmt -p curvine-fuse -p curvine-client --check`: clean.
- `cargo clippy -p curvine-fuse --all-targets -- --deny=warnings --allow clippy::uninlined-format-args`: 0.
- `cargo test -p curvine-fuse --lib` (default concurrency): **130 passed**.
- e2e on a local cluster (rebuilt release fuse binary): `io_*{path_type=curvine}` / `io_dispatch_*` / `stream_lifecycle_*{unknown}` correct under read/write/flush/release load; `stage=stream_io` count == io_duration{read}+{write}; inflight/queue return to zero; negative assertions pass (the two families never cross, no `stream_enqueue` stage). (e2e was re-run against the post-review build; will re-verify against the rebased binary before merge.)


yangcx000:
### e2e re-verified against the rebased binary (`08013d7`)

The PR description noted e2e would be re-run against the rebased binary (the earlier e2e was on the pre-rebase build, before #945/#952 landed on `main`). Done — rebuilt the release fuse binary at `08013d7` and re-ran on the local cluster (fuse log confirms `git version: 08013d7`). Load: write×3 + read×3 + fsync.

**Values correct:**
- `io_*{read,curvine,success}` count=27 (25 MiB), `{write}` count=28 (29 MiB) — reads actually hit the backend, `requests_total == duration_count` on the success path.
- `io_dispatch_duration_us{read}`=27 / `{write}`=28 — one-to-one with `io_duration`.
- `stream_lifecycle_*{flush/fsync/release,unknown}` = 14 / 1 / 7 (fsync=1 matches the single `conv=fsync`).
- `stage_duration_us{stage=stream_io,kind=stream}` count = **55 == io_duration{read}(27) + {write}(28)`.

**Invariants hold:**
- `io_*` never carries flush/fsync/release; `stream_lifecycle_*` never carries read/write (the two families stay disjoint).
- No `stage=stream_enqueue`.
- All `stream_io_inflight` / `stream_lifecycle_inflight` / `stream_write_queue_depth` return to 0 when idle (no negatives).

The rebase (clean field-addition merge with #952 in `fuse_writer.rs`, #952's `is_completed()` reuse-guard preserved) did not alter 2b's metric behaviour. Local gate also green: `cargo fmt --check`, `clippy --deny=warnings`, `cargo test -p curvine-fuse --lib` (130 passed).

szbr00 (approved):

PR #959, opened 2026-07-01
feat(fuse): add cache and readdir metrics (phase 3a)

yangcx000:
## Summary

Phase 3a of the curvine-fuse Prometheus metrics work (stacked after Phase 3b-1 #958, now merged). Adds userspace-cache and readdir observability:

- `curvine_fuse_user_meta_cache_total{cache,status}` — MetaCache hit/miss/put (cache∈{status,list,blocks}). A miss is recorded the moment the cache read returns `None` (before the backend fetch), so a backend error/ENOENT after a miss stays in the hit-rate denominator; `put` only when actually written back (incl. the blocks `put_open` on Cv/Fallback readers).
- `curvine_fuse_user_meta_cache_invalidations_total{cache,reason}` — `invalidate_cache` refactored to carry a static `reason` at all 17 call sites (15-value bounded enum). One inc per affected cache namespace (not per call).
- `curvine_fuse_node_cache_total{operation=lookup,status=hit|miss}` — NodeMap dcache outcome on the **real FUSE Lookup path only**; the hit/miss is decided inside the node-map write lock (before `find_node` materializes) and emitted after the lock is dropped. readdir/mkdir/create/link/symlink pass `record=false`; `.`/`..` not counted; no `put`.
- `curvine_fuse_negative_entry_returned_total` — negative dentry results (backend ENOENT + negative_ttl>0), counted separately from a positive hit.
- `curvine_fuse_readdir_entries{status}` / `curvine_fuse_readdir_duration_us{status}` — observed in `read_dir_common` via a `ReaddirTimer` guard. entries = successful `add_dirent`s in the batch (single readdir syscall's worth, not the directory total); only the success child is observed; error/cancellation records duration{error} with no entries. Excludes opendir/list_stream backend init.

## Design

- All emission gated on the `metrics_enabled` kill-switch (distinct from the `enable_meta_cache` feature gate) via `FuseMetrics::with` — a disabled mount emits nothing and pays no label-lookup cost.
- The node_cache hit/miss is decided in the write lock but emitted out of the lock (no registry/label work under the lock).
- Same-name metric label sets are fixed; no dynamic label changes.

## Testing

- `cargo fmt -p curvine-fuse --check`: clean; `cargo clippy --all-targets -- --deny=warnings`: 0; `cargo test -p curvine-fuse --lib`: 143 passed.
- Tests cover: hit/miss/put (incl. miss-on-backend-error), invalidation multi-namespace counting (+3 root / +4 with parent), node_cache hit/miss + non-lookup paths not counted, readdir success-vs-error (no entries on error), blocks-cache production put wiring, and metrics-disabled no-emission.


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -273,6 +303,8 @@ impl CurvineFileSystem {
                         let path = Path::from_str(&status.path)?;
                         self.state.meta_cache().put_status(&path, status.clone());
```
szbr00, line 304: This `put_status` write is not counted in `user_meta_cache_total{cache="status",status="put"}`. The new metric is documented as counting puts when values are actually written back, but readdir populates the status cache here without calling the same record path used by `get_cached_status()`. That will under-report status-cache puts, especially for directory-heavy workloads. Consider recording `CACHE_STATUS/CACHE_RESULT_PUT` here as well, gated by `metrics_enabled`.

yangcx000: Good catch — fixed. The readdir path now records `user_meta_cache_total{cache="status",status="put"}` right after the `put_status`, matching `get_cached_status()`'s put path and gated by `metrics_enabled` via `record_meta_cache`. Without this, directory-heavy workloads under-reported status-cache puts. Pushed to the branch (amended into the single phase-3a commit). fmt/clippy clean, 143 tests pass.

szbr00 (approved):

PR #961, opened 2026-07-01
feat(fuse): add state-recovery and session-lifecycle metrics (phase 3b)

yangcx000:
## Summary

Phase 3b (final segment) of the curvine-fuse Prometheus metrics work, stacked after Phase 3a #959 (now merged). Adds state-recovery and session-lifecycle observability.

**State persist/restore** (SIGUSR1 / restart-with-env):
- `state_persist_total{status}` / `state_restore_total{status}` — recorded once at the fuse_session persist/restore entry/exit.
- `state_{persist,restore}_stage_duration_us{stage,status}` — per-stage timing via a `StateStageTimer` guard (records `status=error` on drop unless `success()` disarms it, so early-`?`/cancellation counts as error). Four stages span two files: `mount_fds` in `fuse_session.rs`, `node_map`/`file_handles`/`dir_handles` in `NodeState`. A **dedicated** stage domain — values never enter the request `stage_duration_us` enum.
- `state_persist_handle_count{kind}` — node_map/file_handles/dir_handles counts sampled once at the start of persist (best-effort snapshot).
- NodeState magic/version header failure records only `state_restore_total{error}` with no NodeState stage (mount_fds precedes the header in the file format and may already be recorded). Top-level tail I/O errors (open/flush/reload) are stage-less by design. The persist path explicitly `flush()`es the state file before spawning the restore child.

**Session lifecycle**:
- `session_init_total{result}` — `FuseSession::new` wrapped so every early `?` counts as error.
- `session_shutdown_total{reason}` — a `ShutdownOnce` CAS records the first cause exactly once. The `run_all` arm splits its three outcomes into `completed`/`run_all_error`/`run_all_panic`; signal arms record `term_signal`/`sigusr1_persist`; the fd-watcher records `fd_watcher`.
- `curvine_fuse_kernel_fd_health` — single gauge (1=healthy/0=hup-or-exited): set 1 at end of `new`, set 0 by the fd-watcher on POLLHUP/ERR and by a `RunCleanupGuard` on every `run()` exit (incl. the SIGUSR1-persist `?` early return). The fd-watcher returns a JoinHandle held by the cleanup guard and aborted on exit, so a stale watcher can't cross-session-overwrite the process-global gauge.

All emission gated on the `metrics_enabled` kill-switch. `RunCleanupGuard::Drop` is lightweight/synchronous/infallible (only abort + gauge set).

## Testing

- `cargo fmt -p curvine-fuse --check`: clean; `cargo clippy --all-targets -- --deny=warnings`: 0; `cargo test -p curvine-fuse --lib`: 153 passed.
- fd-free seam tests: ShutdownOnce first-cause-once, StateStageTimer success/drop-error, RunCleanupGuard abort + health close-out, state-vs-request stage disjoint, shutdown reason enum.
- e2e verified on a local cluster (rebuilt release binary): SIGUSR1 persist → reload → restore fd handoff succeeds; state_restore_* 4-stage metrics correct; the two stage families stay disjoint; session_init/kernel_fd_health set on startup.


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -831,38 +835,77 @@ impl NodeState {
...
+            let handles = self.all_handles();
+            writer.write_len(handles.len() as u64)?;
+            for handle in &handles {
+                if let Err(e) = handle.persist(writer).await {
+                    error!("node_state::persist: error saving file_handle {:?}", e)
+                }
+            }
+            info!("node_state::persist: {} file_handles saved", handles.len());
+            if let Some(stage) = stage {
+                stage.success();
```
szbr00: This marks the `file_handles` persist stage as successful even if one or more `handle.persist(writer).await` calls failed above, because those errors are only logged and the loop continues. The restore path has the same pattern: failed `FileHandle::restore` entries are skipped, but the stage and top-level restore can still report `success`.

With the new state metrics, this can hide partial/corrupt state recovery as a successful stage. Consider tracking whether any per-handle persist/restore failed and either returning an error or recording the stage/top-level status as error when the restored/saved count is incomplete.

szbr00 (commented):
## Code Review: PR #961

**Scope:** +1,730 / -142 across 6 files — Phase 3a (cache + readdir metrics) and Phase 3b (state-recovery + session-lifecycle metrics) for the Curvine FUSE filesystem.

---

### Critical (must fix before merge)

None.

---

### Suggestions (should consider)

**1. Commit message missing issue ID**

The commit headline `feat(fuse): add state-recovery and session-lifecycle metrics (phase 3b)` is missing the required `(#961)` issue ID. All existing merged commits on `main` include `(#PR-number)` (e.g. `(#958)`, `(#954)`). If this PR will be squash-merged, GitHub will auto-append it — but if `commitlint` runs on the PR branch commits in CI, it will fail. Consider amending the commit or confirming the CI gate runs only on the squash-merge commit.

**2. `read_dir_common_inner` — potential underflow on entry count**

`curvine-fuse/src/fs/curvine_file_system.rs`, `read_dir_common_inner`:

```rust
let entries = index - arg.offset;
```

The subtraction could panic on underflow if `index < arg.offset` (both are `u64`). In practice `index` starts at `arg.offset` and only increases, so this is safe today — but a future refactor could break the invariant silently. Consider `index.saturating_sub(arg.offset)` or add an assertion.

**3. `ShutdownOnce` — fully-qualified paths instead of `use` imports**

`curvine-fuse/src/fuse_metrics.rs`, `ShutdownOnce`:

Using fully-qualified paths (`std::sync::Arc`, `std::sync::atomic::AtomicBool`, `std::sync::atomic::Ordering`) instead of `use` imports adds verbosity. Consider:

```rust
use std::sync::{Arc, atomic::{AtomicBool, Ordering}};
```

at the struct definition or top of the impl block.

**4. `persist_inner` — synchronous I/O in async context**

`curvine-fuse/src/session/fuse_session.rs`, `persist_inner`:

The `writer.flush()?` call is synchronous I/O inside an `async fn`. This is consistent with the existing `writer.write_struct()` pattern and only runs on the SIGUSR1 persist path (not a hot path), so it is an acceptable trade-off. But worth noting that strictly, `std::io::Write::flush` is blocking. If this ever moves to a per-request path, it would need `tokio::task::spawn_blocking`.

---

### Notes (informational)

- **Excellent bug fixes included in this PR:**
  - `sys::fcntl_set(mnt.fd, flags & !libc::FD_CLOEXEC)?` now propagates errors via `?` (previously silently ignored). A silent failure here would produce an unusable state file — the fd could not be inherited by the child process.
  - `writer.flush()?` before `reload_param` spawn is a real correctness fix — the BufWriter would otherwise only flush on drop (after the child has started), so the child could read a truncated state file.

- **Outstanding test coverage** — 20+ tests covering all new code paths, including:
  - Parallel-safe discipline (unique synthetic labels, lower-bound deltas on shared counters)
  - Deterministic gate tests (`ReaddirTimer::start(false)` → `None`, `StateStageTimer` success/drop-error)
  - `ShutdownOnce` CAS first-cause-wins (including disabled-still-dedups)
  - `RunCleanupGuard` abort behavior (fd-free tokio test)
  - `state_stages_disjoint_from_request_stages` — regression guard against label collision
  - `do_lookup_probed` hit/miss/record=false/name=None probe semantics

- **`RunCleanupGuard::Drop` is well-designed** — lightweight, synchronous, infallible. Only `abort()` + gauge set. No `await`/`join`/panic-prone work in `Drop`, which is correct for a destructor.

- **`close_out_health` extraction** is a good pattern — makes the enabled-gate close-out unit-testable against an injected taker without touching the process-global gauge.

- **`do_lookup_probed` design** correctly decides hit/miss inside the write lock (atomic with the lookup) but emits the metric after the lock is dropped. The two-flag split (`is_real_lookup` + `metrics_enabled`) prevents a future caller from accidentally enabling the metric without the gate.

- **Design doc update** is thorough — documents the 6 shutdown reasons, single-gauge `kernel_fd_health`, cross-file stage layout, header-no-stage semantics, and the best-effort snapshot caveat for `state_persist_handle_count`.

- **Architecture fit** — all changes are within `curvine-fuse`, which is the correct crate for FUSE metrics. No server or client crate boundary violations. `orpc::common::Counter`/`Metrics` used correctly as workspace dependencies.

---

*Review generated using the [curvine-code-review](.qoder/skills/curvine-code-review/) skill.*

curvine-fuse/src/fs/state/node_state.rs
yangcx000: Fixed. Both the persist and restore `file_handles` stages now track an `any_failed` flag and only call `stage.success()` when no per-handle failure occurred; on any skipped/failed handle the `StateStageTimer` is left to drop as `status=error`. This keeps the best-effort control flow intact (a corrupt handle is still logged and skipped, the rest are still saved/restored) but stops the metrics from reporting a partial/corrupt recovery as a clean `success` stage. The top-level `state_{persist,restore}_total` still reflects the overall Ok/Err as before. Pushed (amended into the phase-3b commit); fmt/clippy clean, 153 tests pass.

yangcx000:
Thanks for the thorough review. Addressed:

- **file_handles stage marked success despite per-handle failures** (inline): fixed — persist & restore now leave the `StateStageTimer` to drop as `error` when any handle fails, without changing the best-effort skip-and-continue control flow. See inline reply.
- **Suggestion #2 (`index - arg.offset` underflow)**: changed to `index.saturating_sub(arg.offset)` with a comment on the current non-underflow invariant.
- **Suggestion #1 (commit message missing `#961`)**: the CI `commitlint` check is green on this PR, and the repo squash-merges (auto-appending `(#961)`), so no branch-commit amend needed.
- **Suggestion #3 (`ShutdownOnce` fully-qualified paths)**: stylistic; left as-is to keep the diff minimal, can tidy in a follow-up if preferred.

fmt/clippy clean, `cargo test -p curvine-fuse --lib` 153 passed.

szbr00 (approved):

Issue #955, opened 2026-06-30
[BUG] cache_mode worker-failure fallback applies fs_mode consistency semantics, refusing valid S3 data


yangyang233333:
**Describe the bug**

Under a `cache_mode` S3 mount, when a Worker fails mid-read and `FallbackFsReader` degrades to reading directly from S3, the fallback path applies **fs_mode consistency semantics**, which are wrong for cache_mode. A perfectly valid, up-to-date S3 object can be **refused** at exactly the moment fallback is most needed.

`FallbackFsReader` is shared by both `fs_mode` and `cache_mode`, but the two modes have **opposite authority directions**:

| | fs_mode | cache_mode |
|---|---|---|
| Authority | Curvine (UFS is a flushed copy) | **S3/UFS** (Curvine is a read-through cache) |
| `ufs_mtime/len` snapshot means | "what UFS should look like since our last flush" | "what S3 looked like at warm time" |
| Snapshot ≠ UFS means | the UFS copy is untrustworthy (incomplete flush / external edit) | **S3 was updated out-of-band** — i.e. the fresh, authoritative data |

The fallback path uses the fs_mode interpretation unconditionally, so in cache_mode it treats "S3 is newer than our snapshot" as "data is corrupt" and refuses to serve it.

Verified against `main` (commit `c2e5e4e` area; paths below are current `main`).

**Current behavior (code-grounded, `main`)**

`FallbackFsReader::read_chunk0` — `curvine-client/src/unified/fallback_fs_reader.rs:156-172`:

```rust
Err(e) if is_worker_err(&e) => {
    let pos = self.cv_reader.pos();
    // ...
    // Refuse to fall back if UFS data cannot be trusted.
    self.validate_ufs_consistency().await?;   // line 166

    let mut ufs: UfsReader = self.ufs_fs.open_ufs(&self.ufs_path).await?;
    ufs.seek(pos).await?;                      // line 169
    // ...
}
```

`validate_ufs_consistency` (same file, line 63) compares the Curvine snapshot against live S3 via `FileStatus::ufs_valid` (`curvine-common/src/state/file_status.rs:122`, comparing `len` + `ufs_mtime`). If S3 drifted from the snapshot, it returns an error and the read fails.

Two concrete problems in cache_mode:

[...]

PR #963, opened 2026-07-01
feat(fallback): make UFS fallback mode-aware for FsMode vs CacheMode

yangyang233333:
## What

`FallbackFsReader` now carries an `is_fs_mode` flag so worker-failure fallback respects the (opposite) authority direction of the two mount modes.

- **FsMode**: Curvine is authoritative and UFS is only a flushed copy. `validate_ufs_consistency` still gates the fallback and refuses it when the UFS copy has drifted (mtime/len mismatch = incomplete flush or external modification).
- **CacheMode**: S3/UFS is authoritative and Curvine is a read-through cache. Nothing is validated; S3 is read directly at the current `pos`. A still-readable object is served as the current data. A deleted or shrunk-past-`pos` object surfaces the natural `open_ufs`/`seek` error instead of being papered over as a clean EOF. We deliberately do not clamp the seek or pre-check consistency.

Call sites pass `mount.info.is_fs_mode()` instead of a hardcoded bool.

## Tests

New CacheMode fallback integration tests (require `UFS_TEST_PATH`, gated behind a real S3/UFS):

- **TC-17**: S3 shrunk out-of-band, pos 0 -> reads current (shorter) S3 content
- **TC-18**: S3 shrunk past the read pos -> surfaces `Invalid seek position` error (no clamp)
- **TC-19**: S3 unchanged, worker down -> transparent fallback returns the value
- **TC-20**: S3 grown out-of-band, pos 0 -> reads current (longer) S3 content in full

All four pass against a real S3 backend.

### DIFF
--- a/curvine-client/src/unified/fallback_fs_reader.rs
+++ b/curvine-client/src/unified/fallback_fs_reader.rs
@@ -49,7 +49,7 @@ pub struct FallbackFsReader {
     ufs_reader: Option<UfsReader>,
     ufs_path: Path,
     ufs_fs: UfsFileSystem,
-    /// True when the mount is FsMode. Controls whether the snapshot consistency check 
+    /// True when the mount is FsMode. Controls whether the snapshot consistency check
     /// gates the fallback to UFS. CacheMode (false) reads S3 directly without validation.
     is_fs_mode: bool,
 }
--- a/curvine-client/src/unified/fallback_fs_reader.rs
+++ b/curvine-client/src/unified/fallback_fs_reader.rs
@@ -30,8 +30,16 @@ use orpc::sys::DataSlice;
 /// - `cv_reader` is always present and maintains all internal state (pos, chunk, etc.).
 /// - `ufs_reader` is created lazily on the first worker failure; once set, all
 ///   subsequent reads go through UFS (no retry back to Curvine).
-/// - Before falling back, `validate_ufs_consistency` is called to ensure the UFS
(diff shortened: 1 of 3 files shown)


yangyang233333:
close https://github.com/CurvineIO/curvine/issues/955

szbr00 (commented):
After a CacheMode fallback switches to `ufs_reader`, `len()` still reports the cached Curvine length from `cv_reader`. That makes the new “S3 is authoritative” behavior incomplete for callers that size reads from `reader.len()` or use helpers like `read_as_string()`: if S3 grew out-of-band, fallback can read the longer object only when the caller already supplied a larger buffer, while `len()`-based callers will still stop at the stale cached length.

Since `pos()`, `chunk_mut()`, and `chunk_size()` already delegate to the active reader, consider making `len()` delegate to `ufs_reader.len()` once fallback has happened as well.

yangyang233333:
Good catch, thanks! You are right that len() was the odd one out.

Fixed in 3c3d287: len() now delegates to the active ufs_reader once a CacheMode fallback has happened, so S3 is treated as authoritative for size too (len()-based callers like read_as_string now see the current object length). FsMode keeps cv_reader as the metadata authority, matching the existing status()/len() intent, so its behavior is unchanged.

Also added TC-21 as a regression guard: after a CacheMode fallback on a grown S3 object, len() must report the current (longer) length.

szbr00 (approved):

PR #973, opened 2026-07-03
fix(master): make full block reports authoritative

jlon:
## Problem
After worker local storage is replaced or lost, master can keep stale block locations because a full block report only adds reported blocks and an empty worker report is ignored. Reads can then be routed to blocks that no longer exist on that worker.

## Solution
Aggregate chunked full block reports per worker and treat the completed report as the worker's authoritative block set. Remove master-side locations that are absent from the full report, reset partial reports when a worker restarts, accept empty full reports, and return a structured `BlockNotFound` when a worker no longer has a block.

## Test Plan
- `cargo fmt --check`
- `CARGO_TARGET_DIR=/tmp/curvine-pr-split-target cargo clippy -p curvine-common -p curvine-server -p curvine-ufs --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args`
- `CARGO_TARGET_DIR=/tmp/curvine-pr-split-target cargo test -p curvine-common --test fs_error_test block_not_found_round_trips_with_structured_kind --jobs 2 -- --nocapture`

### DIFF
--- a/curvine-server/src/master/fs/master_filesystem.rs
+++ b/curvine-server/src/master/fs/master_filesystem.rs
@@ -26,8 +26,11 @@ use curvine_common::error::FsError;
 use curvine_common::state::*;
 use curvine_common::FsResult;
 use log::warn;
+use orpc::common::LocalTime;
 use orpc::sync::ArcRwLock;
 use orpc::{err_box, err_ext, try_option, CommonResult};
+use parking_lot::Mutex;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 #[derive(Clone)]
@@ -36,7 +39,16 @@ pub struct MasterFilesystem {
     pub worker_manager: SyncWorkerManager,
     pub master_monitor: MasterMonitor,
     pub conf: Arc<MasterConf>,
+    full_block_reports: Arc<Mutex<HashMap<u32, FullBlockReportState>>>,
 }
+
+struct FullBlockReportState {
+    total_len: u64,
+    update_time_ms: u64,
+    reported_blocks: HashSet<i64>,
+}
+
+const FULL_BLOCK_REPORT_TTL_MS: u64 = 60 * 60 * 1000;
 
 impl MasterFilesystem {
     pub fn new(
@@ -50,6 +62,7 @@ impl MasterFilesystem {
             worker_manager,
             master_monitor,
             conf: Arc::new(conf.master.clone()),
+            full_block_reports: Default::default(),
         }
     }
 
@@ -59,6 +72,7 @@ impl MasterFilesystem {
             worker_manager: js.worker_manager(),
             master_monitor: js.master_monitor(),
             conf: Arc::new(conf.master.clone()),
+            full_block_reports: Default::default(),
         }
     }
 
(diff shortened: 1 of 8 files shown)


curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -729,8 +786,28 @@ impl MasterFilesystem {
         }
         drop(wm);
 
+        let mut stale_block_count = 0usize;
         let mut fs_dir = self.fs_dir.write();
-        fs_dir.block_report(batch)
+        if let Some(reported_blocks) = full_reported_blocks {
```
szbr00, line 798: When a completed full report removes stale locations here, those blocks can become under-replicated, but this path only deletes the RocksDB location and does not notify the replication manager. Worker-timeout cleanup calls `report_under_replicated_blocks` after deleting a worker's locations, so this new authoritative-report cleanup should enqueue the same affected block ids or otherwise trigger repair. Without that, a worker losing local storage can stop reads from being routed to missing blocks, but the lost replicas may never be rebuilt.

szbr00 (commented):
The commit message `fix(master): make full block reports authoritative` is missing the mandatory `(#issue-id)` suffix required by the project's commit convention. CI will fail without it. Please amend to include the issue or PR number, e.g. `fix(master): make full block reports authoritative (#973)`.

curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -26,18 +26,30 @@ use curvine_common::error::FsError;
...
 use std::sync::Arc;
+use std::sync::Mutex;
 
 #[derive(Clone)]
 pub struct MasterFilesystem {
     pub fs_dir: SyncFsDir,
     pub worker_manager: SyncWorkerManager,
     pub master_monitor: MasterMonitor,
     pub conf: Arc<MasterConf>,
+    full_block_reports: Arc<Mutex<HashMap<u32, FullBlockReportState>>>,
```
szbr00, line 42: This field uses `std::sync::Mutex`, and both call sites (lines 696 and 729) call `.lock().unwrap()`. The project conventions require `parking_lot::Mutex` over `std::sync` variants and prohibit `unwrap()` in production code. `parking_lot` is already a workspace dependency for `curvine-server`. Since `parking_lot::Mutex::lock()` returns the guard directly (no `Result`), switching to it eliminates both the `unwrap()` calls and the poisoning concern.

curvine-server/src/worker/handler/read_handler.rs
```diff
@@ -59,7 +59,10 @@ impl ReadHandler {
 
     pub fn open(&mut self, msg: &Message) -> FsResult<Message> {
         let mut context = ReadContext::from_req(msg)?;
-        let meta = self.store.get_block(context.block_id)?;
+        let meta = self
+            .store
+            .get_block(context.block_id)
+            .map_err(|_| FsError::block_not_found(context.block_id))?;
```
szbr00: The `map_err(|_| FsError::block_not_found(context.block_id))` discards the original error from `get_block`. Today `get_block_check` only returns a "not found" error, so this is functionally correct. However, the blanket mapping will misreport any future error type (e.g. storage I/O failure) as `BlockNotFound`. Consider checking the error kind or message before mapping, or at minimum preserving the original error as context via `FsError::with_source`.

curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -673,10 +687,53 @@ impl MasterFilesystem {
...
+        });
+        let report = reports
+            .entry(list.worker_id)
+            .or_insert_with(|| FullBlockReportState {
+                total_len: list.total_len,
+                update_time_ms: now,
+                reported_blocks: HashSet::with_capacity(list.total_len as usize),
+            });
+
+        if report.total_len != list.total_len {
```
szbr00, line 708: When `total_len` changes between chunks, the accumulated block IDs are silently cleared and the report restarts from scratch. This is correct behavior (the previous chunks were from a different report), but there is no log when it happens. If a worker's block count keeps fluctuating during a report (e.g. concurrent writes), the report may never complete and the reset would be invisible to operators. Consider adding a `warn!` here similar to the stale-block reconciliation log below.

curvine-server/src/master/fs/master_filesystem.rs
jlon: Fixed by returning the stale block ids from full block report reconciliation and enqueueing them through the existing replication manager path in `MasterHandler::block_report`. The block report RPC still succeeds if enqueue logs an error, matching worker-timeout cleanup semantics.

curvine-server/src/master/fs/master_filesystem.rs
jlon: Fixed by switching `full_block_reports` to `parking_lot::Mutex`. The `lock().unwrap()` calls are gone.

curvine-server/src/worker/handler/read_handler.rs
jlon: Fixed by preserving the original block-store error as context on the `BlockNotFound` response, so future non-not-found failures are no longer silently discarded.

curvine-server/src/master/fs/master_filesystem.rs
jlon: Fixed. A full report now logs a warning when `total_len` changes and the accumulated block ids are discarded.

jlon:
Fixed commit message: `fix(master): make full block reports authoritative (#973)`.

szbr00 (approved):

PR #974, opened 2026-07-03
fix(oss-hdfs): isolate Jindo metadata operation contexts

jlon:
## Problem
Concurrent OSS-HDFS metadata operations shared native Jindo filesystem state. Production master cores pointed at this path: unrelated async operations could reuse callback/context state and corrupt each other's result handling.

## Solution
Keep the Rust async ABI, but run metadata/open calls through a bounded, lazily created pool of Jindo store shards. Each operation gets an isolated handle context, reusable contexts are retained per shard, shutdown waits for in-flight operations without blocking forever, and callback userdata is reference-counted so failed starts do not leak.

## Test Plan
- `cargo fmt --check`
- `CARGO_TARGET_DIR=/tmp/curvine-pr-split-target cargo clippy -p curvine-common -p curvine-server -p curvine-ufs --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args`
- `CARGO_TARGET_DIR=/tmp/curvine-pr-split-target cargo test -p curvine-ufs --all-targets --jobs 2`

### DIFF
--- a/curvine-ufs/ffi/jindosdk_ffi.cpp
+++ b/curvine-ufs/ffi/jindosdk_ffi.cpp
@@ -21,6 +21,13 @@
 #include <cstring>
 #include <cstdlib>
 #include <cstdint>
+#include <chrono>
+#include <algorithm>
+#include <atomic>
+#include <cstdio>
+#include <memory>
+#include <mutex>
+#include <vector>
 
 // Include JindoSDK C API headers
 #include "jdo_api.h"
@@ -36,6 +43,13 @@
 // Thread-local error message storage
 thread_local std::string g_last_error;
 
+static constexpr size_t kDefaultFilesystemStoreShards = 16;
+static constexpr size_t kMaxFilesystemStoreShards = 64;
+static constexpr int64_t kFilesystemStoreShardWaitTimeoutMs = 3 * 1000;
+static constexpr int64_t kFilesystemCloseWaitTimeoutMs = 30 * 1000;
+
+struct JindoFileSystemWrapper;
+
 // =========================
 // Async helper definitions
 // =========================
@@ -50,9 +64,16 @@ static inline JindoStatus map_error_code_to_status(int32_t code) {
     return JINDO_STATUS_ERROR;
 }
 
+static inline bool is_reusable_operation_ctx(JindoStatus status) {
+    return status == JINDO_STATUS_OK || status == JINDO_STATUS_FILE_NOT_FOUND;
+}
+
 struct AsyncBase {
     JdoOperationCall_t call {nullptr};
     JdoOptions_t options {nullptr};
+    JdoHandleCtx_t handle_ctx {nullptr};
+    struct JindoStoreShard* store_shard {nullptr};
+    JindoFileSystemWrapper* fs_wrapper {nullptr};
     void* userdata {nullptr};
 };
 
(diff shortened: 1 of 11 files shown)


szbr00 (commented):
The commit message `fix(oss-hdfs): isolate Jindo metadata operation contexts` is missing the mandatory `(#issue-id)` suffix required by the project's commit convention. CI will fail without it. Please amend to include the issue or PR number, e.g. `fix(oss-hdfs): isolate Jindo metadata operation contexts (#974)`.

curvine-ufs/ffi/jindosdk_ffi.cpp
```diff
@@ -279,85 +580,62 @@ JindoStatus jindo_filesystem_init(
...
+    auto timeout = std::chrono::milliseconds(kFilesystemCloseWaitTimeoutMs);
+    for (auto& shard_owner : wrapper->store_shards) {
+        auto* shard = shard_owner.get();
+        if (!shard->mutex.try_lock_for(timeout)) {
+            // Keep wrapper/stores alive for in-flight native operations. This avoids use-after-free
+            // at shutdown without blocking Rust Drop indefinitely.
+            for (auto* locked : locked_shards) {
+                locked->mutex.unlock();
+            }
+            return;
```
szbr00, line 601: When a shard lock cannot be acquired within the 30s timeout, `jindo_filesystem_free` returns without deleting the wrapper. This is a reasonable trade-off to avoid use-after-free, but the failure is completely silent — operators have no way to detect that a filesystem wrapper and its stores were leaked. Consider writing a message to `g_last_error` or a log statement at this return path so the leak is diagnosable in production.

curvine-ufs/src/oss_hdfs/oss_hdfs_filesystem.rs
```diff
@@ -552,19 +599,41 @@ impl FileSystem<OssHdfsWriter, OssHdfsReader> for OssHdfsFileSystem {
...
+                                userdata,
+                            );
+                        }
+                    }
+                    status
+                })
+                .await?;
+            match start_status {
+                JindoStatus::Ok => {}
+                JindoStatus::FileNotFound => return Ok(false),
```
szbr00: The `JindoStatus::FileNotFound => return Ok(false)` arm in this `match` is unreachable. The C++ shim `jindo_filesystem_exists_async` always returns `JINDO_STATUS_OK` when the operation starts — FileNotFound is handled internally and reported as `(OK, false)` via the callback. This arm can be removed to avoid confusion, or a comment can note it is defensive.

curvine-ufs/ffi/jindosdk_ffi.cpp
```diff
@@ -36,22 +42,38 @@
...
     // See /usr/local/include/jindosdk/jdo_error.h for the full list.
-    if (code == 3001 /* JDO_FILE_NOT_FOUND_ERROR */) {
+    if (code == 3001 /* JDO_FILE_NOT_FOUND_ERROR */
+        || code == 30001 /* JDO_PARENT_NOT_FOUND_ERROR */) {
         return JINDO_STATUS_FILE_NOT_FOUND;
     }
     return JINDO_STATUS_ERROR;
 }
 
+static inline bool is_reusable_operation_ctx(int32_t code, JindoStatus status) {
```
szbr00: `is_reusable_operation_ctx(int32_t code, JindoStatus status)` casts `code` to `(void)` and only checks `status`. The test `filesystem_store_pool_reuses_not_found_contexts` passes different `code` values (0, 3001, 1) but they have no effect. If `code` is intentionally reserved for future use, a brief comment explaining that would help. Otherwise, consider removing the parameter to avoid suggesting it influences the decision.

curvine-ufs/src/oss_hdfs/oss_hdfs_filesystem.rs
jlon: Fixed by removing the unreachable `FileNotFound` start-status arm. Not-found from exists remains reported as `(OK, false)` by the C++ callback path.

curvine-ufs/ffi/jindosdk_ffi.cpp
jlon: Fixed. `jindo_filesystem_free` now sets `g_last_error` and writes a stderr diagnostic before intentionally leaving the wrapper allocated on close-timeout.

curvine-ufs/ffi/jindosdk_ffi.cpp
jlon: Fixed by removing the unused `code` parameter from `is_reusable_operation_ctx` and the test FFI. Reuse is now explicitly status-based.

jlon:
Fixed commit message: `fix(oss-hdfs): isolate Jindo metadata operation contexts (#974)`.

szbr00 (approved):

PR #985, opened 2026-07-05
fix(raft): tolerate invalid response context on apply

jlon:
## Summary

- avoid failing Raft apply when only the response context cannot be decoded
- only decode request context on the leader path where a client response is needed
- log and skip the response when the committed entry context is malformed

## Verification

- cargo fmt --check
- cargo clippy -p curvine-common -p curvine-server --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args
- cargo test -p curvine-server --all-targets --jobs 2
- production rollout smoke check: master/worker ready, restart=0, no panic/FATAL/Crash/backtrace in sampled logs

### DIFF
--- a/curvine-common/src/raft/raft_node.rs
+++ b/curvine-common/src/raft/raft_node.rs
@@ -634,26 +634,42 @@ where
             if entry.get_data().is_empty() {
                 continue;
             }
-            let req_id: i64 = SerdeUtils::deserialize(entry.get_context())?;
 
-            let rep_msg = if let EntryType::EntryConfChange = entry.get_entry_type() {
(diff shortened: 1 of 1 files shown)


szbr00 (approved):
LGTM. Solid correctness fix — moving `req_id` deserialization from an unconditional hard error (`?`) to a leader-only soft error (`warn!` + `continue`) prevents the raft node from crashing when replaying entries with invalid/empty context (e.g. followers replaying from snapshots). The `should_respond` caching and `to_vec()` gating are clean. Minor nit: the `let Some(entry_context) = entry_context else { continue }` is technically unreachable given the current logic, but acceptable as defensive coding.

PR #986, opened 2026-07-05
fix(client): keep local batch writer commit context stable

jlon:


szbr00 (approved):
LGTM. Solid correctness fix — the root cause was that `complete()` generated a fresh `req_id` via `Utils::req_id()` that didn't match the one used during `write_blocks_batch` in `new()`, so the worker couldn't correlate the batch setup with the commit. Reusing the same `BlockClient` connection and `req_id` from `new()` is the right approach. Removing the stored `Arc<FsContext>` is a nice cleanup since it was only needed during construction.

szbr00 (commented):
## Code Review

Reviewed 1 file, 12 lines changed (in `curvine-client/src/block/batch_block_writer_local.rs`).

### Summary

Clean, minimal, and correct fix. The original `BatchBlockWriterLocal` discarded the `BlockClient` and `req_id` after block setup in `new()`, then created a new client and generated a fresh `req_id` in `complete()`. Since the server correlates batch setup and commit by `req_id`, the mismatched ID meant the commit couldn't find the batch session.

The fix stores both `client` and `req_id` in the struct and reuses them in `complete()`. This also avoids a second connection — one `RpcClient` for setup + commit instead of two.

The approach is validated by `BatchBlockWriterRemote`, which already uses this exact pattern (stores `client: BlockClient` and `req_id: i64`, reuses both in `complete()`).

### Findings: 1

| # | Severity | Finding |
| - | -------- | ------- |
| 1 | Minor | No unit test for the `req_id` correlation fix — the core change has no regression test |

### Positive

- Brings `BatchBlockWriterLocal` in line with `BatchBlockWriterRemote` (pattern consistency)
- `BlockClient::drop()` is safe — `pool` is `None` when created via `block_client()`, so no pool leak
- `fs_context` field correctly removed — was only used in `complete()` to create a new client
- Public API unchanged (`new()`, `complete()`, `write()`, `flush()`, `worker_address()` signatures identical)
- `flush()` returns before `write_commit_batch` borrows `&self.client` — no borrow conflict

szbr00 (approved):

PR #987, opened 2026-07-05
fix(master): avoid panics on recoverable service errors

jlon:


curvine-server/src/master/master_server.rs
```diff
@@ -190,61 +196,77 @@ impl Master {
         Self::new(conf)
     }
 
-    pub async fn start(mut self) -> ServerStateListener {
+    pub async fn start(&mut self) -> CommonResult<ServerStateListener> {
         // step 1: Start journal_system, raft server and raft node will be started internally
-        let mut listener = self.journal_system.start().await.unwrap();
-        listener.wait_role().await.unwrap();
+        let journal_system = self
```
szbr00: The `.expect("... must be initialized before startup")` calls still panic. While this guards against a programming error (calling `start()` twice), it is inconsistent with the PR goal of "avoiding panics on recoverable service errors."

If the intent is that double-start is truly unrecoverable, `std::process::abort()` would be more appropriate (consistent with the pattern used in `journal_loader.rs` for fatal errors). If it is recoverable, returning `err_box!` would be more consistent with the rest of the PR.

curvine-common/src/state/mount.rs
```diff
@@ -116,7 +118,66 @@ pub struct MountInfo {
...
 impl MountInfo {
+    pub fn decode_persisted(bytes: &[u8]) -> CommonResult<Self> {
+        match bincode::deserialize::<Self>(bytes) {
+            Ok(info) => Ok(info),
+            Err(e) if Self::is_unexpected_eof(&e) => {
+                let legacy: LegacyMountInfo = bincode::DefaultOptions::new()
+                    .with_fixint_encoding()
+                    .reject_trailing_bytes()
+                    .deserialize(bytes)
+                    .map_err(|_| e)?;
```
szbr00: When both current and legacy deserialization fail, `.map_err(|_| e)?` discards the legacy decode error and returns only the original `UnexpectedEof`. This makes it hard for operators to diagnose corrupted mount data — they see a "missing fields" error that suggests a version mismatch, when the real cause might be data corruption.

Consider logging the legacy error before returning, or chaining both errors:
```rust
.map_err(|legacy_e| {
    warn!("legacy MountInfo decode also failed: {}", legacy_e);
    e
})?
```

curvine-server/src/master/meta/inode/ttl/ttl_bucket.rs
```diff
@@ -141,7 +145,15 @@ impl TtlBucketList {
             return;
         }
 
-        if let Some(expiration_ms) = inode.expiration_ms() {
+        let expiration_ms = match inode.expiration_ms() {
+            Ok(Some(expiration_ms)) => expiration_ms,
+            Ok(None) => return,
+            Err(e) => {
+                warn!("ttl_bucket: skip add for inode {}: {}", inode.id(), e);
```
szbr00, line 153: When `expiration_ms()` returns `Err`, the inode is silently skipped — it will never be tracked in TTL buckets and may never expire. The `warn!` log is the only signal.

If this happens for a file with a delete TTL action, it could leak storage indefinitely. Consider adding a fallback (e.g., a periodic reconciliation that retries skipped inodes) or at minimum tracking the count of skipped inodes in metrics.

szbr00 (commented):
## Code Review

Reviewed 34 files, 1314 lines changed (+867/-447).

### Summary

The PR systematically replaces `.unwrap()`, `panic!`, and `unreachable!()` with proper error propagation (`?`, `err_box!`) or controlled aborts (`std::process::abort()`) throughout the master server. The approach is sound: propagable errors use `err_box!`, truly fatal ones (inconsistent metadata, poisoned mutex) use `abort()`.

Key improvements:
- `Master::get_metrics()` now returns `CommonResult` instead of panicking — all 12 call sites updated correctly
- `MountInfo::decode_persisted` provides backward-compatible deserialization with legacy format fallback
- `MountTable` uses `read_inner()`/`write_inner()` helpers for lock poisoning resilience
- `SenderTask` handles `RecvTimeoutError::Disconnected` gracefully instead of erroring
- `Option`-wrapping in `Master` struct enables clean `take()` semantics in `start(&mut self)`
- Removed `Default` impls that internally used `.unwrap()`/`.expect()`

Verified: all callers of changed function signatures are updated in the diff.

### Findings: 9
- Critical: 0
- Major: 1
- Minor: 6
- Nit: 2

See inline comments for the top 3 findings. Remaining findings:
- **Minor** `mount_table.rs`: `restore_best_effort()` swallows errors with only `warn!` — master starts with incomplete mount table, clients get mount-not-found with no clear root cause
- **Minor** `journal_loader.rs`: `get_fsm_state()` trait method uses `abort()` on poisoned lock — acceptable since trait can't return `Result`, internal callers correctly use `fsm_state_snapshot()`
- **Minor** `journal_test.rs`: `raft_check_quorum = false` and timeout 60s->180s added — worth documenting rationale
- **Nit** `router_handler.rs`: `unwrap_or("N/A")` for missing metrics — acceptable for web UI
- **Nit** `sender_task.rs`: `panic!`->`log::error!`+`abort()` — well-justified for sender thread failure

curvine-common/src/state/mount.rs
jlon: Fixed in fd8b3f296f01ad5bd4e37955ab64bda5b07c228f. The legacy MountInfo fallback now logs the legacy decode error before returning the original current-format EOF error.

curvine-server/src/master/meta/inode/ttl/ttl_bucket.rs
jlon: Fixed in fd8b3f296f01ad5bd4e37955ab64bda5b07c228f. Added ttl_skipped_inodes counter so skipped TTL bucket indexing is visible in metrics in addition to the warning log. I kept reconciliation out of this PR to avoid expanding the behavior beyond the non-fatal panic cleanup.

curvine-server/src/master/master_server.rs
jlon: Fixed in fd8b3f296f01ad5bd4e37955ab64bda5b07c228f. The startup-only Option::take() paths now return err_box! instead of expect(), including block_on_start().

szbr00 (approved):

PR #988, opened 2026-07-05
fix(worker): return recoverable errors instead of panicking

jlon:


szbr00 (commented):
## Code Review

Reviewed 18 files, +477/-239 lines changed.

### Summary

Well-structured PR that replaces worker-side `unwrap()`/`panic`/`expect` paths with structured errors or log-and-skip across I/O, metadata, handlers, and background tasks. Fixes a latent `take()` bug in `complete_batch`, adds proper error cleanup paths, makes `abort`/`finalize` idempotent for retries, and removes several panic paths. The poisoned-lock `abort()` strategy is correct — in-memory state is untrustworthy after poison.

### Findings: 3

| # | Severity | File | Finding | Suggestion |
| - | -------- | ---- | ------- | ---------- |
| 1 | Minor | `block_store.rs`, `dir_state.rs`, `task_context.rs` | `CommonResult` return type with `abort()` is misleading — `write()`/`read()`/`lock_inner()`/`progress()` never return `Err`, they abort. Caller error branches (e.g. `heartbeat_task.rs` `if let Err(e) = store.write().map(...)`) are unreachable dead code. | Add `// SAFETY: never returns Err — aborts on poison` comment, or return the guard directly since the error is unrecoverable. |
| 2 | Minor | `batch_write_handler.rs` `write_batch()` | Behavioral change: batch writes are now atomic (fail-fast) instead of partial-success. Old code pushed `response.is_ok()` per file; new code returns `Err` on first failure. `results` is now always `vec![true; len]`, making the field redundant. | Atomic behavior is correct for consistency. Consider documenting that batch writes are atomic, or simplify the `results` field. |
| 3 | Minor | `worker_server.rs` | `.expect()` introduced on `Option<RpcServer>`/`Option<WebServer>` in `start()`, `block_on_start()`, `start_standalone()`, `service()`. These are invariant checks (programming errors), but the PR's stated goal is replacing panics. | Acceptable for programming errors. For consistency could use `ok_or_else(\|\| err_box!("..."))?`. Not blocking. |

### Positive

- **Fixes `take()` bug in `complete_batch`:** Old code called `self.context.take()` and `self.file.take()` inside the loop, moving the entire `Vec` on iteration 1. Iterations 2+ skipped context checks and file flushes. New code uses `as_ref()`/`as_mut()` and properly iterates all blocks.
- **Error cleanup in `open_batch`:** New `abort_open_contexts()` cleans up already-opened blocks when a subsequent block open fails. Old code leaked them.
- **`block_meta.rs` removes `unwrap()` panic:** Old `check_file` used `Lazy<Regex>` + `parse::<i64>().unwrap()` which panicked on non-numeric block IDs like `blk_abc`. New code uses `strip_prefix` + digit check + `parse().ok()`, correctly returning `None`.
- **`abort_block` idempotency:** Returns `Ok(())` for non-existent blocks (was `err_box!`). Enables safe cleanup paths in `write_handler.rs` and `batch_write_handler.rs`.
- **`finalize_block` idempotency for retries:** Already-finalized blocks return `Ok(meta)` if lengths match. Enables safe retry of `Complete` requests.
- **`worker_handler.rs` handler reuse fix:** Old code always created a new handler for `Complete`/`Cancel`, losing opened files/contexts. New code reuses the existing handler. Latent bug fix.
- **New `flush_batch` method:** Routes `WriteBlock` running requests within a batch to a dedicated flush handler with proper context validation.
- **Poisoned lock strategy is correct:** `abort()` for `BlockStore`/`BdevOffsetAllocator`/`TaskContext` is the right call. Log-and-skip for `HeartbeatTask::run()` is correct for non-fatal heartbeat failures.
- **`replication_manager.rs`:** Replaces `unwrap()` on semaphore acquisition with structured error. Replaces `try_option!` macro with explicit `let-else` and better error messages.

### No issues found in

- `BlockActor::Clone` derivation (fields are `Arc`-backed, cheap to clone)
- `Option<RpcServer>` take pattern in `start()` (prevents double-start, correct ownership)
- `complete_batch` commit order (flush all → close all → commit all; errors leave blocks in `Writing` state for retry)
- Block mismatch validation in `complete_batch` (new check ensures committed block matches opened block)
- Test updates in `dir_list.rs` and `vfs_dataset.rs` (properly propagate `?`)

szbr00 (commented):
## Code Review (Independent Verification)

Reviewed all 20 files, +770/-241 lines. Confirmed all findings from the existing review by @szbr9486 and independently verified the key correctness properties.

### Findings: 3 (all pre-existing)

| # | Severity | File | Finding | Suggestion |
| - | -------- | ---- | ------- | ---------- |
| 1 | Minor | `block_store.rs`, `dir_state.rs`, `task_context.rs` | `write()`/`read()`/`lock_inner()`/`progress()` return `CommonResult`/`MutexGuard` but `abort()` on poison — the `?` at call sites is unreachable dead code. | Add `// SAFETY: aborts on poison, never returns Err` or return the guard directly. |
| 2 | Minor | `batch_write_handler.rs` `write_batch()` | Now fail-fast/atomic. `results: vec![true; len]` is always all-true, making the field redundant. | Document atomicity or simplify in a follow-up. |
| 3 | Minor | `worker_server.rs` | `.expect()` on `Option<RpcServer>`/`Option<WebServer>` — inconsistent with the PR's goal of replacing panics. | Use `ok_or_else(..)? ` for consistency. Not blocking. |

### Verified Correctness Properties

- **Retry safety in `complete_batch`**: `is_commit` is set **after** the commit loop, so partial `finalize_block()` failures allow correct retry (re-executes full flush-close-commit pipeline). Improvement over old code.
- **`WriteHandler::write()` preserves file/context on error**: Uses `try_option_mut!` (mutable borrow, not take), so `write_batch`'s error recovery (take back, push remaining, return error) is sound.
- **Handler cleanup gated on success**: `worker_handler` only takes the handler when `res.is_ok_and(Success) && Complete/Cancel`. Failed Complete keeps handler alive for retry — correct.
- **`handler_matches_code` adds `WriteBlock -> BatchWriter`**: Necessary for routing `WriteBlock` flush requests through the batch handler's `flush_batch` path.
- **Short-circuit vs remote batch separation**: `file: Option<Vec<Option<BlockDevice>>>` correctly represents `None` entries for short-circuit blocks, validated by pre-checks in `write_batch()`.

### Positive

- **`take()` bug fix in `complete_batch`**: Old code consumed `context`/`file` on iteration 1, skipping all subsequent blocks. New code uses `as_ref()`/`as_mut()` — correctly iterates all blocks.
- **Error cleanup in `open_batch`**: `abort_open_contexts()` properly cleans up already-opened blocks on partial failure.
- **`block_meta.rs`**: Regex + `parse().unwrap()` replaced with `strip_prefix` + digit check — no longer panics on non-numeric block IDs.
- **`abort_block` idempotency**: Returns `Ok(())` for non-existent blocks, enabling safe cleanup paths.
- **`finalize_block` idempotency**: Already-finalized blocks return `Ok(meta)` if lengths match — enables safe retry of Complete requests.
- **Handler reuse in `worker_handler.rs`**: Complete/Cancel reuses existing handler instead of creating new (which lost opened files/contexts). Latent bug fix.
- **3 new integration tests** covering short-circuit complete, connection requirement, and full remote batch write/read cycle.
- **Test directory isolation**: Uses `Utils::req_id()` suffix to avoid conflicts under parallel test execution.

### No issues found in

- `BlockActor::Clone` derivation (fields are `Arc`-backed, cheap to clone)
- `worker_server.rs` `thread::spawn().join()` error mapping (correctly maps JoinError to CommonError)
- `mini_cluster.rs` abort-on-error strategy (appropriate for test infrastructure)
- Poisoned lock `abort()` strategy (correct — in-memory state untrustworthy after poison)
- `replication_manager.rs` semaphore and `let-else` replacements (structured errors, better messages)

curvine-server/src/worker/handler/worker_handler.rs
```diff
@@ -69,9 +73,13 @@ impl WorkerHandler {
...
 
-        let need_new_handler = self.handler.is_none()
-            || !matches!(msg.request_status(), RequestStatus::Running)
-            || !Self::handler_matches_code(&self.handler, code);
+        let need_new_handler = match msg.request_status() {
+            RequestStatus::Open => true,
+            RequestStatus::Running | RequestStatus::Complete | RequestStatus::Cancel => {
+                self.handler.is_none() || !Self::handler_matches_code(&self.handler, code)
+            }
+            _ => self.handler.is_none() || !Self::handler_matches_code(&self.handler, code),
```
szbr00, line 78: The `_` arm in `get_handler` produces the same result as the `Running | Complete | Cancel` arm above it. Consider collapsing both arms into a single `_` arm (or removing the `_` arm entirely if future request statuses need different logic):

```rust
let need_new_handler = match msg.request_status() {
    RequestStatus::Open => true,
    _ => self.handler.is_none() || !Self::handler_matches_code(&self.handler, code),
};
```

jlon: Fixed in 2c3751159b45143abb38354f4371ea150de1a4a9. Collapsed the duplicate match arms to Open => true and _ => reuse/create based on handler type.

szbr00 (approved):

PR #989, opened 2026-07-05
fix(server): propagate startup failures in mini cluster

jlon:
## Summary

- propagate mini-cluster master and worker startup failures from spawned threads
- abort test process on startup failure instead of silently losing the error in a detached thread

## Stack

Base PR branch: `jlon/fix-worker-recoverable-errors`.

## Verification

- GitHub CI: Compile and Smoke Test


jlon:
Closing this PR because this split was invalid. The original mini-cluster change cannot compile independently on main: both block_on_start() methods still return () there. I folded the master-side mini-cluster handling into #987 and the worker-side handling into #988.

PR #990, opened 2026-07-05
test(worker): cover non-fatal error handling paths

jlon:
## Summary

- add worker coverage for non-fatal error handling paths
- keep tests in the existing `curvine-server/tests/worker_test.rs` module

## Stack

Base PR branch: `jlon/fix-server-startup-error-propagation`.

## Verification

- GitHub CI: Compile and Smoke Test


jlon:
Closing this PR because the test-only split depends on the worker implementation change. The worker tests and worker-side mini-cluster handling have been folded into #988, so this PR should not stay open as an independent main-based PR.

PR #993, opened 2026-07-06
fix(client): handle channel close gracefully in FsWriterBuffer write_…

bigbigxu:
FsWriterBuffer::write_future` uses a `tokio::select!` loop to
receive either data chunks (`chunk_receiver`) or control tasks
(`task_receiver`) from two channels. Previously, both receivers
called `recv_check()`, which converts a `None` (sender dropped /
channel closed) into an `IOResult<T>` error with the message
"the sender dropped".

This is problematic because a closed channel is the **normal
shutdown path** — when `FsWriterBuffer` is dropped, its senders are
dropped and the channels close. The old code treated this as an
error, causing spurious error propagation and potentially confusing
error logs during clean shutdown.

This change switches both receivers to `recv()` (which returns
`Option<T>`) and explicitly handles `None` by returning `Ok(())`,
treating channel closure as a graceful termination.

### DIFF
--- a/curvine-client/src/file/fs_writer_buffer.rs
+++ b/curvine-client/src/file/fs_writer_buffer.rs
@@ -211,12 +211,18 @@ impl FsWriterBuffer {
             let select_task = tokio::select! {
                 biased;
 
-                chunk = chunk_receiver.recv_check() => {
-                   SelectTask::Data(chunk?)
+                chunk = chunk_receiver.recv() => {
+                   let Some(chunk) = chunk else {
+                        return Ok(())
+                    };
+                   SelectTask::Data(chunk)
                 }
 
-                task = task_receiver.recv_check() => {
-                    SelectTask::Control(task?)
+                task = task_receiver.recv() => {
+                    let Some(task) = task else {
+                        return Ok(())
+                    };
+                    SelectTask::Control(task)
                 }
             };
 


szbr00 (commented):
## Code Review

Reviewed 1 file, 14 lines changed (in `curvine-client/src/file/fs_writer_buffer.rs`).

### Summary

Clean, correct, minimal fix. The original `write_future` used `recv_check()`, which converts channel-close `None` into an error (`"the sender dropped"`). When `FsWriterBuffer` is dropped, both senders drop and both channels close — this is the **normal shutdown path**, not an error. The old code produced spurious error logs during clean shutdown.

The fix switches to `recv()` (returns `Option<T>`) and handles `None` with `return Ok(())`, treating channel closure as graceful termination. The `let-else` pattern is clear and idiomatic.

### Findings: 2 (both Nit)

| # | Severity | Finding |
| - | -------- | ------- |
| 1 | Nit | `try_recv()?` in control handlers (Flush/Complete/Seek/Resize) still propagates `TryRecvError::Closed` as error — pre-existing, inconsistent with this PR's graceful-close philosophy. Unlikely in practice since control tasks run while senders are alive. |
| 2 | Nit | No test verifying that dropping `FsWriterBuffer` doesn't produce spurious error logs. |

### Positive

- Correct fix — `recv_check()` (channel.rs:115-120) converts `None` to `err_box!("the sender dropped")`; channel closure during drop is normal shutdown, not an error
- `return Ok(())` inside `tokio::select!` correctly returns from `write_future`, terminating the background task cleanly
- No data loss — Tokio's `mpsc::recv()` drains remaining items before returning `None`
- `biased` priority is correct — when both channels close simultaneously, `chunk_receiver` is checked first and `Ok(())` is returned immediately
- `ErrorMonitor` interaction is correct — `Ok(())` means no error stored
- `Complete` handler already returns `Ok(())` (line 239), so channel-close handling doesn't interfere with the normal complete flow

szbr00 (approved):

PR #998, opened 2026-07-06
fix(master): decouple auto-cache planning from submit RPC (#984)

### LINKED ISSUE #984: [BUG]:  Auto-cache causes severe metadata performance degradation due to RPC channel contention
## Summary

When `auto_cache` is enabled on a mount point, metadata operations (`GetStatus`, `ListStatus`, etc.) experience severe performance degradation. The root cause is that a burst of `submit_load_job` RPCs contends with metadata RPCs on the shared connection pool (default `conn_size=3`), and master-side `submit_load_job` processing accesses UFS which introduces high latency, causing head-of-line blocking on the master's RPC processing pipeline and slowing down all co-located metadata requests.

### Call Path

When a cache miss occurs during `open()`, `async_cache` is triggered:

```rust
// unified_filesystem.rs L657-659
if mount.info.auto_cache() {
    self.async_cache(&ufs_path)?;
}
```

`async_cache` spawns an async task that submits a load job to the master:

```rust
// unified_filesystem.rs L346-382
pub fn async_cache(&self, source_path: &Path) -> FsResult<()> {
    let client = JobMasterClient::new(self.fs_client());
    // ...
    self.fs_context().rt().spawn(async move {
        let command = LoadJobCommand::builder(source_path.clone()).build();
        let res = client.submit_load_job(command).await;
        // ...
    });
    Ok(())
}
```

### Why It Blocks Metadata

1. **Shared connection pool**: The default `conn_size` is `3`. All RPCs — including `SubmitJob`, `GetStatus`, `ListStatus` — share the same connection pool to the master. Each connection has a `BufferClient` with its own `mpsc::channel`, and `request_future` writes envelopes sequentially:

   ```rust
   // buffer_client.rs L90-166
   loop {
       let env = receiver.recv().await;  // FIFO
       call_map.insert(req_id, env.cb);
       write_frame.send(&env.msg).await; // serial write per connection
   }
   ```

   Round-robin distributes requests across 3 connections, but a burst of `SubmitJob` RPCs can still saturate all connections.

2. **High-latency master-side processing (primary bottleneck)**: The master's `submit_load_job` handler accesses UFS (e.g., HDFS, S3) synchronously, which typically has high latency (tens to hundreds of milliseconds). This blocks the master's RPC processing threads, causing all co-queued metadata RPCs to wait for their turn.

3. **No deduplication**: If the same file is opened multiple times before caching completes, each open triggers a separate `submit_load_job` RPC, multiplying the pressure.

4. **No rate limiting**: There is no concurrency control on the spawned tasks. A burst of cache misses (e.g., first access to a large directory) generates an unbounded number of concurrent `SubmitJob` RPCs.

### Illustration

```
Master RPC processing pipeline:

Without auto_cache:
  [GetStatus 2ms] [ListStatus 3ms] [GetStatus 2ms]  → fast response

With auto_cache (burst of cache misses):
  [SubmitJob 150ms] [SubmitJob 200ms] [GetStatus 2ms] [SubmitJob 180ms] [ListStatus 3ms] ...
                                       ↑ queued behind                    ↑ queued behind
                                         slow SubmitJobs                    slow SubmitJobs

  Each SubmitJob accesses UFS (~50-200ms), blocking master RPC threads.
  Metadata RPCs queued behind SubmitJobs experience high latency.
```

### Solution 1: Client — Use a dedicated connection for `submit_load_job`

Create a separate `FsClient` (with its own connection pool) exclusively for `submit_load_job` RPCs, isolating cache-submission traffic from metadata traffic.

```
Before (shared pool, conn_size=3):
  conn1: [SubmitJob] [GetStatus] [SubmitJob] ...   ← contention
  conn2: [SubmitJob] [ListStatus] [SubmitJob] ...  ← contention
  conn3: [SubmitJob] [SubmitJob] [GetStatus] ...   ← contention

After (dedicated connection):
  conn1: [GetStatus] [ListStatus] [GetStatus] ...  ← metadata only, fast
  conn2: [GetStatus] [ListStatus] [GetStatus] ...  ← metadata only, fast
  conn3: [GetStatus] [ListStatus] [GetStatus] ...  ← metadata only, fast
  conn4: [SubmitJob] [SubmitJob] [SubmitJob] ...   ← cache jobs, isolated
```

```rust
pub struct UnifiedFileSystem {
    cv: CurvineFileSystem,
    // ... existing fields ...
    cache_client: Arc<FsClient>,  // dedicated client for submit_load_job
}

pub fn async_cache(&self, source_path: &Path) -> FsResult<()> {
    let client = JobMasterClient::new(self.cache_client.clone()); // use dedicated client
    let source_path = source_path.clone_uri();
    let log = self.audit_logging_enabled;
    let metrics = self.metrics;

    self.fs_context().rt().spawn(async move {
        // ... unchanged ...
    });

    Ok(())
}
```

**Pros**: Simple client-side change, complete traffic isolation, no server modification needed.
**Cons**: One extra TCP connection per client.

### Solution 2: Master — Enqueue `submit_load_job` without accessing UFS

Change the master's `submit_load_job` handler to write the job into an internal queue immediately and return to the client, deferring UFS access to background processing. This eliminates the high-latency UFS call from the RPC critical path.

```
Before:
  Client → Master.submit_load_job() → access UFS (50-200ms) → response
  (blocks master RPC thread for the entire UFS access duration)

After:
  Client → Master.submit_load_job() → enqueue job → response (~1ms)
  Background thread: dequeue → access UFS → schedule tasks
```

**Pros**: Fundamentally solves the problem, benefits all clients, reduces master RPC thread blocking.
**Cons**: Requires server-side refactoring, introduces async job scheduling complexity.

## Environment

- Component: `curvine-client`
- File: `src/unified/unified_filesystem.rs`
- Related: `orpc/src/client/buffer_client.rs`, `orpc/src/client/client_factory.rs`

## Impact

- **Severity**: High
- **Affected operations**: All metadata operations (`GetStatus`, `ListStatus`, `Exists`, etc.) when `auto_cache` is enabled on a mount point
- **Trigger condition**: Burst of cache misses (e.g., first access to files under a mounted path)

jlon:
# Summary

This PR fixes the auto-cache metadata latency regression reported in #984 by making master `SubmitJob` return after enqueueing the load job instead of synchronously planning it against UFS. UFS status/list planning now runs in bounded background execution, so slow S3/HDFS calls no longer occupy the master RPC critical path for metadata requests.

# Issue Describe / Design

Closes #984

## Root cause

When `auto_cache` is enabled, cache misses trigger `async_cache`, which submits `SubmitJob` RPCs to the master. Before this change, `JobManager::submit_load_job` synchronously called `LoadJobRunner::submit_load_task`, and that path performed UFS work such as `get_status` / recursive task planning before returning.

That made each `SubmitJob` RPC potentially wait on slow S3/HDFS I/O. Because clients share the master connection pool for `SubmitJob`, `GetStatus`, `ListStatus`, and other metadata RPCs, bursts of auto-cache submits caused head-of-line blocking and degraded unrelated metadata operations.

## Reproduction condition

1. Mount a remote UFS path with `auto_cache` enabled.
2. Trigger many first-time reads/cache misses under that mount.
3. Observe a burst of `SubmitJob` RPCs.
4. Metadata RPC latency rises because master-side submit handling waits on UFS planning.

## Fix approach

- Keep the `SubmitJob` RPC path lightweight: validate path/mount, install a `Pending` job in `JobStore`, return `LoadJobResult` immediately.
- Move UFS cache validation, recursive planning, and worker dispatch into `run_queued_load_job` on the master runtime.
- Add `job.master_max_concurrent_load_jobs` with default `16` to bound concurrent master-side UFS planning. Extra accepted jobs wait for permits; they are not dropped.
- Keep the internal `LoadJobRunner::submit_load_task` path synchronous for journal/UFS replay callers that rely on planning and dispatch being complete before returning.
- Add run-scoped task IDs so stale worker reports from an older run cannot update a replaced job. The run id uses timestamp high bits plus local sequence low bits to avoid restart-local id reuse.
- Make terminal job states stable: late worker reports are ignored after `Completed`, `Failed`, or `Canceled`.
- Close cancel/dispatch races by marking worker task contexts as canceled and preventing canceled contexts from being revived by late progress.

# Changes

| `curvine-server/src/master/job/job_manager.rs` | Enqueue load jobs in `submit_load_job`, spawn bounded background planning with a semaphore. | `SubmitJob` returns faster; accepted jobs continue asynchronously. |
| `curvine-common/src/conf/job_conf.rs` | Add `job.master_max_concurrent_load_jobs`, default `16`, reject `0`. | New tunable for master-side load planning concurrency. |
| `curvine-server/src/master/job/job_runner.rs` | Split enqueue/planning paths, preserve synchronous `submit_load_task`, add queued background execution, run-scoped task IDs, and best-effort cancel after races/failures. | Removes UFS I/O from RPC critical path while preserving internal synchronous callers. |
| `curvine-server/src/master/job/job_context.rs` | Track `run_id` and allow installing pre-planned task details. | Supports stale report isolation and async planning. |
| `curvine-server/src/master/job/job_store.rs` | Add run-guarded state updates and ignore progress reports for terminal jobs. | Prevents stale background work or late worker reports from corrupting terminal/replaced jobs. |
| `curvine-server/src/worker/task/*` | Ensure cancel marks task contexts and canceled tasks cannot be completed by late progress. | Makes cancel effective for running load tasks. |
| `curvine-server/tests/load_job_submit_test.rs` | Add regression coverage for async submit return, background planning failure, zero-task completion, stale task reports, terminal late reports, and worker cancel stability. | Covers the key behavior and race protections for this fix. |

# Test verified

| `cargo fmt --check` | PASS | Formatting check. |
| `cargo test -p curvine-server --test load_job_submit_test -- --nocapture` | PASS | 6 tests passed. |
| `CARGO_TARGET_DIR=/tmp/curvine-998-target cargo test -p curvine-server --test load_job_submit_test -- --nocapture` | PASS | 6 tests passed with an isolated target dir. |
| `CARGO_TARGET_DIR=/tmp/curvine-998-target cargo clippy -p curvine-common -p curvine-server --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | No warnings. |
| `git diff --check` | PASS | No whitespace errors. |
| Three independent sub-agent reviews | PASS | Issue-fit, state-machine/race, and performance/config reviews all returned Ready after fixes. |

# Dependencies

- Related PRs: None
- Related issues: #984
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/job/job_runner.rs
+++ b/curvine-server/src/master/job/job_runner.rs
@@ -21,7 +21,8 @@ use curvine_common::conf::ClientConf;
 use curvine_common::error::FsError;
 use curvine_common::fs::{FileSystem, Path};
 use curvine_common::state::{
-    JobTaskState, LoadJobCommand, LoadJobResult, LoadTaskInfo, MountInfo, WorkerAddress,
+    JobTaskState, LoadJobCommand, LoadJobInfo, LoadJobResult, LoadTaskInfo, MountInfo,
+    WorkerAddress,
 };
 use curvine_common::utils::CommonUtils;
 use curvine_common::FsResult;
@@ -30,6 +31,7 @@ use futures::future;
 use log::{debug, error, info, warn};
 use orpc::common::{ByteUnit, FastHashMap, FastHashSet, LocalTime};
 use orpc::err_box;
+use orpc::sync::AtomicCounter;
 use std::collections::LinkedList;
 use std::sync::Arc;
 
@@ -38,20 +40,36 @@ pub struct LoadJobRunner {
     master_fs: MasterFilesystem,
     factory: Arc<UfsFactory>,
     job_max_files: usize,
+    run_seq: Arc<AtomicCounter>,
+}
+
+#[derive(Clone)]
+pub(crate) struct QueuedLoadJob {
+    pub(crate) job_id: String,
+    pub(crate) run_id: u64,
+}
+
+struct PlannedLoadJob {
+    tasks: FastHashMap<String, TaskDetail>,
+    total_size: i64,
 }
 
 impl LoadJobRunner {
+    const RUN_ID_SEQ_MOD: u64 = 1_000_000;
+
     pub fn new(
         jobs: JobStore,
         master_fs: MasterFilesystem,
         factory: Arc<UfsFactory>,
         job_max_files: usize,
+        run_seq: Arc<AtomicCounter>,
     ) -> Self {
         Self {
             jobs,
             master_fs,
             factory,
             job_max_files,
+            run_seq,
         }
     }
 
@@ -66,20 +84,51 @@ impl LoadJobRunner {
         }
     }
 
-    async fn check_job_exists(
+    fn build_job_context(
         &self,
-        job: &JobContext,
+        command: &LoadJobCommand,
+        mnt: &MountInfo,
+    ) -> FsResult<(JobContext, Path, Path)> {
+        let source_path = Path::from_str(&command.source_path)?;
+        let target_path = mnt.toggle_path(&source_path)?;
+        let job_id = CommonUtils::create_job_id(source_path.full_path());
+        let run_id = self.next_run_id();
+        let job_context = JobContext::with_conf(
+            command,
+            job_id,
+            source_path.clone_uri(),
+            target_path.clone_uri(),
+            mnt,
+            &ClientConf::default(),
+            run_id,
+        );
+
+        Ok((job_context, source_path, target_path))
+    }
+
+    fn next_run_id(&self) -> u64 {
+        LocalTime::mills()
+            .saturating_mul(Self::RUN_ID_SEQ_MOD)
+            .saturating_add(self.run_seq.next() % Self::RUN_ID_SEQ_MOD)
+    }
+
+    fn running_job_result(&self, job_id: &str) -> Option<LoadJobResult> {
+        self.jobs.get(job_id).and_then(|exist_job| {
+            let state: JobTaskState = exist_job.state.state();
+            if state.is_running() {
+                Some(LoadJobResult::with_state(&exist_job.info, state))
+            } else {
+                None
+            }
+        })
+    }
+
+    async fn check_already_loaded(
+        &self,
         mnt: &MountValue,
         source_path: &Path,
         target_path: &Path,
-    ) -> FsResult<Option<LoadJobResult>> {
-        if let Some(exist_job) = self.jobs.get(&job.info.job_id) {
-            let state: JobTaskState = exist_job.state.state();
-            if state.is_running() {
-                return Ok(Some(LoadJobResult::with_state(&exist_job.info, state)));
-            }
-        }
-
+    ) -> FsResult<bool> {
         // Data-state fast-path. Applies whether the slot was vacant or held
         // a terminal ctx — e.g. after JobCleanupTask, after master restart +
         // journal replay, or after an earlier run completed the sync.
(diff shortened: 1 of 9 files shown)


curvine-server/src/master/job/job_runner.rs
```diff
@@ -220,6 +313,162 @@ impl LoadJobRunner {
...
+                JobTaskState::Failed,
+                format!("dispatch failed: {}", err),
+            );
+            self.cancel_workers_best_effort(&queued.job_id, &assigned_workers)
+                .await;
+            return Err(err);
+        }
+
+        if !self.is_current_running(&queued) {
+            self.cancel_workers_best_effort(&queued.job_id, &assigned_workers)
```
szbr00: The post-dispatch `cancel_workers_best_effort` can accidentally cancel tasks belonging to a newer run. If the job is canceled and resubmitted while this run is still inside `submit_all_task`, the new run may dispatch tasks to overlapping workers before this cancel fires. Since the worker-side `cancel_job` cancels by `job_id` (not `run_id`), this call would remove the new run's tasks on those workers.

`JobManager::cancel_job` already sends cancel requests to all assigned workers when the job is canceled, making this post-dispatch cancel redundant for the cancel case. Consider removing it, or gating it on a run_id-aware cancel path.

curvine-server/src/master/job/job_runner.rs
```diff
@@ -220,6 +313,162 @@ impl LoadJobRunner {
...
+                    &queued.job_id,
+                    queued.run_id,
+                    JobTaskState::Failed,
+                    format!("plan load job failed: {}", err),
+                );
+                return Err(err);
+            }
+        };
+        let total_size = planned.total_size;
+        let task_count = planned.tasks.len();
```
szbr00, line 402: If `create_all_tasks` returns zero tasks (e.g., source is an empty directory), the job is installed with no tasks and never transitions out of Pending/Loading — no progress reports will ever arrive to trigger completion. Consider checking `task_count == 0` and marking the job as Completed before dispatch.

curvine-server/src/master/job/job_runner.rs
jlon: Fixed in 29ad7957: removed the post-dispatch cancel path for the canceled/replaced-run case, and kept best-effort worker cancel only when dispatch failure is still confirmed to belong to the same run.

curvine-server/src/master/job/job_runner.rs
jlon: Fixed in 29ad7957: zero-task load jobs now transition to Completed immediately with a diagnostic message before dispatch.

szbr00 (approved):

PR #999, opened 2026-07-06
fix(spdk): handle ub underflow and double decrement

thuongle2210:
# Description
- Fix two bugs in poller_callback: UB underflow when pending is empty, and double-decrement of bdev_inflight on duplicate completion calls.

# Key Changes
- `orpc/src/io/spdk_poller.rs`: Early-return guard when pending.is_empty() to prevent underflow; guard fetch_sub behind complete()'s first-call-wins return value to prevent double-decrement.
- Add tests to show the correctness of 2 above cases.

# Impact
- Eliminates UB (wrap to usize::MAX) when a late callback arrives on an already-cleaned-up qpair.
- Prevents bdev_inflight underflow that could stall future I/O or cause incorrect reference counting.

# Future works
1. **UAF prevention** — Pre-push + post-check in `submit_one` to prevent use-after-free
2. **Orphan lifecycle** — Add orphan lifecycle and reclaim stale
3. **Late callback safety** — Orphaned lifecycle with poll and sweep for late SPDK callback safety

### DIFF
--- a/orpc/src/io/spdk_poller.rs
+++ b/orpc/src/io/spdk_poller.rs
@@ -640,6 +640,12 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
     let ctx = Box::from_raw(cb_arg as *mut CallbackCtx);
 
     let qs = &mut *(ctx.qpair_state as *mut QpairState);
+
+    // Defensive guard against UB underflow on empty pending.
+    if qs.pending.is_empty() {
+        return;
+    }
+
     let idx = ctx.pending_idx;
     let last = qs.pending.len() - 1;
     if idx != last {
@@ -650,8 +656,10 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
         qs.pending.pop();
     }
 
-    ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
-    ctx.completion.complete(status);
+    // Guard against double-decrement on repeated complete()
+    if ctx.completion.complete(status) {
+        ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
+    }
 }
 
 impl Drop for SpdkPoller {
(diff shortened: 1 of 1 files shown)


orpc/src/io/spdk_poller.rs
```diff
@@ -640,6 +640,12 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
     let ctx = Box::from_raw(cb_arg as *mut CallbackCtx);
 
     let qs = &mut *(ctx.qpair_state as *mut QpairState);
+
+    // Defensive guard against UB underflow on empty pending.
+    if qs.pending.is_empty() {
```
szbr00, line 646: This guard runs after `Box::from_raw(cb_arg)` and after dereferencing `ctx.qpair_state`, so it only helps the narrow case where the QpairState is still alive but its `pending` Vec happens to be empty. It prevents the `qs.pending.len() - 1` underflow, which is a real win. However, in the "late callback on an already-cleaned-up qpair" scenario described in the PR, `force_complete_qpair` has already reclaimed this ctx via `Box::from_raw` and removed the QpairState from `dead_qpairs` (dropping the Box). In that case `Box::from_raw(cb_arg)` is a double-free and `&mut *(ctx.qpair_state ...)` dereferences freed memory before this guard is ever reached. Since the PR lists UAF/orphan-lifecycle handling as future work, consider a brief inline comment clarifying that this guard is a partial defense against the underflow only, so it isn't later mistaken for full late-callback safety.

orpc/src/io/spdk_poller.rs
```diff
@@ -650,8 +656,10 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
         qs.pending.pop();
     }
 
-    ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
-    ctx.completion.complete(status);
+    // Guard against double-decrement on repeated complete()
+    if ctx.completion.complete(status) {
```
szbr00, line 661: Reordering `complete()` before `fetch_sub` means a waiter woken by `complete()` can observe `bdev_inflight` before it is decremented, whereas previously the decrement was visible first. In practice the drain path (`SpdkBdev::drop`) polls `inflight` with a 1ms sleep rather than reacting to the completion directly, so this is low risk, and it now matches `force_complete_qpair`'s ordering. Flagging it only so the intentional reorder is on record; no change required unless any consumer relies on inflight being settled the instant `wait()` returns.

orpc/src/io/spdk_poller.rs
```diff
@@ -756,4 +764,37 @@ mod test {
...
+        }));
+
+        unsafe { poller_callback(ctx as *mut c_void, 0) };
+
+        assert!(qs.pending.is_empty());
+        assert_eq!(inflight.load(Ordering::Acquire), 1);
+        assert_eq!(completion.wait(1), -libc::ETIMEDOUT);
+    }
+
+    #[test]
```
szbr00, line 792: The name promises "does not decrement inflight," but this test only checks `IoCompletion::complete`'s return value and status idempotence — there is no inflight counter or `poller_callback` involved. Consider either renaming to reflect what it verifies (e.g. `complete_is_idempotent_first_call_wins`), or adding a test that drives `poller_callback` itself to confirm the `fetch_sub` guard runs exactly once, since that production path isn't directly covered by either new test.

thuongle2210: I added and renamed tests to show the correctness. Thanks.

orpc/src/io/spdk_poller.rs
thuongle2210: Acknowledged. The re-order is intentional for the double-decrement guard, and the drain path polls with sleep so there's no practical race.

orpc/src/io/spdk_poller.rs
thuongle2210: I added more comments for explanation and future updates.

szbr00 (approved):

PR #1000, opened 2026-07-06
fix: use pre-push + post-check in submit_one to prevent UAF

thuongle2210:
# Description
- Fix a use-after-free in submit_one when SPDK completes I/O synchronously during the submit call.
- Based on the change of [pr-999](https://github.com/CurvineIO/curvine/pull/999)
# Key Changes
- `orpc/src/io/spdk_poller.rs`: Pre-push cb_ctx_ptr to qs.pending before the SPDK submit call, so poller_callback always finds a valid entry. On rc != 0, use position() to check whether the callback already removed the entry; if found, clean up with proper swap_remove + reindex + free + complete. If not found (callback already fired), skip cleanup entirely. 

# Impact
- Eliminates UAF when SPDK fires poller_callback synchronously during submit_one.
- Ensures pending vec and pending_idx invariants are always correct regardless of callback timing

# Future works
2. **Orphan lifecycle** — Add orphan lifecycle and reclaim stale
3. **Late callback safety** — Orphaned lifecycle with poll and sweep for late SPDK callback safety

### DIFF
--- a/orpc/src/io/spdk_poller.rs
+++ b/orpc/src/io/spdk_poller.rs
@@ -640,6 +640,12 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
     let ctx = Box::from_raw(cb_arg as *mut CallbackCtx);
 
     let qs = &mut *(ctx.qpair_state as *mut QpairState);
+
+    // Defensive guard against UB underflow on empty pending.
+    if qs.pending.is_empty() {
+        return;
+    }
+
     let idx = ctx.pending_idx;
     let last = qs.pending.len() - 1;
     if idx != last {
@@ -650,8 +656,10 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
         qs.pending.pop();
     }
 
-    ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
-    ctx.completion.complete(status);
+    // Guard against double-decrement on repeated complete()
+    if ctx.completion.complete(status) {
+        ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
+    }
 }
 
 impl Drop for SpdkPoller {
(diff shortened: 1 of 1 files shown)


orpc/src/io/spdk_poller.rs
```diff
@@ -526,6 +526,9 @@ impl SpdkPoller {
             );
         }
 
+        // Pre-push: poller_callback needs the entry in pending if SPDK completes synchronously during the submit call.
+        qs.pending.push(cb_ctx_ptr);
```
szbr00: The pre-push is the right fix, but it widens an aliasing-soundness gap. `qs` here is a `&mut Box<QpairState>`, and it is used both before (`qs.pending.push`) and after (`qs.pending.iter().position`) the SPDK submit call. When SPDK completes synchronously, it invokes `poller_callback`, which reconstructs `&mut *(ctx.qpair_state)` pointing at this same QpairState. Holding a `&mut` across a call that creates a second aliasing `&mut` to the same object is UB under Rust's aliasing model (stacked/tree borrows), even though it works in practice today. Consider accessing `pending` exclusively through the raw pointer (`(*qs_ptr).pending...`) across the submit boundary and not retaining the `&mut`, so the reentrant mutation has a single, well-defined provenance.

orpc/src/io/spdk_poller.rs
```diff
@@ -568,15 +571,19 @@ impl SpdkPoller {
...
-            // Submission failed - reclaim allocation and complete with error
-            unsafe { drop(Box::from_raw(cb_ctx_ptr)) };
-            req.bdev_inflight.fetch_sub(1, Ordering::Release);
-            req.completion.complete(rc);
+            // If callback fired during submit and removed this entry from
+            // pending via swap_remove, position() returns None - skip.
+            if let Some(pos) = qs.pending.iter().position(|&p| p == cb_ctx_ptr) {
+                qs.pending.swap_remove(pos);
+                if pos < qs.pending.len() {
+                    unsafe { (*qs.pending[pos]).pending_idx = pos };
```
szbr00: This reindex branch — moving the last entry into `pos` and fixing its `pending_idx` — only runs when the failed entry was not the last one in `pending`. The new test `pre_push_entry_removed_by_callback_skips_cleanup` exercises only the single-entry (pos == last) path, so this reindex logic is untested. Since a stale `pending_idx` here would later drive a wrong `swap_remove` in `poller_callback`, please add a test with two or more pre-pushed entries where a non-last submission fails, asserting the moved entry's `pending_idx` is corrected to `pos`.

orpc/src/io/spdk_poller.rs
```diff
@@ -568,15 +571,19 @@ impl SpdkPoller {
...
+                }
+                unsafe { drop(Box::from_raw(cb_ctx_ptr)) };
+                req.bdev_inflight.fetch_sub(1, Ordering::Release);
```
szbr00: This cleanup path calls `fetch_sub` then `complete(rc)` unconditionally, whereas `poller_callback` now guards `fetch_sub` behind `complete()`'s return value. It's currently safe because reaching this branch means `position()` found the entry, so the callback has not completed the I/O — but that invariant is implicit. Consider mirroring the guarded pattern (`if req.completion.complete(rc) { fetch_sub }`) or adding a one-line comment stating the invariant, so the two decrement sites stay consistent if callback-timing assumptions ever change.

orpc/src/io/spdk_poller.rs
thuongle2210: I understand your concern; it will be covered by a test in an upcoming PR. I prepared it.

orpc/src/io/spdk_poller.rs
thuongle2210: Thanks. I updated the usage.

orpc/src/io/spdk_poller.rs
thuongle2210: I updated it to mirror the poller_callback's pattern

thuongle2210:
Dear @szbr9486 , I've resolved the conflicts with the main branch.

szbr00 (approved):

PR #1001, opened 2026-07-06
feat(spdk): add orphan lifecycle and reclaim stale

thuongle2210:
# Description
- Add orphan lifecycle for late SPDK callbacks. When `force_complete_qpair` signals all outstanding I/Os as failed, the `CallbackCtx` entries are kept alive (moved to `stale`) instead of freed, so the late SPDK callback can safely find its context without a use-after-free. `reclaim_stale()` drains and frees entries at shutdown or qpair unregister.
- Based on the change of [pr-1000](https://github.com/CurvineIO/curvine/pull/1000)
# Key Changes
- `orpc/src/io/spdk_poller.rs`: `QpairState` gains `stale: Vec<*mut CallbackCtx>` field and `reclaim_stale()` method that drains and frees stale entries.
- `force_complete_qpair`: Signals completion in-place without taking ownership. Entries moved from `pending` to `stale` (no longer freed immediately). The `QpairState` itself stays in `dead_qpairs` to keep ctx pointers valid for the late SPDK callback.
- `poller_callback`: Restructured with `complete()` as the outer discriminator. Hot path (`complete()` returns true): removes from pending via swap_remove/pop, signals, decrements inflight, then frees the `CallbackCtx` via `drop(Box::from_raw(...))`. Late path (already signaled by force_complete): entire body skipped - entry lives on in `stale` and is freed by `reclaim_stale()`.
- Shutdown + unregister paths: Call `reclaim_stale()` before `dead_qpairs.clear()` or `.remove()`.
- Minor cleanup: Shorten `std::sync::Arc<AtomicUsize>` to `Arc<AtomicUsize>` in `IoRequest` and `CallbackCtx`.
- `submit_one`: Accesses `pending` through raw pointer (`(*qs_ptr).pending`) instead of `&mut` to avoid aliasing UB with poller_callback reentry.
- Add correctness tests:
    + `force_complete_into_stale_reclaimable`:  Verifies `force_complete_qpair` moves entries to `stale`, keeps `QpairState` alive, and `reclaim_stale` frees them correctly.
    + update `poller_callback_empty_pending_returns_early` to `poller_callback_empty_pending_skips_swap_remove_only`: Underflow guard skips swap_remove when pending is empty, but fetch_sub and free still run on the hot path.
    + `poller_callback_late_path_does_not_double_signal`: Late callback after `force_complete` does not re-signal or double-decrement inflight.
    + Updated `force_complete_reclaims_callback_ctx` test: Now verifies stale lifecycle instead of immediate cleanup.
- update from mutable to immutable variables to avoid warning:
    +  pre_push_entry_removed_by_callback_skips_cleanup
    + poller_callback_empty_pending_skips_swap_remove_only
    + validate_rejects_low_keep_alive
    + validate_accepts_high_keep_alive
    + validate_rejects_inherited_low_keep_alive
# Impact
- Prevents use-after-free when SPDK fires a completion callback after the qpair has already been force-completed.
- Enables safe deferred cleanup of orphaned callbacks without leaking.

# Future Work

### DIFF
--- a/orpc/src/io/spdk_poller.rs
+++ b/orpc/src/io/spdk_poller.rs
@@ -640,6 +640,12 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
     let ctx = Box::from_raw(cb_arg as *mut CallbackCtx);
 
     let qs = &mut *(ctx.qpair_state as *mut QpairState);
+
+    // Defensive guard against UB underflow on empty pending.
+    if qs.pending.is_empty() {
+        return;
+    }
+
     let idx = ctx.pending_idx;
     let last = qs.pending.len() - 1;
     if idx != last {
@@ -650,8 +656,10 @@ unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
         qs.pending.pop();
     }
 
-    ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
-    ctx.completion.complete(status);
+    // Guard against double-decrement on repeated complete()
+    if ctx.completion.complete(status) {
+        ctx.bdev_inflight.fetch_sub(1, Ordering::Release);
+    }
 }
 
 impl Drop for SpdkPoller {
@@ -755,5 +763,38 @@ mod test {
                 unsafe { drop(Box::from_raw(cb_ptr as *mut CallbackCtx)) };
             }
         }
+    }
+
+    #[test]
+    fn poller_callback_empty_pending_returns_early() {
+        let inflight = Arc::new(AtomicUsize::new(1));
+        let completion = IoCompletion::new();
+        let qs = Box::new(QpairState {
+            dead: Arc::new(AtomicBool::new(false)),
+            pending: Vec::new(),
+        });
+        let ctx = Box::into_raw(Box::new(CallbackCtx {
+            completion: completion.clone(),
+            async_ctx: unsafe { std::mem::zeroed() },
+            bdev_inflight: inflight.clone(),
+            qpair_state: &*qs as *const QpairState as *mut QpairState,
+            pending_idx: 0,
+        }));
+
+        unsafe { poller_callback(ctx as *mut c_void, 0) };
+
+        assert!(qs.pending.is_empty());
+        assert_eq!(inflight.load(Ordering::Acquire), 1);
+        assert_eq!(completion.wait(1), -libc::ETIMEDOUT);
+    }
+
+    #[test]
+    fn complete_second_call_does_not_decrement_inflight() {
+        let completion = IoCompletion::new();
+        assert!(completion.complete(42));
+        assert_eq!(completion.wait(0), 42);
+
+        assert!(!completion.complete(99));
+        assert_eq!(completion.wait(0), 42);
     }
 }
--- a/orpc/src/io/spdk_poller.rs
+++ b/orpc/src/io/spdk_poller.rs
@@ -679,26 +679,25 @@ struct CallbackCtx {
 unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
     let ctx = &*(cb_arg as *mut CallbackCtx);
 
-    let qs = &mut *(ctx.qpair_state as *mut QpairState);
-
-    // Defensive guard against UB underflow on empty pending.
-    if qs.pending.is_empty() {
-        return;
-    }
-
-    let idx = ctx.pending_idx;
-    let last = qs.pending.len() - 1;
-    if idx != last {
-        let last_ptr = qs.pending[last];
-        qs.pending.swap_remove(idx);
-        (*last_ptr).pending_idx = idx;
-    } else {
-        qs.pending.pop();
-    }
-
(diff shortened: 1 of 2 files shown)


orpc/src/io/spdk_poller.rs
```diff
@@ -635,11 +670,22 @@ struct CallbackCtx {
...
-/// C callback invoked by SPDK when NVMe command completes.
+/// SPDK NVMe completion callback.
+///
+/// **Hot path** — first completion, qpair_state alive. Remove from pending, decrement inflight,
+///   then free the CallbackCtx.
+/// **Late path** — completion already signaled by force_complete, qpair_state may be freed.
+///   Don't free — entry is in QpairState::stale and will be freed by reclaim_stale.
 unsafe extern "C" fn poller_callback(cb_arg: *mut c_void, status: i32) {
-    let ctx = Box::from_raw(cb_arg as *mut CallbackCtx);
+    let ctx = &*(cb_arg as *mut CallbackCtx);
```
szbr00, line 680: Switching from `Box::from_raw` to `&*` here removes the only place the hot path reclaimed the CallbackCtx, and no compensating free was added — so every normally-completed I/O now leaks its CallbackCtx along with the `Arc<IoCompletion>` and `Arc<AtomicUsize>` it holds. On the hot path the entry is `swap_remove`/`pop`-ed out of `pending` and never inserted into `stale`, so `reclaim_stale()` will never collect it either. The doc comment directly above even specifies the hot path should "remove from pending, decrement inflight, then free the CallbackCtx" — but the code never frees. Since the empty-pending guard already diverts the late path with an early return, the hot path (past the guard) owns the ctx and can safely reclaim it — e.g. `drop(Box::from_raw(cb_arg))` after `complete()`/`fetch_sub` — while the late path stays free-less and is reclaimed via `reclaim_stale()`.

orpc/src/io/spdk_poller.rs
```diff
@@ -466,6 +471,9 @@ impl SpdkPoller {
         if let IoOp::UnregisterQpair { qpair, ack } = &req.op {
             active_qpairs.retain(|&qp| qp != *qpair);
             let key = *qpair as usize;
+            if let Some(qs) = dead_qpairs.get_mut(&key) {
+                qs.reclaim_stale();
```
szbr00, line 475: `reclaim_stale()` frees the stale CallbackCtx entries and then the QpairState is dropped via `dead_qpairs.remove(&key)` (same pattern in the shutdown path with `dead_qpairs.clear()`). This is only safe if SPDK is guaranteed not to deliver any further completion for this qpair afterward; otherwise a late callback dereferences both a freed CallbackCtx (`&*ctx`) and a freed `qpair_state`, reintroducing the exact UAF this PR prevents. Does the UnregisterQpair path disconnect/drain the qpair at the SPDK level before this runs? If not, reclaiming here may be premature — this seems to depend on the "Deferred qpair cleanup" and "Unregister ctrlr + detach target" future-work items. Consider gating reclaim on a confirmed qpair disconnect so no callback can be in flight.

orpc/src/io/spdk_poller.rs
```diff
@@ -756,4 +814,268 @@ mod test {
...
+            async_ctx: unsafe { std::mem::zeroed() },
+            bdev_inflight: inflight.clone(),
+            qpair_state: &mut *qs as *mut QpairState,
+            pending_idx: 0,
+        });
+        let cb_ctx_ptr = Box::into_raw(cb_ctx);
+        qs.pending.push(cb_ctx_ptr);
+
+        // Simulate callback firing during submit via the real callback.
+        unsafe { poller_callback(cb_ctx_ptr as *mut c_void, 0) };
```
szbr00, line 972: This call drives the hot path (status 0, entry present in pending), which after the borrow change no longer frees `cb_ctx_ptr` — and the test doesn't free it either, so it leaks silently. That's why the CallbackCtx leak isn't caught by CI. Consider adding a leak-detecting assertion for the hot path (e.g. a Drop-counting payload reachable from CallbackCtx, or running these under Miri/leak-sanitizer) so the "hot path frees exactly once" invariant is actually verified rather than assumed.

orpc/src/io/spdk_poller.rs
thuongle2210: This PR introduces the stale + reclaim_stale approach and fixes the double-free. I'm aware of the leak concern, so I updated the poller_callback's approach.

orpc/src/io/spdk_poller.rs
thuongle2210: I understand your concern. The next PR moves QpairState to an orphaned map instead of reclaiming here, reclaim_stale is called later when SPDK is guaranteed done. SPDK-level detach comes in a later PR as well.

orpc/src/io/spdk_poller.rs
thuongle2210: I updated the poller callback in this PR (previously I had planned it for the next PR). For your idea about leak-detecting, I'll consider making it a new PR - useful instrumentation to have. Thanks.

thuongle2210:
Hi @szbr9486 , I resolved conflicts with main branch.

szbr00 (approved):

PR #1002, opened 2026-07-07
fix(fuse): prevent pipe poisoning from short splice transfers

### LINKED ISSUE #965: [BUG]: FuseSender pipe permanently poisoned by a single short splice/vmsplice transfer (mount hangs in D state)
## Summary

A `FuseSender` can be **permanently poisoned** by a single short (partial) `splice`/`vmsplice` transfer. Once poisoned, that sender's long-lived, reused pipe retains a partial response forever: every subsequent response either fails to be delivered or is delivered corrupted, and the requests it owns are never answered. On the kernel side the fuse connection accumulates un-answered requests and any application touching that mount hangs in uninterruptible sleep (`D` state).

Observed in production: `ls /mnt/curvine/<dir>` hung in `D` state, ~104 requests backlogged on the fuse connection, the process alive but the tokio runtime completely idle, logs frozen with a single `warn`.

## Current behavior (as of `main`)

The response path uses a per-sender, long-lived pipe that is reused for the lifetime of the sender and is never reset or drained on error.

- Each `FuseSender` owns one `Pipe2` created once in `FuseSender::new` and never rebuilt — `curvine-fuse/src/session/channel/fuse_sender.rs:52`. The sender loops forever over its receiver — `fuse_sender.rs:70`.
- Sending a response is a two-step splice through that pipe — `fuse_sender.rs:196-209`:
  1. `write_iov` → `vmsplice` (userspace iovec → pipe)
  2. `read_io` → `splice` (pipe → `/dev/fuse`)
- Both underlying calls pass `SPLICE_F_NONBLOCK`, so **partial transfers are allowed**:
  - `splice`: `orpc/src/sys/sys_libc.rs:165` (`SPLICE_F_NONBLOCK | SPLICE_F_MOVE`)
  - `vmsplice`: `orpc/src/sys/sys_libc.rs:436` (`SPLICE_F_NONBLOCK`)
- `Pipe2::write_iov` / `Pipe2::read_io` issue a **single** splice and return the raw byte count with **no completion loop and no length check** — `orpc/src/sys/pipe/pipe2.rs:98-104` and `pipe2.rs:107-113`. A short transfer returns `Ok(short_len)`, not `WouldBlock`.
- A helper that *does* loop to completion, `splice_out_full` (`orpc/src/sys/sys_libc.rs:172`), exists but **has zero callers** — the sender never uses it.
- On a short transfer, `fuse_sender.rs:199` / `:204` see `len_transferred != len` and `err_box!` early-return. But the bytes already pushed into the pipe are left there — the pipe is not drained and not rebuilt.
- The `start` loop only logs a `warn` and continues to the next task — `fuse_sender.rs:102-112` (and `:167-174`). The sender does not exit and the pipe is not reset.

### Why it becomes permanent (two failure modes)

Because the pipe is a FIFO that survives across requests and is never cleared after the error:

1. **Space starvation → every later `write_iov` short-writes.** The residual R bytes occupy the pipe. Subsequent `write_iov` can only push `capacity - R` bytes → `!= len` → `err_box` → request never answered. This matches the reported `res:903552 expect:1048592` (delta ≈ residual). Eventually the pipe fills, `writable()` (`orpc/src/sys/pipe/async_fd.rs:79`) never becomes ready, the sender parks on `.await`, and the runtime goes idle — explaining the single frozen log line and idle tokio.
2. **Corruption even when lengths happen to match.** The pipe is FIFO with the stale R bytes at the head. The next `read_io` splices `len` bytes starting from that stale head into `/dev/fuse`, so the kernel receives a mis-aligned / corrupted response — a protocol-level failure, not just a stuck one.

### Note on the retry path

tokio's `async_io` only retries when the closure returns `WouldBlock` (`orpc/src/sys/pipe/async_fd.rs:105`). A **partial** `vmsplice`/`splice` returns `Ok(short)`, which is *not* an error, so it is surfaced as a successful short result and caught by the `!= len` check — it is never retried. This is why the non-blocking short transfer is the trigger.

## Root cause

The splice response path lacks two things:
1. **Short-transfer completion** — a partial `vmsplice`/`splice` under `SPLICE_F_NONBLOCK` is not looped to completion (the `splice_out_full`-style loop exists but is unused here).
2. **Error-path pipe reset** — after any splice error, the reused pipe is not drained/rebuilt, so residual bytes poison every future response on that sender.

## Proposed solution (open to discussion)

1. **Complete the transfer instead of erroring on short counts.** In `Pipe2::read_io` (and the `write_iov` step), loop until `len` bytes are transferred, treating a partial transfer as "continue" and `WouldBlock`/`EAGAIN` as "await writable/readable then retry" — i.e. wire the existing `splice_out_full` / `splice_in_full` completion pattern into the sender path rather than issuing a single splice.
2. **Reset the pipe on the error path.** If a splice ultimately fails, drain the residual bytes (or rebuild the `Pipe2`) before `continue`, so a single failure cannot poison all subsequent responses on that sender.
3. Optionally, treat a non-recoverable delivery failure on a sender as fatal for that sender (rebuild it) rather than silently `continue`-ing with a dirty pipe.

Steps 1 and 2 are complementary: (1) prevents most poisoning; (2) is the safety net for any residual error path.

## Environment

- curvine: `main` (paths/line numbers cited against current `main`)
- fuse: splice read/write enabled (`FUSE_SPLICE_WRITE | FUSE_SPLICE_READ`), pipe sized to the fuse buffer (`FUSE_MAX_MAX_PAGES * page_size + FUSE_BUFFER_HEADER_SIZE`)
- Linux

I'm happy to open a PR if the direction looks right — likely wiring the completion loop into `Pipe2::read_io`/`write_iov` plus an error-path pipe reset in `FuseSender`. Would appreciate a sanity check on the preferred approach (loop-to-complete vs. rebuild-on-error vs. both) before I do.

bigbigxu:
### Summary

Fixes #965.

A single short (partial) `splice`/`vmsplice` transfer under `SPLICE_F_NONBLOCK` could permanently poison the FuseSender's long-lived, reused pipe. The residual bytes left in the FIFO would either starve the pipe (every subsequent `write_iov` short-writes) or corrupt responses (stale bytes spliced ahead of the real data). Once poisoned, all requests on that sender were never answered, causing the FUSE mount to hang in D state with ~100+ backlogged requests.

### Root Cause

1. `Pipe2::write_iov` and `Pipe2::read_io` issued a **single** `vmsplice`/`splice` syscall and returned the raw byte count. Under `SPLICE_F_NONBLOCK`, the kernel may transfer fewer bytes than requested (short transfer). The caller treated `transferred != expected` as an error and returned early — but the bytes already pushed into the pipe were **left there**.

2. The pipe is a FIFO reused across requests. Residual bytes from a failed transfer:
   - **Space starvation**: occupy pipe capacity, causing every subsequent `write_iov` to short-write → error → more residual bytes → pipe eventually fills → sender parks on `writable().await` forever.
   - **Corruption**: stale bytes at the FIFO head get spliced into `/dev/fuse` ahead of the next response, causing protocol-level misalignment.

3. The sender's error path only logged a `warn!` and continued to the next task — the pipe was never drained or rebuilt.

### Changes

| File | Change |
|------|--------|
| `orpc/src/sys/pipe/pipe2.rs` | `write_iov` and `read_io` now loop to completion; added `skip_iov_bytes` to resume partial vmsplice from the correct offset; added buffer-size validation |
| `curvine-fuse/src/session/channel/fuse_sender.rs` | Added `drain_pipe()` to clear residual bytes on splice error; `pipe2` is now `Option<Pipe2>`; small responses (< 8 KB) use `writev` instead of splice; `write()` checks for short writev |
| `curvine-fuse/src/session/channel/fuse_receiver.rs` | `pipe2` is now `Option<Pipe2>`; `receive()` chooses between `splice()` and `read()`; `read()` uses proper error handling instead of `.unwrap()` |
| `curvine-fuse/src/session/channel/mod.rs` | Pass `enable_splice` config to sender/receiver; `buf_size` now accounts for `max_readahead_kb` |
| `curvine-common/src/conf/fuse_conf.rs` | New `enable_splice: bool` config (default: `true`) |

### How the Fix Works

**1. Loop-to-completion (primary defense)**

`Pipe2::write_iov` and `Pipe2::read_io` now loop until all requested bytes are transferred:

```rust
// write_iov: vmsplice loop
let mut written = 0;
while written < len {
    let cur_iov = Self::skip_iov_bytes(iov, written);
    let res = self.writer.async_write(|fd| sys::vm_splice(fd.fd(), &cur_iov)).await?;
    if res == 0 { return err_box!("vmsplice returned 0"); }
    written += res as usize;
}

// read_io: splice loop
let mut remaining = len;
while remaining > 0 {
    let res = fd_out.async_write(|fd| sys::splice(fd_in, None, fd.fd(), None, remaining)).await?;
    if res == 0 { return err_box!("splice returned 0"); }
    remaining -= res as usize;
}
```

A partial transfer under `SPLICE_F_NONBLOCK` is treated as "continue" — the loop resumes from the partial offset (`skip_iov_bytes` for vmsplice, `remaining` counter for splice). This eliminates the condition that created residual bytes in the first place.

**2. drain_pipe (safety net)**

If a splice ultimately fails (e.g. I/O error, broken pipe), `drain_pipe` reads and discards any residual bytes from the pipe before the sender continues:

```rust
fn drain_pipe(pipe2: &Pipe2) {
    let fd = pipe2.read_raw_fd();
    let mut buf = [0u8; 8192];
    loop {
        match sys::read(fd, &mut buf) {
            Ok(n) if n > 0 => continue,
            _ => break,
        }
    }
}
```

This ensures that even if the completion loop itself fails, the pipe is clean for the next response.

**3. Small-response writev fast path**

Responses smaller than 8 KB now use `writev` (1 syscall) instead of `vmsplice + splice` (2 syscalls). This eliminates pipe involvement entirely for small metadata responses, reducing both latency and the surface area for pipe poisoning.

**4. enable_splice runtime switch**

A new `fuse.enable_splice` config (default: `true`) allows disabling splice entirely as a fallback. When disabled, both receiver and sender use plain `read`/`writev` — extra memory copy but zero pipe management overhead.

### Test Plan

- [ ] Compile and run existing FUSE tests
- [ ] Mount with `enable_splice=true`, run `fio` sequential read/write (128 KB block) — verify no hangs
- [ ] Mount with `enable_splice=false`, run same workload — verify fallback path works
- [ ] Stress test: concurrent `ls`, `cat`, `dd` on the same mount for 10+ minutes — verify no D-state hangs
- [ ] Verify small metadata responses (e.g. `stat`, `lookup`) use writev path (check logs with `debug=true`)
- [ ] Verify large data responses still use splice path

### Issue Reference

Closes #965

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -29,7 +29,7 @@ use orpc::runtime::{RpcRuntime, Runtime};
 use orpc::sync::channel::AsyncSender;
 use orpc::sync::FastDashMap;
 use orpc::sys::pipe::{AsyncFd, Pipe2, PipeFd};
-use orpc::{err_box, sys};
+use orpc::{err_box, sys, try_option_ref};
 use std::sync::Arc;
 use tokio::sync::{watch, Notify};
 use tokio_util::bytes::BytesMut;
@@ -43,7 +43,7 @@ pub struct FuseReceiver<T> {
     fs: Arc<T>,
     rt: Arc<Runtime>,
     sender: AsyncSender<FuseTask>,
-    pipe2: Pipe2,
+    pipe2: Option<Pipe2>,
     buf: BytesMut,
     fuse_len: usize,
     debug: bool,
@@ -64,8 +64,13 @@ impl<T: FileSystem> FuseReceiver<T> {
         audit_logging_enabled: bool,
         metrics_enabled: bool,
         pending_requests: Arc<FastDashMap<u64, Arc<Notify>>>,
+        enable_splice: bool,
     ) -> IOResult<Self> {
-        let pipe2 = Pipe2::new(PipeFd::new(buf_size, false, false)?)?;
+        let pipe2 = if enable_splice {
+            Some(Pipe2::new(PipeFd::new(buf_size, false, false)?)?)
+        } else {
+            None
+        };
         let buf = BytesMut::zeroed(buf_size);
 
         let client = Self {
@@ -87,32 +92,43 @@ impl<T: FileSystem> FuseReceiver<T> {
 
     // Read a data from fuse.
     pub async fn receive(&mut self) -> IOResult<BytesMut> {
-        self.splice().await
+        if self.pipe2.is_some() {
+            self.splice().await
+        } else {
+            self.read().await
+        }
     }
 
-    // Use libc::read to read data, test it, and there are multiple memory copies.
+    // Use libc::read to read data directly into the buffer (no splice).
     pub async fn read(&mut self) -> IOResult<BytesMut> {
+        self.buf.reserve(self.fuse_len);
+        unsafe {
+            self.buf.set_len(self.fuse_len);
+        }
+
         let len = self
             .kernel_fd
             .async_read(|fd| sys::read(fd.fd(), &mut self.buf))
-            .await
-            .unwrap();
-        Ok(BytesMut::from(&self.buf[..len as usize]))
+            .await?;
+        let len = len as usize;
+        if len < FUSE_IN_HEADER_LEN {
+            return err_box!("short read on fuse device");
+        }
+
+        Ok(self.buf.split_to(len))
     }
 
     pub async fn splice(&mut self) -> IOResult<BytesMut> {
-        let write_len = self
-            .pipe2
-            .write_io(&self.kernel_fd, None, self.fuse_len)
-            .await
-            .unwrap();
+        let pipe2 = try_option_ref!(self.pipe2);
+
+        let write_len = pipe2.write_io(&self.kernel_fd, None, self.fuse_len).await?;
 
         self.buf.reserve(write_len);
         unsafe {
             self.buf.set_len(write_len);
         }
 
-        let read_len = self.pipe2.read_buf(&mut self.buf[..write_len]).await?;
+        let read_len = pipe2.read_buf(&mut self.buf[..write_len]).await?;
         if write_len != read_len {
             return err_box!(
                 "splice read and write lengths are inconsistent, write len {}, read len {}",
@@ -124,8 +140,7 @@ impl<T: FileSystem> FuseReceiver<T> {
             return err_box!("short read on fuse device");
         };
 
-        let req_buf = self.buf.split_to(read_len);
-        Ok(req_buf)
+        Ok(self.buf.split_to(read_len))
     }
 
     /// Build a reply handle for `unique`. When `labels` is `Some`, a metrics
--- a/curvine-fuse/src/session/channel/fuse_sender.rs
+++ b/curvine-fuse/src/session/channel/fuse_sender.rs
@@ -24,8 +24,13 @@ use orpc::io::IOResult;
 use orpc::runtime::Runtime;
 use orpc::sync::channel::AsyncReceiver;
 use orpc::sys::pipe::{AsyncFd, Pipe2, PipeFd};
-use orpc::{err_box, sys};
+use orpc::{err_box, sys, try_option_ref};
 use std::sync::Arc;
+
+/// Responses smaller than this threshold use writev (1 syscall) instead of
+/// vmsplice + splice (2 syscalls).  The zero-copy benefit of splice only
+/// outweighs the extra syscall for larger payloads.
+const SPLICE_THRESHOLD: usize = 8192;
 
 /// FuseSender
 /// Reads data from queue and writes to fuse fd.
@@ -36,7 +41,7 @@ pub struct FuseSender<T> {
     rt: Arc<Runtime>,
     kernel_fd: Arc<AsyncFd>,
     receiver: AsyncReceiver<FuseTask>,
-    pipe2: Pipe2,
+    pipe2: Option<Pipe2>,
     debug: bool,
 }
 
@@ -48,8 +53,13 @@ impl<T: FileSystem> FuseSender<T> {
         receiver: AsyncReceiver<FuseTask>,
         buf_size: usize,
         debug: bool,
+        enable_splice: bool,
     ) -> IOResult<Self> {
-        let pipe2 = Pipe2::new(PipeFd::new(buf_size, false, false)?)?;
+        let pipe2 = if enable_splice {
+            Some(Pipe2::new(PipeFd::new(buf_size, false, false)?)?)
+        } else {
+            None
+        };
         let fuse_rx = Self {
             fs,
             rt,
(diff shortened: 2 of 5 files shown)


curvine-fuse/src/session/channel/mod.rs
```diff
@@ -34,7 +34,8 @@ pub struct FuseChannel<T> {
 
 impl<T: FileSystem> FuseChannel<T> {
     pub fn new(fs: Arc<T>, rt: Arc<Runtime>, mnt: &FuseMnt, conf: &FuseConf) -> FuseResult<Self> {
-        let buf_size = FuseUtils::get_fuse_buf_size();
+        let max_readahead = conf.max_readahead_kb.unwrap_or(0).saturating_mul(1024);
+        let buf_size = FuseUtils::get_fuse_buf_size().max(max_readahead as usize);
```
szbr00: This sizes the pipe to `max_readahead` bytes exactly, but read replies include a FUSE response header in addition to the data payload. With the default 1 MiB readahead this is masked by `FuseUtils::get_fuse_buf_size()` including extra header space, but if a user configures a larger `max_readahead_kb`, a full-size read reply becomes `max_readahead + fuse_out_header`, which is larger than `buf_size`. The new `Pipe2::write_iov` validation then rejects the response before writing it, so large configured reads can fail. Please include response header slack here, e.g. max against `max_readahead as usize + FUSE_BUFFER_HEADER_SIZE` or the exact out-header allowance used by the FUSE buffer sizing convention.

curvine-fuse/src/session/channel/fuse_sender.rs
```diff
@@ -178,33 +188,62 @@ impl<T: FileSystem> FuseSender<T> {
...
 
     pub async fn write(&mut self, rep: ResponseData) -> IOResult<()> {
-        let (_, iovec) = rep.as_iovec()?;
-        self.kernel_fd
+        let (len, iovec) = rep.as_iovec()?;
+        let written = self
+            .kernel_fd
             .async_write(|fd| sys::writev(fd.fd(), &iovec))
             .await?;
+        if written as usize != len {
```
szbr00: The new direct `writev` path still assumes the kernel writes the full iovec in one syscall. If `writev` returns a positive short count, some bytes may already have been accepted by `/dev/fuse`, but this returns an error and drops the remaining bytes. That makes the `enable_splice=false` fallback vulnerable to the same class of "request never receives a complete response" failure, and the small-response fast path now uses it by default. Consider looping to completion with the same offset/resume logic as `Pipe2::write_iov` instead of treating a short positive write as terminal.

curvine-fuse/src/session/channel/fuse_sender.rs
```diff
@@ -178,33 +188,62 @@ impl<T: FileSystem> FuseSender<T> {
...
+
+    /// Drain any residual bytes left in the pipe after a failed transfer.
+    /// This prevents the pipe from being permanently poisoned (issue #965):
+    /// without draining, stale bytes at the head of the FIFO would corrupt
+    /// every subsequent response.
+    fn drain_pipe(pipe2: &Pipe2) {
+        let fd = pipe2.read_raw_fd();
+        let mut buf = [0u8; 8192];
+        loop {
+            match sys::read(fd, &mut buf) {
```
szbr00, line 247: `drain_pipe` breaks on every read error. If `read` is interrupted (`EINTR`) while residual bytes are still in the pipe, the pipe can remain partially poisoned and the next response can still see stale FIFO contents. Please retry interrupted reads and only stop once the nonblocking pipe is empty (`EAGAIN`/`EWOULDBLOCK`) or EOF/error conditions that really mean no more data should be drained.

orpc/src/sys/pipe/pipe2.rs
```diff
@@ -94,22 +94,59 @@ impl Pipe2 {
...
+                self.buf_size
+            );
+        }
         let fd_in = self.reader.raw_fd();
-        let res = fd_out
-            .async_write(|fd| sys::splice(fd_in, None, fd.fd(), None, len))
-            .await?;
-        Ok(res as usize)
+        let mut remaining = len;
+        while remaining > 0 {
```
szbr00, line 140: The loop-to-completion here is on the pipe → /dev/fuse direction. Unlike vmsplice into a pipe, a write/splice to /dev/fuse is message-oriented: the kernel expects each reply to arrive as a single complete write. If a splice under SPLICE_F_NONBLOCK returns a short count and we resume with the remaining bytes in a second splice, the kernel may interpret the two fragments as two separate (malformed) replies rather than one, which is a different failure mode than the pipe-poisoning this PR targets. Could you confirm the FUSE device accepts a resumed partial reply here? If it does not, the sender side likely needs to guarantee the whole reply is handed to /dev/fuse in one splice (fits since len <= buf_size <= pipe capacity), treating a short splice as an error to retry from the pipe head rather than advancing `remaining`.

curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -87,32 +92,43 @@ impl<T: FileSystem> FuseReceiver<T> {
...
+        }
     }
 
-    // Use libc::read to read data, test it, and there are multiple memory copies.
+    // Use libc::read to read data directly into the buffer (no splice).
     pub async fn read(&mut self) -> IOResult<BytesMut> {
+        self.buf.reserve(self.fuse_len);
+        unsafe {
+            self.buf.set_len(self.fuse_len);
+        }
```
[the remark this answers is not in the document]
bigbigxu: This understanding is accurate; each read operation overwrites the buf data, so there is no undefined issue.

curvine-fuse/src/session/channel/fuse_sender.rs
```diff
@@ -178,33 +188,62 @@ impl<T: FileSystem> FuseSender<T> {
...
+                Ok(n) if n > 0 => continue,
+                _ => break,
+            }
         }
     }
 }
```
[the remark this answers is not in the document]
bigbigxu: fixed

curvine-fuse/src/session/channel/mod.rs
bigbigxu: fixed

curvine-fuse/src/session/channel/fuse_sender.rs
bigbigxu: fixed

orpc/src/sys/pipe/pipe2.rs
bigbigxu: /dev/fuse is message-based, and each write operation must be completed completely, so writev and readv do not need to loop through each other.

szbr00 (approved):

PR #1006, opened 2026-07-07
refactor(worker): introduce BlockLayout and BlockMetaStore

gongxun0928:
# Summary

This PR introduces two new abstraction boundaries for the worker block storage path: `BlockLayout` (encapsulates filesystem vs SPDK block lifecycle operations) and `BlockMetaStore` (provides a unified metadata store interface for the in-memory index and optional SPDK RocksDB persistence). The goal is to isolate SPDK-specific branching from `BlockStore`, `VfsDataset`, and `VfsDir` so that future storage layouts or metadata backends can be added safely.

# Issue Describe / Design

- **Root cause:** The worker block path did not have a real abstraction boundary. `VfsDir` mixed media accounting with block lifecycle work such as allocate, finalize, scan, and remove. `VfsDataset` owned both the live block map and SPDK RocksDB persistence details, so callers still had to understand SPDK-specific metadata. As SPDK support grew, `StorageType::SpdkDisk` branches spread through the worker storage path, making future layouts or worker-side metadata stores hard to add safely.
- **Fix approach:**
  - Add `BlockLayout` trait with `FileLayout` (filesystem blocks) and `BdevLayout` (SPDK blocks) implementations for the block lifecycle operations covered by this first PR.
  - Add `BlockLayouts` as the dataset-local registry that maps `StorageType` to the correct layout without passing SPDK state through non-SPDK paths.
  - Add `BlockMetaStore` trait plus `MemMetaStore` (in-memory index) and `VfsMetaStore` (in-memory + SPDK persistence coordinator), and adapt `SpdkMetaStore` to the new metadata-store boundary while preserving the existing 29-byte RocksDB value format.
  - Move filesystem block lifecycle code out of `VfsDir` into `FileLayout`.
  - Move SPDK allocation, restore, and deallocation lifecycle code into `BdevLayout`.
  - Route `BlockStore::async_remove_block` through `VfsDataset` instead of directly manipulating storage internals.
- **Design decisions:**
  - `VfsDataset` remains the orchestrator for block lifecycle consistency.
  - `VfsDir` is reduced toward media state, capacity accounting, and health checks.
  - `VfsMetaStore` coordinates the in-memory index with optional SPDK persistence, so callers do not need to branch on SPDK metadata details.
  - The handler data path is intentionally unchanged in this PR: `BlockMeta::create_reader/create_writer` and the handler-owned `BlockDevice` flow stay as-is.

# Changes

| `curvine-server/src/worker/block/block_store.rs` | Route `async_remove_block` through `VfsDataset` instead of manipulating storage internals | None |
| `curvine-server/src/worker/storage/layout/mod.rs` | Add `BlockLayout` trait, `BlockLayoutKind` enum, and `BlockLayouts` registry | New module |
| `curvine-server/src/worker/storage/layout/file_layout.rs` | `FileLayout`: allocate/finalize/scan/release/deallocate for filesystem blocks | New module |
| `curvine-server/src/worker/storage/layout/bdev_layout.rs` | `BdevLayout`: allocate/finalize/scan/release/deallocate for SPDK blocks | New module |
| `curvine-server/src/worker/storage/meta_store/mod.rs` | Add `BlockMetaStore` trait | New module |
| `curvine-server/src/worker/storage/meta_store/mem_meta_store.rs` | `MemMetaStore`: in-memory block index | New module |
| `curvine-server/src/worker/storage/meta_store/vfs_meta_store.rs` | `VfsMetaStore`: coordinates in-memory index with SPDK persistence | New module |
| `curvine-server/src/worker/storage/spdk_meta_store.rs` | Implement `BlockMetaStore` trait, preserving 29-byte RocksDB value format | None |
| `curvine-server/src/worker/storage/vfs_dataset.rs` | Replace `block_map` + `spdk_meta` with `VfsMetaStore`, delegate lifecycle to `BlockLayouts`, add `remove_block_state_by_id`/`release_block_space`/`remove_block_by_id` | None |
| `curvine-server/src/worker/storage/vfs_dir.rs` | Remove `create_block`, `finalize_block`, `scan_blocks`, `scan_spdk_blocks` (moved to layouts) | None |
| `curvine-server/src/worker/storage/mod.rs` | Export `meta_store` and `layout` modules | None |

# Test verified

| `cargo test -p curvine-server` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/storage/layout/file_layout.rs
+++ b/curvine-server/src/worker/storage/layout/file_layout.rs
@@ -55,6 +55,7 @@ impl BlockLayout for FileLayout {
         Ok(blocks)
     }
 
+    // Filesystem blocks own no under-lock state; offset/file teardown is in deallocate().
     fn release(&self, _meta: &BlockMeta) -> CommonResult<()> {
         Ok(())
     }
--- a/curvine-server/src/worker/storage/layout/file_layout.rs
+++ b/curvine-server/src/worker/storage/layout/file_layout.rs
@@ -18,7 +18,7 @@ use crate::worker::storage::VfsDir;
 use curvine_common::state::ExtendedBlock;
 use orpc::common::FileUtils;
 use orpc::CommonResult;
-use std::fs::{self, File};
+use std::fs::File;
 
 #[derive(Clone, Copy)]
 pub struct FileLayout;
@@ -60,7 +60,7 @@ impl BlockLayout for FileLayout {
     }
 
     fn deallocate(&self, meta: &BlockMeta) -> CommonResult<()> {
-        fs::remove_file(meta.get_block_path()?)?;
+        FileUtils::delete_path(meta.get_block_path()?, false)?;
         Ok(())
     }
 }
--- a/curvine-server/src/worker/storage/layout/file_layout.rs
+++ b/curvine-server/src/worker/storage/layout/file_layout.rs
@@ -1,0 +1,66 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::worker::block::{BlockMeta, BlockState};
+use crate::worker::storage::layout::BlockLayout;
+use crate::worker::storage::VfsDir;
+use curvine_common::state::ExtendedBlock;
+use orpc::common::FileUtils;
+use orpc::CommonResult;
+use std::fs::{self, File};
+
+#[derive(Clone, Copy)]
+pub struct FileLayout;
+
+impl BlockLayout for FileLayout {
+    fn allocate(&self, dir: &VfsDir, block: &ExtendedBlock) -> CommonResult<BlockMeta> {
+        let meta = BlockMeta::with_tmp(block, dir);
+        let file = meta.get_block_path()?;
+        File::create(file)?;
+        Ok(meta)
+    }
+
+    fn finalize(&self, meta: &BlockMeta, _committed_len: i64) -> CommonResult<BlockMeta> {
+        BlockMeta::with_final(meta)
+    }
+
+    fn scan(&self, dir: &VfsDir) -> CommonResult<Vec<BlockMeta>> {
+        let active_dir = FileUtils::list_files(&dir.active_dir, true)?;
+        let staging_dir = FileUtils::list_files(&dir.staging_dir, true)?;
+
+        let mut blocks = vec![];
+        for file in active_dir {
+            if let Ok(meta) = BlockMeta::from_file(&file, BlockState::Finalized, dir) {
+                blocks.push(meta);
+            }
+        }
+
+        for file in staging_dir {
+            if let Ok(meta) = BlockMeta::from_file(&file, BlockState::Recovering, dir) {
+                blocks.push(meta);
+            }
+        }
+
+        Ok(blocks)
+    }
+
+    fn release(&self, _meta: &BlockMeta) -> CommonResult<()> {
+        Ok(())
+    }
+
+    fn deallocate(&self, meta: &BlockMeta) -> CommonResult<()> {
+        fs::remove_file(meta.get_block_path()?)?;
+        Ok(())
+    }
+}
--- a/curvine-server/src/worker/storage/meta_store/vfs_meta_store.rs
+++ b/curvine-server/src/worker/storage/meta_store/vfs_meta_store.rs
@@ -1,0 +1,88 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::worker::block::BlockMeta;
+use crate::worker::storage::meta_store::{BlockMetaStore, MemMetaStore};
+use crate::worker::storage::SpdkMetaStore;
+use curvine_common::state::StorageType;
+use log::warn;
+use std::sync::Arc;
+
+pub struct VfsMetaStore {
+    mem: MemMetaStore,
+    spdk: Option<Arc<SpdkMetaStore>>,
+}
+
+impl VfsMetaStore {
+    pub fn new(spdk: Option<Arc<SpdkMetaStore>>) -> Self {
+        Self {
+            mem: MemMetaStore::new(),
+            spdk,
+        }
+    }
+
+    pub fn get(&self, id: i64) -> Option<&BlockMeta> {
+        self.mem.get(id)
+    }
+
+    pub fn put(&mut self, meta: BlockMeta) -> Option<BlockMeta> {
+        self.persist_spdk_put(&meta);
+        self.mem.put(meta)
+    }
+
+    pub fn remove(&mut self, id: i64) -> Option<BlockMeta> {
+        let meta = self.mem.remove(id);
+        if let Some(meta) = meta.as_ref() {
+            self.persist_spdk_remove(meta);
+        }
+        meta
+    }
+
+    pub fn block_count(&self) -> usize {
+        self.mem.block_count()
+    }
+
+    pub fn all_blocks(&self) -> Vec<BlockMeta> {
+        self.mem.values().cloned().collect()
+    }
+
+    #[cfg(test)]
+    pub fn contains_key(&self, id: i64) -> bool {
+        self.mem.contains_key(id)
+    }
+
+    fn persist_spdk_put(&self, meta: &BlockMeta) {
+        if meta.storage_type() != StorageType::SpdkDisk {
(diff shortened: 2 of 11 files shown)


curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        Ok(blocks)
+    }
+
+    fn release(&self, _meta: &BlockMeta) -> CommonResult<()> {
+        Ok(())
+    }
+
+    fn deallocate(&self, meta: &BlockMeta) -> CommonResult<()> {
+        fs::remove_file(meta.get_block_path()?)?;
```
szbr00: This changes delete semantics on the `async_remove_block` path. Previously that path used `FileUtils::delete_path(path, false)`, which is guarded by `if path.exists()` and is a no-op when the backing file is already gone. `fs::remove_file` instead returns `NotFound`, so `async_remove_block` now propagates an error for an already-missing file.

Because `async_remove_block` runs `layout.deallocate(&meta)?` before `release_block_space(&meta)`, that error also skips `dir.release_space(...)` — the block is already removed from the meta index and `blocks_to_delete` is already decremented, so the dir's used-space accounting is never reversed. Net effect: a delete that used to succeed silently now errors and leaks capacity.

Consider preserving the tolerant semantics here, e.g. ignore `ErrorKind::NotFound` in `deallocate` (or keep using `delete_path`), so a missing file stays a no-op and space is still released.

curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -240,6 +191,37 @@ impl VfsDataset {
...
+        id: i64,
+    ) -> CommonResult<(BlockMeta, BlockLayoutKind)> {
+        let meta = match self.meta.remove(id) {
+            None => return err_box!("Not found block {}", id),
+            Some(meta) => meta,
+        };
+
+        let layout = self.layouts.get(meta.storage_type());
+        if meta.storage_type() == StorageType::SpdkDisk {
+            self.find_dir(meta.dir_id())?;
```
szbr00: The old delete path tolerated a missing dir (`if let Ok(dir) = find_dir(...)`), silently skipping the offset free. Here `find_dir(...)?` runs after `self.meta.remove(id)` has already deleted the in-memory entry and the RocksDB record, so a missing dir now returns an error with the metadata already removed but the offset not freed — a partial, inconsistent teardown. If a missing dir is genuinely not expected, an assertion/log would be clearer; if it can happen, consider freeing the offset from `meta.dir` (as `release` does) without requiring `find_dir` to succeed.

curvine-server/src/worker/storage/layout/file_layout.rs
gongxun0928: To address this issue, we replaced the direct call to fs::remove_file with FileUtils::delete_path(path, false) for file deletion. This restores the tolerant semantics where deleting a file that is already missing is treated as a no-op and does not return an error.

Additionally, we unified the file deletion logic in abort_block to also use FileUtils::delete_path(path, false), ensuring consistent behavior across all related code paths.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Gate layout.release() on find_dir(...) success for all storage types instead of requiring SPDK dirs up front, preserving the old offset-free skip when the dir is absent while keeping FileLayout::release as a no-op.

curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -301,78 +281,68 @@ impl Dataset for VfsDataset {
...
+        dir.release_space(false, actual_len);
         dir.reserve_space(true, final_meta.actual_len);
-        self.block_map.insert(final_meta.id(), final_meta.clone());
-        self.spdk_put(&final_meta);
+        self.meta.put(final_meta.clone());
 
         Ok(final_meta)
     }
 
     fn abort_block(&mut self, block: &ExtendedBlock) -> CommonResult<()> {
```
szbr00, line 341: A few points on this `abort_block` path (all minor):

**Lock-held file IO (Minor):** `remove_block_by_id` runs `layout.deallocate` (file deletion for `FileLayout`) while the dataset `&mut` write lock is held, whereas `async_remove_block` was deliberately restructured to drop the lock before `deallocate`. The nice lock-free-IO pattern isn't applied here. This isn't a regression (the old `abort_block` also deleted under the lock), but consider mirroring the `async_remove_block` split — release under lock, deallocate after — or add a comment noting the lock is intentionally held.

**Lenient delete (Minor):** the FS deallocate now uses `FileUtils::delete_path(.., false)`, which silently ignores a missing file, versus the old `try_err!(fs::remove_file(..))` that errored if the file was gone. This is arguably a better idempotent-abort behavior, just flagging it as a deliberate semantic change.

**Double lookup (Nit):** `self.meta.get(block.id).is_none()` followed by `remove_block_by_id` → `self.meta.remove(id)` does two hash lookups. Since `self` is `&mut`, you could early-return on the removed `Option` directly.

curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -0,0 +1,66 @@
...
+        for file in staging_dir {
+            if let Ok(meta) = BlockMeta::from_file(&file, BlockState::Recovering, dir) {
+                blocks.push(meta);
+            }
+        }
+
+        Ok(blocks)
+    }
+
+    fn release(&self, _meta: &BlockMeta) -> CommonResult<()> {
```
szbr00, line 59: **Nit:** `FileLayout::release` (and symmetrically `BdevLayout::deallocate`) is an empty no-op. The release/deallocate contract is well-documented on the trait, but a one-line comment at each no-op impl (e.g. "filesystem blocks own no under-lock state; nothing to release") would make the asymmetry obvious to future readers.

curvine-server/src/worker/storage/meta_store/vfs_meta_store.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn all_blocks(&self) -> Vec<BlockMeta> {
+        self.mem.values().cloned().collect()
+    }
+
+    #[cfg(test)]
+    pub fn contains_key(&self, id: i64) -> bool {
+        self.mem.contains_key(id)
+    }
+
+    fn persist_spdk_put(&self, meta: &BlockMeta) {
```
szbr00, line 65: **Nit (pre-existing, out of scope):** `persist_spdk_put`/`persist_spdk_remove` swallow RocksDB errors with only a `warn!`, so a failed persist still updates the in-memory index — leaving mem and RocksDB divergent (the block is lost from RocksDB on restart). This is carried over verbatim from the old `spdk_put`/`spdk_delete`, so no change is needed in this refactor PR; noting it as a candidate for a follow-up that surfaces or handles persistence failures.

szbr00 (commented):
## Code Review

Reviewed 11 files, +610 / −301. Pure worker-storage refactor; no behavior change intended.

### Summary
The extraction of SPDK-vs-filesystem branching into `BlockLayout` (`FileLayout`/`BdevLayout` + `BlockLayouts` registry) and metadata handling into `BlockMetaStore`/`VfsMetaStore` is clean and behavior-preserving. I traced each moved path against the original: allocate/finalize/scan/release/deallocate reproduce the old `VfsDir` and SPDK logic exactly; `async_remove_block` keeps its lock scoping (offset + RocksDB freed under lock, file IO outside); the `finalize_block` borrow-scoping rework and the `open_block` `.cloned()` are correct fixes for `&`/`&mut` conflicts; the 29-byte SPDK RocksDB format is preserved. The change is self-contained — the removed `VfsDir` methods and `block_map`/`spdk_put`/`spdk_delete` have no external callers, and `Dataset::finalize_block` (used by the write handlers) is untouched. Static enum dispatch avoids `dyn` overhead. Tests were updated to use the public API instead of the private `block_map`.

### Findings: 5 (no Critical/Major)
- Critical: 0
- Major: 0
- Minor: 2 (abort_block holds the dataset lock across file deletion; abort file-delete is now lenient about a missing file)
- Nit: 3 (abort_block double lookup; no-op layout methods lack a comment; pre-existing SPDK persist errors are swallowed)

Sound and merge-ready from a correctness standpoint. See inline comments for details.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Re: abort_block holds write lock across file deletion (Minor)

Agreed that holding the dataset write lock across deallocate is suboptimal. Traced the call chain: all production callers go through BlockStore::abort_block → self.write()?.abort_block(...), so the RwLockWriteGuard spans the entire VfsDataset::abort_block call, including file IO. Splitting deallocate/release inside VfsDataset alone would not shrink the lock scope for this path.

async_remove_block already uses the intended 3-phase pattern at the BlockStore layer (write lock for meta/offset release → unlock → deallocate → read lock for space accounting). The same pattern needs to be applied at BlockStore::abort_block, not in the layout refactor.

Since this PR is a behavior-preserving storage abstraction refactor and abort is a low-frequency error path, we'd prefer to land the BlockStore-level lock narrowing in a follow-up PR rather than mixing lock-orchestration changes into this one. Happy to file/track that separately.

Re: abort double lookup (Nit)
Good catch. Abort currently does get then remove because remove_block_state_by_id treats missing blocks as an error (needed by async_remove_block), while abort must be idempotent (Ok(()) when absent). Legacy code used a single remove for both behaviors.
The extra lookup is an in-memory HashMap op on a low-frequency error path, so we’re leaving it as-is for this refactor PR. We can collapse it to a single remove when we do the follow-up BlockStore-level abort teardown/lock-scope work.

curvine-server/src/worker/storage/meta_store/vfs_meta_store.rs
gongxun0928: we’re leaving it as-is for this refactor PR.

curvine-server/src/worker/storage/layout/file_layout.rs
gongxun0928: Added one-line comments on FileLayout::release and BdevLayout::deallocate to document the intentional split: filesystem blocks defer teardown to deallocate, while bdev blocks release offsets under lock in release.

szbr00 (approved):

PR #1010, opened 2026-07-08
fix(cli): remove redundant conf arg shadowing global config in load subcommands (#1008)

### LINKED ISSUE #1008: [BUG]: load-status read error config.
When using `cv load-status job_id` to check the job status, it read the wrong port of the master.    

```

Checking status for job_47924dc96d807517c75ccf9fedd60bdb (refreshing every 5 seconds)
Press Ctrl+C to stop watching.
2026/07/07 09:59:41.059 WARN cluster_connector.rs:312 rpc(-4666363644864835651) node localhost:8995 is unavailable: Cannot assign requested address (os error 99):
2026/07/07 09:59:41.159 WARN cluster_connector.rs:312 rpc(-4666363644864835651) node localhost:8995 is unavailable: Cannot assign requested address (os error 99):
2026/07/07 09:59:41.361 WARN cluster_connector.rs:312 rpc(-4666363644864835651) node localhost:8995 is unavailable: Cannot assign requested address (os error 99):
2026/07/07 09:59:41.761 WARN cluster_connector.rs:312 rpc(-4666363644864835651) node localhost:8995 is unavailable: Cannot assign requested address (os error 99):
```

bigbigxu:
# Summary

Remove the redundant `--conf` argument from `load`, `load-status`, and `cancel-load` subcommands that shadowed the global `--conf` argument on `CurvineArgs`, causing the CLI to ignore user-provided config files and fall back to the default `localhost:8995` master address.

# Issue Describe / Design

Closes #1008

**Root cause:** `CurvineArgs` in `main.rs` defines a global `--conf` argument (`global = true`). However, the `LoadCommand`, `LoadStatusCommand`, and `CancelLoadCommand` subcommands each also defined their own `--conf` argument with `default_value = "${CURVINE_CONF_FILE}"`. In clap, a subcommand-level argument shadows the global argument. When users passed `--conf /path/to/config` after the subcommand (e.g., `cv load-status <job_id> --conf /path`), the value was captured by the subcommand's `conf` field, while `CurvineArgs.conf` remained `None`. Since `get_conf()` only reads `self.conf` (the global arg), the config file was never loaded, and the CLI fell back to the default configuration with `localhost:8995`.

Additionally, the `default_value = "${CURVINE_CONF_FILE}"` is a literal string — clap does not expand environment variables in `default_value`. The subcommand `conf` fields were dead code, never used for configuration loading.

**Fix approach:** Remove the redundant `conf` field from all affected subcommands so the global `--conf` argument on `CurvineArgs` is the single source of truth. The `get_conf()` method already handles the priority chain: CLI `--conf` -> `CURVINE_CONF_FILE` env var -> default config.

# Changes

| `curvine-cli/src/cmds/load_status.rs` | Removed `conf` field and updated `new()` signature | `--conf` now correctly routes to the global argument |
| `curvine-cli/src/cmds/load.rs` | Removed `conf` field and updated `LoadStatusCommand::new()` call | Same as above |
| `curvine-cli/src/cmds/load_cancel.rs` | Removed `conf` field | Same as above |

# Test verified

| `cargo check -p curvine-cli` | PASS | Compilation verified |

# Dependencies

- Related issues: #1008
- Related PRs: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-cli/src/cmds/load.rs
+++ b/curvine-cli/src/cmds/load.rs
@@ -23,9 +23,6 @@ use orpc::CommonResult;
 pub struct LoadCommand {
     path: String,
 
-    #[arg(long, default_value = "${CURVINE_CONF_FILE}")]
-    conf: String,
-
     /// Watch load job status after submission
     #[arg(long, short = 'w')]
     watch: bool,
@@ -46,12 +43,8 @@ impl LoadCommand {
         println!("{}", rep);
 
         if self.watch {
-            let status_command = LoadStatusCommand::new(
-                rep.job_id.clone(),
-                false,
-                "1s".to_string(),
-                self.conf.clone(),
-            );
+            let status_command =
+                LoadStatusCommand::new(rep.job_id.clone(), false, "1s".to_string());
 
             status_command.execute(client).await?;
         }
--- a/curvine-cli/src/cmds/load_cancel.rs
+++ b/curvine-cli/src/cmds/load_cancel.rs
@@ -21,9 +21,6 @@ use crate::util::*;
 #[derive(Parser, Debug)]
 pub struct CancelLoadCommand {
     job_id: String,
-
-    #[arg(long, default_value = "${CURVINE_CONF_FILE}")]
-    conf: String,
 }
 
 impl CancelLoadCommand {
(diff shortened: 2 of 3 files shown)


szbr00 (commented):
The fix is correct and complete (only these three subcommands carried the shadowing `--conf`). One suggestion: this class of bug is invisible to `cargo check` and regressed silently before, so a small regression test that parses `cv load-status <job> --conf /path` and asserts the value lands on the global `CurvineArgs.conf` (and that the subcommands expose no local `conf`) would lock the behavior. Non-blocking if CLI parse-test scaffolding doesn't exist yet.

szbr00 (approved):

PR #1005, opened 2026-07-07
fix(master): harden load jobs after auto-cache isolation (#984)

jlon:
# Summary

This PR is a hotfix follow-up to #998 for #984. It hardens load/auto-cache submission end to end: master-side load planning runs behind an isolated bounded queue, queue pressure is returned as a structured `ResourceExhausted` error, and clients use their own bounded auto-cache submit queue with retry instead of spawning unbounded submit RPCs.

# Issue Describe / Design

Related to #984. Follow-up to #998.

Root cause:
- #998 moved auto-cache load planning off the submit RPC path, but background load work still needed a bounded pressure point and a structured queue-full signal.
- The client auto-cache path submitted one async task per cache miss. When the master queue was full, the failure was only a stringly `Common` error, so the client could not make a stable retry decision.
- Public `cv load` on a CV path had ambiguous direction: UFS-only metadata should hydrate Curvine from the mounted UFS source, while a CV path that already has Curvine data should keep exporting/flushing Curvine data to UFS.
- When direct UFS reads were disabled, valid auto-cache opportunities could be skipped before the load job was submitted.

Reproduction steps:
- Drive many cache misses or `cv load` requests while master-side load planning is saturated.
- Observe that queue-full rejection is not distinguishable by error kind and client auto-cache has no local queue/retry policy.
- Mount an external UFS path in FsMode, sync UFS-only metadata, then run `cv load -w <cv-path>`.
- Disable direct UFS reads and open a valid auto-cache miss.

Fix approach:
- Add a dedicated `master-load-job` runtime plus bounded FIFO queue for master-side load planning.
- Return `FsError::ResourceExhausted` when the master load-job queue is full.
- Add a client-side bounded auto-cache submit queue with local pending-path dedupe and finite retry on `ResourceExhausted`.
- Preserve duplicate-job semantics and add explicit shutdown handling for queued master workers.
- Select CV-path load direction from real Curvine metadata: UFS-only metadata imports from UFS, existing Curvine data exports to UFS.
- Keep submit/wait/get-status job-id derivation consistent for UFS-only metadata.

# Changes

| Module / File | Change | Impact on existing behavior |
| ------------- | ------ | --------------------------- |
[...]

PR #1012, opened 2026-07-08
build(docker): narrow runtime image build scope

jlon:
# Summary

Split out the Docker/runtime build changes from PR #1005. This PR narrows runtime image builds to Curvine runtime binaries and copies the complete `etc/` directory into the distribution.

# Issue Describe / Design

Related to #984 and #1005 split work.

Root cause / motivation:
- Runtime images do not need SDK build targets by default.
- Dockerfiles carried local WebUI patch steps that are better avoided in the image build.
- `build.sh` copied only top-level files from `etc/`, which can miss nested config directories.

# Changes

| `build/build.sh` | Copy `etc/` recursively into dist conf | Preserves nested config files/directories |
| `curvine-docker/deploy/Dockerfile_rocky9` | Build runtime packages via `BUILD_ARGS`; remove WebUI patch workaround | Runtime image build is narrower by default |
| `curvine-docker/deploy/Dockerfile_ubuntu24` | Add configurable `BUILD_ARGS`; remove WebUI patch workaround | Runtime image build is narrower by default |

# Test verified

| `bash -n build/build.sh` | PASS | Shell syntax check |
| `git diff --cached --check` | PASS | No whitespace errors before commit |

# Dependencies

- Related PRs: #1005
- Related issues: #984
- Blocks / blocked by: None

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -290,8 +290,8 @@ mkdir -p "$DIST_DIR"/lib
 mkdir -p "$DIST_DIR"/tests
 
 
-# Copy configuration files and bin
-cp "$FS_HOME"/etc/* "$DIST_DIR"/conf
+# Copy configuration files and directories.
+cp -R "$FS_HOME"/etc/. "$DIST_DIR"/conf/
 
 cp "$FS_HOME"/build/bin/* "$DIST_DIR"/bin
 chmod +x "$DIST_DIR"/bin/*
--- a/curvine-docker/deploy/Dockerfile_rocky9
+++ b/curvine-docker/deploy/Dockerfile_rocky9
@@ -77,7 +77,7 @@ ENTRYPOINT []
 # Examples:
 #   -p core --ufs opendal-s3 --spdk-rdma                   # Core + S3 + SPDK-RDMA
 #   -p server -p client -p tests --ufs opendal-s3 --spdk   # Bench + SPDK-TCP only
-ARG BUILD_ARGS="--skip-java-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs --spdk-rdma"
+ARG BUILD_ARGS="-p server -p client -p cli -p fuse -p web --skip-java-sdk --skip-python-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs --spdk-rdma"
 
 # Copy and apply build configurations
 COPY curvine-docker/deploy/config /build-config/config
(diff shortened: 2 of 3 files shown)


curvine-docker/deploy/Dockerfile_rocky9
```diff
@@ -77,7 +77,7 @@ ENTRYPOINT []
 # Examples:
 #   -p core --ufs opendal-s3 --spdk-rdma                   # Core + S3 + SPDK-RDMA
 #   -p server -p client -p tests --ufs opendal-s3 --spdk   # Bench + SPDK-TCP only
-ARG BUILD_ARGS="--skip-java-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs --spdk-rdma"
+ARG BUILD_ARGS="-p server -p client -p cli -p fuse -p web --skip-java-sdk --skip-python-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs-native --ufs oss-hdfs --spdk-rdma"
```
szbr00, line 80: **Major (verification):** This runtime `BUILD_ARGS` keeps `-p web`, but the two `RUN` blocks that previously worked around the `vue-cli-service` wrapper failure (`npm install --legacy-peer-deps` + the `node_modules/package.json` symlink, and the `sed` patch of `build.sh`'s `npm run build`) are removed in this PR. The image now builds the WebUI via `build.sh`'s plain `npm install && npm run build`.

If that workaround was genuinely required, the WebUI build will break at Docker build time. The PR test plan only ran `bash -n build.sh` and `git diff --check` — neither image was actually built. Please build at least one runtime image end-to-end (or confirm #1005 already verified WebUI builds without the workaround) before merge.

**Minor:** `--skip-java-sdk --skip-python-sdk` are redundant here since the explicit `-p` list already excludes `java`/`python`. Harmless, but keep or drop consciously.

curvine-docker/deploy/Dockerfile_ubuntu24
```diff
@@ -175,11 +158,11 @@ RUN echo "Building Curvine from source..." && \
...
     echo "Using TMPDIR=$TMPDIR for compilation temporary files" && \
-    echo "Running 'make all' with --skip-java-sdk and all UFS support..." && \
-    # Build with retry mechanism and all UFS support
+    echo "Running runtime build with retry mechanism: make all ARGS=\"${BUILD_ARGS}\"" && \
+    # Build with retry mechanism and runtime UFS support
     for attempt in 1 2 3; do \
         echo "Build attempt $attempt..." && \
-        if make all ARGS='--skip-java-sdk --ufs opendal-s3 --ufs opendal-oss --ufs opendal-azblob --ufs opendal-gcs --ufs opendal-hdfs --ufs opendal-webhdfs --ufs oss-hdfs'; then \
+        if make all ARGS="${BUILD_ARGS}"; then \
             echo "✓ Build succeeded with all UFS support" && \
```
szbr00, line 166: **Nit:** The success branch still logs `"✓ Build succeeded with all UFS support"`, but the build is now the narrower runtime set defined by `BUILD_ARGS` above. Consider updating the message to reflect the runtime scope so logs aren't misleading.

**Nit (pre-existing):** The UFS sets diverge between images — `Dockerfile_rocky9` uses `opendal-hdfs-native` + `--spdk-rdma`, while this file uses `opendal-hdfs` + `opendal-webhdfs` and no SPDK. Not introduced by this PR, but now that both are parameterized the divergence is easy to overlook; consider aligning or documenting why they differ.

build/build.sh
```diff
@@ -290,8 +290,8 @@ mkdir -p "$DIST_DIR"/lib
 mkdir -p "$DIST_DIR"/tests
 
 
-# Copy configuration files and bin
-cp "$FS_HOME"/etc/* "$DIST_DIR"/conf
+# Copy configuration files and directories.
+cp -R "$FS_HOME"/etc/. "$DIST_DIR"/conf/
```
szbr00, line 294: Nice fix — this is a real bug, not just a nicety. `etc/` contains a nested `grafana/` directory, and with `set -e` active the old `cp "$FS_HOME"/etc/* ...` would hit `cp: etc/grafana is a directory (not copied)`, return non-zero, and abort the whole build. The recursive `cp -R .../etc/.` correctly copies nested config directories.

szbr00 (commented):
## Code Review

Reviewed 3 files, +14 / −53. Infra/build change; overall clean and the intent is sound.

### Summary
The change to narrow runtime image `BUILD_ARGS` and copy `etc/` recursively is a good improvement — the recursive copy is actually a latent bug fix (nested `etc/grafana/` would abort the build under `set -e`). The `BUILD_ARGS` plumbing is correct: both Dockerfiles consume the ARG in `make all ARGS="${BUILD_ARGS}"`, and all package/flag tokens are valid per `build.sh`.

Main risk: the PR removes the WebUI (`vue-cli-service`) workaround `RUN` steps while still building `-p web`, but neither Docker image was actually built in the test plan. If the workaround was required, the web build will fail at image build time. Recommend an end-to-end image build (or confirmation from #1005) before merge.

### Findings: 5
- Critical: 0
- Major: 1 (unverified WebUI build after removing patch steps)
- Minor: 1 (redundant `--skip-*` flags)
- Nit: 3 (stale success log message; pre-existing UFS-set divergence; verified-correct note on `build.sh`)

szbr00 (approved):

PR #1013, opened 2026-07-08
fix(master): avoid TTL fs_dir read-lock reentry (#1011)

### LINKED ISSUE #1011: [BUG]: Master metadata operations can stall behind the global fs_dir RwLock under file-operation storms
**Describe the bug**

The current master-side isolation work in PR #998 / PR #1005 protects the RPC runtime from an unbounded load-job submission storm, but it does not protect the master control plane from other metadata-operation storms.

The reason is that all master metadata operations eventually converge on the same global `fs_dir` lock:

- `fs_dir` is `ArcRwLock<FsDir>`.
- `ArcRwLock` is backed by `std::sync::RwLock` in `orpc/src/sync/lock.rs`.
- `FileStatus`, `ListStatus`, `GetBlockLocations`, mkdir/create/delete/rename/set_attr/free, block report reconciliation, background TTL cleanup, and related master metadata paths all acquire this same lock directly or indirectly.

Executor/runtime isolation can isolate entry execution resources, but it cannot isolate a shared global lock. Under a file-operation storm, the global `fs_dir` lock becomes the control-plane bottleneck. If any background task or request path holds or recursively reacquires the lock incorrectly, the active master can become Ready at the pod level but stop replying to normal metadata RPCs.

A concrete production failure mode was observed around inode TTL cleanup:

1. The master started TTL cleanup and logged that expired buckets were found.
2. No later `inode ttl cleanup completed` or `inode ttl cleanup failed` log appeared.
3. The active master still accepted TCP connections, but simple metadata commands such as `cv fs --cache-only stat /` timed out.
4. Follower masters still returned `Not a leader` quickly, proving the client/RPC stack was not globally broken.
5. The active master audit log was dominated by StarRocks-style `GetBlockLocations` requests, but CPU was low and worker threads were mostly sleeping, consistent with lock waiting rather than CPU saturation.

The specific deadlock pattern was:

- TTL path resolution held `fs_dir.read()`.
- The old recursive path resolver attempted to acquire `fs_dir.read()` again while the first read guard was still alive.
- With a pending writer from concurrent metadata operations, `std::sync::RwLock` can block new readers.
- The TTL thread then waited for a second read lock while still holding the first read lock.
- The pending writer waited for the first read lock to drop.
- Later readers such as `FileStatus`, `ListStatus`, and `GetBlockLocations` were also blocked behind the pending writer.

This turns a local TTL cleanup bug into a master-wide metadata outage because the lock is global.

**To Reproduce**

One concrete reproduction path:

1. Start a Curvine master with TTL cleanup enabled.
2. Create nested files/directories with expired TTL policies so the TTL checker has expired buckets to process.
3. Generate a concurrent metadata storm that includes writes, for example create/delete/mkdir/rename or client workloads that miss many metadata paths and trigger repeated metadata operations.
4. Wait for the TTL cleanup interval.
5. Observe that TTL cleanup starts and reports expired buckets, but does not complete.
6. Run a simple command against the active master, for example:

```bash
cv fs --cache-only stat /
```

7. The active master can time out even though the pod remains Ready.

A smaller synthetic form of the same issue is:

1. Hold `fs_dir.read()`.
2. Have another thread wait on `fs_dir.write()`.
3. While still holding the first read guard, try to acquire `fs_dir.read()` again.
4. The second read can block behind the pending writer, creating a self-deadlock.

**Expected behavior**

The master control plane should not become globally unavailable because one metadata background task or one operation path holds/re-enters the global metadata lock.

At minimum:

- No metadata path should recursively acquire `fs_dir` while already holding it.
- No path should perform blocking I/O, slow work, or long recursive traversal while holding `fs_dir`.
- Background maintenance such as TTL cleanup must always finish, fail, or yield; it must not block all metadata RPCs.
- Under file-operation storms, the master should return bounded errors/backpressure where appropriate instead of hanging unrelated metadata operations indefinitely.
- Worker heartbeat/block-report control paths should remain observable and should not be indirectly starved by unrelated metadata storms.

- Curvine deployment: Kubernetes, 3 masters, 20 workers.
- Workload: StarRocks clients issuing high-volume metadata reads, especially `GetBlockLocations` on many metadata file paths.
- Curvine branch context: current PR #1005 hotfix branch plus local TTL executor fix.
- Relevant implementation:
  - `orpc/src/sync/lock.rs`: `ArcRwLock<T>` wraps `std::sync::RwLock<T>`.
  - `curvine-server/src/master/fs/master_filesystem.rs`: master metadata APIs acquire the shared `fs_dir` read/write lock.
  - `curvine-server/src/master/meta/inode/ttl/ttl_executor.rs`: TTL cleanup resolves inode paths and then calls delete/free.
  - `curvine-server/src/master/meta/inode/ttl/ttl_manager.rs`: cleanup should log completion or failure after `cleanup_once()` returns.

**Additional context**

PR #998 / PR #1005 are still useful because they prevent load-job submission storms from unboundedly spawning work on the master RPC runtime. However, they should not be described as a complete fix for master unavailability under file-operation storms.

They solve the entry-resource problem, not the shared-metadata-lock problem.

Recommended follow-up work:

1. Short term hardening:
   - Remove all recursive/reentrant `fs_dir` acquisitions.
   - Keep `fs_dir` lock scopes small and explicit.
   - Never perform UFS I/O, network I/O, blocking waits, or heavy recursive work while holding `fs_dir`.
   - Add slow lock wait / slow lock hold logging around `fs_dir`.
   - Add metrics for `fs_dir` read/write wait time, hold time, and queue pressure.

2. Correctness checks:
   - Add regression coverage for TTL cleanup on nested expired inodes.
   - Add tests that background maintenance does not block basic metadata reads indefinitely.
   - Audit all master background tasks for `fs_dir` lock ordering and lock nesting.

3. Medium/long term scalability:
   - Consider splitting the global metadata lock by inode/subtree or separating read snapshots from write mutation.
   - Consider an RCU/snapshot-style read path for hot metadata reads such as `GetBlockLocations` and `FileStatus`.
   - Consider a single-writer metadata state machine plus lock-free or snapshot-based readers, if that better matches Curvine master semantics.

Acceptance criteria:

- Active master remains responsive to basic metadata commands during a metadata storm.
- TTL cleanup always logs completion or failure for each execution.
- No known master background task can self-deadlock on `fs_dir`.
- Lock wait/hold metrics make future metadata lock contention visible before it becomes an outage.
- PR documentation clearly distinguishes RPC-runtime isolation from metadata-lock scalability.

jlon:
# Summary

Fixes a master-wide stall caused by TTL path resolution recursively re-acquiring the global `fs_dir` read lock. TTL now resolves an inode path with a single read-lock acquisition and then releases that lock before executing delete/free.

# Issue Describe / Design

Closes #1011.

Root cause:
- `InodeTtlExecutor::build_path_recursive` held `fs_dir.read()` while recursively resolving the parent path.
- `fs_dir` is backed by `std::sync::RwLock`; when a writer is queued, a second read acquisition by the same thread can block.
- Under metadata write pressure, TTL cleanup can self-deadlock while holding the first read guard, blocking later readers and writers behind the global metadata lock.

Reproduction shape:
1. Create a nested expired TTL inode.
2. Keep concurrent metadata writes queued.
3. Run TTL cleanup/path resolution.
4. Old code can block on recursive read-lock acquisition; the fixed code completes.

Fix approach:
- Replace recursive path building with an iterative parent-chain walk under one read guard.
- Preserve the old `FileEntry` fallback behavior.
- Keep delete/free outside the read-lock scope.

# Changes

| `curvine-server/src/master/meta/inode/ttl/ttl_executor.rs` | Resolve inode paths iteratively under one `fs_dir` read lock | Removes reentrant lock deadlock risk; TTL delete/free semantics preserved |
| `curvine-server/tests/master_fs_test.rs` | Add nested TTL delete and concurrent-writer regression coverage | Covers the production freeze shape without changing runtime behavior |

# Test verified

| `cargo test -p curvine-server --test master_fs_test ttl_executor_deletes_nested_expired_inode -- --nocapture --test-threads=1` | PASS | |
| `cargo test -p curvine-server --test master_fs_test ttl_path_resolution_no_reentrant_deadlock_under_writers -- --nocapture --test-threads=1` | PASS | |
| `cargo test -p curvine-server --test master_fs_test -- --test-threads=1` | PASS | 23 passed |
| Production validation | PASS | Active master responded to repeated `cv fs --cache-only stat /`; master/worker pods Ready; no recent critical ERROR/WARN/panic matched |

# Dependencies

- Related PRs: #1005
- Related issues: #1011
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
+++ b/curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
@@ -72,7 +72,13 @@ impl InodeTtlExecutor {
 
             let inode_view = match fs_dir.store.get_inode(current_id, None)? {
                 Some(inode_view) => inode_view,
-                None => return err_box!("Cannot resolve path for inode {}", inode_id),
+                None => {
+                    return err_box!(
+                        "Cannot resolve path for inode {} (missing ancestor {})",
+                        inode_id,
+                        current_id
+                    );
+                }
             };
 
             match &inode_view {
@@ -90,10 +96,6 @@ impl InodeTtlExecutor {
                     break;
                 }
             }
-        }
-
-        if components.is_empty() {
-            return Ok("/".to_string());
         }
 
         components.reverse();
--- a/curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
+++ b/curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
@@ -14,6 +14,7 @@
 
 use crate::master::fs::MasterFilesystem;
 use crate::master::meta::inode::{Inode, InodeView, ROOT_INODE_ID};
+use crate::master::meta::FsDir;
 use curvine_common::state::TtlAction;
 use curvine_common::FsResult;
 use log::debug;
@@ -49,44 +50,54 @@ impl InodeTtlExecutor {
     }
 
     fn resolve_inode_path(&self, inode_id: i64) -> FsResult<String> {
-        self.build_path_recursive(inode_id)
+        let fs_dir = self.filesystem.fs_dir();
+        let fs_dir_guard = fs_dir.read();
+        Self::build_path_from_store(&fs_dir_guard, inode_id)
     }
 
-    fn build_path_recursive(&self, inode_id: i64) -> FsResult<String> {
+    fn build_path_from_store(fs_dir: &FsDir, inode_id: i64) -> FsResult<String> {
         if inode_id == ROOT_INODE_ID {
             return Ok("/".to_string());
         }
 
-        let fs_dir = self.filesystem.fs_dir();
-        let fs_dir_guard = fs_dir.read();
-        if let Ok(Some(inode_view)) = fs_dir_guard.store.get_inode(inode_id, None) {
+        let mut current_id = inode_id;
+        let mut components = Vec::new();
+        let mut visited = Vec::new();
+
+        while current_id != ROOT_INODE_ID {
+            if visited.contains(&current_id) {
+                return err_box!("Cycle detected while resolving inode path {}", inode_id);
+            }
+            visited.push(current_id);
+
+            let inode_view = match fs_dir.store.get_inode(current_id, None)? {
+                Some(inode_view) => inode_view,
+                None => return err_box!("Cannot resolve path for inode {}", inode_id),
+            };
+
             match &inode_view {
                 InodeView::File(f) => {
-                    let parent_path = self.build_path_recursive(f.parent_id())?;
-                    let file_path = if parent_path == "/" {
-                        format!("/{}", f.name)
-                    } else {
-                        format!("{}/{}", parent_path, f.name)
-                    };
-                    return Ok(file_path);
+                    components.push(f.name.clone());
+                    current_id = f.parent_id();
                 }
                 InodeView::Dir(d) => {
-                    let parent_path = self.build_path_recursive(d.parent_id())?;
-                    let dir_path = if parent_path == "/" {
-                        format!("/{}", d.name)
-                    } else {
-                        format!("{}/{}", parent_path, d.name)
-                    };
-                    return Ok(dir_path);
+                    components.push(d.name.clone());
+                    current_id = d.parent_id();
                 }
                 InodeView::FileEntry(e) => {
-                    // For empty files, we can't determine parent_id, so return a basic path
-                    return Ok(format!("/{}", e.name));
+                    // FileEntry does not carry parent_id, so preserve the previous fallback.
+                    components.push(e.name.clone());
+                    break;
                 }
             }
         }
 
-        err_box!("Cannot resolve path for inode {}", inode_id)
+        if components.is_empty() {
+            return Ok("/".to_string());
+        }
+
+        components.reverse();
+        Ok(format!("/{}", components.join("/")))
     }
 
     pub fn get_inode_from_store(&self, inode_id: i64) -> FsResult<Option<InodeView>> {
--- a/curvine-server/tests/master_fs_test.rs
+++ b/curvine-server/tests/master_fs_test.rs
@@ -22,12 +22,14 @@ use curvine_common::proto::{
 use curvine_common::raft::storage::{AppStorage, ApplyMsg};
 use curvine_common::state::MountOptions;
 use curvine_common::state::{
-    BlockLocation, ClientAddress, CommitBlock, CreateFileOpts, FileAllocOpts, WorkerInfo,
+    BlockLocation, ClientAddress, CommitBlock, CreateFileOpts, CreateFileOptsBuilder,
+    FileAllocOpts, TtlAction, WorkerInfo,
 };
 use curvine_common::state::{OpenFlags, RenameFlags, SetAttrOptsBuilder};
 use curvine_common::utils::SerdeUtils;
 use curvine_server::master::fs::{FsRetryCache, MasterFilesystem, OperationStatus};
 use curvine_server::master::journal::{JournalBatch, JournalEntry, JournalLoader, JournalSystem};
+use curvine_server::master::meta::inode::ttl::InodeTtlExecutor;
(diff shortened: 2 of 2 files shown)


curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
```diff
@@ -49,44 +50,54 @@ impl InodeTtlExecutor {
...
+
+        while current_id != ROOT_INODE_ID {
+            if visited.contains(&current_id) {
+                return err_box!("Cycle detected while resolving inode path {}", inode_id);
+            }
+            visited.push(current_id);
+
+            let inode_view = match fs_dir.store.get_inode(current_id, None)? {
+                Some(inode_view) => inode_view,
+                None => return err_box!("Cannot resolve path for inode {}", inode_id),
```
szbr00: **Nit:** This error interpolates the original `inode_id`, but the node that actually failed to resolve is the current ancestor `current_id`. When a mid-chain parent link is broken, the message points at the wrong id. Consider including both, e.g. `"Cannot resolve path for inode {} (missing ancestor {})", inode_id, current_id`.

curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
```diff
@@ -49,44 +50,54 @@ impl InodeTtlExecutor {
...
-                    return Ok(format!("/{}", e.name));
+                    // FileEntry does not carry parent_id, so preserve the previous fallback.
+                    components.push(e.name.clone());
+                    break;
                 }
             }
         }
 
-        err_box!("Cannot resolve path for inode {}", inode_id)
+        if components.is_empty() {
```
szbr00: **Nit:** This branch looks unreachable — the early `inode_id == ROOT_INODE_ID` return plus the fact that every loop iteration pushes a component means `components` is never empty here. Harmless defensive code; consider dropping it or adding a comment noting it's a guard.

szbr00 (commented):
## Code Review

Reviewed 2 files, +135 / −23. Well-targeted concurrency fix with strong regression coverage.

### Summary
The fix is correct. The old `build_path_recursive` re-acquired the writer-preferring `std::sync::RwLock` read lock on each recursion level, which self-deadlocks once a writer is queued. The new `build_path_from_store` walks the parent chain iteratively under a single read guard passed by reference, and delete/free run after the guard is dropped — no reentrancy remains. I verified path semantics are identical across `File`/`Dir`/`FileEntry`/root, the `&guard → &FsDir` coercion and import are valid, and the removed method had no other callers. Cycle detection is a nice defensive addition, and the regression test reproduces the real freeze shape (deep path + continuous writers + 10s timeout) so it would deadlock on the old code.

### Findings: 3 (no Critical/Major/Minor)
- Critical: 0
- Major: 0
- Minor: 0
- Nit: 3 (imprecise error id on broken ancestor; unreachable `components.is_empty()` guard; pre-existing by-path delete TOCTOU, not a regression)

Overall this looks good to merge. See inline comments for the nits.

curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
jlon: Thanks. Fixed in df709b76: the missing-inode error now reports both the requested inode id and the missing ancestor id.

curvine-server/src/master/meta/inode/ttl/ttl_executor.rs
jlon: Thanks. Fixed in df709b76: removed the unreachable empty-components fallback after the non-root path walk.

jlon:
Thanks for the review. I also noted the by-path delete TOCTOU concern from the summary. I am keeping it out of this split PR because it is pre-existing behavior and broader than the TTL fs_dir read-lock reentry fix; it should be handled separately if we want to change delete/free semantics.

szbr00 (approved):

PR #1014, opened 2026-07-08
feat(master): add fs_dir lock stall watchdog (#1011)

jlon:
# Summary

Add a master-side watchdog for the global `fs_dir` metadata lock. The watchdog probes the lock with non-blocking `try_read()` and exports metrics/logs when the lock stays unavailable beyond the stall threshold.

# Issue Describe / Design

Related to #1011.

Root cause: production file-operation storms can make the master appear dead when all metadata RPCs block behind the single global `fs_dir` lock. Before this change, that condition had no direct master metric or log, so operators only saw secondary symptoms such as heartbeat loss or RPC timeouts.

Fix approach: start a lightweight scheduled task with `MasterActor`. It performs a non-blocking read-lock probe once per second, records probe latency, reports a stall only after 15 seconds, and clears the signal after the lock becomes available again. It never blocks the scheduler, never takes a write lock, and never restarts or aborts the process.

# Changes

| `curvine-server/src/master/fs/fs_dir_watchdog.rs` | Add `FsDirWatchdog` loop task using `try_read()` | Adds observability only |
| `curvine-server/src/master/fs/master_actor.rs` | Start the watchdog with the existing scheduled executor | Adds one lightweight background task |
| `curvine-server/src/master/fs/mod.rs` | Register watchdog module | None |
| `curvine-server/src/master/master_metrics.rs` | Add `fs_dir_stalled`, `fs_dir_stall_total`, and `fs_dir_probe_acquire_us` metrics | Adds metrics only |

# Test verified

# Dependencies

- Related PRs: None
- Related issues: #1011
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/fs/fs_dir_watchdog.rs
+++ b/curvine-server/src/master/fs/fs_dir_watchdog.rs
@@ -37,6 +37,7 @@ pub struct FsDirWatchdog {
     stall_threshold_ms: i64,
     first_unavailable_ms: AtomicI64,
     stall_reported: AtomicBool,
+    poison_reported: AtomicBool,
 }
 
 impl FsDirWatchdog {
@@ -47,6 +48,16 @@ impl FsDirWatchdog {
             stall_threshold_ms,
             first_unavailable_ms: AtomicI64::new(0),
             stall_reported: AtomicBool::new(false),
+            poison_reported: AtomicBool::new(false),
+        }
+    }
+
+    fn reset_unavailable_state(&self) {
+        self.first_unavailable_ms.store(0, Ordering::SeqCst);
+        self.stall_reported.store(false, Ordering::SeqCst);
+        self.poison_reported.store(false, Ordering::SeqCst);
+        if let Some(metrics) = self.metrics {
+            metrics.fs_dir_stalled.set(0);
         }
     }
 
@@ -54,6 +65,7 @@ impl FsDirWatchdog {
         if let Some(metrics) = self.metrics {
             metrics.fs_dir_probe_acquire_us.set(probe_us);
         }
+        self.poison_reported.store(false, Ordering::SeqCst);
         if self.stall_reported.swap(false, Ordering::SeqCst) {
             let stalled_ms =
                 LocalTime::mills() as i64 - self.first_unavailable_ms.load(Ordering::SeqCst);
@@ -68,7 +80,11 @@ impl FsDirWatchdog {
         self.first_unavailable_ms.store(0, Ordering::SeqCst);
     }
 
-    fn on_unavailable(&self, poisoned: bool) {
+    fn on_unavailable(&self, probe_us: i64, poisoned: bool) {
+        if let Some(metrics) = self.metrics {
+            metrics.fs_dir_probe_acquire_us.set(probe_us);
+        }
+
         let now = LocalTime::mills() as i64;
         let first = self.first_unavailable_ms.load(Ordering::SeqCst);
         let first = if first == 0 {
@@ -79,7 +95,7 @@ impl FsDirWatchdog {
         };
 
         let unavailable_ms = now - first;
-        if poisoned {
+        if poisoned && !self.poison_reported.swap(true, Ordering::SeqCst) {
             error!("fs_dir metadata lock is poisoned; a holder panicked while mutating metadata");
         }
         if unavailable_ms >= self.stall_threshold_ms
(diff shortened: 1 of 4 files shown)


curvine-server/src/master/fs/fs_dir_watchdog.rs
```diff
@@ -0,0 +1,118 @@
...
+                metrics.fs_dir_stall_total.inc();
+            }
+        }
+    }
+}
+
+impl LoopTask for FsDirWatchdog {
+    type Error = FsError;
+
+    fn run(&self) -> FsResult<()> {
```
szbr00, line 120: **Major (alert fidelity):** `run()` probes unconditionally, so the watchdog also runs on **follower/standby masters** (`MasterActor.start()` runs on all roles — the sibling `HeartbeatChecker` relies on a per-tick guard rather than not being started). A follower doing heavy journal replay can hold `fs_dir` write locks long enough to cross the 15s threshold, which then emits `"control plane metadata RPCs are likely blocked on the global lock"` and sets `fs_dir_stalled=1` on a node that serves no metadata RPCs — a false alert that undermines the feature's purpose.

`HeartbeatChecker::run()` guards this exact way:

```rust
fn run(&self) -> FsResult<()> {
    if !self.monitor.is_active() {
        return Ok(());
    }
    ...
}
```

Suggestion: gate the probe/report on `self.fs.master_monitor.is_active()` and reset the stall state while inactive, or document why follower monitoring is intended and drop the leader-specific wording.

curvine-server/src/master/fs/fs_dir_watchdog.rs
```diff
@@ -0,0 +1,118 @@
...
+        // Non-blocking shared probe. It never blocks the watchdog thread.
+        match self.fs.fs_dir().try_read() {
+            Ok(_guard) => self.on_available(spent.used_us() as i64),
+            Err(TryLockError::WouldBlock) => self.on_unavailable(false),
+            Err(TryLockError::Poisoned(_)) => self.on_unavailable(true),
+        }
+        Ok(())
+    }
+
+    fn terminate(&self) -> bool {
```
szbr00, line 136: **Minor:** `terminate()` always returns `false`, so the watchdog thread never exits on graceful master stop, unlike the sibling `HeartbeatChecker` which returns `self.monitor.is_stop()`. This leaks the thread on in-process shutdown/restart (tests, mini-cluster). Suggest `self.fs.master_monitor.is_stop()` for a clean shutdown and consistency.

curvine-server/src/master/fs/fs_dir_watchdog.rs
```diff
@@ -0,0 +1,118 @@
...
+        let first = if first == 0 {
+            self.first_unavailable_ms.store(now, Ordering::SeqCst);
+            now
+        } else {
+            first
+        };
+
+        let unavailable_ms = now - first;
+        if poisoned {
+            error!("fs_dir metadata lock is poisoned; a holder panicked while mutating metadata");
```
szbr00, line 99: **Minor (log spam):** this `error!` is outside the `stall_reported` gate, so a poisoned lock — which is a permanent state — logs this error on **every tick, once per second, indefinitely**. Consider rate-limiting it the same way as the stall warning (e.g. behind a `poison_reported` flag / once per episode).

curvine-server/src/master/fs/fs_dir_watchdog.rs
```diff
@@ -0,0 +1,118 @@
...
+            metrics: Master::get_metrics().ok(),
+            stall_threshold_ms,
+            first_unavailable_ms: AtomicI64::new(0),
+            stall_reported: AtomicBool::new(false),
+        }
+    }
+
+    fn on_available(&self, probe_us: i64) {
+        if let Some(metrics) = self.metrics {
+            metrics.fs_dir_probe_acquire_us.set(probe_us);
```
szbr00, line 66: **Nit:** `fs_dir_probe_acquire_us` is only updated on the success path, so during a stall the latency gauge holds a stale "healthy" value. Optional: also set it on the unavailable path (probe duration or a sentinel), or document that it's only meaningful alongside `fs_dir_stalled`.

szbr00 (commented):
## Code Review

Reviewed 4 files, +160 / −0. Observability-only `fs_dir` lock stall watchdog.

### Summary
The core mechanism is sound and safe: `try_read()` is non-blocking, the guard is dropped immediately (never holds the lock or blocks writers), `ArcRwLock` derefs to `std::sync::RwLock` so the `TryLockError` matching is valid, and the single-threaded episode state machine (arm on first miss → report once at threshold → clear on recovery) is correct. `ScheduledExecutor::start` consumes `self`, so there's no early-drop bug, and metrics wiring is fine.

My main concern is where/when it reports: the watchdog has no `is_active()` gate, so it also probes and reports on follower/standby masters, where a heavy journal-replay write burst can trip the 15s stall and raise a misleading "control plane blocked" alert (`fs_dir_stalled=1`) on a node that serves no metadata RPCs. The sibling `HeartbeatChecker` guards this with `is_active()`. Two smaller issues: `terminate()` never returns true (thread not stopped on shutdown, unlike `HeartbeatChecker`'s `is_stop()`), and the poisoned-lock `error!` logs once per second forever.

### Findings: 4 (no Critical)
- Critical: 0
- Major: 1 (no `is_active()` gate → false stall alerts on followers)
- Minor: 2 (`terminate()` never stops the thread; poisoned-lock error logs every tick)
- Nit: 1 (`fs_dir_probe_acquire_us` stale during a stall)

Observability-only — nothing risks data/correctness. Finding #1 is worth resolving before merge for alert fidelity. See inline comments.

curvine-server/src/master/fs/fs_dir_watchdog.rs
jlon: Fixed in c38b20f5. The watchdog now returns early when the master is inactive and resets the unavailable/stalled/poison episode state (including fs_dir_stalled) while inactive. I also added a unit test that holds the fs_dir write lock and verifies inactive masters reset without probing/reporting.

curvine-server/src/master/fs/fs_dir_watchdog.rs
jlon: Fixed in c38b20f5. terminate() now follows the master monitor stop state via self.fs.master_monitor.is_stop(), matching the HeartbeatChecker shutdown behavior. Covered by a unit test.

curvine-server/src/master/fs/fs_dir_watchdog.rs
jlon: Fixed in c38b20f5. Poisoned-lock logging is now guarded by a poison_reported episode flag and resets on recovery or while inactive, so a permanently poisoned lock will not emit the same error every tick.

curvine-server/src/master/fs/fs_dir_watchdog.rs
jlon: Fixed in c38b20f5. The unavailable paths now update fs_dir_probe_acquire_us with the probe duration as well, so the gauge no longer keeps the last healthy success value during a stall.

szbr00 (approved):

PR #1015, opened 2026-07-08
fix(orpc): return async handler errors as responses

jlon:
# Summary

Return async `MessageHandler` failures as normal RPC error responses instead of turning them into stream-level errors.

# Issue Describe / Design

Root cause: `StreamHandler::call` already converts sync handler errors into response messages, but the async path returned `Err(...)` directly. That closed or failed the stream and prevented the caller from receiving the standard RPC error payload.

Reproduction steps: install an async handler that returns an error and call `StreamHandler::call`. Before this fix, `call` returned an error and sent no response. After this fix, `call` succeeds at the transport level and sends one failed response message.

Fix approach: build the response header before moving the request into `async_handle`, then encode async handler errors with `error_ext`, matching the sync path.

# Changes

| `orpc/src/handler/stream_handler.rs` | Convert async handler errors to RPC error responses | Async business errors reach clients as structured responses instead of closing the stream |
| `orpc/tests/stream_handler_test.rs` | Add regression coverage for async handler errors | Test-only |

# Test verified

| `cargo test -p orpc --test stream_handler_test` | PASS | |
| `cargo clippy -p orpc --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/orpc/src/handler/stream_handler.rs
+++ b/orpc/src/handler/stream_handler.rs
@@ -12,10 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::err_box;
 use crate::handler::{Frame, MessageHandler};
 use crate::io::IOResult;
-use crate::message::Message;
+use crate::message::{Builder, Message};
 use crate::runtime::{RpcRuntime, Runtime};
 use crate::server::ServerConf;
 use crate::sys::RawPtr;
@@ -86,9 +85,19 @@ impl<F: Frame, M: MessageHandler> StreamHandler<F, M> {
                 })
                 .await?
         } else {
+            let error_response_base = Builder::new()
+                .code(request.code())
+                .request(request.request_status())
+                .req_id(request.req_id())
+                .seq_id(request.seq_id())
+                .build();
+            let req_id = request.req_id();
             match self.handler.as_mut().async_handle(request).await {
                 Ok(v) => v,
-                Err(e) => return err_box!("{}", e),
+                Err(e) => {
+                    debug!("handler request {} error: {}", req_id, e);
+                    error_response_base.error_ext(&e)
+                }
             }
         };
 
--- a/orpc/src/handler/stream_handler.rs
+++ b/orpc/src/handler/stream_handler.rs
@@ -85,17 +85,20 @@ impl<F: Frame, M: MessageHandler> StreamHandler<F, M> {
                 })
                 .await?
         } else {
-            let error_response_base = Builder::new()
-                .code(request.code())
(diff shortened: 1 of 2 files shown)


orpc/src/handler/stream_handler.rs
```diff
@@ -86,9 +85,19 @@ impl<F: Frame, M: MessageHandler> StreamHandler<F, M> {
                 })
                 .await?
         } else {
+            let error_response_base = Builder::new()
```
szbr00: **Nit:** `error_response_base` is constructed unconditionally on every async request but consumed only in the `Err` arm; on the success path it's built and dropped. The cost is negligible (`Builder::build()` is a small stack struct with `header: None` and empty data — no heap allocation), so this is pure micro-optimization.

Optional: capture the four primitive fields (`code`/`request_status`/`req_id`/`seq_id`) as locals before moving `request` into `async_handle`, and construct the error message only inside the `Err` arm.

szbr00 (commented):
## Code Review

Reviewed 2 files, +94 / −3. Aligns the async `MessageHandler` error path with the existing sync path.

### Summary
The fix is correct, minimal, and well-tested. Because `async_handle(request)` moves `request`, the async path pre-builds `error_response_base` from `request`'s fields and calls `error_ext(&e)` on it. I traced `error_ext` → `response`: the error message's routing is derived solely from `{code, request_status, req_id, seq_id}` (the complete set of `Protocol` fields; header/data of `self` are ignored, data is replaced with `err.encode()`, status becomes `Error`), and the Builder copies exactly those four fields from `request` — so the result is equivalent to `request.error_ext(&e)`. The behavior change is the intended one: instead of returning `Err` (which propagated through `run()` and closed the stream with no response), it now sends one `Error` response and returns `Ok`, matching the sync path. Logging is consistent, and the new test asserts exactly one non-success response carrying the decoded error.

### Findings: 1 (no Critical/Major/Minor)
- Critical: 0
- Major: 0
- Minor: 0
- Nit: 1 (`error_response_base` built unconditionally but used only on error — negligible, no heap alloc)

Effectively LGTM. See the inline comment for the optional nit.

orpc/src/handler/stream_handler.rs
jlon: Thanks. Fixed in 01551973: async_handle now captures the routing fields first and builds the error response base only inside the Err arm.

szbr00 (approved):

PR #1016, opened 2026-07-08
fix(common): add resource exhausted fs error

jlon:
# Summary

Add a structured `FsError::ResourceExhausted` error kind for server/client backpressure and resource-limit failures.

# Issue Describe / Design

Root cause: queue-full and resource-pressure paths need a stable error kind so callers can retry or throttle without parsing error strings. `FsError` did not have such a kind, so these failures would have to fall back to `Common` or ad-hoc text matching.

Fix approach: add `ErrorKind::ResourceExhausted`, `FsError::ResourceExhausted`, a helper constructor, and encode/decode/ctx mapping. This PR only adds the common error type; it does not change any master or client behavior by itself.

# Changes

| `curvine-common/src/error/fs_error.rs` | Add structured `ResourceExhausted` kind and mapping | Existing errors are unchanged |
| `curvine-common/tests/fs_error_test.rs` | Add round-trip encoding test | Test-only |

# Test verified

| `cargo test -p curvine-common --test fs_error_test` | PASS | |
| `cargo clippy -p curvine-common --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-common/src/error/fs_error.rs
+++ b/curvine-common/src/error/fs_error.rs
@@ -64,6 +64,7 @@ pub enum ErrorKind {
     NotADirectory = 27,
     InvalidArgument = 28,
     BlockNotFound = 29,
+    ResourceExhausted = 30,
 
     #[num_enum(default)]
     Common = 10000,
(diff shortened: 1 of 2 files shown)


jlon:
Superseded by #1026, which keeps the structured ResourceExhausted error together with the complete client auto-cache backpressure flow.

PR #1018, opened 2026-07-08
feat(common): add auto-cache submit backpressure config

jlon:
# Summary

Add client-side auto-cache submit backpressure configuration fields and validation.

# Issue Describe / Design

Root cause: client-side retry/queue behavior for auto-cache submit needs explicit local capacity and retry timing. Without named config fields, the client implementation would have to hard-code queue size and backoff policy.

Fix approach: add `ClientConf` fields for local submit queue size, retry count, retry min sleep, and retry max sleep. Defaults keep existing client startup compatible, and `init()` rejects impossible values such as zero queue size or max sleep lower than min sleep.

This PR only adds config. It does not change auto-cache submit behavior by itself.

# Changes

| `curvine-common/src/conf/client_conf.rs` | Add auto-cache submit queue/retry config with defaults and validation | No behavior change until client code consumes these fields |
| `curvine-common/tests/client_conf_cli_test.rs` | Add validation coverage | Test-only |

# Test verified

| `cargo test -p curvine-common --test client_conf_cli_test validates_auto_cache_submit_backpressure_conf` | PASS | |
| `cargo clippy -p curvine-common --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-common/src/conf/client_conf.rs
+++ b/curvine-common/src/conf/client_conf.rs
@@ -18,7 +18,7 @@ use curvine_common_macros::ClientCliArgs;
 use orpc::client::ClientConf as RpcConf;
 use orpc::common::{ByteUnit, DurationUnit, Utils};
 use orpc::io::net::InetAddr;
-use orpc::CommonResult;
+use orpc::{err_box, CommonResult};
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


jlon:
Superseded by #1026, which keeps the auto-cache submit configuration together with the complete client queue/retry/source-validation flow.

PR #1020, opened 2026-07-08
fix(master): separate load and export path direction

jlon:
# Summary

Separate client load/import path resolution from fs_mode journal export path resolution.

# Issue Describe / Design

Root cause: `LoadJobRunner` used one `build_job_context` path toggle for both user load requests and fs_mode journal replay. That is wrong for CV paths whose metadata is `ufs_only`: a user load for that CV path should read the mounted UFS source and populate Curvine, while journal replay must export committed Curvine data back to UFS.

Reproduction steps: create a mounted CV path with UFS-only metadata, then submit a load command using the CV path. Before this fix, the job id and source path are derived from the CV path and the runner treats the operation like a CV-to-UFS export. After this fix, the job id/source are derived from the mounted UFS path and the target stays the CV path. A separate export regression covers journal replay semantics.

Fix approach: make the direction explicit with two modes: `LoadUfsOnlyFromUfs` for normal load/import and `ExportCurvineToUfs` for fs_mode journal replay. `UfsLoader` now calls `submit_export_task` so replay is not affected by the load/import UFS-only fallback.

# Changes

| `curvine-server/src/master/job/job_runner.rs` | Add explicit load/export CV source modes and `submit_export_task` | CV UFS-only load uses UFS source; journal export remains CV source |
| `curvine-server/src/master/journal/ufs_loader.rs` | Call `submit_export_task` for fs_mode replay | Preserves durable CV-to-UFS replay semantics |
| `curvine-server/tests/load_job_submit_test.rs` | Add regression tests for UFS-only load and journal export direction | Test-only |

# Test verified

| `cargo test -p curvine-server --test load_job_submit_test fs_mode_cv_path_load_uses_ufs_source_when_metadata_is_ufs_only` | PASS | |
| `cargo test -p curvine-server --test load_job_submit_test direct_export_task_keeps_cv_source_for_fs_mode_journal_sync` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/job/job_runner.rs
+++ b/curvine-server/src/master/job/job_runner.rs
@@ -97,26 +97,31 @@ impl LoadJobRunner {
         cv_source_mode: CvSourceMode,
     ) -> FsResult<(JobContext, Path, Path)> {
         let command_source = Path::from_str(&command.source_path)?;
-        let load_cv_from_ufs = if !command_source.is_cv() {
-            false
-        } else {
-            match cv_source_mode {
-                CvSourceMode::LoadUfsOnlyFromUfs => {
-                    match self.master_fs.file_status(command_source.path()) {
-                        Ok(status) => status.storage_policy.ufs_only(),
-                        Err(FsError::FileNotFound(_)) => false,
-                        Err(err) => return Err(err),
-                    }
-                }
-                CvSourceMode::ExportCurvineToUfs => false,
+        let (source_path, target_path) = match (cv_source_mode, command_source.is_cv()) {
+            (CvSourceMode::LoadUfsOnlyFromUfs, false) => {
+                let target_path = mnt.get_cv_path(&command_source)?;
+                (command_source, target_path)
             }
-        };
-
-        let (source_path, target_path) = if load_cv_from_ufs {
-            (mnt.get_ufs_path(&command_source)?, command_source)
-        } else {
-            let target_path = mnt.toggle_path(&command_source)?;
-            (command_source, target_path)
+            (CvSourceMode::LoadUfsOnlyFromUfs, true) => {
+                let status = self.master_fs.file_status(command_source.path())?;
+                if !status.storage_policy.ufs_only() {
+                    return err_box!(
+                        "load job for CV path {} requires UFS-only metadata",
+                        command_source.full_path()
+                    );
+                }
+                (mnt.get_ufs_path(&command_source)?, command_source)
+            }
+            (CvSourceMode::ExportCurvineToUfs, true) => {
+                let target_path = mnt.get_ufs_path(&command_source)?;
+                (command_source, target_path)
+            }
+            (CvSourceMode::ExportCurvineToUfs, false) => {
+                return err_box!(
+                    "export job source path {} must be a CV path",
+                    command_source.full_path()
+                );
+            }
         };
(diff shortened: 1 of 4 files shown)


curvine-server/src/master/job/job_runner.rs
```diff
@@ -219,7 +247,32 @@ impl LoadJobRunner {
         command: LoadJobCommand,
         mnt: MountInfo,
     ) -> FsResult<LoadJobResult> {
-        let (mut job_context, source_path, target_path) = self.build_job_context(&command, &mnt)?;
+        self.submit_direct_task(command, mnt, CvSourceMode::LoadUfsOnlyFromUfs, "load")
```
szbr00, line 255: @/tmp/pr1020_f1.md

curvine-server/src/master/job/job_runner.rs
```diff
@@ -88,9 +94,30 @@ impl LoadJobRunner {
...
+                    }
+                }
+                CvSourceMode::ExportCurvineToUfs => false,
+            }
+        };
+
+        let (source_path, target_path) = if load_cv_from_ufs {
+            (mnt.get_ufs_path(&command_source)?, command_source)
+        } else {
+            let target_path = mnt.toggle_path(&command_source)?;
```
szbr00: @/tmp/pr1020_f2.md

curvine-server/tests/load_job_submit_test.rs
```diff
@@ -128,6 +142,89 @@ fn submit_load_job_returns_before_ufs_planning() -> CommonResult<()> {
...
+    let running = JobContext::with_conf(
+        &command,
+        expected_job_id.clone(),
+        cv_path.to_string(),
+        expected_target.clone(),
+        &mount.info,
+        &ClientConf::default(),
+        1,
+    );
+    job_manager.jobs().insert(expected_job_id.clone(), running);
```
szbr00: @/tmp/pr1020_f3.md

curvine-server/src/master/job/job_runner.rs
jlon: Thanks. Fixed in 8438475d: public load jobs now only import UFS into Curvine; ordinary CV paths no longer fall back to CV-to-UFS export, which is reserved for submit_export_task.

curvine-server/src/master/job/job_runner.rs
jlon: Thanks. Fixed in 8438475d: build_job_context now selects direction explicitly. UFS sources map to CV targets, UFS-only CV metadata maps to its UFS source, and export mode requires a CV source.

curvine-server/tests/load_job_submit_test.rs
jlon: Thanks. Fixed in 8438475d: the export regression test no longer pre-installs a running job. It now exercises submit_export_task directly via an empty CV directory, and I added a regression test that ordinary CV load is rejected instead of exported.

jlon:
Verification update for 8438475d: `cargo fmt --check` and `git diff --check` passed locally. I attempted the targeted `load_job_submit_test` cases with heartbeat output, but this environment killed the compile with SIGKILL (exit 137) while building dependencies around `librocksdb-sys`, so I do not have a completed local cargo test result for this push.

szbr00 (approved):

PR #1021, opened 2026-07-08
feat(orpc): detect ArcRwLock reentry in debug builds (#1011)

jlon:
# Summary

Add debug-build reentrant-acquire detection for `ArcRwLock`.

# Issue Describe / Design

Related to #1011.

Root cause: `std::sync::RwLock` can deadlock when a thread that already holds a guard re-acquires the same lock after a writer is queued. This is exactly the kind of mistake that can turn the master metadata lock into a silent control-plane stall during tests or debug builds.

Fix approach: assign each `ArcRwLock` a debug-only id and track currently held lock ids in thread-local state. `read()` and `write()` panic in debug/test builds if the current thread already holds a guard for the same lock. Release builds do not maintain this state. The wrapper guards implement `Deref`/`DerefMut`, and `curvine-server` clippy verifies existing master call sites still compile.

# Changes

| `orpc/src/sync/lock.rs` | Add debug-only reentry tracking and wrapper guards for `ArcRwLock` | Release behavior unchanged; debug/test builds fail fast on same-thread reentry |
| `orpc/tests/arc_rw_lock_test.rs` | Add reentry and reacquire-after-drop coverage | Test-only |

# Test verified

| `cargo test -p orpc --test arc_rw_lock_test` | PASS | |
| `cargo clippy -p orpc --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | |
| `cargo clippy -p curvine-server --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | verifies master callers compile with wrapper guards |

# Dependencies

- Related PRs: None
- Related issues: #1011
- Blocks / blocked by: None

### DIFF
--- a/orpc/src/sync/lock.rs
+++ b/orpc/src/sync/lock.rs
@@ -52,7 +52,7 @@ mod reentry_guard {
     }
 
     pub(super) fn unmark_held(id: u64) {
-        HELD_LOCKS.with(|held| {
+        let _ = HELD_LOCKS.try_with(|held| {
             let mut held = held.borrow_mut();
             if let Some(pos) = held.iter().rposition(|x| *x == id) {
                 held.remove(pos);
--- a/orpc/src/sync/lock.rs
+++ b/orpc/src/sync/lock.rs
@@ -12,28 +12,104 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 use std::sync::{Arc, Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
 
-pub struct ArcRwLock<T>(Arc<RwLock<T>>);
+// The reentrancy guard only exists in debug builds. `std::sync::RwLock` is
+// writer-preferring on Linux: once a writer is queued, a new `read()` blocks.
+// A thread that already holds a read guard and re-acquires the same lock can
+// deadlock. Release builds pay zero cost; debug/test builds panic loudly at the
+// reentrant acquisition instead of silently wedging the whole lock.
+#[cfg(debug_assertions)]
+mod reentry_guard {
+    use std::cell::RefCell;
+    use std::sync::atomic::{AtomicU64, Ordering};
+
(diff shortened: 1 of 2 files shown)


szbr00 (commented):
## Code Review

Reviewed 2 files, ~182 lines changed.

### Summary
Solid, well-scoped change. The debug-only reentrancy detection is correct: id tracking is shared across clones (matching the shared underlying `RwLock`), each id appears at most once in the thread-local held-set, the acquire/mark/drop-unmark ordering is sound, and there's no `RefCell` borrow-panic hazard during the reentry `panic!`. Release builds keep the plain `Arc<RwLock>` path at zero cost, and the wrapper guards' `Deref`/`DerefMut` keep all existing master call sites working.

Main things to consider: the two `#[should_panic]` tests will fail under `cargo test --release` (detection is compiled out), and `try_read`/`try_write` via `Deref` bypass tracking. Both are non-blocking for the core goal.

### Findings: 4
- Critical: 0
- Major: 0
- Minor: 2
- Nit: 2

orpc/src/sync/lock.rs
```diff
@@ -12,36 +12,160 @@
...
+                );
+            }
+        });
+    }
+
+    pub(super) fn mark_held(id: u64) {
+        HELD_LOCKS.with(|held| held.borrow_mut().push(id));
+    }
+
+    pub(super) fn unmark_held(id: u64) {
```
szbr00, line 54: `unmark_held` uses `HELD_LOCKS.with`, which can panic if the thread-local has already been destroyed. This can occur when an `ArcRwLock` guard is stored in another thread-local: `HELD_LOCKS` may be initialized later and therefore destroyed first during thread teardown. Dropping the guard would then panic, potentially aborting the process if teardown is already unwinding. Consider using `HELD_LOCKS.try_with(...)` here and treating an unavailable TLS value as already cleaned up.

jlon: Thanks. Fixed in 0afd9288: unmark_held now uses HELD_LOCKS.try_with and treats unavailable TLS during teardown as already cleaned up.

szbr00 (approved):

PR #1017, opened 2026-07-08
fix(master): bound load job admission queue (#1011)

jlon:

PR #1022, opened 2026-07-08
fix(master): fail queued load jobs fast when source is missing

jlon:
# Summary

Fail queued load jobs immediately when the source path is missing.

# Issue Describe / Design

Root cause: for missing UFS/CV sources, the queued load path could continue into cache validation and recursive planning before discovering the same `FileNotFound`. Under a load-job storm, these deterministic invalid jobs still consume the load-job execution path longer than necessary.

Reproduction steps: submit a load job for a mounted file path that does not exist. The job returns `Pending` first, then should quickly become `Failed` with a source-not-found diagnostic. This PR makes that failure happen before cache checks, directory traversal, task planning, or worker dispatch.

Fix approach: add a source-root existence check at the start of `run_queued_load_job`. `FileNotFound` is converted into a failed job state with a stable `source file not found` message and returns `Ok(())` so the background worker consumes the invalid job without noisy retry semantics. Non-FileNotFound source check errors still fail the job and return the error.

# Changes

| `curvine-server/src/master/job/job_runner.rs` | Check queued load source before cache validation/planning | Missing sources fail faster; valid source and empty-directory load semantics are unchanged |
| `curvine-server/tests/load_job_submit_test.rs` | Assert source-not-found diagnostic for missing source | Test-only |

# Test verified

| `cargo test -p curvine-server --test load_job_submit_test submit_load_job_returns_before_ufs_planning` | PASS | |
| `cargo test -p curvine-server --test load_job_submit_test empty_directory_load_completes_without_worker_reports` | PASS | verifies directories are not treated as missing files |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/job/job_runner.rs
+++ b/curvine-server/src/master/job/job_runner.rs
@@ -56,6 +56,7 @@ struct PlannedLoadJob {
 
 impl LoadJobRunner {
     const RUN_ID_SEQ_MOD: u64 = 1_000_000;
+    const SOURCE_NOT_FOUND_PREFIX: &'static str = "source file not found";
 
     pub fn new(
         jobs: JobStore,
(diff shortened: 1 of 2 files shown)


jlon:
Superseded by #1017. The missing-source fast-fail logic is now part of the complete master load-job admission fix, together with bounded queues, background runtime, ResourceExhausted, metrics, shutdown handling, and regression tests.

PR #1024, opened 2026-07-08
fix(master): reconcile stale block reports safely (#1011)

jlon:
# Summary

This PR makes master block report handling resilient to stale or invalid reported blocks and removes the foreground cost of full-report stale-location reconciliation.

# Issue Describe / Design

Related to #1011.

- Root cause: workers can report blocks whose inode no longer exists or is not a file, and completed full block reports previously forced the master to scan and remove stale worker locations in the foreground `WorkerBlockReport` RPC path.
- Fix approach: return `DeleteBlock` commands for missing/non-file reported blocks so workers can clean them, apply normal report entries in bounded chunks, and enqueue full-report stale-location reconciliation to a background lane. New finalized/writing incremental reports invalidate older full-report reconcile generations so delayed full reports cannot remove newer locations.
- Semantics: `DeleteBlock` commands are returned in the block report response. Under-replicated reporting for stale full-report locations is preserved in the background reconcile path instead of the foreground RPC path.

# Changes

| `curvine-server/src/master/fs/master_filesystem.rs` | Add block inode state classification, `DeleteBlock` result collection, background full-report reconcile, generation invalidation, and chunked fs_dir apply. | Invalid worker-reported blocks are cleaned by worker command; full-report stale cleanup no longer blocks the RPC foreground path. |
| `curvine-server/src/master/master_handler.rs` | Convert block report `delete_blocks` into worker commands and pass replication handler for background reconcile. | Worker receives cleanup commands; under-replicated reporting is preserved. |
| `curvine-server/src/master/meta/fs_dir.rs` | Distinguish missing block inode from non-file inode. | Lets master schedule worker cleanup instead of treating non-file inode as a normal file block. |
| `curvine-server/src/worker/block/block_actor.rs` | Handle delete commands returned by full block report. | Worker applies master cleanup commands during full report. |
| `curvine-server/src/worker/block/heartbeat_task.rs` | Make delete-block task callable from full-report path and handle commands returned by incremental report. | Worker applies master cleanup commands during heartbeat/incremental report. |
| `curvine-server/tests/master_fs_test.rs` | Cover non-file reported block cleanup and async full-report stale-location reconcile. | Test-only. |
| `curvine-server/tests/lock_order_deadlock_stress_test.rs` | Update direct block report calls for the new API. | Test-only. |

# Test verified

| `cargo test -p curvine-server --test master_fs_test block_report_for_non_file_inode_schedules_worker_delete -- --test-threads=1` | PASS | 1 passed. |
| `cargo test -p curvine-server --test master_fs_test full_block_report_reconcile_removes_stale_location_async -- --test-threads=1` | PASS | 1 passed. |
| `cargo test -p curvine-server --test lock_order_deadlock_stress_test -- --test-threads=1` | PASS | 2 passed. |
| `cargo test -p curvine-server --test master_fs_test -- --test-threads=1` | PASS | 23 passed. |

# Dependencies

- Related PRs: #1005, #1031
- Related issues: #1011
- Blocks / blocked by: None


curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -769,55 +811,107 @@ impl MasterFilesystem {
...
         // @todo check cluster.
+        let invalidate_full_reconcile = !list.full_report
+            && list.blocks.iter().any(|block| {
+                matches!(
+                    block.status,
+                    BlockReportStatus::Finalized | BlockReportStatus::Writing
+                )
+            });
+        if invalidate_full_reconcile {
+            self.invalidate_full_block_reconcile(list.worker_id);
```
szbr00: Invalidating only `full_block_reconciles` does not protect an in-progress full report that has not yet produced a reconcile job. For example, a full-report snapshot starts, an incremental report adds block X and calls this invalidation, and then the remaining full-report chunks complete. That completion creates a new reconcile generation whose snapshot does not contain X, so it can remove the newer location. Please also invalidate/version the accumulated `full_block_reports` session so a reconcile created after an incremental update cannot represent an older snapshot.

curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -828,31 +922,183 @@ impl MasterFilesystem {
...
+        let mut batch = Vec::new();
+        for block_id in existing_blocks {
+            if !reported_blocks.contains(&block_id) {
+                batch.push((false, block_id, BlockLocation::with_id(worker_id)));
+                stale_block_ids.push(block_id);
+            }
+        }
+
+        if !batch.is_empty() {
+            if !self.is_full_block_reconcile_current(worker_id, generation) {
```
szbr00: This generation check has a TOCTOU race with applying the removal batch. After the check succeeds, an incremental report can invalidate the generation and add a location before this stale batch acquires `fs_dir`; the stale reconcile then removes the newly added location. Chunking extends the race across every chunk. The generation validation and stale removals need serialization with incremental invalidation/application, or the metadata updates need an epoch check that is enforced atomically when each removal is applied.

curvine-server/src/master/fs/master_filesystem.rs
jlon: Fixed in 904c1edf. Incremental finalized/writing reports now invalidate both the pending reconcile generation and the accumulated full_block_reports session via an invalidated tombstone, so later chunks from the same stale full-report session cannot complete a reconcile snapshot that predates the incremental update. I also added a regression test for this interleaving. Validation: cargo fmt --check passed; git diff --check passed. The targeted curvine-server test could not complete locally because librocksdb-sys compilation repeatedly timed out before the test binary ran.

curvine-server/src/master/fs/master_filesystem.rs
jlon: Fixed in 904c1edf. The stale reconcile apply path now rechecks the generation while holding full_block_reconciles across the removal batch application, so incremental invalidation/application and stale removals are serialized around the generation decision. Validation: cargo fmt --check passed; git diff --check passed. The targeted curvine-server test could not complete locally because librocksdb-sys compilation repeatedly timed out before the test binary ran.

szbr00 (approved):

PR #1029, opened 2026-07-08
fix(worker): harden block report scheduling

jlon:


curvine-server/src/worker/block/heartbeat_task.rs
```diff
@@ -96,8 +98,59 @@ impl HeartbeatTask {
...
+        blocks: Vec<BlockReportInfo>,
+    ) {
         for block in blocks {
-            self.report_blocks.insert(block.id, block);
+            report_blocks.insert(block.id, block);
+        }
+    }
+
+    fn submit_block_report(&self, report_blocks: Vec<BlockReportInfo>) {
+        if self.report_in_flight.swap(true, Ordering::AcqRel) {
```
szbr00: The in-flight check occurs after `run()` has drained `report_blocks`. If another report for the same block is inserted between that drain and this reinsertion, `put_missing_report` overwrites the newer status with the older drained value. Unlike the previous synchronous flow, this can happen on every heartbeat while a successful RPC is still in flight, potentially losing transitions such as `Deleted`. Please acquire/check the in-flight guard before draining the pending map so it remains untouched while another report is running.

curvine-server/src/worker/block/block_actor.rs
```diff
@@ -162,9 +168,42 @@ impl BlockActor {
...
+    actor: BlockActor,
+    complete: Arc<AtomicBool>,
+}
+
+impl LoopTask for FullBlockReportTask {
+    type Error = FsError;
+
+    fn run(&self) -> FsResult<()> {
+        let spend = TimeSpent::new();
+        match self.actor.full_block_report() {
```
szbr00: Each retry rebuilds the full report from a new `all_blocks()` snapshot, but the master retains successfully received chunks from the failed attempt. If the block set changes while keeping the same `total_len`, the master merges IDs from different snapshots and may complete with their union. For example, partial `{A,B}` followed by `{B,C}` can reconcile as `{A,B,C}`, preserving a stale location for A. Please retain the original snapshot/progress across retries or explicitly start a new report generation that clears the master's partial state.

jlon: Fixed in f6a3326: full block report retries now retain the original block snapshot for the retry task instead of rebuilding from a fresh `all_blocks()` snapshot on each attempt.

curvine-server/src/worker/block/heartbeat_task.rs
jlon: Fixed in f6a3326: the heartbeat task now acquires the in-flight guard before draining pending reports, and failed report reinsertion no longer overwrites newer pending states for the same block.

szbr00 (approved):

PR #1019, opened 2026-07-08
fix(master): isolate control-plane rpc lanes (#1011)

jlon:
# Summary

This PR isolates several master control-plane RPC classes into bounded executor lanes. It keeps the main RPC sync pool from being occupied by heartbeat, block-report, list, and get-block-locations storms while preserving the existing handler semantics.

# Issue Describe / Design

Related to #1011.

- Root cause: the master previously let more request types run through the shared RPC sync path. Under request storms, metadata reads, worker heartbeat, and full block report handling could occupy the shared processing resource and delay unrelated control-plane requests.
- Fix approach: add nonblocking `try_spawn` support to ORPC executors, then route heartbeat, block report, master info, list, and get-block-locations through dedicated bounded `GroupExecutor` lanes. Async handler errors are returned as normal RPC error responses, including standby/not-leader responses.
- Trade-off: this does not remove the global `fs_dir` lock bottleneck. It limits blast radius at the RPC execution layer and keeps worker/control traffic from sharing the same saturated sync pool.

# Changes

| Module / File | Change | Impact on existing behavior |
| ------------- | ------ | --------------------------- |
| `orpc/src/runtime/single_executor.rs` | Add nonblocking `try_spawn` that returns an error when the bounded queue is full or stopped. | Existing blocking `spawn` behavior is unchanged. |
| `orpc/src/runtime/group_executor.rs` | Expose `try_spawn` and `fixed_try_spawn` for grouped executors. | Enables bounded admission without blocking callers. |
| `curvine-server/src/master/master_server.rs` | Create dedicated master RPC lanes for heartbeat, block report, control, list, and get-block-locations. | New bounded executors isolate these request classes from the main sync pool. |
| `curvine-server/src/master/master_handler.rs` | Move selected RPCs to async handling and run their existing logic on the dedicated lanes. | Existing RPC response semantics are preserved; errors are encoded as RPC error responses. |
| `curvine-server/tests/master_fs_test.rs` | Cover async lane routing and standby async error response behavior. | Prevents regression back to sync-pool handling. |
| `orpc/tests/executor_test.rs` | Cover queue-full behavior for `SingleExecutor::try_spawn`. | Verifies nonblocking admission semantics. |

# Test verified

| Test case | Result | Notes |
| --------- | ------ | ----- |
| `cargo fmt --check` | PASS | |
| `cargo test -p orpc --test executor_test single_executor_try_spawn_returns_error_when_queue_is_full` | PASS | 1 passed. |
[...]

PR #1030, opened 2026-07-08
fix(worker): submit incremental block reports asynchronously (#1011)

jlon:
# Summary

This PR moves worker incremental block report RPCs out of the heartbeat loop. A worker keeps at most one incremental block report in flight; new report entries are put back into the pending map and retried by the next heartbeat.

# Issue Describe / Design

Related to #1011.

- Root cause: `HeartbeatTask::run` sent `client.incr_block_report` synchronously after heartbeat. When the master was slow or overloaded, the worker heartbeat loop could block on block report RPC latency instead of continuing its regular heartbeat cadence.
- Fix approach: submit incremental block reports to the existing worker executor and guard them with `report_in_flight`. Failed executor submission or failed master RPC requeues the report entries into `report_blocks`.
- Trade-off: this PR intentionally does not depend on #1019, so it uses the existing `executor.spawn` API. That keeps the PR independent from the ORPC `try_spawn` PR, but queue-full backpressure can still block briefly while enqueueing.

# Changes

| `curvine-server/src/worker/block/heartbeat_task.rs` | Add single in-flight async submission for incremental block reports and requeue on failure. | Heartbeat no longer waits for incremental block report RPC completion. |
| `curvine-server/src/worker/block/block_actor.rs` | Initialize the per-heartbeat-task in-flight guard. | No behavior change outside worker block report scheduling. |

# Test verified

# Dependencies

- Related PRs: #1005, #1029
- Related issues: #1011
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/block/block_actor.rs
+++ b/curvine-server/src/worker/block/block_actor.rs
@@ -23,6 +23,7 @@ use orpc::common::TimeSpent;
 use orpc::runtime::{GroupExecutor, Runtime};
 use orpc::sync::StateCtl;
 use orpc::CommonResult;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
 /// Worker block management role.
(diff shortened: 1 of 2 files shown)


jlon:
Superseded by #1029, which now combines full block report retry and incremental block report async scheduling into one worker block report scheduling PR.

PR #1031, opened 2026-07-08
fix(master): reconcile full block reports asynchronously (#1011)

jlon:
# Summary

This PR moves full block report stale-location reconciliation out of the foreground `block_report` path. The master still applies the current report entries immediately, but the expensive stale-location scan and delete batch runs in a bounded background lane.

# Issue Describe / Design

Related to #1011.

- Root cause: a completed full block report made the master scan all existing block locations for that worker and apply stale-location removals while serving the block report RPC. For large workers this holds or repeatedly contends on the global `fs_dir` lock and can delay other control-plane metadata operations.
- Fix approach: keep incremental block report updates synchronous, enqueue only the completed full-report reconcile work, and keep at most the latest pending reconcile job per worker. New finalized/writing incremental reports invalidate stale full-report reconcile generations so a delayed full report cannot remove a newer worker location.
- Semantics: under-replicated reporting is preserved by moving it into the background reconcile path. The foreground RPC no longer waits for stale-location reconciliation to finish.

# Changes

| `curvine-server/src/master/fs/master_filesystem.rs` | Add per-worker full-report reconcile state, bounded background executor, generation invalidation, and chunked `fs_dir.block_report` apply. | Full-report stale cleanup becomes asynchronous; current report entries are still applied before RPC response. |
| `curvine-server/src/master/master_handler.rs` | Pass the replication handler into `block_report` so background reconcile can preserve under-replicated reporting. | Existing under-replicated reporting is preserved after async reconciliation. |
| `curvine-server/tests/master_fs_test.rs` | Add behavior test for async stale-location removal after full block report. | Covers visible metadata result. |
| `curvine-server/tests/lock_order_deadlock_stress_test.rs` | Update direct test calls to pass no replication handler. | Test-only API adjustment. |

# Test verified

| `cargo test -p curvine-server --test master_fs_test full_block_report_reconcile_removes_stale_location_async -- --test-threads=1` | PASS | 1 passed. |
| `cargo test -p curvine-server --test lock_order_deadlock_stress_test -- --test-threads=1` | PASS | 2 passed. |
| `cargo test -p curvine-server --test master_fs_test -- --test-threads=1` | PASS | 22 passed. |

# Dependencies

- Related PRs: #1005, #1019, #1024, #1029, #1030
- Related issues: #1011
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/fs/master_filesystem.rs
+++ b/curvine-server/src/master/fs/master_filesystem.rs
@@ -20,13 +20,15 @@ use crate::master::meta::FsDir;
 
 use crate::master::fs::DeleteResult;
 use crate::master::meta::parse_glob_pattern;
+use crate::master::replication::master_replication_handler::MasterReplicationHandler;
 use crate::master::{Master, MasterMonitor, SyncFsDir, SyncWorkerManager};
 use curvine_common::conf::{ClusterConf, MasterConf};
 use curvine_common::error::FsError;
 use curvine_common::state::*;
 use curvine_common::FsResult;
-use log::warn;
+use log::{error, info, warn};
 use orpc::common::LocalTime;
+use orpc::runtime::GroupExecutor;
 use orpc::sync::ArcRwLock;
 use orpc::{err_box, err_ext, try_option, CommonResult};
 use parking_lot::Mutex;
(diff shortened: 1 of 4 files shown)


jlon:
Superseded by #1024, which now combines DeleteBlock cleanup commands and async full-block-report stale-location reconciliation into one master block report PR.

PR #1037, opened 2026-07-09
fix(orpc): spdk EAL: Cannot use IOVA as 'PA' since physical addresses are not available in  Linux VM

gongxun0928:
# Summary

We have been running SPDK smoke on the example cluster (malloc NVMe-oF target + worker initiator), and sometimes the worker dies at `spdk_env_init()` with:

```
Cannot use IOVA as `PA` since physical addresses are not available
Cannot set up PCI infrastructure
```

Curvine never set `spdk_env_opts.iova_mode` before; it relied on SPDK/DPDK auto-detection. That works on bare metal, but on VMs, containers, or some AMD IOMMU setups it often picks `pa` when physical addresses are not actually available. Same class of problem as [spdk/spdk#2683](https://github.com/spdk/spdk/issues/2683).

This PR adds an **opt-in** `iova_mode` knob instead of forcing a global default, so prod bare-metal behavior stays unchanged while constrained envs can opt into `"va"`.

# Issue Describe / Design

Related to [spdk/spdk#2683](https://github.com/spdk/spdk/issues/2683)

**Root cause:** `spdk_env_opts.iova_mode` was left unset; DPDK auto-detection mis-picks `pa` on hosts without usable physical addresses.

**Repro:** Run worker SPDK initiator on a Linux VM / container against a local NVMe-oF target, for example malloc `nvmf_tgt` on `127.0.0.1:4420`. `spdk_env_init()` fails intermittently with the PA IOVA error above. Our `nvmf_tgt` already uses `--iova-mode=va`, but the worker had no equivalent config.

**Fix approach:**
- Add `SpdkConf.iova_mode` + `curvine_spdk_env_opts_set_iova_mode()` FFI
- **Empty (default):** skip `set_iova_mode`, keep auto-detect, no behavior change for existing deployments
- **`"va"` / `"pa"`:** explicit override when auto-detect breaks
- Integration test helpers default to `"va"` and remain overridable via `SPDK_IOVA_MODE`; empty string exercises auto-detect

**Why not default everything to `va`?** Auto-detect is usually right on real hardware. Forcing `va` globally would paper over environment issues. Opt-in override is safer.

# Changes

| `orpc/csrc/spdk_opts_helper.c` | Add `curvine_spdk_env_opts_set_iova_mode()` | None |
| `orpc/src/io/spdk_ffi.rs` | FFI declaration | None |
| `orpc/src/io/spdk_env.rs` | `SpdkConf.iova_mode`, validation, pass-through in `env_init`, `Display` output, and orpc test helper | None when unset; explicit `va`/`pa` when configured |
| `orpc/src/io/spdk_bdev.rs`, `orpc/src/io/spdk_bdev_test.rs` | Use the shared orpc SPDK IOVA test helper | Test-only |
| `curvine-server/tests/common/mod.rs` | `spdk_iova_mode_for_test()` helper, default `"va"` | Test-only |
| `curvine-server/tests/spdk_*_test.rs` | Use helper in test worker conf | Test-only |
| `curvine-docker/deploy/spdk/curvine-cluster-spdk.toml` | Document optional `iova_mode = "va"` for VM/container deployments | Example config only |

# Test verified

| Unit tests: empty = auto-detect, `va`/`pa` accepted, invalid values rejected | PASS | `orpc/src/io/spdk_env.rs` |
| Example-cluster SPDK smoke | PASS | malloc NVMe-oF target + worker initiator; `fs blocks` shows `SpdkDisk` |
| `cargo fmt --check` | PASS | Local after review fixes |
| `cargo test -p orpc` | PASS | Local after review fixes |

**Bottom line for prod:** if you do not set `iova_mode`, nothing changes from today.

# Dependencies

- Related PRs: None
- Related issues: [spdk/spdk#2683](https://github.com/spdk/spdk/issues/2683)
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/tests/common/mod.rs
+++ b/curvine-server/tests/common/mod.rs
@@ -13,3 +13,15 @@ pub fn require_spdk_test_env() {
         std::process::exit(0);
     }
 }
+
+/// IOVA mode for SPDK integration tests.
+///
+/// Defaults to `"va"` because VM/container/AMD hosts often fail auto-detect
+/// (spdk/spdk#2683). Set `SPDK_IOVA_MODE` to override; use empty string to
+/// exercise SPDK auto-detection.
+pub fn spdk_iova_mode_for_test() -> String {
+    match std::env::var("SPDK_IOVA_MODE") {
+        Ok(v) => v,
+        Err(_) => "va".to_string(),
+    }
+}
--- a/orpc/src/io/spdk_bdev.rs
+++ b/orpc/src/io/spdk_bdev.rs
@@ -773,6 +773,7 @@ mod test {
         let subnqn = std::env::var("SPDK_TARGET_NQN")
             .unwrap_or_else(|_| "nqn.2024-01.io.curvine:test".into());
         let trtype = std::env::var("SPDK_TRANSPORT_TYPE").unwrap_or_else(|_| "tcp".into());
+        conf.iova_mode = std::env::var("SPDK_IOVA_MODE").unwrap_or_else(|_| "va".into());
         conf.targets = vec![crate::io::spdk_env::NvmeTarget {
             traddr,
             trsvcid,
--- a/orpc/src/io/spdk_bdev.rs
+++ b/orpc/src/io/spdk_bdev.rs
@@ -773,7 +773,7 @@ mod test {
         let subnqn = std::env::var("SPDK_TARGET_NQN")
             .unwrap_or_else(|_| "nqn.2024-01.io.curvine:test".into());
         let trtype = std::env::var("SPDK_TRANSPORT_TYPE").unwrap_or_else(|_| "tcp".into());
-        conf.iova_mode = std::env::var("SPDK_IOVA_MODE").unwrap_or_else(|_| "va".into());
+        conf.iova_mode = crate::io::spdk_env::spdk_iova_mode_for_test();
         conf.targets = vec![crate::io::spdk_env::NvmeTarget {
             traddr,
             trsvcid,
--- a/orpc/src/io/spdk_env.rs
+++ b/orpc/src/io/spdk_env.rs
@@ -429,15 +429,26 @@ impl Default for SpdkConf {
 
 impl Display for SpdkConf {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let iova_mode_label = if self.iova_mode.is_empty() {
+            "auto"
+        } else {
+            self.iova_mode.as_str()
+        };
         write!(
             f,
-            "SpdkConf(enabled={}, hugepage={}MB, reactor_mask={}, targets={})",
+            "SpdkConf(enabled={}, hugepage={}MB, reactor_mask={}, iova_mode={}, targets={})",
             self.enabled,
             self.hugepage_mb,
             self.reactor_mask,
+            iova_mode_label,
             self.targets.len()
         )
     }
+}
+
+#[cfg(test)]
+pub(crate) fn spdk_iova_mode_for_test() -> String {
+    std::env::var("SPDK_IOVA_MODE").unwrap_or_else(|_| "va".to_string())
 }
 
 // Bdev descriptor (discovered after init)
--- a/orpc/src/io/spdk_env.rs
+++ b/orpc/src/io/spdk_env.rs
@@ -256,6 +256,10 @@ pub struct SpdkConf {
     #[serde(skip)]
     pub hugepage_mb: u32, // parsed by init()
     pub reactor_mask: String, // hex, e.g. "0x3"
+    /// DPDK IOVA mode override. Empty = let SPDK/DPDK auto-detect (default).
+    /// Set to "va" when auto-detect fails (VM/virtio/AMD IOMMU, spdk/spdk#2683),
+    /// or "pa" for bare metal that requires physical addresses.
+    pub iova_mode: String,
     pub targets: Vec<NvmeTarget>,
     pub io_queue_depth: u32,
     pub io_queue_requests: u32,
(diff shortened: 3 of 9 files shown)


szbr00 (commented):
## Code Review

Reviewed all 8 changed files (+92/-2) in full, plus surrounding SPDK env-init, FFI, and config plumbing.

### Summary
Clean, low-risk, well-tested change. The opt-in design (empty = auto-detect) preserves existing behavior exactly, `#[serde(default)]` on `SpdkConf` keeps existing tomls working, validation rejects bad values, and the new unit tests cover empty/valid/invalid. No Critical or Major issues.

I specifically verified the one real risk — the C helper stores the raw pointer (`opts->iova_mode = mode;`). On the Rust side the `iova_mode` `CString` is bound in the function scope and outlives the `curvine_spdk_env_init` call, and SPDK consumes `iova_mode` synchronously during env init. This matches the existing `name`/`core_mask` handling, so the pointer is valid for its entire use. Safe.

### Findings: 3 (all Nit / optional)

| # | Severity | File | Finding | Suggestion |
| - | -------- | ---- | ------- | ---------- |
| 1 | Nit | `curvine-docker/deploy/spdk/curvine-cluster-spdk.toml` | Example SPDK toml documents `reactor_mask` but not `iova_mode`. VM/container deploys (the exact scenario this fixes) won't discover the knob from the example config. | Optionally add `# iova_mode = "va"  # set on VM/container hosts if env_init fails with PA IOVA error`. |
| 2 | Nit | `orpc/src/io/spdk_bdev.rs`, `orpc/src/io/spdk_bdev_test.rs`, `curvine-server/tests/common/mod.rs` | The `SPDK_IOVA_MODE` → default `"va"` logic is duplicated in three places. | Acceptable given the crate boundary; noting for future consolidation. |
| 3 | Nit | `orpc/src/io/spdk_env.rs` | `validate()` returns early when `!enabled`, so an invalid `iova_mode` with `enabled=false` isn't caught. | Consistent with all other fields here; no change needed. |

Bottom line: LGTM — only optional nits.

szbr00 (commented):
## Code Review

Reviewed 8 files, +92 / -2 lines changed.

### Summary
This PR adds a configurable `iova_mode` field to `SpdkConf` to work around SPDK's IOVA auto-detection failure in VMs/containers/AMD hosts (spdk/spdk#2683). The change is well-scoped, properly validated, backward-compatible via `#[serde(default)]`, and includes good test coverage. The FFI layer correctly delegates to a C helper to handle SPDK version-dependent struct layout.

### Findings: 1
- Critical: 0
- Major: 0
- Minor: 0
- Nit: 1

**[Nit] `orpc/src/io/spdk_env.rs` — `Display` impl missing `iova_mode`**
The `Display` impl for `SpdkConf` does not include the new `iova_mode` field, while `env_init()` log was updated to show it. For consistency, consider adding `iova_mode={}` to the `Display` format string, or leave as-is if intentionally omitting from the summary.

### What was checked
- **Correctness:** Validation rejects unknown modes; accepts "va", "pa", and empty (auto-detect). CString lifetime is correct — lives across the `unsafe` block containing `spdk_env_init`. Serde backward compatibility via `#[serde(default)]` on the struct. Default impl field ordering matches struct definition.
- **Safety:** No unsafe concerns — the C helper `curvine_spdk_env_opts_set_iova_mode` simply assigns the pointer to `opts->iova_mode`, consumed by `spdk_env_init` before the CString is dropped.
- **Design:** Consistent with existing patterns (`reactor_mask` uses the same CString-via-C-helper approach). Test helper `spdk_iova_mode_for_test()` is reusable.
- **Tests:** Three new unit tests cover default, rejection, and acceptance. Integration tests updated consistently.

Overall: looks good to merge.

orpc/src/io/spdk_bdev.rs
```diff
@@ -773,6 +773,7 @@ mod test {
         let subnqn = std::env::var("SPDK_TARGET_NQN")
             .unwrap_or_else(|_| "nqn.2024-01.io.curvine:test".into());
         let trtype = std::env::var("SPDK_TRANSPORT_TYPE").unwrap_or_else(|_| "tcp".into());
+        conf.iova_mode = std::env::var("SPDK_IOVA_MODE").unwrap_or_else(|_| "va".into());
```
szbr00: The SPDK_IOVA_MODE env-var lookup with a "va" fallback is now duplicated here, in spdk_bdev_test.rs, and in curvine-server/tests/common/mod.rs (as spdk_iova_mode_for_test). Since these live in different crates a shared helper isn't trivial, but it's worth consolidating the two orpc copies into one test helper so the default value can't drift between them. Non-blocking.

gongxun0928: Thanks. Fixed in 52f91d8: consolidated the two orpc SPDK_IOVA_MODE test defaults behind spdk_iova_mode_for_test().

gongxun0928:
Thanks for the review. Fixed in 52f91d8: added the optional iova_mode example config comment, included iova_mode in SpdkConf Display, and consolidated the two orpc SPDK_IOVA_MODE test defaults. Kept enabled=false validation behavior unchanged because it is consistent with the existing SpdkConf validation semantics.

szbr00 (approved):

orpc/src/io/spdk_env.rs
```diff
@@ -351,6 +355,13 @@ impl SpdkConf {
             );
         }
 
+        if !self.iova_mode.is_empty() && self.iova_mode != "va" && self.iova_mode != "pa" {
```
thuongle2210, line 358: We should lowercase this config value before comparison and add 'va' and 'pa' to the allowlist for a more concise comparison (nice to have).

curvine-server/tests/common/mod.rs
```diff
@@ -13,3 +13,15 @@ pub fn require_spdk_test_env() {
         std::process::exit(0);
     }
 }
+
+/// IOVA mode for SPDK integration tests.
+///
+/// Defaults to `"va"` because VM/container/AMD hosts often fail auto-detect
+/// (spdk/spdk#2683). Set `SPDK_IOVA_MODE` to override; use empty string to
+/// exercise SPDK auto-detection.
+pub fn spdk_iova_mode_for_test() -> String {
```
thuongle2210, line 22: Thought: Could we reuse spdk_iova_mode_for_test, which is defined in orpc?

PR #1038, opened 2026-07-09
feat(fuse): enforce readonly mount mode (#896)

### LINKED ISSUE #896: [FEATURE] Support both read-write and readonly mount modes for buckets
## Problem

Currently, when a bucket is mounted via curvine-fuse, it is exposed in read-write mode only. There is no way to mount a bucket as read-only.

This is inconvenient or unsafe in several scenarios:
- Production workloads that only consume data (training, inference, analytics) want a hard guarantee that writes cannot happen, even by accident.
- Buckets backed by object stores where the user only has read credentials.
- Multi-tenant environments where one process writes and others should only read.

Today, the only protection is filesystem-level permissions, which is not equivalent to a true read-only mount and does not cover all paths (e.g. truncate / setattr / unlink).

## Proposed Solution

Add a mount-time option (e.g. `--readonly` / `mount.readonly = true`) per bucket that:

1. Marks the FUSE mount as `MS_RDONLY` (so the kernel itself rejects write syscalls early).
2. Causes curvine-fuse to short-circuit any write-side FUSE ops (`write`, `create`, `mkdir`, `unlink`, `rmdir`, `rename`, `setattr` size changes, `truncate`, `link`, `symlink`, `fallocate`, etc.) with `EROFS`.
3. Skips opening write paths to the master/worker for that mount, reducing resource usage.

The mode should be configurable per-bucket so the same curvine-fuse instance can mount one bucket RW and another RO.

## Impact

- Stronger correctness guarantee for read-only consumers (kernel-enforced).
- Aligns with common object-store semantics where a credential may be read-only.
- Reduces accidental data corruption / deletion in shared environments.
- No behavior change for existing RW mounts (opt-in flag).

ligeng8848:
## Summary
- Enforce `access_mode=read_only` bucket mounts in FUSE write-side operations with `EROFS`.
- Add `FsError::ReadOnly` and map it to POSIX `EROFS`.
- Add global `--readonly` / `fuse.readonly=true` support for whole-FUSE `ro` mount options.

## Issue Describe / Design
Fixes #896.

This reuses the existing bucket-level `access_mode=read_only/read_write` metadata. Since OS `MS_RDONLY` only applies to the whole FUSE mount, per-bucket readonly behavior is enforced in curvine-fuse by checking write-side FUSE ops against mount metadata before they reach Curvine/UFS writers. The FUSE guard constructs the readonly error and maps it to EROFS; the unified layer retains a ReadOnly fallback for non-FUSE callers.

## Changes
| Area | Change |
| --- | --- |
| common error | Add `ReadOnly` error kind and encode/decode coverage |
| fuse error | Map readonly failures to `EROFS` |
| fuse config/CLI | Add `fuse.readonly` and `--readonly` for whole-mount `ro`; avoid duplicate `ro` mount options |
| unified fs | Return readonly errors for read-only cache-mode write RPCs |
| curvine-fuse | Short-circuit create/mkdir/unlink/rmdir/rename/setattr/xattr/link/symlink/fallocate/locks in the FUSE guard with `EROFS`; avoid per-write mount lookup in the data hot path |
| curvine-lancedb | Preserve `FsError::ReadOnly` as object-store `NotSupported` instead of `Generic` |

## Test verified
- `cargo fmt`
- `git diff --check`
- Remote Linux: `cargo test -p curvine-common readonly -- --nocapture` passed
- Remote Linux PR branch S3/FUSE probe: read-only cache-mode write returns `EROFS(30) / Read-only file system`
- Local macOS: `cargo test -p curvine-common readonly -- --nocapture` blocked by local `librocksdb-sys` C++ toolchain failure: missing standard headers like `cstdint`, `array`, `memory`.
- Remote Linux: `cargo test -p curvine-lancedb readonly_fs_error_maps_to_object_store_not_supported -- --nocapture` blocked by crates.io download timeout for `openssl-probe v0.2.1`.

### DIFF
--- a/curvine-client/src/unified/unified_filesystem.rs
+++ b/curvine-client/src/unified/unified_filesystem.rs
@@ -152,7 +152,7 @@ impl UnifiedFileSystem {
         let state = self.mount_cache.get_mount(self, path).await?;
         if let Some(mnt) = state {
             if mnt.info.is_read_only_cache_mode() && Self::is_mount_write_rpc(rpc_code) {
-                return err_ext!(FsError::unsupported(format!(
+                return err_ext!(FsError::read_only(format!(
                     "{} on read_only cache_mode mount {}",
                     rpc_code, path
                 )));
@@ -284,6 +284,7 @@ impl UnifiedFileSystem {
 
     pub async fn link(&self, src_path: &Path, dst_path: &Path) -> FsResult<()> {
         let fut = async {
+            let _ = self.get_mount_checked(dst_path, RpcCode::Link).await?;
             match self.get_mount_checked(src_path, RpcCode::Link).await? {
                 None => self.cv.link(src_path, dst_path).await,
                 Some(_) => err_ext!(FsError::unsupported("link")),
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -98,6 +98,9 @@ pub struct FuseConf {
     // Mount the configuration, needs to be passed to the linux kernel.
     pub fuse_opts: Vec<String>,
 
+    // Mount the whole FUSE filesystem read-only at the kernel level.
+    pub readonly: bool,
+
     // Overwrite the permission bits set by the file system in st_mode.
     // The generated permission bit is the missing permission bit in the given umask value.This value is given in octal representation.
     // Default value 022
@@ -320,7 +323,14 @@ impl FuseConf {
     }
 
     pub fn set_fuse_opts(&self, mount_options: &mut String) {
+        if self.readonly {
+            mount_options.push_str(",ro");
+        }
+
         self.fuse_opts.iter().for_each(|opt| match opt.as_str() {
+            "ro" => {
+                mount_options.push_str(",ro");
+            }
             "default_permissions" => {
                 mount_options.push_str(",default_permissions");
             }
@@ -360,6 +370,7 @@ impl Default for FuseConf {
             fuse_channel_size: 0,
             stream_channel_size: 0,
             fuse_opts: vec![],
+            readonly: false,
             umask: Self::DEFAULT_UMASK,
             uid: sys::get_uid(),
             gid: sys::get_gid(),
@@ -460,6 +471,23 @@ max_readahead_kb = 1024
             conf.max_readahead_kb,
             Some(FuseConf::DEFAULT_MAX_READAHEAD_KB)
         );
+    }
+
+    #[test]
+    fn readonly_adds_ro_mount_option() {
+        let conf = FuseConf {
+            readonly: true,
+            ..Default::default()
+        };
+        let mut mount_options = String::new();
+        conf.set_fuse_opts(&mut mount_options);
+        assert!(mount_options.split(',').any(|opt| opt == "ro"));
+    }
+
+    #[test]
+    fn toml_readonly_is_parsed() {
+        let conf: FuseConf = toml::from_str("readonly = true").expect("parse");
+        assert!(conf.readonly);
     }
 
     #[test]
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -323,13 +323,18 @@ impl FuseConf {
     }
 
     pub fn set_fuse_opts(&self, mount_options: &mut String) {
+        let mut ro_added = false;
         if self.readonly {
             mount_options.push_str(",ro");
+            ro_added = true;
         }
 
         self.fuse_opts.iter().for_each(|opt| match opt.as_str() {
             "ro" => {
-                mount_options.push_str(",ro");
+                if !ro_added {
+                    mount_options.push_str(",ro");
+                    ro_added = true;
+                }
             }
             "default_permissions" => {
                 mount_options.push_str(",default_permissions");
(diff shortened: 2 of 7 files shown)


curvine-client/src/unified/unified_filesystem.rs
```diff
@@ -152,7 +152,7 @@ impl UnifiedFileSystem {
         let state = self.mount_cache.get_mount(self, path).await?;
         if let Some(mnt) = state {
             if mnt.info.is_read_only_cache_mode() && Self::is_mount_write_rpc(rpc_code) {
-                return err_ext!(FsError::unsupported(format!(
+                return err_ext!(FsError::read_only(format!(
```
szbr00, line 155: **[Major]** Changing this shared `get_mount` error from `FsError::unsupported` to `FsError::read_only` reclassifies **non-FUSE** callers too, not just the FUSE path this PR targets.

`curvine-lancedb/src/object_store.rs:1874` explicitly matches `FsError::Unsupported(_) | FsError::UnsupportedUfsRead(_)` → `OsError::NotSupported`. With `ReadOnly` it now falls through to the `e => OsError::Generic` arm, silently changing the object-store error taxonomy for writes on read-only cache mounts.

Suggestion: add `FsError::ReadOnly(_)` to that `NotSupported` arm in `object_store.rs` (or keep `get_mount` returning `Unsupported` and translate to `EROFS` only at the FUSE layer). The Python SDK is unaffected — its `_ =>` default already yields `PyException`, same as `Unsupported`.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1343,6 +1365,9 @@ impl fs::FileSystem for CurvineFileSystem {
 
     async fn write(&self, op: Write<'_>, reply: FuseResponse) -> FuseResult<()> {
         let handle = self.state.find_handle(op.header.nodeid, op.arg.fh)?;
+        let path = self.state.get_path(op.header.nodeid)?;
```
szbr00: **[Minor / perf]** `write()` is a data hot path, and this adds a `get_path` (inode-tree walk + allocation) plus `ensure_writable_path` (`get_mount`: `get_possible_mounts` allocation + RwLock read) on every write op. It's redundant with the enforcement already done at `open`/`create`, except for the rare case where a mount flips to read-only mid-write.

Consider resolving writability once at `open`/`create` and caching it on the handle, so `write()` avoids the per-call cost while still handling the mid-flight case cheaply.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1624,6 +1664,7 @@ impl fs::FileSystem for CurvineFileSystem {
 
     async fn set_lk(&self, op: SetLk<'_>) -> FuseResult<()> {
         let path = self.state.get_path(op.header.nodeid)?;
+        self.ensure_writable_path(&path, RpcCode::SetLock).await?;
```
szbr00, line 1672: **[Minor / behavior]** This rejects **all** advisory lock ops with `EROFS` on a read-only mount, including shared/read locks (`F_RDLCK`) and unlock (`F_UNLCK`). POSIX advisory read locks are normally permitted on read-only filesystems, so applications that take shared locks on read-only-opened files may break (same applies to `set_lkw`).

This is consistent with the existing treatment of `SetLock` as a write RPC for read-only cache mounts, so it may be acceptable — just flagging it as a deliberate decision to confirm.

curvine-common/src/conf/fuse_conf.rs
```diff
@@ -320,7 +323,14 @@ impl FuseConf {
     }
 
     pub fn set_fuse_opts(&self, mount_options: &mut String) {
+        if self.readonly {
```
szbr00, line 327: **[Nit]** If both `conf.readonly = true` and `fuse_opts` contains `"ro"`, `,ro` is appended twice (here and in the `"ro"` match arm below). Harmless since the kernel dedups, but consider guarding so `ro` is only added once.

szbr00 (commented):
## Code Review

Reviewed all 6 changed files (+118/-2) in full, plus the cross-crate impact of the shared error change.

### Summary
Core design is sound and well-tested for the FUSE path — the error plumbing (`ErrorKind::ReadOnly = 30`, all `kind`/`ctx`/`encode`/`decode` arms + round-trip test), the `EROFS` mapping, and the `ensure_writable_path` helper are all consistent. RPC codes at each call site fall within `is_mount_write_rpc`, and I verified the new `get_path` in `write()` does not regress the "write to an unlinked fd" pattern (path resolves from `NodeAttr.parent/name`, kept alive by `n_lookup` until `forget`).

The one item worth addressing before merge is the shared `get_mount` error change (finding #1), which leaks beyond FUSE into `curvine-lancedb`'s error taxonomy. The rest are Minor/Nit.

### Findings: 4
- Major: 1 (shared error variant change reclassifies `curvine-lancedb` object-store errors from `NotSupported` to `Generic`)
- Minor: 2 (`write()` hot-path cost; advisory read-lock rejection on read-only mounts)
- Nit: 1 (possible double `,ro` mount option)

curvine-client/src/unified/unified_filesystem.rs
ligeng8848: Thanks, good catch. Fixed in 2d69f7c by mapping FsError::ReadOnly to ObjectStore::NotSupported in curvine-lancedb, with a regression test for the error taxonomy.

curvine-fuse/src/fs/curvine_file_system.rs
ligeng8848: Thanks. Fixed in 2d69f7c by removing the per-write get_path/get_mount check from the data hot path. Writable handles are already gated at open/create, so write now uses the existing handle directly.

curvine-fuse/src/fs/curvine_file_system.rs
ligeng8848: Thanks for flagging this. I kept the current behavior deliberately for this PR: UnifiedFileSystem still treats SetLock as a write RPC for read-only cache-mode mounts, so allowing only shared/unlock at the FUSE layer would not be sufficient by itself and would change the cross-layer semantics. We can split read-lock/unlock behavior in a follow-up if we want POSIX advisory locks to be usable on read-only bucket mounts.

curvine-common/src/conf/fuse_conf.rs
ligeng8848: Thanks. Fixed in 2d69f7c by de-duplicating the ro mount option when both fuse.readonly=true and fuse_opts contains ro, with a unit test.

szbr00 (approved):

PR #1036, opened 2026-07-09
refactor(fuse): clean up FuseConf constants and document cache layers (#1023)

yangyang233333:
# Summary

Cleans up `FuseConf`: removes dead associated constants and one dead param, renames misnamed constants to the `DEFAULT_*` convention, and documents the three easy-to-conflate cache layers. This is the merged PR-4 of the incremental `FuseConf` cleanup tracked in #1023 (§5 + §4 + the `node_cache_size` removal from §1). Purely internal — no TOML keys or CLI flags that anything actually consumes.

# Issue Describe / Design

Related to #1023 (§5 "Dead / misnamed associated constants", §4 "Overlapping cache parameters lack documented boundaries", and the `node_cache_size` item from §1 "Dead parameters"). The original standalone PR-1 (#1034) was closed and its `node_cache_size` removal folded in here; the `read_dir_fill_ino` / `remember` removals are dropped for now.

Design decisions:
- **Dead constants removed** (grepped repo-wide, zero readers): `FuseConf::MAX_READ`, `MAX_WRITE`, `MAX_READ_AHEAD`, and `curvine-fuse/src/lib.rs::FUSE_DEFAULT_UMASK`. The live, unrelated `read_handler.rs::MAX_READ_AHEAD` (16 MiB, `i64`) is intentionally left untouched.
- **`TTR_TIMEOUT` split, not just renamed.** The old single `TTR_TIMEOUT` (undefined 'TTR' acronym) fed four fields. Renaming it to one name would be misleading, so it becomes two semantically-honest constants — `DEFAULT_ENTRY_TIMEOUT` (for `entry_timeout`) and `DEFAULT_ATTR_TIMEOUT` (for `attr_timeout` / `ac_attr_timeout` / `ac_attr_timeout_set`, which FUSE convention keeps equal to `attr_timeout`). Both remain `1.0`, so defaults are unchanged.
- **`UMASK` -> `DEFAULT_UMASK`**, matching the well-formed `DEFAULT_MAX_READAHEAD_KB` / `client_conf::DEFAULT_FILE_SYSTEM_UMASK`.
- **`node_cache_size` removed** (dead param): declared, CLI-exposed and TOML-parsed but never consumed. `NodeMap.nodes` is evicted by `node_cache_timeout` (TTL) only via `CleanerTask -> clean_cache`; the capacity was never enforced. Distinct from the *live* `meta_cache_capacity` (consumed by `MetaCache::new`), which is untouched.
[...]

Issue #1023, opened 2026-07-08
[FEATURE] Refactor curvine-fuse FuseConf: remove dead params, fix misnomers, unify TTL representation


yangcx000:
## Is your feature request related to a problem? Please describe.

`FuseConf` (`curvine-common/src/conf/fuse_conf.rs`) has accumulated several config-hygiene issues that hurt usability and maintainability. Every finding below is **verified against the current code** — consumption sites were grepped across `curvine-fuse` / `curvine-common` / `curvine-client`, excluding the field definition, CLI wiring, and tests.

### 1. Dead parameters — declared, CLI-exposed, TOML-parsed, but never consumed

| Param | Default | Evidence |
| --- | --- | --- |
| `read_dir_fill_ino` | true | zero reads — only a CLI setter in `mount_args.rs`; comment claims it controls readdir ino-fill, no code path consumes it |
| `remember` | false | zero reads — only a CLI setter; comment claims "remember opened inodes", never wired |
| `node_cache_size` | 200000 | zero reads — `NodeMap.nodes` is an unbounded map evicted only by TTL (`node_cache_ttl`); the capacity is never enforced. (Note: distinct from the *live* `meta_cache_capacity`, which **is** consumed by `MetaCache::new` — easy to confuse.) |

These pay a cost (CLI surface, TOML schema, reader confusion) for no behavior.

### 2. Misnamed parameter

- `mnt_per_task` reads as "mnt per task" but actually means **tasks per mount point**: `FuseChannel::new` does `for _ in 0..mnt_per_task { create sender/receiver }`, and `FuseChannel::new` is called once per mount. The field controls how many sender/receiver tasks each mount gets. Should be `tasks_per_mnt`.

### 3. TTL is represented as two fields each (raw + derived) with inconsistent naming and types

Every timeout/TTL exists **twice**: a serializable "raw" field the user sets, and a `#[serde(skip_serializing)]` derived `Duration` computed once in `init()`. Business logic only ever reads the derived field; the raw field is read only by `init()`.

| Raw field (in TOML) | Type | → derived field (runtime) | Business reads (raw / derived) |
| --- | --- | --- | --- |
| `entry_timeout` | `f64` (bare seconds) | `entry_ttl: Duration` | 0 / 4 |
| `attr_timeout` | `f64` (bare seconds) | `attr_ttl: Duration` | 0 / 8 |
[...]

PR #1041, opened 2026-07-09
refactor(fuse): restructure metadata cache into a dedicated dcache mo…

bigbigxu:
## Summary

This change refactors the curvine-fuse client-side metadata caching layer. The
previous implementation spread inode state across `NodeMap` and `NodeAttr`
inside `NodeState`, with a large amount of caching and directory logic living
directly in `CurvineFileSystem`. This PR extracts that logic into a dedicated,
testable `dcache` module and reduces `CurvineFileSystem` to a thin FUSE-op
dispatcher.

## Design

The new `fs/dcache` module models the client-side directory/inode cache as a
single owned tree instead of loosely coupled maps:

- **`DirTree`** — the single source of truth for cached inodes. Owns a
  `FastHashMap<u64, Inode>`, the inode id allocator, and cache TTLs. All
  lookup / insert / evict / rename operations go through it, so it is the only
  structure that needs `RwLock` protection in `NodeState`.
- **`Inode`** — merges the former `NodeAttr` + node metadata. Derefs to
  `FileStatus`, tracks `Lifecycle` (Cached / Invalid / Dirty), lookup and
  reference counters, last-access time, and an optional boxed `DirEntry` for
  directories.
- **`DirEntry`** — per-directory child index (`children`, `deleted_children`,
  `scan_complete`) that supports negative-entry tracking for consistent
  readdir/lookup results.
- **`Lifecycle` / `OpState`** — explicit cache validity and in-flight
  operation states, replacing ad-hoc boolean flags.
- **`BackendHandle`** — unifies the reader/writer/status backing of an open
  file descriptor, including flock/plock owner bookkeeping and a read-version
  counter used to invalidate reader state when a writer advances.

`NodeState` becomes a coordinator: it holds one `RwLock<DirTree>`, the writer
and file/dir handle maps, striped path locks, and cache TTL settings, and
exposes `dir_read()` / `dir_write()` accessors instead of the old NodeMap API.

Cache behavior is now configurable via `FuseConf`:

- `enable_meta_cache` — toggle metadata caching (default off).
- `meta_cache_timeout` (default `60s`) — attribute cache TTL.
- `node_cache_timeout` (default `1h`) — node/dir-tree cache TTL.

## Changes

| `curvine-fuse/src/fs/dcache/` (new) | New `dcache` module: `dir_tree.rs`, `inode.rs`, `dir_entry.rs`, `op_state.rs`, `mod.rs` (`Lifecycle`), moved `cleaner_task.rs` | New internal cache layer; no external API change |
| `curvine-fuse/src/fs/state/node_map.rs`, `node_attr.rs` (removed) | Deleted; replaced by `dcache::DirTree` / `Inode` | Internal refactor |
| `curvine-fuse/src/fs/state/node_state.rs` | Rewritten to own `RwLock<DirTree>` and delegate inode bookkeeping | Internal refactor |
| `curvine-fuse/src/fs/state/backend_handle.rs` (new) | Unified reader/writer/status handle backing with lock owners + read version | Consolidates handle state |
| `curvine-fuse/src/fs/state/file_handle.rs`, `mod.rs` | Adapted to new state layout and re-exports | Internal refactor |
| `curvine-fuse/src/fs/curvine_file_system.rs` | Reduced ~800 lines; delegates to `NodeState` | Behavior preserved, logic relocated |
| `curvine-fuse/src/fs/fuse_writer.rs`, `fuse_reader.rs` | Adjusted to new handle/inode APIs | Internal refactor |
| `curvine-common/src/conf/fuse_conf.rs` | Add `enable_meta_cache`, `meta_cache_timeout`, `node_cache_timeout` and derived TTL durations | New config with safe defaults |
| `curvine-fuse/src/cli/mount_args.rs` | Wire up new cache options | New CLI flags |
| `curvine-fuse/src/fuse_metrics.rs` | Add write-back memory gauges and cache hit/miss/put counters | Observability only |
| `curvine-fuse/src/fuse_utils.rs`, `web_server.rs`, `session/*` | Follow-on adjustments for the new layout | Internal refactor |

## Test verified

| `cargo build -p curvine-fuse` | TODO | Compile the refactored crate |
| `cargo test -p curvine-fuse` | TODO | dcache / node_state unit tests |
| `cargo test -p curvine-common` | TODO | fuse_conf parsing of new options |
| FUSE mount smoke test (read / write / readdir / rename / unlink) | TODO | Verify no behavior regression |

## Dependencies

- Related PRs: builds on `#1023` (FuseConf constants and cache-layer docs), `#1036`
- Related issues: None
- Blocks / blocked by: None


curvine-fuse/src/fs/dcache/dir_tree.rs
```diff
@@ -0,0 +1,1001 @@
...
+                        if inode.is_deleted() {
+                            return err_fuse!(
+                                libc::ENOENT,
+                                "inode {} marked for deletion, suppress lookup revive",
+                                inode.ino
+                            );
+                        }
+                        inode.ino
+                    }
+
```
szbr00, line 216: **[Major]** When `(parent, name)` has no dentry but an inode with the same server id already exists, this branch takes `inode.ino` and then unconditionally `self.inodes.insert(ino, Inode::with_status(ino, parent, name, status))`, replacing the live inode with a fresh one.

`with_status` resets `n_lookup = 1` / `ref_ctr = 1` and drops the existing `dir` (DirEntry), `locs`, `mark_delete`, and `Dirty` lifecycle. Consequences:

- `ref_ctr` under-counts when the same inode is reachable via a second dentry (server-side hard link, or an external rename re-looked-up by the kernel): a later FORGET can drop the inode while the kernel still references it -> stale nodeid / ENOENT.
- If the cached inode is a directory with children, they are orphaned (the parent chain in `try_get_path` breaks).
- If the inode was `Dirty` (uncommitted write-back), the dirty state and `locs` are silently discarded.

This path is reachable from every FUSE LOOKUP (`NodeState::do_lookup` -> `DirTree::lookup`), and the inode table is maintained regardless of `enable_meta_cache`.

Note: the bundled test `lookup_existing_server_id_updates_inode_path_without_reinsert` asserts `ref_ctr == 2`, `n_lookup == 2`, and "without reinsert" — the opposite of what this code does, so that test would fail against this revision.

Suggested fix: in the existing-inode branch, update `parent` / `name`, call `add_ref(1)` + `add_lookup(1)` + `update_status(status)`, and skip the reinsert.

curvine-fuse/src/fs/dcache/dir_tree.rs
```diff
@@ -0,0 +1,1001 @@
...
+    }
+
+    pub fn clear(&mut self, has_handle: impl Fn(u64) -> bool) {
+        let now = LocalTime::mills();
+        let ttl = self.cache_ttl;
+
+        let removed: Vec<(u64, u64, String)> = self
+            .inodes
+            .values()
+            .filter(|inode| inode.can_evict(ttl) && !has_handle(inode.ino))
```
szbr00, line 543: **[Minor]** Eviction here filters on `can_evict(ttl) && !has_handle(ino)`, and `can_evict` only checks `!is_root`, `last_access + ttl < now`, and empty children — it ignores `n_lookup` / `ref_ctr`. So an inode the kernel still references by nodeid (`n_lookup > 0`) can be evicted once its TTL lapses (the test's Case 1 evicts an inode with `n_lookup = 1`). After eviction, a kernel op on that cached dentry hits `get_inode_check` -> ENOENT until a re-lookup. The `has_handle` guard only covers open file handles, not kernel dentry references.

Consider also skipping eviction when `n_lookup > 0`, or documenting why transient ENOENT for kernel-referenced-but-evicted inodes is acceptable here.

curvine-fuse/src/fs/dcache/op_state.rs
```diff
@@ -0,0 +1,24 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Default, Deserialize, Serialize)]
+#[repr(i8)]
+pub enum OpState {
```
szbr00: **[Nit]** `OpState` is defined here but not referenced anywhere in the crate (the tree uses `Lifecycle` instead). Either wire it in or drop the file to avoid dead code.

curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -0,0 +1,224 @@
...
+        reply: FuseResponse,
+    ) -> FuseResult<()> {
+        let reader = match &self.reader {
+            Some(v) => v,
+            None => return err_fuse!(libc::EIO),
+        };
+
+        if let Some(writer) = state.find_writer(self.ino).await {
+            let writer_ver = writer.write_ver();
+            if self.read_ver.get() != writer_ver {
```
szbr00, line 80: **[Minor]** The read-after-write reconciliation reads `self.read_ver.get()`, then (if stale) flushes, builds a new reader, `reader.replace(...)`, and `self.read_ver.set(writer_ver)` — all on `&self` with no CAS. Two concurrent reads on the same fh could both observe staleness and double-flush/replace. If concurrent reads per handle are possible, consider a guard or compare-and-swap on `read_ver`; if they are always serialized, a comment noting that invariant would help.

szbr00 (commented):
## Code Review

Reviewed the core of this refactor at head `273c30b`: the new `dcache` module (`dir_tree.rs`, `inode.rs`, `dir_entry.rs`, `op_state.rs`, `mod.rs`), `state/backend_handle.rs`, `state/node_state.rs`, the `curvine_file_system.rs` dispatcher, and `fuse_conf.rs`.

### Summary
This is a solid structural refactor — a single-owner `DirTree` inode table, clear separation of `Inode` / `DirEntry` / `BackendHandle`, and good unit-test coverage. The main risk is one concrete correctness bug that a bundled test actually contradicts, plus the fact that the test plan is entirely `TODO`.

**[Process] Please build and run tests before merge.** The test plan lists `cargo build -p curvine-fuse` and all test suites as TODO. For a ~6k-line rewrite of the FUSE inode table this is high risk. Concretely, the bundled unit test `lookup_existing_server_id_updates_inode_path_without_reinsert` appears to contradict the current `DirTree::lookup` implementation (see inline comment on `dir_tree.rs`), which strongly suggests the suite has not been run against this revision. Recommend `cargo build -p curvine-fuse` + `cargo test -p curvine-fuse` / `-p curvine-common`.

### Findings: 4 inline
- Major: 1 — `DirTree::lookup` reinserts a fresh `Inode` on an existing server id, resetting ref/lookup counters and dropping cached children / dirty state.
- Minor: 2 — eviction can drop kernel-referenced inodes (`n_lookup` ignored); non-atomic read-version reconciliation in `BackendHandle::read`.
- Nit: 1 — `OpState` is dead code.

See inline comments for details. The headline item is the `lookup` reinsert (`dir_tree.rs`), which I'd treat as a merge blocker.

szbr00 (approved):

PR #1042, opened 2026-07-09
refactor(fuse): make FuseConf time fields unit-explicit and default meta cache on

yangyang233333:
## Summary

Make the `FuseConf` kernel-cache timeout fields unit-explicit (seconds `f64` → milliseconds `u64`), remove two dead config-only fields, and default the userspace metadata cache on.

## Changes

| Area | Change |
|---|---|
| `curvine-common/src/conf/fuse_conf.rs` | Rename `attr_timeout`/`entry_timeout`/`negative_timeout` (`f64` seconds) → `*_ms` (`u64` milliseconds); `init()` uses `Duration::from_millis`; add `DEFAULT_NEGATIVE_TIMEOUT_MS`; default `enable_meta_cache = true`; drop dead `ac_attr_timeout` / `ac_attr_timeout_set` |
| `curvine-fuse/src/cli/mount_args.rs` | Rename CLI flags to `--*-timeout-ms`; remove `--ac-attr-timeout-ms` |
| `etc/curvine-cluster.toml`, `curvine-docker/deploy/example/conf/curvine-cluster.toml` | Document renamed keys in `[fuse]` templates |

## BREAKING CHANGE

The kernel-cache timeout fields are renamed and their unit changes from **seconds (f64)** to **milliseconds (u64)**:

```
attr_timeout     = 1.0  ->  attr_timeout_ms     = 1000
entry_timeout    = 1.0  ->  entry_timeout_ms    = 1000
negative_timeout = 0.0  ->  negative_timeout_ms = 0
```

The old keys are **no longer recognized and are silently ignored** (`FuseConf` is `#[serde(default)]` without `deny_unknown_fields`), so stale configs fall back to defaults. **No serde alias is added on purpose**: a second→millisecond alias would reinterpret an old value like `5.0` as `5ms` instead of `5000ms`, which is worse than a clean fallback. **Migrate by renaming to the `*_ms` keys and multiplying the old value by 1000.**

### Removed dead fields

`ac_attr_timeout` / `ac_attr_timeout_set` mirror libfuse's `auto_cache` mechanism. curvine-fuse talks the `/dev/fuse` protocol directly and never implements or reads them — they were config-only no-ops (like the previously removed `node_cache_size`). No migration needed.

## Test verified

- `cargo check -p curvine-common -p curvine-fuse` — passes
- `cargo fmt -p curvine-common -p curvine-fuse -- --check` — passes
- `cargo test -p curvine-common --lib conf::fuse_conf` — 7 passed

### DIFF
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -28,10 +28,9 @@ use std::time::Duration;
 //
 // 1. Kernel-side caching (the kernel caches on our behalf, controlled via the
 //    FUSE reply `entry_valid` / `attr_valid` fields):
-//    - `entry_timeout`   -> how long the kernel trusts a name->inode lookup.
-//    - `negative_timeout`-> how long the kernel caches a negative lookup (ENOENT).
-//    - `attr_timeout`    -> how long the kernel trusts cached file/dir attributes.
-//    - `ac_attr_timeout*`-> attr-cache timeout used for auto-cache refresh.
+//    - `entry_timeout_ms`   -> how long the kernel trusts a name->inode lookup.
+//    - `negative_timeout_ms`-> how long the kernel caches a negative lookup (ENOENT).
+//    - `attr_timeout_ms`    -> how long the kernel trusts cached file/dir attributes.
 //    These trade metadata freshness for fewer upcalls into user space.
 //
 // 2. User-side caching (maintained inside curvine-fuse itself):
@@ -116,33 +115,22 @@ pub struct FuseConf {
     // The default value is true
     pub read_dir_fill_ino: bool,
 
-    // Name search cache time.
+    // Name search cache time, in milliseconds.
     // After performing a name search, if the same name is requested again, the kernel will check the cache first.
     // If the buffer record is still valid, the cache result will be returned directly, unlike user space for requests.
-    // Default 1.0 seconds
-    pub entry_timeout: f64,
+    // Default 1000ms (1 second). Sub-second granularity is supported (e.g. 500 = 0.5s).
+    pub entry_timeout_ms: u64,
 
-    // The timeout (in seconds) of cache negative lookups.This means that if the file does not exist (find returns ENOENT)
+    // The timeout (in milliseconds) of cache negative lookups. This means that if the file does not exist (find returns ENOENT)
     // Then the search will only be redone after the timeout, and the file/directory will be assumed to not exist before this.
-    // The default value is 0.0 seconds, which means cache negative lookup is disabled.
-    pub negative_timeout: f64,
+    // The default value is 0ms, which means cache negative lookup is disabled.
+    pub negative_timeout_ms: u64,
 
-    // Cache time for file and directory attributes.
+    // Cache time for file and directory attributes, in milliseconds.
     // This means that after a file or directory attribute search, if the same attribute is requested again, the kernel will first check the cache.
     // If the record in the cache is still valid (i.e. the timeout time has not exceeded), the cached result will be returned directly without making a request to the user space again
-    // Default is 1.0 seconds.
-    pub attr_timeout: f64,
-
-    // Parameters are used to specify the file attribute cache timeout for automatic cache refresh.
-    // By default, it is set to the same value as attr_timeout (i.e. 1.0 seconds).
-    //
-    // When the file is opened, if the automatic cache (auto_cache) needs to refresh the file data, the kernel will check whether the cache of the file attributes has expired.
-    // If the cache record of file attributes is still valid (i.e. the timeout time has not exceeded), the automatic cache will refresh the file data.
-    pub ac_attr_timeout: f64,
-
-    // Parameters are used to specify the file attribute cache timeout for automatic cache refresh.
-    // By default, it is set to the same value as attr_timeout (i.e. 1.0 seconds).
-    pub ac_attr_timeout_set: f64,
+    // Default is 1000ms (1 second). Sub-second granularity is supported (e.g. 500 = 0.5s).
+    pub attr_timeout_ms: u64,
 
     // Parameters are used to specify whether the file system should remember the opened files and directories.
     // By default, the FUSE file system clears the cache when a file or directory is closed.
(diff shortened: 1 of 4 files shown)


curvine-common/src/conf/fuse_conf.rs
```diff
@@ -364,18 +354,16 @@ impl Default for FuseConf {
...
+            negative_timeout_ms: FuseConf::DEFAULT_NEGATIVE_TIMEOUT_MS,
+            attr_timeout_ms: FuseConf::DEFAULT_ATTR_TIMEOUT_MS,
             remember: false,
             web_port: ClusterConf::DEFAULT_FUSE_WEB_PORT,
 
             max_background: 256,
             congestion_threshold: 192,
 
-            enable_meta_cache: false,
+            enable_meta_cache: true,
```
szbr00: **[Major - behavior]** This flips the default from `false` to `true`, which silently enables the userspace metadata cache for every existing deployment that doesn't explicitly set `enable_meta_cache`. That flag gates real caching paths (`node_state.rs` lookup/attr, `curvine_file_system.rs` getattr/readdir/lookup), so clients will now serve cached metadata with a TTL of up to `meta_cache_ttl` (default 120s) — a meaningful change to metadata-consistency semantics for a distributed FS.

Two concerns:

1. This behavior change isn't in the "BREAKING CHANGE" section (which only documents the field rename). It's arguably a bigger user-visible change than the rename and should be called out explicitly, with the staleness window and how to revert (`enable_meta_cache = false`).
2. Bundling a behavioral default flip into an otherwise-mechanical rename hurts reviewability and bisectability. Consider splitting the default change into its own commit/PR so it can be evaluated (and reverted) independently.

Please also confirm the metadata cache layer is considered production-ready to be on by default.

curvine-common/src/conf/fuse_conf.rs
```diff
@@ -116,33 +115,22 @@ pub struct FuseConf {
...
     pub read_dir_fill_ino: bool,
 
-    // Name search cache time.
+    // Name search cache time, in milliseconds.
     // After performing a name search, if the same name is requested again, the kernel will check the cache first.
     // If the buffer record is still valid, the cache result will be returned directly, unlike user space for requests.
-    // Default 1.0 seconds
-    pub entry_timeout: f64,
+    // Default 1000ms (1 second). Sub-second granularity is supported (e.g. 500 = 0.5s).
+    pub entry_timeout_ms: u64,
```
bigbigxu, line 122: Is it better to use the DurationUnit type directly?

curvine-common/src/conf/fuse_conf.rs
yangyang233333: Thanks — agreed on both points. I've split the behavioral default flip out of this PR:

- This PR is now a **pure mechanical rename** (`*_timeout` → `*_timeout_ms`, seconds `f64` → ms `u64`, plus removing the dead `ac_attr_timeout*` no-ops). `enable_meta_cache` is back to its previous default (`false`) here, so this PR no longer changes metadata-consistency semantics.
- The `false` → `true` default flip now lives in its own PR: #1043. It documents the staleness window (up to `meta_cache_ttl`, default 120s), the revert (`enable_meta_cache = false`), and raises the production-readiness question explicitly so it can be evaluated and reverted independently.

curvine-common/src/conf/fuse_conf.rs
```diff
@@ -364,11 +354,9 @@ impl Default for FuseConf {
...
             gid: sys::get_gid(),
             read_dir_fill_ino: true,
-            entry_timeout: FuseConf::DEFAULT_ENTRY_TIMEOUT,
-            negative_timeout: 0.0,
-            attr_timeout: FuseConf::DEFAULT_ATTR_TIMEOUT,
-            ac_attr_timeout: FuseConf::DEFAULT_ATTR_TIMEOUT,
-            ac_attr_timeout_set: FuseConf::DEFAULT_ATTR_TIMEOUT,
+            entry_timeout_ms: FuseConf::DEFAULT_ENTRY_TIMEOUT_MS,
+            negative_timeout_ms: FuseConf::DEFAULT_NEGATIVE_TIMEOUT_MS,
+            attr_timeout_ms: FuseConf::DEFAULT_ATTR_TIMEOUT_MS,
```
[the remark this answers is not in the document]
yangyang233333: Resolved: the default flip has been removed from this PR, so `FuseConf::default()` intentionally keeps `enable_meta_cache: false` and now matches the (rename-only) title/description. The default change is handled separately in #1043.

curvine-common/src/conf/fuse_conf.rs
```diff
@@ -243,9 +233,9 @@ impl FuseConf {
     pub const DEFAULT_MAX_READAHEAD_KB: u32 = 1024;
 
     pub fn init(&mut self) -> CommonResult<()> {
-        self.attr_ttl = Duration::from_secs_f64(self.attr_timeout);
-        self.entry_ttl = Duration::from_secs_f64(self.entry_timeout);
-        self.negative_ttl = Duration::from_secs_f64(self.negative_timeout);
+        self.attr_ttl = Duration::from_millis(self.attr_timeout_ms);
+        self.entry_ttl = Duration::from_millis(self.entry_timeout_ms);
+        self.negative_ttl = Duration::from_millis(self.negative_timeout_ms);
```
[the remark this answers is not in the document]
yangyang233333: Good suggestion. I'll add a focused deserialization test covering (1) parsing of the new `*_timeout_ms` keys and (2) legacy second-based keys (`attr_timeout` etc.) being ignored via `#[serde(default)]` and falling back to the `*_MS` constants — which also guards against a unit/alias regression being reintroduced.

curvine-common/src/conf/fuse_conf.rs
yangyang233333: Thanks for the push here — I dug into the codebase to decide, and I'd actually like to keep these as raw `u64` milliseconds rather than `DurationUnit` strings, for consistency.

`u64` ms is the dominant convention for timeout config fields in this repo, e.g.:
- `ClientConf`: `conn_timeout_ms`, `rpc_timeout_ms`, `data_timeout_ms`, `pipeline_timeout_ms` (all `u64`)
- `JournalConf`: `conn_timeout_ms`, `io_timeout_ms`, `raft_tick_interval_ms` (all `u64`)
- `FilesystemConf` (libsdk): `conn_timeout_ms`, `rpc_timeout_ms`, `data_timeout_ms` (all `u64`)

So `attr_timeout_ms` / `entry_timeout_ms` / `negative_timeout_ms` as `u64` ms actually line up with how timeouts are configured everywhere else. The `DurationUnit` string form (`meta_cache_ttl`, `node_cache_timeout`) is the minority pattern here and tends to be used for coarser, human-scale durations (`"1h"`, `"120s"`), whereas these three are sub-second kernel knobs where an explicit integer ms is unambiguous and matches the neighboring `*_timeout_ms` fields.

Given that, I'd lean toward leaving them as `u64` ms in this PR. Happy to reconsider if you feel strongly, but I think it's the more consistent choice overall.

szbr00 (approved):

PR #1043, opened 2026-07-09
refactor(fuse): enable userspace metadata cache by default

yangyang233333:
## Summary

Flip the default of `enable_meta_cache` from `false` to `true`, so mounts that do not explicitly configure it now use the userspace metadata cache.

This is intentionally a standalone, single-line change: the behavioral default flip is kept separate from any mechanical/config refactors so it can be evaluated, bisected, and reverted independently.

## Behavior change (user-visible)

Mounts without an explicit `enable_meta_cache` setting will now serve filesystem metadata (lookup / getattr / readdir, plus file blocks) from the userspace `MetaCache` instead of issuing an RPC to the master every time.

- **Staleness window:** up to `meta_cache_ttl` (default `120s`). A metadata change made by *another* client can take up to that long to become visible on a mount that has the entry cached.
- **Trade-off:** fewer master RPCs / lower metadata latency, at the cost of the freshness guarantee above.
- **Scope of the cache:** gated on `enable_meta_cache` at the real caching paths — `node_state.rs` (lookup/attr/blocks) and `curvine_file_system.rs` (getattr/readdir/lookup).

## How to revert to previous behavior

```toml
[fuse]
enable_meta_cache = false
```

## Consistency notes

- The cache is **bounded**: `meta_cache_capacity` (default 100000 entries).
- **Local** mutating operations invalidate the affected path synchronously (setattr, setxattr/removexattr, resize, mkdir, and the other write paths call `invalidate_cache`). The staleness window therefore applies to **cross-client** changes, not to writes issued through the same mount.

## Changes

| File | Change |
| --- | --- |
| `curvine-common/src/conf/fuse_conf.rs` | `Default::default()` for `FuseConf` sets `enable_meta_cache: true` (was `false`) |

## Open question for reviewers

Please confirm the metadata cache layer is considered **production-ready to be on by default**. If there are known correctness or eviction gaps, we can hold this PR and keep the opt-in default until they're resolved.

### DIFF
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -375,7 +375,7 @@ impl Default for FuseConf {
             max_background: 256,
             congestion_threshold: 192,
 
-            enable_meta_cache: false,
+            enable_meta_cache: true,
             meta_cache_capacity: 100000,
             meta_cache_ttl: "120s".to_string(),
 


curvine-common/src/conf/fuse_conf.rs
```diff
@@ -375,7 +375,7 @@ impl Default for FuseConf {
             max_background: 256,
             congestion_threshold: 192,
 
-            enable_meta_cache: false,
+            enable_meta_cache: true,
```
szbr00, line 378: Flipping this default changes the out-of-the-box consistency model for metadata. With the cache on, `lookup`/`getattr`/`readdir`/blocks are served from the userspace `MetaCache`, and there is no cluster-wide invalidation — a metadata change made by another client (or a non-FUSE client) only becomes visible after `meta_cache_ttl` (default 120s). So the default behavior silently shifts from "read-your-cluster-writes" to eventual consistency bounded by 120s for any multi-client / multi-mount deployment. This matches the open question in the PR description: please confirm this eventual-consistency default is acceptable for the general deployment case, or consider keeping it opt-in until cross-client invalidation exists. At minimum, this staleness window should be called out prominently in user-facing docs, not just the PR body.

szbr00, line 378: Enabling this by default surfaces a local (same-mount) staleness gap: on child mutations (`create`/`unlink`/`mkdir`/`rmdir`), `invalidate_cache` only clears the parent's `list` slot via `invalidate_list`, preserving the parent's cached `status`. A subsequent `getattr` on the parent directory can therefore return a stale `mtime`/`ctime` even though the change was made through the same mount. Consider fully invalidating (or refreshing) the parent `status` on child add/remove before making this the default.

szbr00, line 378: Another same-mount gap that becomes default-on with this flip: `write()` does not invalidate the cached status; invalidation only happens at `flush`/`release`/`fsync`. Between `open(write)` and flush, a `getattr`/`lookup` on the file being written can return stale size/mtime from a status entry populated after open. Worth confirming this is acceptable (or invalidating on write) before enabling by default.

szbr00, line 378: Note: `get_cached_list` is currently never called — `readdir` uses a live backend stream (`DirHandle`), not the list cache. So the parent `invalidate_list` performed on every mutation provides no readdir-freshness benefit today. Not a blocker for this flip, but the list-cache path is effectively dead and the parent-list invalidation is a no-op; either wire it up or drop it to avoid implying a guarantee that isn't there.

bigbigxu:
I think metadata cache is valuable, but I would prefer to keep it opt-in for now rather than enabling it by default.

The main concern is consistency: enabling it by default improves lookup/getattr/readdir performance, but it can make metadata updates from other clients less immediately visible. That is a risky default for general-purpose FUSE usage.

I’d suggest we keep `fuse.enable_meta_cache=false` by default, document when users should enable it, and maybe revisit the default after we have clear benchmarks, consistency notes, and enough validation around invalidation behavior.

PR #1046, opened 2026-07-09
fix(client): make block client pool session-safe

jlon:
# Summary

Fix block client reuse so local block open/read-write/commit stays on the same worker connection, and prevent closed raw RPC connections from being returned to or reused by the block client pool.

# Issue Describe / Design

Related to client-side small-file block connection reuse.

- Root cause 1: local short-circuit block read/write opens a stateful worker-side block handler, but `complete`/`cancel` reacquired a block client from the pool. Under reuse, the commit could be sent on a different worker connection than the open request.
- Root cause 2: raw `RpcClient` returned `Connection ... is closed` on send/receive EOF without marking `ClientState` closed, so the pool could treat an EOF connection as active.
- Root cause 3: reused worker write handlers did not reset `is_commit` on a new open, so a reused handler could keep stale commit state.

Fix approach:
- Keep the `BlockClient` acquired during local block open inside `BlockReaderLocal`/`BlockWriterLocal` and use that same client for commit/cancel.
- Filter inactive clients when acquiring/releasing pooled block clients.
- Mark raw RPC clients closed on real send/receive errors and EOF.
- Reset worker write/batch-write commit state when a new block open starts.

# Changes

| `curvine-client/src/block/block_reader_local.rs` | Store the opened `BlockClient` and use it for read commit. | Preserves worker-side session affinity for local reads. |
| `curvine-client/src/block/block_writer_local.rs` | Store the opened `BlockClient` and use it for write commit/cancel. | Preserves worker-side session affinity for local writes. |
| `curvine-client/src/block/block_client_pool.rs` | Drop inactive pooled clients on acquire/release. | Prevents reuse of closed block connections. |
| `curvine-client/src/block/block_client.rs` | Expose active-state check from the underlying RPC client. | Enables pool health filtering. |
| `orpc/src/client/rpc_client.rs` | Mark raw RPC clients closed on send/receive error or EOF. | Keeps RPC state consistent with network reality. |
| `curvine-server/src/worker/handler/write_handler.rs` | Reset commit state on new write open. | Avoids stale state across reused worker handlers. |
| `curvine-server/src/worker/handler/batch_write_handler.rs` | Reset commit state on new batch write open. | Avoids stale state across reused batch handlers. |

# Test verified

| `CARGO_TARGET_DIR=/tmp/curvine-client-pool-hotfix-target cargo build --release -p curvine-server -p curvine-client -p curvine-tests --jobs 2` | PASS | |
| `CARGO_TARGET_DIR=/tmp/curvine-client-pool-hotfix-target cargo clippy -p orpc -p curvine-client -p curvine-server --all-targets --jobs 2 -- --deny=warnings --allow clippy::uninlined-format-args` | PASS | |
| Local E2E, pool enabled: `fs.write` 2048 files, 64 threads, 4KB files | PASS | Cold local cluster, `block_size_str=1MB`. |
| Local E2E, pool enabled: `fs.write` 4096 files, 64 threads, 4KB files | PASS | Warm local cluster. |
| Local E2E, pool enabled: `fs.read` 4096 files, 64 threads, 4KB files | PASS | Warm local cluster. |
| Local E2E, no-pool control: `fs.write`/`fs.read` 4096 files, 64 threads, 4KB files | PASS | Warm local cluster. |

Note: 4096-file cold-start writes can still hit local startup/connection pressure in this environment even with the block pool disabled; that is not caused by this pool-session fix and should be tracked separately.

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-client/src/block/block_client_pool.rs
+++ b/curvine-client/src/block/block_client_pool.rs
@@ -112,12 +112,12 @@ impl BlockClientPool {
             let idle_duration = current_time.saturating_sub(client.uptime());
             self.cur_idle_size.decr();
 
-            if idle_duration < self.idle_time_ms {
+            if idle_duration < self.idle_time_ms && client.is_active() {
                 debug!("acquiring connection for worker {} from pool", addr);
                 return Ok(client);
             } else {
                 debug!(
-                    "connection for worker {} expired (idle={}ms), removing from pool",
+                    "connection for worker {} is inactive or expired (idle={}ms), removing from pool",
                     addr, idle_duration
                 );
                 client.clear_pool();
@@ -140,7 +140,13 @@ impl BlockClientPool {
             Some(pool) if pool.id == self.id => {
                 let addr = client.worker_addr().clone();
 
-                if self.cur_idle_size.get() >= self.max_idle_size {
+                if !client.is_active() {
+                    debug!(
+                        "connection for worker {} is closed, not returning to pool",
+                        client.worker_addr()
+                    );
+                    client.clear_pool();
+                } else if self.cur_idle_size.get() >= self.max_idle_size {
                     debug!(
                         "pool full(max={}), closing connection for worker {}",
                         self.max_idle_size,
--- a/curvine-client/src/block/block_client.rs
+++ b/curvine-client/src/block/block_client.rs
@@ -93,7 +93,13 @@ impl BlockClient {
 
     pub async fn rpc(&self, msg: Message) -> FsResult<Message> {
         let client = try_option_ref!(self.client);
-        let rep_msg = client.timeout_rpc(self.timeout, msg).await?;
+        let rep_msg = match client.timeout_rpc(self.timeout, msg).await {
+            Ok(rep_msg) => rep_msg,
+            Err(err) => {
+                client.set_closed();
+                return Err(FsError::from(err));
+            }
+        };
         match rep_msg.check_error_ext::<FsError>() {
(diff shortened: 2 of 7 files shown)


curvine-client/src/block/block_client_pool.rs
```diff
@@ -140,7 +140,13 @@ impl BlockClientPool {
             Some(pool) if pool.id == self.id => {
                 let addr = client.worker_addr().clone();
 
-                if self.cur_idle_size.get() >= self.max_idle_size {
+                if !client.is_active() {
```
szbr00, line 143: `is_active()` still returns true after an RPC timeout. `BlockClient::rpc` uses `RpcClient::timeout_rpc`, and when the Tokio timeout expires it cancels the raw send/receive future without calling `set_closed()`. The connection may then contain a late response for the timed-out request, but this release path returns it to the pool and the next session can consume that stale response. Please mark raw clients closed on timeout (and other protocol-desynchronizing errors) or ensure `BlockClient` removes the client from the pool whenever `timeout_rpc` fails.

jlon: Fixed in 5453db5: `BlockClient::rpc` now marks the underlying `RpcClient` closed when `timeout_rpc` returns a transport/protocol error, so timed-out raw connections are not returned to the pool for reuse.

szbr00 (approved):

PR #1047, opened 2026-07-09
feat(spdk): add orphaned lifecycle with poll and sweep for late spdk callback safety

thuongle2210:
# Description
Introduce the orphaned map as a separate lifecycle layer between `dead_qpairs` and final deallocation. When a qpair is unregistered or fails, its `QpairState` is moved to the orphaned map instead of being immediately freed. This keeps the stale `CallbackCtx` entries alive for late SPDK callbacks that may fire during `free_io_qpair`. A new `poll_and_sweep` function consolidates the duplicated poll-and-fail logic and moves failed qpairs into the orphaned map.

# Key Changes
- `handle_unregister`: Now signals pending entries with `-ESHUTDOWN`, moves them to stale, then moves the entire `QpairState` to the orphaned map (instead of freeing it). Handles collision by merging stale lists if the qpair was already orphaned.
- `poll_and_sweep`: Extracted from duplicated inline poll-fail logic in Active and Idle paths. Polls all active qpairs; on error, calls `force_complete_qpair` and moves the `QpairState` from `dead_qpairs` to the orphaned map. Accepts a `context` string for differentiated logging ("poller" vs "keep-alive").
- Shutdown path: Shutdown path: Drains remaining dead_qpairs into the orphaned map instead of reclaiming. Reclamation is deferred to future works.
- Poisoned lock safety: `force_complete_qpair` is called before acquiring the orphaned lock in `poll_and_sweep`, so I/O completion is guaranteed even if the lock is poisoned. All `if let Ok(guard) = orphaned.lock()` sites are replaced with `into_inner()` recovery instead of silently skipping cleanup.
 - Add correctness tests:
  - `handle_unregister_orphans_pending_entries`: Happy path: 2 pending entries are signaled with `-ESHUTDOWN`, moved to stale, qpair state moved to orphaned.
  - `handle_unregister_orphan_collision_merges_entries`: Collision: pre-populated orphaned entry + new pending entry are merged into stale lists.
  - `force_complete_qpair_preserves_existing_stale`: Stale entries are not re-signaled; pending entries are moved and signaled with `-EIO`.

# Impact
- Enables safe `free_io_qpair` in the env layer without racing with late SPDK callbacks.
- Keeps `QpairState` alive in memory until explicit reclamation, preventing use-after-free during controller detach.
- Force-complete before orphaned lock guarantees no stranded I/Os even if the mutex is poisoned.
# Future Work

### DIFF
--- a/orpc/src/io/spdk_poller.rs
+++ b/orpc/src/io/spdk_poller.rs
@@ -732,6 +732,9 @@ struct QpairState {
     /// Force-completed entries kept alive for late callbacks. Freed by reclaim_stale().
     stale: Vec<*mut CallbackCtx>,
 }
+// SAFETY: QpairState moves from the poller thread into the orphaned
+// Mutex after force_complete_qpair sets dead=true, preventing late
+// callbacks from following the stale raw pointers.
 unsafe impl Send for QpairState {}
 
 impl QpairState {
--- a/orpc/src/io/spdk_poller.rs
+++ b/orpc/src/io/spdk_poller.rs
@@ -161,6 +161,9 @@ pub struct SpdkPoller {
     /// Whether poller is blocked on eventfd (idle). Bdevs check this to
     /// skip eventfd write syscall when poller is already active.
     is_sleeping: Arc<AtomicBool>,
+    /// Orphaned QpairState entries keyed by qpair address.
+    /// Kept alive for late SPDK callbacks during free_io_qpair.
+    orphaned: Arc<Mutex<HashMap<usize, Box<QpairState>>>>,
 }
 
 impl SpdkPoller {
@@ -178,10 +181,20 @@ impl SpdkPoller {
         let eventfd_raw = eventfd.as_raw_fd();
         let eventfd_arc = Arc::new(eventfd);
 
+        let orphaned = Arc::new(Mutex::new(HashMap::new()));
+        let orphaned_clone = orphaned.clone();
+
         let handle = std::thread::Builder::new()
             .name("spdk-poller".to_string())
             .spawn(move || {
-                Self::poller_loop(rx, shutdown_clone, is_sleeping_clone, eventfd_raw, config);
+                Self::poller_loop(
+                    rx,
+                    shutdown_clone,
+                    is_sleeping_clone,
+                    eventfd_raw,
+                    config,
+                    orphaned_clone,
+                );
             })
             .expect("Failed to spawn SPDK poller thread");
 
@@ -191,6 +204,7 @@ impl SpdkPoller {
             shutdown,
             handle: Some(handle),
             is_sleeping,
+            orphaned,
         }
     }
 
@@ -257,6 +271,7 @@ impl SpdkPoller {
         is_sleeping: Arc<AtomicBool>,
         eventfd: RawFd,
         mut config: PollerConfig,
+        orphaned: Arc<Mutex<HashMap<usize, Box<QpairState>>>>,
     ) {
         let mut active_qpairs: Vec<*mut spdk_ffi::spdk_nvme_qpair> = Vec::new();
         let active_ctrlrs: Vec<CtrlHandle> = std::mem::take(&mut config.ctrlrs);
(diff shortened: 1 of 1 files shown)


thuongle2210:
Dear @szbr9486, could you pls help me review this change?

orpc/src/io/spdk_poller.rs
```diff
@@ -178,10 +178,19 @@ impl SpdkPoller {
         let eventfd_raw = eventfd.as_raw_fd();
         let eventfd_arc = Arc::new(eventfd);
 
+        let orphaned = Arc::new(Mutex::new(HashMap::new()));
```
szbr00, line 181: The `orphaned` map is created locally here in `start()` and only moved into the poller thread — it's never stored on `SpdkPoller` nor exposed through an accessor. Combined with `reclaim_stale` now being `#[cfg(test)]`, nothing can reclaim orphaned entries in production.

Since `handle_unregister` moves a `QpairState` into `orphaned` on *every* qpair unregister (normal bdev/controller teardown, not just shutdown), and `poll_and_sweep` does the same on every poll failure, the map grows unbounded over the process lifetime. Dropping the map on thread exit won't help either: `Box<QpairState>` has no `Drop` that frees the raw `*mut CallbackCtx` entries in `pending`/`stale`, so each one leaks.

The PR description states "the env layer is responsible for calling `reclaim_stale` after `qpair_pool::drain_all`," but that path isn't wired up in this PR — the map is inaccessible and the method is test-gated. Please either expose `orphaned` (store it on `SpdkPoller` + accessor, keep `reclaim_stale` non-test) and land the env-layer reclamation here, or clarify how orphaned entries get reclaimed.

orpc/src/io/spdk_poller.rs
```diff
@@ -641,6 +656,50 @@ impl SpdkPoller {
...
+                } else {
+                    None
+                }
+            })
+            .collect();
+        if err_keys.is_empty() {
+            return;
+        }
+        active_qpairs.retain(|&qp| !err_keys.contains(&(qp as usize)));
+        if let Ok(mut guard) = orphaned.lock() {
```
szbr00: `force_complete_qpair` is now only invoked inside `if let Ok(mut guard) = orphaned.lock()`. If the mutex is ever poisoned, the failed qpairs are removed from `active_qpairs` but their stranded I/Os are never force-completed, so the callers block forever. The previous code force-completed unconditionally.

Consider force-completing before acquiring the lock (it only touches `dead_qpairs`), so cleanup can't be skipped by a lock failure. The same "silently skip on lock failure" pattern also appears in `handle_unregister` and the shutdown drain.

orpc/src/io/spdk_poller.rs
```diff
@@ -650,8 +709,10 @@ struct QpairState {
     /// Force-completed entries kept alive for late callbacks. Freed by reclaim_stale().
     stale: Vec<*mut CallbackCtx>,
 }
+unsafe impl Send for QpairState {}
```
szbr00, line 738: `unsafe impl Send for QpairState {}` has no SAFETY comment, unlike the other `unsafe impl Send` blocks in this file. Since `QpairState` now crosses threads via the `orphaned` map and holds raw `*mut CallbackCtx`, please document why this is sound (e.g., ownership is transferred only after late SPDK callbacks have quiesced).

orpc/src/io/spdk_poller.rs
thuongle2210: I updated the PR description with a small change. I intentionally leak it and will handle the reclamation in a future PR.

orpc/src/io/spdk_poller.rs
thuongle2210: This is unlikely in practice. the orphaned lock is only held by the poller thread and the critical section has no unwrap()/expect() SPDK callbacks never acquire the orphaned lock, so only OOM during a HashMap/Vec allocation inside the lock could trigger it, which kills the process anyway. However, as defense-in-depth, I refactored by 
- 1: moving force_complete_qpair before the lock in poll_and_sweep
- 2: replacing if let Ok(guard) with into_inner() recovery in all three sites.

orpc/src/io/spdk_poller.rs
thuongle2210: I added the explanation.

szbr00 (approved):

PR #1050, opened 2026-07-11
fix(client): preserve per-file batch write outcomes (#1045)

### LINKED ISSUE #1045: [BUG]: Batch small-file writes collapse per-file outcomes and leak on item-local failures
### Describe the bug

The small-file batch write path does not preserve the result of each logical file write from Worker I/O through Master metadata completion.

A batch API must be a vectorized execution of the existing single-file write path. For every input file `i`, the following invariant must hold:

```text
batch_write(files)[i] == write_string(files[i])
```

This equality covers the success/failure decision and externally observable side effects. Batching may reduce RPC and scheduling overhead, but it must not introduce a different success condition or collapse independent file outcomes.

There are two related defects on `main`:

1. **Worker batch completion short-circuits on a block-local failure.**

   `BatchWriteHandler::complete_batch` flushes all files and then finalizes/aborts all blocks with `?`. A flush failure terminates the RPC before any block is completed. A finalize failure terminates the loop after earlier blocks may already have been finalized, leaving the failing and later blocks unprocessed. The response therefore cannot report one outcome per block, and unprocessed `Writing` blocks can retain metadata, backing storage, and reserved capacity.

2. **Per-block and per-file outcomes are collapsed before they reach the original caller.**

   `BlocksBatchCommitResponse.results` and `CompleteFilesBatchResponse.results` are ordered per-item results. Ignoring them, or immediately converting any `false` into one aggregate `FsResult<()>`, prevents the caller from knowing which original small-file writes succeeded. If a batch partially succeeds, the caller can neither retry only failed files nor safely avoid retrying successful files.

The problem is therefore end-to-end; changing only the Worker loop is necessary for isolation and cleanup but is not sufficient.

### To reproduce

1. Submit at least two independent small files through `write_batch_string`.
2. Inject a write/flush/finalize failure for one file's block while allowing the other file's block to succeed.
3. Complete the batch.
4. Observe one or both of the following on `main`:
   - The Worker RPC stops at the first block-local error and does not return a complete ordered result vector.
   - After per-block Worker results are produced, the client collapses them into one aggregate error, so the original caller cannot distinguish the successful file from the failed file.

### Required behavior

For each input file, batch execution must use the same success condition as the standalone file writer:

1. The file data write succeeds.
2. The file/block flush succeeds where the standalone path requires it.
3. Every required Worker replica completes that file's block successfully.
4. Master `complete_file` succeeds for that file.

Only when all four conditions hold is that file successful.

Additional requirements:

- A failure for file `i` must not stop processing file `j`.
- Worker block-local failures must produce an ordered result for every block and perform the same best-effort cleanup as the standalone path.
- Client layers must preserve ordered results instead of collapsing them into `FsResult<()>` before the original file boundary.
- Replica results must be collected from all Workers. File `i` is Worker-complete only when every required replica reports success for item `i`.
- Only Worker-complete files may be submitted to Master completion. Master results must be mapped back to the original input indices.
- The public batch result must contain one structured outcome per original file, in input order.
- Batch-wide `Err` is reserved for failures that cannot be reliably attributed to individual inputs, such as malformed protocol responses or result-count mismatches.
- If local short-circuit flush produces mixed outcomes, the commit protocol must be able to finalize successful items and abort failed items within the same batch; a single batch-wide commit/cancel decision is insufficient for that case.

Explicit cancellation is an operation of its own: a successful per-item abort is a successful cancel operation. It is not a successful file write.

### API expectation

`write_batch_string` currently has no production callers outside tests, so its aggregate-only return type does not need to be preserved. It should return ordered per-file outcomes, for example:

```rust
pub struct BatchWriteOutcome {
    pub path: Path,
    pub result: FsResult<()>,
}

pub async fn write_batch_string(
    &self,
    files: &[(Path, &str)],
) -> FsResult<Vec<BatchWriteOutcome>>;
```

The exact type may differ, but it must preserve structured per-file success/failure information. Error strings are not a substitute for structured results.

### Implementation notes

- Worker batch write/commit handlers should execute item-local state transitions and continue after item-local errors.
- `BlockClient::write_commit_batch` should return the ordered Worker results instead of reducing them to `()`.
- Batch replica aggregation must collect all Worker responses rather than return on the first failed future.
- Master batch completion should receive only files whose Worker replicas completed successfully, then map its subset response back to original indices.
- The existing single-block `WriteHandler::complete` is the behavioral path to compare against, but it must not be described as aborting on flush failure unless the implementation actually does so.
- Mid-session `flush_batch` may still return an RPC-level error while the session remains open; this issue concerns preserving independent logical file outcomes through the terminal batch write/commit flow.

### Test plan

1. Compare batch and standalone outcomes for the same files on the success path.
2. Inject one remote Worker write/flush/finalize failure and verify only the corresponding file fails.
3. Inject one local short-circuit flush failure and verify the failed item is aborted while other items finalize.
4. Verify multi-replica aggregation: one replica failure fails only the corresponding file, while all Worker futures are still collected.
5. Inject one Master `complete_file` failure and verify successful files remain successful and the failed file is reported separately.
6. Verify returned outcomes are the same length and order as the input files, including when `write_batch_string` splits inputs into multiple internal batches or handles large files individually.
7. Add an example-cluster fault-injection E2E scenario that proves the final per-file outcomes reach the original caller.

### Environment

- OS: N/A; the bug is platform-independent.
- Affected branch: `main` and earlier implementations of the small-file batch path.
- Storage: DISK and SPDK semantics must be equivalent.

gongxun0928:
# Summary

Make small-file batch writes a vectorized form of standalone file writes. Every input now receives one ordered `FsResult<()>`, and a failure in one file no longer stops, hides, or changes the outcome of another file.

# Issue Describe / Design

Closes #1045

The old path had two coupled problems:

1. Worker batch open/write/complete processing could stop at the first item-local error, so later blocks were not processed and the response could not describe every block.
2. Client and Master layers collapsed per-block responses into one aggregate `FsResult<()>`, so the original caller could not identify which file succeeded.

The fix treats batch execution as an indexed collection of standalone writes. File `i` is successful only when its create, block allocation, data write/flush, every Worker replica completion, and Master completion all succeed. Item-local failures remain attached to index `i`; batch-wide errors are reserved for failures that cannot be attributed safely.

Key design points:

- Worker open/write/complete loops continue after item-local errors and return one ordered result per block.
- Batch and standalone completion share `WriteHandler::complete_block`, including validation, flush, finalize, and best-effort abort behavior.
- The existing `BlocksBatchCommitRequest` / `BlocksBatchCommitResponse` RPC is retained. An additive `cancel_flags` field allows successful items to finalize and failed items to abort in the same request; the existing `results[i]` reports each requested operation.
- `BatchBlockWriter` groups blocks by their actual assigned Worker, collects all Worker responses, and ANDs replica outcomes back onto original file indices.
- Only Worker-complete files are sent to Master completion, and Master results are restored to the original input order.
- The inode returned by batch create is carried through add-block and complete-file requests, so a rename between phases cannot retarget metadata operations by path.
- Remote completion relies on the standalone Worker completion path; the obsolete remote batch-flush path is removed. Short-circuit local files are still flushed before completion.
- `write_batch_string` returns `FsResult<Vec<FsResult<()>>>`. Oversized inputs continue through `write_string`, while smaller internal batches preserve the original global order.

## Compatibility

- The public Rust return type changes from `FsResult<()>` to `FsResult<Vec<FsResult<()>>>`. The repository has no production caller of this API outside tests.
- Worker commit keeps the same RPC and protobuf message. `cancel_flags` is additive, and a new Worker still accepts the legacy batch-wide `cancel` field. However, a new client must not use mixed per-item completion against an old Worker because the old Worker ignores `cancel_flags`; update client and Worker together for the batch-write path.
- Master response details and inode IDs are additive protobuf fields, but the new client expects ordered structured batch results. A coordinated client/server upgrade is recommended for this currently unused batch API.

# Changes

| Worker write handlers | Process open/write/complete per item; reuse standalone completion; remove obsolete remote batch flush | One block failure no longer short-circuits later blocks |
| Worker protocol/client | Retain the existing batch commit RPC, add per-item `cancel_flags`, and consume ordered `results` | Mixed finalize/abort is expressible without replacing the RPC |
| `BatchBlockWriter` | Group by actual Worker assignment, collect all responses, and AND replica results by original index | Heterogeneous placements and partial replica failures are handled correctly |
| Master batch RPCs | Return structured create/add/complete results and carry `inode_id` after create | Per-file metadata failures remain attributable; rename cannot lose inode identity |
| `CurvineFileSystem` | Return ordered `Vec<FsResult<()>>`, split internal batches safely, and complete only Worker-successful files | Callers can retry only failed files |
| Tests | Cover Worker item isolation, cleanup, grouping, replica aggregation, inode identity, and public partial-failure outcomes | Verifies the full Worker → Client → Master → public-result path |

# Test verified

| `make format` | PASS | Full release formatting and Clippy checks; `bdi.rs` is unchanged |
| `cargo test -p curvine-client -p curvine-server --lib` | PASS | Client: 52 passed; Server: 83 passed, 1 ignored |
| `cargo test -p curvine-server --test worker_test -- --test-threads=1` | PASS | 13 Worker integration tests, including partial completion and cleanup |
| `cargo test -p curvine-tests --test fs_test test_filesystem_end_to_end_operations_on_cluster -- --test-threads=1` | PASS | Includes ordered outcomes and inode identity across rename |
| `cargo test -p curvine-tests --test fs_test test_batch_worker_partial_failure_end_to_end -- --test-threads=1` | PASS | Real Worker validation failure reaches the public result as `[Err, Ok]`; successful file reads back |

# Dependencies

- Related PRs: None
- Related issues: #1045
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/handler/batch_write_handler.rs
+++ b/curvine-server/src/worker/handler/batch_write_handler.rs
@@ -19,9 +19,9 @@ use crate::worker::storage::BlockWriteContext;
 use curvine_common::error::FsError;
 use curvine_common::fs::RpcCode;
 use curvine_common::proto::{
-    BlockWriteRequest, BlockWriteResponse, BlocksBatchCommitRequest, BlocksBatchCommitResponse,
-    BlocksBatchWriteRequest, BlocksBatchWriteResponse, DataHeaderProto, FilesBatchWriteRequest,
-    FilesBatchWriteResponse,
+    BlockBatchOpenResult, BlockWriteRequest, BlockWriteResponse, BlocksBatchCommitRequest,
+    BlocksBatchCommitResponse, BlocksBatchWriteRequest, BlocksBatchWriteResponse,
+    FilesBatchWriteRequest, FilesBatchWriteResponse,
 };
 use curvine_common::utils::ProtoUtils;
 use curvine_common::FsResult;
@@ -33,9 +33,8 @@ use orpc::CommonResult;
 
 pub struct BatchWriteHandler {
     pub(crate) store: BlockStore,
-    pub(crate) context: Option<Vec<WriteContext>>,
+    pub(crate) context: Option<Vec<Option<WriteContext>>>,
     pub(crate) file: Option<Vec<Option<BlockWriteContext>>>,
-    pub(crate) is_commit: bool,
     pub(crate) write_handler: WriteHandler,
 }
 
@@ -46,24 +45,12 @@ impl BatchWriteHandler {
             store,
             context: None,
             file: None,
-            is_commit: false,
             write_handler: WriteHandler::new(store_clone)?,
         })
     }
 
-    fn check_context(context: &WriteContext, expected_req_id: i64) -> FsResult<()> {
-        if context.req_id != expected_req_id {
-            return err_box!(
-                "Request id mismatch, expected {}, actual {}",
-                context.req_id,
-                expected_req_id
-            );
-        }
-        Ok(())
-    }
-
-    fn abort_open_contexts(store: &BlockStore, contexts: &[WriteContext]) {
-        for context in contexts {
+    fn abort_open_contexts(store: &BlockStore, contexts: &[Option<WriteContext>]) {
+        for context in contexts.iter().flatten() {
             if let Err(e) = store.abort_block(&context.block) {
                 log::warn!(
                     "failed to abort batch-opened block {} after open error: {}",
@@ -74,12 +61,19 @@ impl BatchWriteHandler {
         }
     }
 
+    fn abort_batch_state(&mut self) {
+        // Close all open devices before releasing their block allocations.
+        drop(self.file.take());
+        if let Some(contexts) = self.context.take() {
+            Self::abort_open_contexts(&self.store, &contexts);
+        }
+    }
+
     pub fn open_batch(&mut self, msg: &Message) -> FsResult<Message> {
         let header: BlocksBatchWriteRequest = msg.parse_header()?;
-        let mut responses = Vec::with_capacity(header.blocks.len());
+        let mut results = Vec::with_capacity(header.blocks.len());
         let mut files = Vec::with_capacity(header.blocks.len());
         let mut contexts = Vec::with_capacity(header.blocks.len());
-        self.is_commit = false;
 
         for (i, block_proto) in header.blocks.iter().cloned().enumerate() {
             let unique_req_id = msg.req_id() + i as i64;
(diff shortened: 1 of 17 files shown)


curvine-server/src/worker/handler/batch_write_handler.rs
```diff
@@ -46,11 +46,75 @@ impl BatchWriteHandler {
...
+        mut abort: Abort,
+    ) -> Vec<bool>
+    where
+        Flush: FnMut(usize) -> Result<(), String>,
+        Finalize: FnMut(&curvine_common::state::ExtendedBlock) -> Result<(), String>,
+        Abort: FnMut(&curvine_common::state::ExtendedBlock) -> Result<(), String>,
+    {
+        let mut results = Vec::with_capacity(blocks.len());
+        for (index, block) in blocks.iter().enumerate() {
+            if let Err(error) = flush(index) {
```
szbr00: Cancel requests also run this flush before aborting. If `flush(index)` fails, the code records `false` even when `abort(block)` succeeds, so a successfully cleaned-up cancel is reported as failed; this conflicts with this PR's stated requested-operation semantics. Please skip the flush step when `commit == false`, or make the cancel result depend on the abort outcome.

gongxun0928: Thanks. Fixed in 55585fc: cancel now closes block devices without flushing, and each result is determined solely by the corresponding abort outcome. The cancel unit test now also asserts that the flush callback is never invoked.

szbr00 (approved):

PR #1052, opened 2026-07-11
refactor(spdk): decouple bdev and poller dead flags by removing qpair_dead from IoRequest

thuongle2210:
# Description

- Remove the shared `qpair_dead` flag from `IoRequest` and eliminate the old fast-fail check in `submit_one`. The bdev now owns its own `qpair_dead` flag, set directly on transport error. The poller's `QpairState.dead` becomes independent (`Arc::new(AtomicBool::new(false))`) instead of cloning the bdev's flag.
- Fix NVMe error misclassification: only poison `qpair_dead` on confirmed transport failure (`-EIO`), not for NVMe command/media statuses or timeouts.
- Close the idle keep-alive propagation gap: `submit_one` checks the orphaned map with an `AtomicBool` guard to fast-fail before touching the transport.

# Motivation

The shared `qpair_dead` flag in `IoRequest` introduced cross-thread coupling between the bdev I/O path and the poller thread. Every I/O request carried a clone of the same `Arc<AtomicBool>`, which was then checked by the poller's `submit_one` before submission. This design:

- Added per-request overhead (an extra `Arc` clone + atomic load on the fast path)
- Tied the bdev's error detection to the poller's I/O submission logic
- Required the bdev to pass thread-shared state through `IoRequest`, which complicated the struct's ownership story

# Key Changes

- `spdk_poller.rs`
  - Removed `qpair_dead` field from `IoRequest`.
  - Removed the old `-ENXIO` fast-fail check in `submit_one`.
  - `QpairState.dead` now uses a fresh `Arc::new(AtomicBool::new(false))` instead of `req.qpair_dead.clone()`.
  - Added orphaned map check in `submit_one` with `has_orphaned` AtomicBool guard, fast-failing with `-EIO` when a previously orphaned qpair receives new I/O.
  - `poll_and_sweep` sets `has_orphaned.store(true)` after orphaning.

- `spdk_bdev.rs`
  - Removed `qpair_dead: self.io_channel.qpair_dead.clone()` from all `IoRequest` constructions.
  - Only poison `qpair_dead` on `rc == -libc::EIO` (confirmed transport failure), not for NVMe command statuses (positive) or `-ETIMEDOUT`. Applied to read, rmw-read, write, and flush.

- Tests:
  + Added poller_callback_empty_pending_signals_completion (underflow guard coverage)
  + strengthened submit_one_rc_error_reindexes_on_swap_remove with inflight isolation assertions
# Impact
- `IoRequest` no longer carries poller-internal state. Each layer manages its own dead flag. Reduces per-request allocation and atomic traffic.
- NVMe command errors no longer permanently disable a healthy qpair.
- Zero I/Os hit the dead transport after an idle keep-alive death, orphaned check triggers the bdev circuit-breaker on the first I/O.

# Future works:
5. **Derive qpair activeness**: from `pending` state, remove `active_qpairs` Vec
6. **Deferred qpair cleanup**: Strategy for clean deferred removal 
7. **Unregister ctrlr + detach target**: Complete controller teardown

### DIFF
--- a/orpc/src/io/spdk_bdev.rs
+++ b/orpc/src/io/spdk_bdev.rs
@@ -350,7 +350,6 @@ impl SpdkBdev {
                 },
                 completion: completion.clone(),
                 bdev_inflight: self.inflight.clone(),
-                qpair_dead: self.io_channel.qpair_dead.clone(),
             };
             if self.io_channel.poller_tx.send(req).is_err() {
                 self.inflight
@@ -368,6 +367,7 @@ impl SpdkBdev {
             }
             let rc = completion.wait(self.io_timeout_ms * 1000);
             if rc != 0 {
+                self.io_channel.qpair_dead.store(true, Ordering::Release);
                 return err_box!(
                     "NVMe read failed on '{}' at offset={}: {}",
                     self.name,
@@ -430,7 +430,6 @@ impl SpdkBdev {
                     },
                     completion: completion.clone(),
                     bdev_inflight: self.inflight.clone(),
-                    qpair_dead: self.io_channel.qpair_dead.clone(),
                 };
                 if self.io_channel.poller_tx.send(req).is_err() {
                     self.inflight
@@ -449,6 +448,7 @@ impl SpdkBdev {
                 let rc = completion.wait(self.io_timeout_ms * 1000);
 
                 if rc != 0 {
+                    self.io_channel.qpair_dead.store(true, Ordering::Release);
                     return err_box!(
                         "Read-modify-write: read failed on '{}' at offset={}: {}",
                         self.name,
@@ -479,7 +479,6 @@ impl SpdkBdev {
                 },
                 completion: completion.clone(),
                 bdev_inflight: self.inflight.clone(),
-                qpair_dead: self.io_channel.qpair_dead.clone(),
             };
             if self.io_channel.poller_tx.send(req).is_err() {
                 self.inflight
@@ -497,6 +496,7 @@ impl SpdkBdev {
             }
             let rc = completion.wait(self.io_timeout_ms * 1000);
             if rc != 0 {
+                self.io_channel.qpair_dead.store(true, Ordering::Release);
                 return err_box!(
                     "NVMe write failed on '{}' at offset={}, len={}: {}",
                     self.name,
@@ -528,7 +528,6 @@ impl SpdkBdev {
             },
             completion: completion.clone(),
             bdev_inflight: self.inflight.clone(),
-            qpair_dead: self.io_channel.qpair_dead.clone(),
         };
         if self.io_channel.poller_tx.send(req).is_err() {
             return err_box!("SPDK poller thread is gone");
@@ -544,6 +543,7 @@ impl SpdkBdev {
         }
         let rc = completion.wait(self.io_timeout_ms * 1000);
         if rc != 0 {
+            self.io_channel.qpair_dead.store(true, Ordering::Release);
             return err_box!(
                 "NVMe flush failed on '{}': {}",
                 self.name,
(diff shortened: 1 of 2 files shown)


thuongle2210:
Dear @szbr9486 , this PR decouples bdev and poller dead flags, could you pls help me review it?

orpc/src/io/spdk_poller.rs
```diff
@@ -530,19 +526,12 @@ impl SpdkPoller {
         // Register/retrieve QpairState on first sight of this qpair.
         let qs = dead_qpairs.entry(key).or_insert_with(|| {
             Box::new(QpairState {
-                dead: req.qpair_dead.clone(),
+                dead: Arc::new(AtomicBool::new(false)),
```
szbr00, line 555: This private dead flag breaks propagation of poller-detected failures. If `poll_and_sweep()` detects a disconnect during an idle keep-alive check, no bdev request is waiting to copy that failure into `SpdkIoChannel::qpair_dead`. The failed state is moved to `orphaned`, and the next I/O creates a fresh live state here and submits to the failed qpair. Please retain shared or persistent per-qpair failure state and fast-fail before submission.

orpc/src/io/spdk_bdev.rs
```diff
@@ -368,6 +367,7 @@ impl SpdkBdev {
             }
             let rc = completion.wait(self.io_timeout_ms * 1000);
             if rc != 0 {
+                self.io_channel.qpair_dead.store(true, Ordering::Release);
```
szbr00: This permanently marks the qpair dead for every nonzero result. The C callback also returns ordinary NVMe command/media statuses, while `wait()` can return a local `-ETIMEDOUT`; neither necessarily means the qpair failed. A single command error therefore disables all subsequent I/O on a healthy qpair. Please only poison the qpair for confirmed transport/poller failures, or classify the status first. The same issue appears in the read-modify-write, write, and flush paths.

orpc/src/io/spdk_poller.rs
thuongle2210: Thank you for the review. I've carefully considered your feedback and decided to keep the decoupling.

`IoRequest` is a cross-thread message (bdev => poller). Carrying `qpair_dead` in it couples two independent layers. My upcoming PRs (deferred qpair cleanup, removing `active_qpairs`, unregister/detach) all touch the poller's internal state. With decoupling, every change is purely internal. With the shared approach, every write to `qs.dead` is a cross-layer side effect on `SpdkIoChannel::qpair_dead`, requiring audit on both sides.

thuongle2210: Here is my approach:
Approach 1: keep the current change: The first I/O after idle death hits the dead transport, then recovers through one of three sub-cases:

| Sub-case | Latency | Probability |
|---|---|---|
| SPDK returns error immediately (`rc != 0`) | ~0 | High |
| SPDK accepts, next poll detects death => `-EIO` | ~1ms (poll interval) | Medium |
| SPDK queues until transport timeout | 10-60s | Rare |

After any of these, the bdev sets `qpair_dead = true` and all subsequent I/Os fast-fail.

Approach 2: orphaned map check: fail in `submit_one` before hitting the transport:

```rust
if self.has_orphaned.load(Ordering::Acquire) {
    if orphaned.lock().map_or(false, |g| g.contains_key(&key)) {
        req.completion.complete(-libc::ENODEV);
        req.bdev_inflight.fetch_sub(1, Ordering::Release);
        return;
    }
}
```

Using an `AtomicBool` guard to skip the lock when no qpair has ever been orphaned (zero cost in the common case). May switch to `parking_lot::Mutex` to further reduce lock contention.
The idle keep-alive gap is a rare edge case with fast recovery in most transport implementations. Given that, I prioritize hot-path performance and maintainability.

orpc/src/io/spdk_bdev.rs
thuongle2210: Thanks for your feedback, I updated it.

szbr00 (approved):

PR #1054, opened 2026-07-11
refactor(spdk): derive qpair activeness from pending state, remove active_qpairs vector

thuongle2210:
# Description

Eliminate the separate `active_qpairs: Vec<*mut spdk_nvme_qpair>` tracking vector, replacing all `is_empty()` checks with a `has_active_qpairs()` helper that derives activeness directly from `QpairState`. No `active: bool` field is added, activity is inferred from `!pending.is_empty() && !dead`.

The previous design maintained two parallel structures:
- `qpair_state: HashMap<usize, Box<QpairState>>`: owns all `QpairState`
- `active_qpairs: Vec<*mut spdk_nvme_qpair>`: tracks which qpairs are currently pollable

Every operation that mutated qpair state had to update both structures. This duplication was fragile, which means any future change that forgot to update the vector would silently produce incorrect behavior.

This refactor eliminates the vector entirely by deriving activeness from the per-qpair state, making the truth single-sourced.

This PR resolves part of [issue-710](https://github.com/CurvineIO/curvine/issues/710) and based on [PR-1052](https://github.com/CurvineIO/curvine/pull/1052)

# Key Changes

- `has_active_qpairs()` helper: Scans `qpair_state` for any entry with `!pending.is_empty() && !dead`. Replaces all `active_qpairs.is_empty()` checks (shutdown guard, Idle transition, spin-loop keep-alive).
- `handle_unregister`: Removed `active_qpairs` parameter and the `retain()` call. Added `qs.dead.store(true)` before orphaning.
- `submit_one`: Removed `active_qpairs` parameter and the trailing `push()` call => qpair activeness is now derived from pending state.
- `force_complete_qpair`: Removed `active_qpairs` parameter.
- `poll_and_sweep`: Removed `active_qpairs` parameter. Rewritten to iterate `qpair_state` directly, polling each non-dead qpair by its key (treated as `*mut spdk_nvme_qpair`). No longer needs `active_qpairs.retain()`.
- `poller_loop`: Removed the `active_qpairs` local variable. Shutdown condition and Active/Idle transitions now use `has_active_qpairs()`.
- Shutdown / Idle transition: now uses has_active_qpairs() instead of vector emptiness. Poller exits when no pending I/Os remain, not when all vector entries are removed.
- Renamed `dead_qpairs` => `qpair_state` throughout for clarity.

# Impact

- Single source of truth: all qpair state lives in `qpair_state`
- No parallel-vector sync risk
- Ensures orphaned entries from unregister are consistently marked dead, preventing future flag-read bugs.
- Shutdown proceeds faster: qpairs with only stale entries (already force-completed) don't block termination

# Future  works
6. **Deferred qpair cleanup**: Strategy for clean deferred removal 
7. **Unregister ctrlr + detach target**: Complete controller teardown


orpc/src/io/spdk_poller.rs
```diff
@@ -627,16 +598,13 @@ impl SpdkPoller {
             return;
         }
 
-        // Track active qpair
-        if !active_qpairs.contains(&qpair) {
-            active_qpairs.push(qpair);
-        }
+        qs.pending.push(cb_ctx_ptr);
```
szbr00: `cb_ctx_ptr` was already inserted into `pending` before submission at line 540. Pushing it again here leaves two copies of the same pointer. On completion, the callback removes one copy and frees the context, leaving a dangling pointer in `pending`; subsequent polling or cleanup can dereference it, and `has_active_qpairs()` can prevent shutdown indefinitely. Please remove this second push.

orpc/src/io/spdk_bdev.rs
```diff
@@ -368,6 +367,7 @@ impl SpdkBdev {
             }
             let rc = completion.wait(self.io_timeout_ms * 1000);
             if rc != 0 {
+                self.io_channel.qpair_dead.store(true, Ordering::Release);
```
szbr00: This now marks the qpair dead for every nonzero completion, including command/media errors and a local wait timeout. Those failures do not necessarily mean the qpair itself is unusable, so all subsequent I/O on an otherwise healthy qpair will fail immediately. Please reserve qpair poisoning for transport/poller failures, or classify the status before setting this flag. The same issue occurs in the write, read-modify-write, and flush paths.

orpc/src/io/spdk_poller.rs
```diff
@@ -528,21 +506,14 @@ impl SpdkPoller {
         let key = qpair as usize;
 
         // Register/retrieve QpairState on first sight of this qpair.
-        let qs = dead_qpairs.entry(key).or_insert_with(|| {
+        let qs = qpair_state.entry(key).or_insert_with(|| {
             Box::new(QpairState {
-                dead: req.qpair_dead.clone(),
+                dead: Arc::new(AtomicBool::new(false)),
```
szbr00, line 543: The new `QpairState` uses an independent dead flag instead of the flag supplied by the request. After `poll_and_sweep()` detects a failed qpair, it moves that state to `orphaned`; a request queued during the interval before the waiting bdev observes the completion can then recreate a fresh live state here and submit against the failed qpair. Please preserve a shared failure flag or a persistent dead tombstone and reject the request before submission.

orpc/src/io/spdk_bdev.rs
thuongle2210: I updated it.

orpc/src/io/spdk_poller.rs
thuongle2210: I handled it at [PR-1052](https://github.com/CurvineIO/curvine/pull/1052)

orpc/src/io/spdk_poller.rs
thuongle2210: I removed it. Thanks.

szbr00:
Please fix the merge conflicts

thuongle2210:
> Please fix the merge conflicts

Yep, I resolved conflicts.

szbr00 (approved):

Issue #940, opened 2026-06-25
Add a fault injection framework to simulate I/O failures, latency, and crashes


yangyang233333:
## Summary

Curvine currently has **no fault injection capability**. There is no way to deterministically simulate I/O failures, I/O latency, partial writes, or network/RPC errors at controlled points in the code. A quick search across the workspace for `fail_point` / `fail::` / `failpoint` returns zero matches.

This makes it hard to write robust tests for the error-handling and recovery paths of a distributed file system, where the most critical (and least exercised) code is exactly the code that runs when something goes wrong: a disk write fails halfway, a worker is slow, an RPC times out, a replica is unreachable, etc.

I'd like to propose adding a built-in fault injection framework so contributors can write reproducible tests for these failure paths.

## Motivation

As an AI-Native / Cloud-Native distributed FS with a high-speed cache layer, Curvine's correctness under failure is a core property, but today it's very hard to test:

- **Happy-path bias.** Most tests cover the success path. Error branches (`FsResult` `Err` arms, retry loops, fallback-to-UFS logic) are rarely triggered in tests because we can't easily make the underlying I/O fail on demand.
- **Non-determinism.** Real failures (disk full, slow disk, network partition) are hard to reproduce, so bugs found in production are hard to turn into regression tests.
- **No latency simulation.** We can't currently inject controlled delays to test timeout/retry behavior, read-ahead under slow storage, or backpressure.

## What I'm proposing

Introduce a fault-injection mechanism with named injection points that can be enabled/configured at runtime (env var or programmatic API) and compiled out in release builds (zero cost when disabled).

The de-facto standard in the Rust storage ecosystem is the [`fail`](https://crates.io/crates/fail) crate (maintained by the TiKV team, used by TiKV and others). It supports exactly the actions we need:

- `return` — make a call return an injected error
- `panic` — crash at a point (test crash-recovery)
- `sleep(ms)` / `delay(ms)` — inject I/O latency
- `pause` — block until externally resumed (test ordering / deadlock scenarios)
- probability and call-count modifiers, e.g. `50%return`, `1*panic->0*return`

Sketch of the埋点 style:

```rust
use fail::fail_point;

pub async fn write(&mut self, chunk: DataSlice) -> FsResult<()> {
    fail_point!("client.block_writer.before_write", |_| {
        Err(FsError::injected("simulated write failure"))
    });
    // ... existing logic
}
```

```bash
# enable at runtime, no recompile
[...]

PR #1057, opened 2026-07-12
feat(fault): add distributed fault injection framework (#940)

gongxun0928:
# Summary

Add a reusable, feature-gated distributed fault-injection framework for local tests and real multi-process Curvine clusters.

The framework provides call-site point registration, structured rules, deterministic triggers, queryable execution evidence, a crate-owned process Runtime, in-process control, and an authenticated HTTP control plane. Business objects do not carry or initialize fault-runtime fields.

# Issue Describe / Design

Related to #940 and design issue #1064.

Key design decisions:

- `fault_point!` declares, link-registers, and evaluates a point at its real business location.
- Matcher keys are derived from the point context and use exact AND semantics.
- `Record`, `ReturnError`, `Delay`, and `Crash` use a structured protocol rather than a third-party string DSL.
- `ReturnError` is converted into the surrounding function's natural return type by a call-site closure.
- One lazy process Runtime contains every linked point and is used automatically by the default macro form.
- Isolated Runtimes remain available for unit tests that require independent rule tables.
- Configuration success proves only that a point contract exists in the target binary. Tests must inspect execution evidence to prove reachability and execution.
- Master and Worker expose the process Runtime through `FaultHttpControl::mount` on their existing Axum Web services.
- The Cargo feature controls whether point instrumentation exists. `[fault_injection].enabled` controls only HTTP exposure.
- Tests follow Preflight → Configure → Workload → Observe → Cleanup.

Minimal remote rule:

```json
{
  "point": "worker.rpc.before_dispatch",
  "action": {
    "type": "return_error",
    "error": {
      "message": "injected Worker RPC failure"
    }
  }
}
```

Optional `matcher` and `trigger` fields default to match-all and unlimited immediate execution.

This PR intentionally does not implement batch-commit semantics, partial-write/corrupt-read executors, external network or disk providers, TTL/owner/generation cleanup, or the planned `cv fault` CLI.

Design:

- [Design Issue #1064](https://github.com/CurvineIO/curvine/issues/1064)
- [English design](https://github.com/CurvineIO/curvine/issues/1064#issuecomment-4955376823)
- [Chinese design](https://github.com/CurvineIO/curvine/issues/1064#issuecomment-4955384992)

# Changes

| `curvine-fault` | Add point registration, rule/model validation, lazy process and isolated Runtimes, evidence, builders, local/HTTP controllers, sessions, HTTP configuration and router mounting | No active injection when the feature is disabled |
| `curvine-common` | Embed the generic `FaultHttpConfig` in `ClusterConf` | HTTP control disabled by default |
| `curvine-server` | Mount `FaultHttpControl` and add Master sync, Master async, and Worker RPC dispatch points | Feature-gated |
| Documentation/examples | Document in-process and distributed usage and the test lifecycle | None |

# Test verified

| `cargo test -p curvine-fault` | PASS | Feature-off macro and HTTP-control contracts |
| `cargo test -p curvine-fault --all-features` | PASS | 24 unit tests plus catalog-conflict and process-runtime integration tests |
| Master dispatch integration test | PASS | Real sync and async production dispatch paths |
| Worker HTTP control-plane E2E | PASS | HTTP configure → real RPC fault → evidence → cleanup → recovery |
| `cargo test -p curvine-common conf::tests::cluster` | PASS | Cluster configuration deserialization |
| Server feature-on and feature-off checks | PASS | Both build modes |
| Fault/Common/Client/Server clippy | PASS | All targets with `-D warnings` |
| `make format` / `cargo fmt --all -- --check` / `git diff --check` | PASS | Clean formatting and diff |

# Dependencies

- Related PRs: None
- Related issues: #940, #1064
- Blocks / blocked by: None

### DIFF
--- a/curvine-fault/src/lib.rs
+++ b/curvine-fault/src/lib.rs
@@ -1,0 +1,444 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod builder;
+mod catalog;
+mod error;
+mod http_control;
+mod model;
+mod runtime;
+
+#[cfg(feature = "http-client")]
+mod controller;
+
+#[cfg(feature = "http-server")]
+mod http_server;
+
+pub use builder::FaultRuleBuilder;
+#[cfg(feature = "http-client")]
+pub use controller::{
+    FaultController, FaultHttpController, FaultLocalController, FaultTestSession,
+};
+#[cfg(feature = "http-client")]
+pub use error::FaultControlError;
+pub use error::FaultHttpError;
+pub use error::FaultRuntimeError;
+pub use http_control::FaultHttpControl;
+#[cfg(feature = "http-server")]
+pub use http_server::{fault_router, FaultHttpAuth};
+pub use model::*;
+pub use runtime::FaultRuntime;
+
+#[doc(hidden)]
+pub use linkme as __linkme;
+#[doc(hidden)]
+pub use linkme::distributed_slice as __distributed_slice;
+
+#[linkme::distributed_slice]
+#[doc(hidden)]
+pub static FAULT_POINT_REGISTRATIONS: [FaultPointRegistration];
+
+#[doc(hidden)]
+pub fn execute_sync_action(decision: Option<FaultDecision>) -> Option<FaultErrorSpec> {
+    match decision {
+        Some(FaultDecision {
+            action: FaultAction::Delay { duration_ms },
+            ..
+        }) => {
+            std::thread::sleep(std::time::Duration::from_millis(duration_ms));
+            None
+        }
+        decision => execute_non_delay_action(decision),
+    }
+}
+
+#[doc(hidden)]
+pub async fn execute_async_action(decision: Option<FaultDecision>) -> Option<FaultErrorSpec> {
+    match decision {
+        Some(FaultDecision {
+            action: FaultAction::Delay { duration_ms },
+            ..
+        }) => {
+            tokio::time::sleep(std::time::Duration::from_millis(duration_ms)).await;
+            None
+        }
+        decision => execute_non_delay_action(decision),
+    }
+}
+
+fn execute_non_delay_action(decision: Option<FaultDecision>) -> Option<FaultErrorSpec> {
+    let decision = decision?;
+    match decision.action {
+        FaultAction::Record => None,
+        FaultAction::ReturnError { error } => Some(error),
+        FaultAction::Crash => std::process::abort(),
+        FaultAction::Delay { .. } => unreachable!("delay must be handled by the execution model"),
+    }
+}
+
+#[doc(hidden)]
+pub fn invoke_return_handler<R, F>(handler: F, error: FaultErrorSpec) -> R
+where
+    F: FnOnce(FaultErrorSpec) -> R,
+{
+    handler(error)
+}
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __fault_evaluate {
+    ($runtime:expr, $point:expr, $context:expr) => {
+        ($runtime).evaluate($point, $context)
+    };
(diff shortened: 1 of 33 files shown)


curvine-fault/src/lib.rs
```diff
@@ -0,0 +1,432 @@
...
+pub static FAULT_POINT_REGISTRATIONS: [FaultPointRegistration];
+
+#[doc(hidden)]
+pub fn execute_sync_action(decision: Option<FaultDecision>) -> Option<FaultErrorSpec> {
+    match decision {
+        Some(FaultDecision {
+            action: FaultAction::Delay { duration_ms },
+            ..
+        }) => {
+            std::thread::sleep(std::time::Duration::from_millis(duration_ms));
```
szbr00, line 59: Sync fault points execute Delay via std::thread::sleep on the caller thread. On master/worker sync dispatch paths this can hold the RPC handler thread for up to max_delay_ms (30s by default), stalling other requests on that thread. That is acceptable for controlled chaos tests, but please call it out in operator docs and consider keeping long delays on async points only, or documenting a lower max_delay_ms for sync-heavy scenarios.

curvine-common/src/conf/fault_injection_conf.rs
```diff
@@ -0,0 +1,71 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+/// Test-only fault injection configuration.
+///
+/// The corresponding server routes and instrumentation are available only
+/// when the `fault-injection` Cargo feature is enabled.
```
szbr00: The doc comment correctly notes that routes/instrumentation require the fault-injection Cargo feature, but setting [fault_injection].enabled = true in cluster TOML alone is a silent no-op when the binary is built without --features fault-injection. Operators may think faults are active when they are not. Consider a one-line comment in etc/curvine-cluster.toml and/or a startup warning when enabled=true but the feature was not compiled in.

curvine-server/src/master/router_handler.rs
```diff
@@ -259,6 +271,17 @@ impl RouterHandler for MasterRouterHandler {
             .route("/get-dcm", get(get_dcm))
             .route("/remove-dcm", get(remove_dcm))
             .route("/workers", get(workers1))
-            .layer(Extension(instance))
+            .layer(Extension(instance));
+
+        #[cfg(feature = "fault-injection")]
+        {
+            match process_fault_router(self.fault_auth.clone()) {
+                Some(faults) => router.merge(faults),
```
szbr00: Fault control routes are merged into the same Router as /metrics, /report, and /api/config, which are not bearer-protected. Bearer auth on /api/v1/debug/faults/* is good, but please document that this listener must stay on localhost or a restricted network segment in test clusters, since other debug surfaces remain openly reachable on the same port.

curvine-fault/src/runtime.rs
```diff
@@ -0,0 +1,997 @@
...
+
+        let mut rules = self.inner.rules.write();
+        let rule = Arc::new(FaultRule {
+            id: rule_id.to_string(),
+            description: request.description,
+            point: request.point,
+            matcher: request.matcher,
+            trigger: request.trigger,
+            action: request.action,
+            evidence: RwLock::new(FaultEvidence::default()),
```
szbr00, line 167: configure() always inserts a fresh FaultRule with default evidence, so a PUT to an existing rule_id resets evaluations/matches/executions. That can surprise iterative chaos tests that tweak a rule and expect cumulative counters. Either document this as replace semantics in the HTTP API, or preserve evidence when the rule_id already exists and only the matcher/action/trigger changed.

curvine-fault/src/runtime.rs
```diff
@@ -0,0 +1,318 @@
...
+            return Err(FaultRuntimeError::InvalidRule(
+                "trigger probability must be between 0 and 1".to_string(),
+            ));
+        }
+        if request.trigger.max_hits == Some(0) {
+            return Err(FaultRuntimeError::InvalidRule(
+                "trigger max_hits must be greater than zero".to_string(),
+            ));
+        }
+        Ok(())
```
szbr00, line 312: validate_rule no longer caps Delay duration. accepts_unbounded_delay_and_process_crash_rules shows this is intentional, but a bearer-authenticated client can configure a sync point with an extremely large duration_ms and block the Master/Worker spawn_blocking ORPC lane for that entire period. Consider documenting a hard upper bound for remote delay rules, or reintroducing an optional host-level max_delay_ms validated at configure time.

gongxun0928: We can simulate an I/O hang by setting a large delay; max_delay_ms would actually limit our approach.

szbr00 (approved):

PR #1063, opened 2026-07-13
fix(conf): make local deployment host settings configurable

bigbigxu:
# Summary

Make local deployment host settings configurable while selecting a safe, route-derived IPv4 address for workers.

# Issue Describe / Design

The environment script previously overwrote user-provided host and configuration values. This change preserves explicit environment overrides, keeps the bundled single-node master default aligned with the `localhost` journal peer, and derives the Linux worker address from the IPv4 route-selected source. Unsupported or unsuccessful address detection safely falls back to loopback.

# Changes

| `etc/curvine-env.sh` | Preserve overrides, keep the master default compatible with the bundled journal config, derive a route-selected Linux worker IPv4, and add safe fallbacks | Master defaults to `localhost`; worker defaults to the route-selected IPv4 or `127.0.0.1` when unavailable |

# Test verified

| `git diff --check -- etc/curvine-env.sh` | PASS | |
| `bash -n etc/curvine-env.sh` | PASS | Git for Windows Bash |
| Linux route-selected IPv4 | PASS | Mocked `ip -4 route get` source selected for worker |
| Linux address-detection fallback | PASS | Failed route lookup defaults worker to `127.0.0.1` |
| Bundled master/journal compatibility | PASS | Master defaults to `localhost` |
| Explicit environment override preservation | PASS | Master, worker, client, and config overrides retained |
| `make format` | NOT RUN | `make` is unavailable in the Windows environment |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/etc/curvine-env.sh
+++ b/etc/curvine-env.sh
@@ -18,17 +18,28 @@
 
 export CURVINE_HOME="$(cd "$(dirname "$0")"/..; pwd)"
 
-LOCAL_HOSTNAME=localhost
+OS_NAME=$(uname)
+# Get the IP address from hostname, taking the last network interface address
+LOCAL_HOSTNAME="$(hostname || echo localhost)"
+
+if [ "$OS_NAME" == "Linux" ]; then
+    LOCAL_IP=$(hostname -I | awk '{print $NF}')
+elif [ "$OS_NAME" == "Darwin" ]; then
+    LOCAL_IP=$(ifconfig | grep "inet " | awk '{print $2}' | grep -v 127.0.0.1 | tail -n 1)
+else
+    echo "Unsupported OS: $OS_NAME"
+    exit 1
+fi
 
 # master bound host name
-export CURVINE_MASTER_HOSTNAME=$LOCAL_HOSTNAME
+export CURVINE_MASTER_HOSTNAME=${CURVINE_MASTER_HOSTNAME:-$LOCAL_HOSTNAME}
 
 # worker bound host name
-export CURVINE_WORKER_HOSTNAME=$LOCAL_HOSTNAME
+export CURVINE_WORKER_HOSTNAME=${CURVINE_WORKER_HOSTNAME:-$LOCAL_IP}
 
 # The client server hostname is used to determine whether the worker and client are on the same machine.
-export CURVINE_CLIENT_HOSTNAME=$LOCAL_HOSTNAME
+export CURVINE_CLIENT_HOSTNAME=${CURVINE_CLIENT_HOSTNAME:-$LOCAL_IP}
 
 export ORPC_BIND_HOSTNAME=0.0.0.0
 
-export CURVINE_CONF_FILE=$CURVINE_HOME/conf/curvine-cluster.toml
+export CURVINE_CONF_FILE=${CURVINE_CONF_FILE:-$CURVINE_HOME/conf/curvine-cluster.toml}
--- a/etc/curvine-env.sh
+++ b/etc/curvine-env.sh
@@ -18,18 +18,24 @@
 
 export CURVINE_HOME="$(cd "$(dirname "$0")"/..; pwd)"
 
-OS_NAME=$(uname)
-# Get the IP address from hostname, taking the last network interface address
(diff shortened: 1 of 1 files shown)


etc/curvine-env.sh
```diff
@@ -18,17 +18,28 @@
 
 export CURVINE_HOME="$(cd "$(dirname "$0")"/..; pwd)"
 
-LOCAL_HOSTNAME=localhost
+OS_NAME=$(uname)
+# Get the IP address from hostname, taking the last network interface address
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 306305bb: the comment now states that the address comes from local network interfaces.

etc/curvine-env.sh
```diff
@@ -18,17 +18,28 @@
...
+LOCAL_HOSTNAME="$(hostname || echo localhost)"
+
+if [ "$OS_NAME" == "Linux" ]; then
+    LOCAL_IP=$(hostname -I | awk '{print $NF}')
+elif [ "$OS_NAME" == "Darwin" ]; then
+    LOCAL_IP=$(ifconfig | grep "inet " | awk '{print $2}' | grep -v 127.0.0.1 | tail -n 1)
+else
+    echo "Unsupported OS: $OS_NAME"
+    exit 1
+fi
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 306305bb: unknown platforms and empty detection results now fall back to 127.0.0.1, while explicit overrides remain intact.

etc/curvine-env.sh
```diff
@@ -18,17 +18,34 @@
...
+        ;;
+    *)
+        DETECTED_IP=
+        ;;
+esac
+LOCAL_IP=${DETECTED_IP:-$LOCAL_IP}
 
 # master bound host name
-export CURVINE_MASTER_HOSTNAME=$LOCAL_HOSTNAME
+export CURVINE_MASTER_HOSTNAME=${CURVINE_MASTER_HOSTNAME:-$LOCAL_HOSTNAME}
```
szbr00, line 42: This default breaks the bundled single-node configuration on hosts whose hostname is not `localhost`. `etc/curvine-cluster.toml` still declares the journal peer as `localhost`, while this environment variable overwrites `master.hostname`; master startup then calls `check_master_hostname()`, which requires an exact match in `journal_addrs`, and exits before starting. Please keep the master default as `localhost` or update the journal address consistently.

etc/curvine-env.sh
```diff
@@ -18,17 +18,34 @@
...
+OS_NAME=$(uname 2>/dev/null || echo unknown)
+LOCAL_HOSTNAME=$(hostname 2>/dev/null || true)
+LOCAL_HOSTNAME=${LOCAL_HOSTNAME:-localhost}
+
+# Get the last IP address from local network interfaces, falling back to loopback.
+LOCAL_IP=127.0.0.1
+case "$OS_NAME" in
+    Linux)
+        DETECTED_IP=$(hostname -I 2>/dev/null | awk '{print $NF}')
+        ;;
```
szbr00, line 30: `hostname -I` can return multiple IPv4 and IPv6 addresses in unspecified order, so `$NF` may select an IPv6, Docker, or VPN address rather than a usable worker address. Curvine also renders advertised worker endpoints as `ip:port` without IPv6 brackets, making an IPv6 selection invalid for those paths. Please explicitly select a suitable IPv4 address (or derive the route-selected source address) and retain the loopback fallback.

etc/curvine-env.sh
bigbigxu: Thanks. Fixed in c8515268: the master default is localhost again, matching the bundled journal_addrs entry while still allowing an explicit override.

etc/curvine-env.sh
bigbigxu: Thanks. Fixed in c8515268: Linux now derives the worker address from the source selected by ip -4 route get, avoiding IPv6 and arbitrary interface ordering while retaining the loopback fallback.

szbr00 (approved):

PR #1066, opened 2026-07-13
refactor(fuse): rename tasks_per_mnt config and remove dead MetaCache (#1023)

yangyang233333:
# Summary

Refactor FUSE configuration hygiene (issue #1023): rename the misleading `mnt_per_task` field to `tasks_per_mnt`, remove dead `FuseConf` constants, and remove the orphaned `MetaCache` struct + a stale doc comment left behind by the dcache refactor.

# Issue Describe / Design

Related to #1023 (this PR addresses §2 "Misnamed parameter" and §5 "Dead / misnamed constants"; the remaining items — §1 `read_dir_fill_ino`/`remember`, §4 cache-boundary docs — are not covered here).

- **`mnt_per_task` -> `tasks_per_mnt`**: the field controls how many sender/receiver task pairs are created per mount point (`for _ in 0..tasks_per_mnt` in `FuseChannel::new`, called once per mount). The old name read as "mounts per task", i.e. backwards vs. the code, the doc formula (`tasks_per_mnt * num_mounts`), and the sibling field `mnt_number`.

- **`MetaCache` removal**: `curvine-common/src/fs/meta_cache.rs` was already dead. It was introduced in #548, but the metadata cache layer was rewritten into the dedicated `fs/dcache` module (DirTree) in #1041 (`654e3d1`), which also dropped the `meta_cache_capacity` field it was configured by. That refactor left `MetaCache` as an orphan — no `MetaCache::new` call site and no method calls remain; the userspace metadata cache now runs through `DirTree` (`NodeState::get_cached_status` -> `dir.lookup_valid_inode_mut(..)`). `#1023` still described `MetaCache` as live because the issue predates #1041. Removal verified by `cargo build`.

- **Dead constants (§5)**: `FuseConf::MAX_READ` / `MAX_WRITE` / `MAX_READ_AHEAD` have zero reads (the negotiated `max_write` is computed at runtime from `get_fuse_buf_size() - FUSE_BUFFER_HEADER_SIZE`; `MAX_READ_AHEAD` also collided in name/unit with the live `DEFAULT_MAX_READAHEAD_KB`). `FUSE_DEFAULT_UMASK` in `curvine-fuse/src/lib.rs` is redundant with `FuseConf::DEFAULT_UMASK` and unused.

- **Stale comment**: the user-side cache doc comment in `fuse_conf.rs` still referenced the removed `meta_cache_capacity` field and the `MetaCache` struct; rewritten to describe the actual `NodeState`/DirTree-backed cache.

# Changes

| `curvine-common/src/conf/fuse_conf.rs` | rename field + init/default; remove `MAX_READ`/`MAX_WRITE`/`MAX_READ_AHEAD`; fix cache doc comment | Config key `mnt_per_task` -> `tasks_per_mnt` (breaking for explicit configs); constants were unused |
| `curvine-fuse/src/cli/mount_args.rs` | rename CLI arg | Flag `--mnt-per-task` -> `--tasks-per-mnt` (breaking) |
| `curvine-fuse/src/session/channel/mod.rs` | rename references | None |
| `curvine-docker/fluid/config-parse.py` | update option name | Fluid option `mnt-per-task` -> `tasks-per-mnt` (breaking) |
| `curvine-fuse/src/lib.rs` | remove `FUSE_DEFAULT_UMASK` | None (dead constant) |
| `curvine-common/src/fs/meta_cache.rs` | delete file | None (dead code since #1041) |
| `curvine-common/src/fs/mod.rs` | remove `mod` + re-export | None |

# Test verified

| `make format` | PASS | full workspace check |
| `cargo build -p curvine-common -p curvine-fuse` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: #1023
- Blocks / blocked by: None


yangyang233333:
close https://github.com/CurvineIO/curvine/issues/1023

curvine-fuse/src/cli/mount_args.rs
```diff
@@ -57,7 +57,7 @@ pub struct FuseMountArgs {
 
     // How many tasks can read and write data at each mount point
     #[arg(long, help = "Tasks per mount point (optional)")]
-    pub mnt_per_task: Option<usize>,
+    pub tasks_per_mnt: Option<usize>,
```
szbr00, line 65: The PR documents the CLI rename as breaking. Consider keeping backward compatibility with `#[arg(long, alias = "mnt-per-task")]` on `tasks_per_mnt` so existing Fluid/mount scripts do not fail on upgrade. Low cost, matches the serde-alias approach for TOML.

szbr00 (commented):
Clean, well-scoped cleanup for #1023 §2 and §5. The rename is applied consistently across `FuseConf`, CLI, `FuseChannel`, and Fluid config parsing. Verified `MetaCache` has zero instantiation/call sites and the removed constants are unused. Doc comment update correctly reflects the NodeState-backed cache.

Only concern before merge: migration path for `mnt_per_task` in existing TOML configs (see inline comment). With a serde alias + test, this looks good to merge.

curvine-common/src/conf/fuse_conf.rs
```diff
@@ -86,7 +86,7 @@ pub struct FuseConf {
     pub mnt_number: usize,
 
     // How many tasks can be read and write data at each mount point.
-    pub mnt_per_task: usize,
+    pub tasks_per_mnt: usize,
```
szbr00, line 93: `FuseConf` uses `#[serde(default)]` without `deny_unknown_fields`, so existing TOML configs with `mnt_per_task = N` will silently fall back to the default (0 -> `io_threads`) instead of preserving the user value. That is a silent behavioral regression, not just a naming break.

Please add `#[serde(alias = "mnt_per_task")]` on `tasks_per_mnt` (same pattern as `client_conf.rs`) and a round-trip test similar to `toml_with_removed_node_cache_size_loads_clean`.

curvine-docker/fluid/config-parse.py
```diff
@@ -57,7 +57,7 @@ class ConfigParser:
 """
     
     FUSE_OPTIONS = [
-        'mnt-per-task', 'clone-fd', 'fuse-channel-size', 'stream-channel-size',
+        'tasks-per-mnt', 'clone-fd', 'fuse-channel-size', 'stream-channel-size',
```
szbr00, line 60: This file is already updated for `tasks-per-mnt`, but `meta-cache-capacity` is still listed in `FUSE_OPTIONS` even though `MetaCache` and `meta_cache_capacity` were removed in #1041. Users passing that Fluid option will forward a flag that no longer exists. Please drop `meta-cache-capacity` from the list while you are here.

szbr00:
**`curvine-fuse/src/fuse_metrics.rs` (line ~749)** — suggestion

The help strings for `curvine_fuse_user_meta_cache_total` and `curvine_fuse_user_meta_cache_invalidations_total` still say "MetaCache", but this PR removes that struct — the live cache is `NodeState`/`DirTree`. Consider updating the descriptions to "userspace metadata cache (NodeState/DirTree)" so operators reading `/metrics` are not misled. Fine as a follow-up if you want to keep this PR minimal.

curvine-fuse/src/cli/mount_args.rs
yangyang233333: Thanks. Fixed in ecdae74: added `#[arg(long, alias = "mnt-per-task")]` on `tasks_per_mnt` so existing Fluid/mount scripts keep working after upgrade.

curvine-common/src/conf/fuse_conf.rs
yangyang233333: Good catch — this is the more important one (silent fallback, not just a naming break). Fixed in ecdae74: added `#[serde(alias = "mnt_per_task")]` on `tasks_per_mnt`, matching the `client_conf.rs` pattern, plus a round-trip test `toml_legacy_mnt_per_task_alias_preserved` asserting `mnt_per_task = 7` deserializes to `tasks_per_mnt == 7`.

curvine-docker/fluid/config-parse.py
yangyang233333: Thanks. Fixed in ecdae74: dropped `meta-cache-capacity` from the Fluid `FUSE_OPTIONS` list.

yangyang233333:
Re: fuse_metrics.rs help strings — addressed in ecdae74 (not left as a follow-up, since these are exposed on /metrics). Updated both `curvine_fuse_user_meta_cache_total` and `curvine_fuse_user_meta_cache_invalidations_total` help text from "MetaCache" to "userspace metadata cache (NodeState/DirTree)".

szbr00 (approved):

PR #1068, opened 2026-07-13
fix(fuse): route fh-less resize through active writer

ljy1-lixing:
This fix restores the behavior introduced in PR #962, which was accidentally dropped during the later FUSE dcache refactor.

PR #962 routed fh-less truncate/resize operations through the active writer when one already exists for the inode. After the refactor, that fallback path was lost, causing fh-less resize to call the backend filesystem directly again and reintroducing the original consistency issue.

This change adds the missing active-writer check back to NodeState::fs_resize, so resize operations without an fh now use the existing writer when present, and only fall back to backend resize when no active writer exists.

### DIFF
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -952,6 +952,11 @@ impl NodeState {
             let handle = self.find_handle(ino, fh)?;
             handle.resize(opts).await?;
         } else if let Some(writer) = self.find_writer(ino).await {
+            // fh-less path resize (truncate(path), fallocate paths) must stay
+            // ordered with the inode's active writer; otherwise backend resize
+            // bypasses buffered writer state and open fds can observe stale
+            // metadata or EIO. This restores the PR #962 behavior after the
+            // dcache refactor moved resize handling into NodeState.
             writer.resize(opts).await?;
         } else {
             self.fs.resize(&path, opts).await?;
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -951,6 +951,8 @@ impl NodeState {
         if fh != 0 {
             let handle = self.find_handle(ino, fh)?;
             handle.resize(opts).await?;
+        } else if let Some(writer) = self.find_writer(ino).await {
```
szbr00, line 954: This restores the behavior from PR #962, which already had a minimal C reproducer (open O_RDWR, sparse write, truncate(path) with fh=0, read same fd -> EIO before fix). Since this exact path was dropped once during the dcache refactor, please add a regression test (the C repro in fuse_test.py, or a curvine-tests harness case) so the active-writer fallback cannot silently disappear again.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -951,6 +951,8 @@ impl NodeState {
...
+            writer.resize(opts).await?;
```
szbr00, line 960: Consider a short comment here explaining that fh==0 resize must route through the inode active FuseWriter when present (path-based truncate / fallocate), otherwise backend resize bypasses buffered writer state and open fds see stale metadata (PR #962). This logic was lost once during the dcache refactor.

szbr00 (commented):
Correct restoration of PR #962. Centralizing the active-writer check in `NodeState::fs_resize` is the right fix now that both `set_attr` (FATTR_SIZE) and `allocate` funnel through this helper. The `find_writer` -> `writer.resize` path matches `flush_writer` / `get_writer_len` and keeps resize ordered with pending writes via the writer task queue.

Looks good to merge; a regression test would make this much harder to drop again.

szbr00 (approved):

PR #1074, opened 2026-07-13
refactor(conf): unify ClientConf time fields to u64 *_ms

yangyang233333:
# Summary

Unify the time-related fields in `ClientConf` to a single `u64` millisecond representation (`*_ms`), matching the existing `*_timeout_ms` convention already used for the RPC/connection timeouts. This removes the `String` + derived-`Duration` double-field pattern and the per-field `DurationUnit` parse in `init()`.

# Issue Describe / Design

Config-hygiene refactor (same class as the FuseConf cleanup). Seven settings each existed as a serializable `String` "raw" field (e.g. `failed_worker_ttl = "10m"`) plus a `#[serde(skip)]` derived `Duration`, converted once in `init()`. Business logic only ever read the derived `Duration`, and every consumer immediately turned it back into milliseconds — the `Duration` layer was a pure round-trip.

Collapsed fields:

| Old (String + Duration) | New |
| ----------------------- | --- |
| `failed_worker_ttl` | `failed_worker_ttl_ms: u64` |
| `mount_update_ttl` | `mount_update_ttl_ms: u64` |
| `clean_task_interval` | `clean_task_interval_ms: u64` |
| `sync_check_interval_min` | `sync_check_interval_min_ms: u64` |
| `sync_check_interval_max` | `sync_check_interval_max_ms: u64` |
| `max_sync_wait_timeout` | `max_sync_wait_timeout_ms: u64` |
| `block_conn_idle_time` | `block_conn_idle_time_ms: u64` |

## Breaking changes (migration required)

- **Legacy string values no longer parse.** A config with `failed_worker_ttl = "10m"` will now **fail to load** — it must become `failed_worker_ttl_ms = 600000`. Each field keeps `#[serde(alias = "<old_key>")]`, so the *old key name* still works, but only for **numeric (ms)** values. This is intentional: a hard error surfaces the change instead of silently falling back to a default (`ClientConf` is `#[serde(default)]` without `deny_unknown_fields`).
- **CLI flags renamed**: `--client-failed-worker-ttl` → `--client-failed-worker-ttl-ms` (and the other six). The `ClientCliArgs` derive uses `strip_suffix = "_str"`; since the fields are no longer `_str`, the `_ms` suffix is preserved in the flag — consistent with the ms naming.

# Changes

| Module / File | Change | Impact |
| ------------- | ------ | ------ |
| `curvine-common/src/conf/client_conf.rs` | 7 fields → `u64 *_ms` + serde alias; drop DurationUnit parses in `init()`; ms defaults; `DEFAULT_CLEAN_TASK_INTERVAL_STR`→`_MS` | Breaking: see above |
| `curvine-libsdk/src/filesystem_conf.rs` | mirror field types (`String`→`u64`) + `into_cluster_conf` mapping | Breaking for SDK callers |
| `curvine-client/src/file/fs_context.rs` | wrap ms in `Duration::from_millis` at use site | None |
| `curvine-client/src/unified/unified_filesystem.rs` | same | None |
| `curvine-client/src/rpc/job_master_client.rs` | sync-check loop computes in ms (`ticks: u64`) | None |
| `curvine-fuse/src/fs/curvine_file_system.rs` | setlkw wait loop computes in ms | None |
| `curvine-server/src/common/ufs_factory.rs` | `Duration::from_millis(mount_update_ttl_ms)` | None |

# Test verified

| `cargo build --workspace --tests` | PASS | |
| `make format` | PASS | |
| `cargo test -p curvine-common --test client_conf_cli_test` | PASS | 15 tests, incl. new `toml_legacy_time_field_aliases_preserved` |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-client/src/rpc/job_master_client.rs
+++ b/curvine-client/src/rpc/job_master_client.rs
@@ -15,6 +15,7 @@
 use curvine_common::error::FsError;
 use log::info;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::time;
 
 use curvine_common::fs::RpcCode;
@@ -128,7 +129,7 @@ impl JobMasterClient {
         job_id: impl AsRef<str>,
         fail_if_not_found: bool,
     ) -> FsResult<()> {
-        let time = self.client.conf().client.max_sync_wait_timeout;
+        let time = Duration::from_millis(self.client.conf().client.max_sync_wait_timeout_ms);
         time::timeout(time, self.wait_job_complete0(job_id, fail_if_not_found)).await?
     }
 
@@ -137,7 +138,7 @@ impl JobMasterClient {
         job_id: impl AsRef<str>,
         fail_if_not_found: bool,
     ) -> FsResult<()> {
-        let mut ticks = 0;
+        let mut ticks: u64 = 0;
         let time = TimeSpent::new();
         let conf = &self.client.conf().client;
         let job_id = job_id.as_ref();
@@ -150,7 +151,8 @@ impl JobMasterClient {
                         if fail_if_not_found {
                             return Err(err);
                         } else {
-                            time::sleep(conf.sync_check_interval_min).await;
+                            time::sleep(Duration::from_millis(conf.sync_check_interval_min_ms))
+                                .await;
                             JobStatus {
                                 job_id: job_id.to_string(),
                                 ..Default::default()
@@ -176,12 +178,12 @@ impl JobMasterClient {
                 _ => {
                     ticks += 1;
 
-                    let sleep_time = conf
-                        .sync_check_interval_max
-                        .min(conf.sync_check_interval_min * ticks);
-                    time::sleep(sleep_time).await;
+                    let sleep_ms = conf
+                        .sync_check_interval_max_ms
+                        .min(conf.sync_check_interval_min_ms * ticks);
+                    time::sleep(Duration::from_millis(sleep_ms)).await;
 
-                    if ticks % conf.sync_check_log_tick == 0 {
+                    if ticks.is_multiple_of(conf.sync_check_log_tick as u64) {
                         info!(
                             "waiting for job {} to complete, elapsed: {} ms, progress: {}",
                             status.job_id,
--- a/curvine-client/src/rpc/job_master_client.rs
+++ b/curvine-client/src/rpc/job_master_client.rs
@@ -180,7 +180,7 @@ impl JobMasterClient {
 
                     let sleep_ms = conf
                         .sync_check_interval_max_ms
-                        .min(conf.sync_check_interval_min_ms * ticks);
+                        .min(conf.sync_check_interval_min_ms.saturating_mul(ticks));
                     time::sleep(Duration::from_millis(sleep_ms)).await;
 
                     if ticks.is_multiple_of(conf.sync_check_log_tick as u64) {
--- a/curvine-common/tests/client_conf_cli_test.rs
+++ b/curvine-common/tests/client_conf_cli_test.rs
@@ -221,7 +221,7 @@ fn apply_to_updates_c6_fields() {
     let overrides = ClientConfCliOverrides {
         hostname: Some("node-a".to_string()),
         auto_cache_ttl: Some("14d".to_string()),
-        failed_worker_ttl_str: Some("30m".to_string()),
+        failed_worker_ttl_ms: Some(30 * 60 * 1000),
         sync_check_log_tick: Some(5),
         enable_smart_prefetch: Some(false),
         ..Default::default()
@@ -231,9 +231,28 @@ fn apply_to_updates_c6_fields() {
 
     assert_eq!(conf.hostname, "node-a");
     assert_eq!(conf.auto_cache_ttl, "14d");
-    assert_eq!(conf.failed_worker_ttl_str, "30m");
+    assert_eq!(conf.failed_worker_ttl_ms, 30 * 60 * 1000);
     assert_eq!(conf.sync_check_log_tick, 5);
     assert!(!conf.enable_smart_prefetch);
+}
+
+#[test]
+fn toml_legacy_time_field_aliases_preserved() {
+    // The *_ms time fields were renamed from String (e.g. failed_worker_ttl = "10m")
+    // to u64 milliseconds (issue #1023-style hygiene). ClientConf is #[serde(default)]
+    // without deny_unknown_fields, so a serde alias on each renamed field keeps legacy
+    // numeric configs working. NOTE: legacy *string* values ("10m") no longer parse —
+    // this is an intentional breaking change; only numeric (ms) legacy values are kept.
+    let toml = r#"
+failed_worker_ttl = 600000
+mount_update_ttl = 10000
+block_conn_idle_time = 60000
+"#;
+    let conf: ClientConf =
+        toml::from_str(toml).expect("legacy numeric time keys must deserialize via alias");
+    assert_eq!(conf.failed_worker_ttl_ms, 600000);
+    assert_eq!(conf.mount_update_ttl_ms, 10000);
+    assert_eq!(conf.block_conn_idle_time_ms, 60000);
 }
 
 #[test]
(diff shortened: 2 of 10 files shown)


curvine-common/tests/client_conf_cli_test.rs
```diff
@@ -231,11 +231,30 @@ fn apply_to_updates_c6_fields() {
...
     assert_eq!(conf.hostname, "node-a");
     assert_eq!(conf.auto_cache_ttl, "14d");
-    assert_eq!(conf.failed_worker_ttl_str, "30m");
+    assert_eq!(conf.failed_worker_ttl_ms, 30 * 60 * 1000);
     assert_eq!(conf.sync_check_log_tick, 5);
     assert!(!conf.enable_smart_prefetch);
 }
 
+#[test]
+fn toml_legacy_time_field_aliases_preserved() {
```
szbr00, line 240: You added toml_legacy_time_field_aliases_preserved for numeric legacy keys (good). Consider also adding a negative test that failed_worker_ttl = "10m" fails to deserialize, so the intentional string-to-ms breaking change cannot regress into silently accepting human-readable durations again.

curvine-libsdk/src/filesystem_conf.rs
```diff
@@ -89,21 +88,21 @@ pub struct FilesystemConf {
     pub display_thread: bool,
     pub display_position: bool,
 
-    pub failed_worker_ttl: String,
+    pub failed_worker_ttl_ms: u64,
```
szbr00, line 92: ClientConf carries #[serde(alias = "failed_worker_ttl")] (etc.) for legacy numeric keys, but this libsdk FilesystemConf does not. Any TOML still using the old key names — including output from the Java toToml() until it is fixed — will silently drop those values instead of erroring. Consider adding the same aliases here for parity, or document that only *_ms keys are accepted in libsdk TOML.

szbr00 (commented):
Clean refactor: collapsing the String+Duration double-fields removes a pure round-trip in `init()` and aligns with the existing `*_timeout_ms` convention. Rust consumer updates look correct (`Duration::from_millis` at use sites, sync-check loops now compute in ms), and the serde aliases on `ClientConf` are a good migration guard for numeric legacy keys.

Blocking concern: `curvine-libsdk/java/.../FilesystemConf.java` was not updated while `filesystem_conf.rs` was. The Java `toToml()` -> JNI -> `FilesystemConf::from_str` path will not carry user time settings after this change. Please fix Java (and ideally add a libsdk/JNI test) before merge.

szbr00:
**`curvine-libsdk/java/src/main/java/io/curvine/FilesystemConf.java` (line ~74)** — must-fix

Rust `FilesystemConf` was renamed to `*_ms: u64`, but this Java class still declares String fields like `failed_worker_ttl = "10m"` and `toToml()` emits those old keys with quoted string values. `FilesystemConf::from_str` in Rust has no serde aliases and expects numeric `failed_worker_ttl_ms`, so Java/Hadoop configs are silently ignored (falling back to `#[serde(default)]` zeros) or fail to apply user tuning.

Please mirror the Rust rename here (`long failed_worker_ttl_ms = 600_000`, etc.) and update `toToml()` / `toString()` accordingly so the JNI path (`CurvineNative.newFilesystem(conf.toToml())`) stays compatible.

bigbigxu:
Would it be better to support human-readable size and duration values instead of exposing raw millisecond or byte counts?

Mature configuration systems commonly use explicit units. For example, Kubernetes-style resource quantities such as `512Mi` or `2Gi`, and duration strings such as `250ms`, `30s`, or `5m`. This approach is self-documenting and reduces mistakes caused by misreading numeric magnitudes or assuming the wrong unit.

For example:

```rust
pub bytes: ByteUnit,
pub time: DurationUnit,
```

The configuration could then be written as:

```toml
cache_size = "512MiB"
attr_timeout = "1s"
negative_timeout = "250ms"
```

Internally, these values can still be normalized to bytes and milliseconds. This would make the configuration easier to read and use while keeping the runtime representation unchanged.

If backward compatibility is required, we could continue accepting existing numeric values and additionally support unit-suffixed strings.

curvine-common/tests/client_conf_cli_test.rs
yangyang233333: Done in 4deb3b8: added `toml_legacy_string_duration_is_rejected`, asserting `failed_worker_ttl = "10m"` fails to deserialize.

curvine-libsdk/src/filesystem_conf.rs
yangyang233333: Fixed in 4deb3b8: added `#[serde(alias = "...")]` on all seven `*_ms` fields in libsdk FilesystemConf for parity with ClientConf.

curvine-client/src/rpc/job_master_client.rs
```diff
@@ -176,12 +178,12 @@ impl JobMasterClient {
                 _ => {
                     ticks += 1;
 
-                    let sleep_time = conf
-                        .sync_check_interval_max
-                        .min(conf.sync_check_interval_min * ticks);
-                    time::sleep(sleep_time).await;
+                    let sleep_ms = conf
+                        .sync_check_interval_max_ms
+                        .min(conf.sync_check_interval_min_ms * ticks);
```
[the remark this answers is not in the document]
yangyang233333: Fixed in 570a204: switched to `saturating_mul` before the `.min(max)` clamp, so overflow saturates to u64::MAX and is clamped instead of wrapping.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1179,10 +1179,10 @@ impl fs::FileSystem for CurvineFileSystem {
             }
 
             ticks += 1;
-            let sleep_time = check_interval_max.min(check_interval_min * ticks);
-            tokio::time::sleep(sleep_time).await;
+            let sleep_ms = check_interval_max_ms.min(check_interval_min_ms * ticks);
```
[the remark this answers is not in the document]
yangyang233333: Fixed in 570a204: same `saturating_mul` fix applied to the FUSE setlkw wait loop.

curvine-libsdk/src/filesystem_conf.rs
```diff
@@ -89,21 +88,21 @@ pub struct FilesystemConf {
...
-    pub max_sync_wait_timeout: String,
+    pub sync_check_interval_min_ms: u64,
+    pub sync_check_interval_max_ms: u64,
+    pub max_sync_wait_timeout_ms: u64,
     pub sync_check_log_tick: u32,
 
     pub enable_block_conn_pool: bool,
     pub block_conn_idle_size: usize,
-    pub block_conn_idle_time: String,
+    pub block_conn_idle_time_ms: u64,
```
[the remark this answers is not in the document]
yangyang233333: Fixed in 4deb3b8: Java `FilesystemConf` duration fields changed from String to `long *_ms`, so `toToml()` now emits numeric keys matching the Rust side; `toString()` updated too. Rust libsdk also gained serde aliases for legacy numeric keys.

szbr00 (approved):

yangyang233333:
Thanks @bigbigxu — this is a fair point, and unit-suffixed values like `"250ms"` / `"512MiB"` are genuinely more self-documenting.

The reason I went with raw `*_ms` here is consistency with what is *already exposed to users*. A large number of `ClientConf` fields are already surfaced as raw numeric values today — the existing `conn_timeout_ms`, `rpc_timeout_ms`, `data_timeout_ms`, the retry `*_duration_ms` / `*_sleep_ms` fields, etc. The goal of this PR was to bring the remaining `String + DurationUnit` stragglers into line with that established convention, not to introduce a second style. Having half the timeouts written as `"5m"` strings and the other half as `_ms` numbers is exactly the inconsistency this refactor set out to remove.

Standardizing everything on `DurationUnit` / `ByteUnit` is a reasonable direction, but it is the *opposite* migration and a much larger, project-wide change (it would touch all the existing raw-ms fields too). I think that is better handled as its own dedicated discussion/PR rather than folded in here. For now, to keep this change focused and the user-facing config uniform, I would prefer to expose the raw ms values consistently. Happy to file a follow-up issue to track a project-wide move to unit-suffixed config if the team wants to go that way.

PR #1078, opened 2026-07-13
build(docker): merge Fluid runtime support into main image

jlon:
# Summary

Merge Fluid runtime support into the main Curvine Docker image, remove the separate `curvine-fluid` image build path, and keep CacheRuntime DataLoad working with current Fluid `1.1.0-alpha.10` / image `v1.1.0-676f47a`.

# Issue Describe / Design

The old Fluid integration kept a standalone `curvine-fluid` image, its own Dockerfile, local Make targets, and a GitHub Actions workflow. After the Fluid scripts are included in the main Curvine image, that second build path no longer has a clear purpose and can drift from the main image.

Design in this PR:

- Put Fluid scripts directly into the main runtime image.
- Keep normal Curvine containers on the existing entrypoint path; only Fluid runtime signals delegate to `/fluid-entrypoint.sh`.
- Delete the local `make docker-build-fluid*` targets.
- Delete the standalone `curvine-fluid` Dockerfile and GitHub Actions workflow.
- Remove the runtime-image workflow job that used to trigger `build-fluid-image.yml`.
- Remove the release workflow check for `ghcr.io/curvineio/curvine-fluid`, because that image is no longer produced.
- Update CacheRuntime and ThinRuntime examples to use the main Curvine image.
- Keep a legacy DataLoad fallback for older/custom Fluid DataLoad charts that do not mount runtime config.
- For current Fluid DataLoad jobs, read mounted runtime config and infer the runtime namespace from env, config fields, `targetPath`, or serviceaccount namespace. This fixes DataLoad pods that mount `/etc/fluid/config/runtime.json` but do not set `FLUID_DATASET_NAMESPACE`.

Related Fluid issue: https://github.com/fluid-cloudnative/fluid/issues/6121
Related Fluid fix: https://github.com/fluid-cloudnative/fluid/pull/6113

# Changes

| Area | Change |
| --- | --- |
| Main image | Copy Fluid integration scripts into Rocky 9 and Ubuntu 24 runtime images. |
| Entrypoint | Detect Fluid mode by argument/env/config file and delegate to `/fluid-entrypoint.sh`; normal Curvine startup remains unchanged. |
| Fluid image | Delete the standalone Fluid Dockerfile and image build workflow. |
| GitHub Actions | Remove the runtime workflow job that built `curvine-fluid` and stop release checks from waiting for that image. |
| Makefile | Remove `docker-build-fluid`, `docker-build-fluid-cache`, and `docker-build-fluid-thin`. |
| CacheRuntime | Switch examples to `curvine:latest`, add default demo/e2e volumes, and document that `emptyDir` loses data on pod recreation. |
| DataLoad | Use mounted Fluid runtime config on current Fluid, keep the missing-config path as a legacy fallback, resolve Curvine paths back to mounted UFS sources before `cv load`, and infer namespace when Fluid omits `FLUID_DATASET_NAMESPACE`. |
| ThinRuntime | Switch sample image to the main Curvine image. |

# Test verified

| Test | Result |
| --- | --- |
| `make docker-build` with Rocky 9 option | Passed; built `curvine:latest` image `9c25908fb773`. |
| Load `curvine:latest` into local Kubernetes node runtime | Passed; kind node image id matched `9c25908fb773`. |
| `docker run --rm curvine:latest echo normal-entrypoint-ok` | Passed in earlier validation; normal main-image entrypoint path still works. |
| Local Fluid CacheRuntime e2e on Fluid `1.1.0-alpha.9` | Passed in earlier validation; master/worker became Ready with the main image. |
| External S3 MountUFS e2e | Passed in earlier validation; mounted the user-provided S3 path to `/flink/user` through Kubernetes Secret-based credentials. |
| Fluid `1.1.0-alpha.10` chart with `v1.1.0-676f47a` images | Passed; `cacheruntime-controller`, `dataset-controller`, `fluid-csi`, `fluid-webhook`, and `fluidapp-controller` ran with `v1.1.0-676f47a`. |
| CacheRuntime e2e on Fluid `1.1.0-alpha.10` / `v1.1.0-676f47a` | Passed; Dataset/PV/PVC became Bound and master/worker ran from the rebuilt main image. |
| DataLoad runtime config mount verification | Passed; DataLoad pod mounted `fluid-runtime-...` ConfigMap at `/etc/fluid/config` read-only. |
| Small-file DataLoad e2e on Fluid `1.1.0-alpha.10` / `v1.1.0-676f47a` | Passed; DataLoad completed in 10s for `/flink/user/.probe_write`, Curvine load job completed 17/17 bytes, and `cv fs stat` reported `Is complete: true`, `Len: 17`, storage state `Both`. |
| `generate_config.py` DataLoad namespace inference check | Passed; with no `FLUID_DATASET_NAMESPACE`, runtime config `targetPath` generated master endpoint in the DataLoad namespace instead of `default`. |
| `bash -n curvine-docker/deploy/entrypoint.sh curvine-docker/fluid/entrypoint.sh curvine-docker/fluid/mountUfs.sh` | Passed. |
| `python3 -m py_compile curvine-docker/fluid/generate_config.py curvine-docker/fluid/config-parse.py` | Passed. |
| YAML parse for Fluid sample manifests | Passed. |
| `git diff --check` | Passed. |
| `make format` | Passed. |
| `make -n docker-build-fluid` | Expected failure: target no longer exists. |

# Dependencies

- Related upstream Fluid issue: https://github.com/fluid-cloudnative/fluid/issues/6121
- Related upstream Fluid fix: https://github.com/fluid-cloudnative/fluid/pull/6113
- Validated with Fluid chart `fluid-1.1.0-alpha.10`; the chart appVersion still reports `1.1.0-9b3c2e2`, so the e2e explicitly overrode component image tags to `v1.1.0-676f47a`.
- No Curvine code PR dependency.
- After this PR, Fluid users should use the main Curvine image instead of `curvine-fluid`.


curvine-docker/fluid/cache-runtime/curvine-cache-runtime-class.yaml
```diff
@@ -76,10 +79,10 @@ dataOperationSpecs:
         with urllib.request.urlopen(req, context=context, timeout=5) as response:
             pod = json.load(response)
     except urllib.error.HTTPError as err:
-        print(f"Unable to read current pod labels from Kubernetes API: HTTP {err.code} {err.reason}", file=sys.stderr)
+        print(f"Unable to read current pod labels from Kubernetes API: HTTP {err.code} {err.reason}; falling back to DataLoad name inference", file=sys.stderr)
```
szbr00: This stderr says "falling back to DataLoad name inference", but the Python helper still exits 1. The outer bash only reaches hostname inference because of `dataset="$(get_dataset_from_pod_label ...)" || dataset=""`, so operators may read this as an in-function fallback that never happens. Consider either exiting 0 with no dataset output on API failure, or rewording the message to make clear the caller will try hostname inference next.

szbr00 (commented):
LGTM overall. Merging Fluid scripts into the main image with deploy-time delegation to `/fluid-entrypoint.sh` is a clean simplification, normal `docker run curvine:latest ...` behavior stays intact, and the DataLoad fallback path (pod labels → hostname/DNS inference → minimal config) matches the Fluid #6121 workaround described in the PR. Rocky9 build, CacheRuntime e2e, MountUFS, and DataLoad validation give good confidence. Minor follow-up on the misleading API-failure log line above; otherwise ready to merge.

szbr00:
This PR deletes `.github/workflows/build-fluid-image.yml` and removes the runtime workflow job that called it, but `.agents/skills/cv-run-workflow/SKILL.md` and `references/workflow-catalog.md` still list `build-fluid-image.yml`. Please update those skill docs in this PR or immediately after merge so automation/docs do not point at a removed workflow.

thuongle2210 (approved):

curvine-docker/fluid/cache-runtime/curvine-cache-runtime-class.yaml
jlon: Fixed in 6eb6e938: reworded the stderr so it says the caller will try DataLoad name inference after the helper exits.

jlon:
Fixed in 6eb6e938: removed the deleted `build-fluid-image.yml` workflow from the workflow skill docs and catalog.

szbr00 (approved):

PR #1092, opened 2026-07-13
refactor(worker): centralize block-relative I/O handling

gongxun0928:
# Summary

Refactor worker block I/O so handlers operate only on block-relative
offsets. Layout-specific open logic and absolute device addressing are
moved behind `BlockLayout` and `BlockRead/WriteContext`, without changing
the intended remote read/write behavior.

# Issue Describe / Design

No tracking issue yet.

### Why

Worker read/write handlers previously mixed two coordinate systems:

- local files used offsets relative to the block/file start
- SPDK bdevs used absolute device offsets, forcing handlers to special-case
  `bdev_offset`, `supports_short_circuit()`, and device `len()`/`pos()`

That made layout knowledge leak into the RPC session layer and made every
new storage backend a handler change.

### Approach

1. Keep `BlockMeta` as metadata only.
2. Make `BlockLayout` responsible for opening a block-scoped I/O session.
3. Introduce `BlockWriteContext` / `BlockReadContext` as the single place that:
   - stores `device_base`
   - translates block offsets to absolute device offsets
   - enforces block-size boundaries
4. Let handlers call `seek_to` / `read_region` / `write_region` without knowing
   whether the backend is a file or a bdev.

### Trade-offs

- Short-circuit remains a path handoff to the co-located client; it does not use
  `Block*Context` for data I/O, because the client already opens a per-block file
  and operates in block-relative coordinates.
- Capability probes such as `supports_resize()` / `supports_send_file()` /
  `as_local_mut()` remain as intentional escape hatches for filesystem-only
  features.

# Changes

| `curvine-server/.../block_io_context.rs` | New block-relative write/read session wrappers | None |
| `curvine-server/.../layout/mod.rs` | Extend `BlockLayout` with open/short-circuit APIs and shared offset validation | None |
| `curvine-server/.../layout/file_layout.rs` | File-backed open + short-circuit path | None |
| `curvine-server/.../layout/bdev_layout.rs` | SPDK-backed open; short-circuit returns `None` | None |
| `curvine-server/.../block_store.rs` | Store-level open/short-circuit delegation | None |
| `curvine-server/.../read_handler.rs` | Consume `BlockReadContext` | None for valid traffic; clearer offset handling |
| `curvine-server/.../write_handler.rs` | Consume `BlockWriteContext`; better flush/length-check abort | Cleanup on invalid complete is safer |
| `curvine-server/.../batch_write_handler.rs` | Track `BlockWriteContext` instead of raw devices | None |
| `curvine-server/.../worker_replication_manager.rs` | Use store open + actual read length | More correct under short reads |
| `curvine-server/tests/worker_test.rs` | Add abort and short-circuit resize regressions | Coverage only |
| `orpc/src/io/block_io.rs` | Add `supports_resize()` | None |

# Test verified

| `make format` | PASS | |
| `cargo test -p curvine-server --test worker_test -- --test-threads=1` | PASS | Includes oversized-complete abort and short-circuit resize coverage |
| `cargo test -p curvine-server --lib block_io -- --test-threads=1` | PASS | Block-relative offset translation unit tests |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/handler/write_handler.rs
+++ b/curvine-server/src/worker/handler/write_handler.rs
@@ -282,7 +282,7 @@ impl WriteHandler {
                 );
             }
             return err_box!(
-                "Invalid write offset: {}, block size: {}",
+                "Invalid block length: {}, block size: {}",
                 context.block.len,
                 context.block_size
             );
--- a/curvine-server/src/worker/handler/write_handler.rs
+++ b/curvine-server/src/worker/handler/write_handler.rs
@@ -14,6 +14,7 @@
 
 use crate::worker::block::BlockStore;
 use crate::worker::handler::WriteContext;
+use crate::worker::storage::BlockWriteContext;
 use crate::worker::{Worker, WorkerMetrics};
 use curvine_common::error::FsError;
 use curvine_common::proto::{BlockWriteResponse, DataHeaderProto};
@@ -22,7 +23,6 @@ use curvine_common::FsResult;
 use log::{info, warn};
 use orpc::common::{ByteUnit, TimeSpent};
 use orpc::handler::MessageHandler;
-use orpc::io::{BlockDevice, BlockIO};
 use orpc::message::{Builder, Message, RequestStatus};
 use orpc::{err_box, ternary, try_option_mut, CommonResult};
 use std::mem;
@@ -30,7 +30,7 @@ use std::mem;
 pub struct WriteHandler {
     pub(crate) store: BlockStore,
     pub(crate) context: Option<WriteContext>,
-    pub(crate) file: Option<BlockDevice>,
+    pub(crate) file: Option<BlockWriteContext>,
     pub(crate) is_commit: bool,
     pub(crate) io_slow_us: u64,
     pub(crate) metrics: &'static WorkerMetrics,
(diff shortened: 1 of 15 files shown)


curvine-server/src/worker/handler/write_handler.rs
```diff
@@ -266,22 +257,37 @@ impl WriteHandler {
...
         if context.block.len > context.block_size {
+            if let Err(abort_err) = self.store.abort_block(&context.block) {
+                log::warn!(
+                    "failed to abort block {} after length check failed: {}",
+                    context.block.id,
+                    abort_err
+                );
+            }
             return err_box!(
                 "Invalid write offset: {}, block size: {}",
```
szbr00: This failure path compares `context.block.len` against `context.block_size`, but the error still says "Invalid write offset". Consider rewording to something like "Invalid block length" so operators and tests are not misled about which field failed validation.

szbr00 (commented):
LGTM. Centralizing block-relative offset translation in `BlockReadContext`/`BlockWriteContext` cleanly removes the SPDK-vs-file branching from handlers, and the layout-owned `open_reader`/`open_writer`/`short_circuit` APIs are a solid boundary. The abort-on-flush-failure and abort-on-oversized-complete cleanup are meaningful hardening, and advancing replication by actual `read_len` is more correct than always subtracting the requested chunk size.

Regression coverage for oversized-complete abort and short-circuit resize is good. If you have SPDK builds available in CI/dev, worth also running `cargo test -p curvine-server --test spdk_worker_test --features spdk -- --test-threads=1` since `bdev_layout` open paths were moved.

szbr00 (approved):

PR #1100, opened 2026-07-14
fix(fuse): preserve zero permission mode

ljy1-lixing:
## Summary

Fix Curvine FUSE preserving permission mode `0000`.

When a file was changed to mode `0000` through `chmod` or `fchmod`, or created under a full `umask(0777)`, Curvine could report the file with a default mode such as `0755` instead of `0000`. The operation succeeded, but later `stat`/`fstat` did not reflect the requested permissions.

## Minimal Reproducer

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <sys/stat.h>
#include <unistd.h>

    char path[4096];
    struct stat st;

    if (argc != 2) {
        fprintf(stderr, "usage: %s <mount-dir>\n", argv[0]);
        return 2;
    }

    snprintf(path, sizeof(path), "%s/mode_zero.dat", argv[1]);
    unlink(path);

    int fd = open(path, O_CREAT | O_RDWR, 0644);
    if (fd < 0) {
        perror("open");
        return 1;
    }
    close(fd);

    if (chmod(path, 0000) != 0) {
        perror("chmod");
        return 1;
    }

    if (stat(path, &st) != 0) {
        perror("stat");
        return 1;
    }

    if ((st.st_mode & 0777) != 0000) {
        fprintf(stderr, "expected mode 0000, got %03o\n",
                (unsigned)(st.st_mode & 0777));
        return 1;
    }

    unlink(path);
    return 0;
}
```

```bash
gcc /tmp/mode_zero.c -o /tmp/mode_zero
/tmp/mode_zero /curvine-fuse
```

## Root Cause

Curvine FUSE treated `mode == 0` as an unset value.

That affected multiple paths:

- `create` skipped ACL/mode initialization when the computed mode was `0000`.
- `mkdir` had the same `mode != 0` guard.
- `status_to_attr` converted persisted `mode == 0` back to the configured default mode.

As a result, valid POSIX mode `0000` was lost or hidden behind the default permission mode.

## Changes

- Always apply the kernel-provided create mode after umask, even when it becomes `0000`.
- Always apply the kernel-provided mkdir mode after umask.
- Preserve `status.mode == 0` when converting metadata to FUSE attributes.
- Set default permissions only for synthetic `.` and `..` directory entries.
- Add a regression test covering `chmod`, `fchmod`, and full-umask create.

## Verification

- Minimal reproducer passes after the fix.
- `python3 build/tests/scripts/mode_zero_permissions_repro.py --dir /tmp/curvine-mode-zero-repro --allow-non-mount` passes.
- `cargo fmt --check` passes.

### DIFF
--- a/curvine-fuse/src/fuse_utils.rs
+++ b/curvine-fuse/src/fuse_utils.rs
@@ -174,16 +174,13 @@ impl FuseUtils {
     }
 
     pub fn create_opts(op: &Create<'_>, fs: &UnifiedFileSystem) -> CreateFileOpts {
-        let mut builder = CreateFileOptsBuilder::with_conf(&fs.conf().client);
-        if op.arg.mode != 0 {
-            builder = builder.acl(
+        CreateFileOptsBuilder::with_conf(&fs.conf().client)
+            .acl(
                 op.header.uid,
                 op.header.gid,
                 op.arg.mode & 0o7777 & !op.arg.umask,
             )
-        }
-
-        builder.build()
+            .build()
     }
 
     pub fn check_xattr(name: &str, read: bool) -> FuseResult<()> {
@@ -262,11 +259,12 @@ impl FuseUtils {
             }
         };
 
-        let mode = if status.mode != 0 {
-            FuseUtils::get_mode(status.mode, status.file_type)
+        let perm = if status.id == FUSE_UNKNOWN_INO as i64 && FuseUtils.is_dot(&status.name) {
+            FUSE_DEFAULT_MODE & !conf.umask
         } else {
-            FuseUtils::get_mode(FUSE_DEFAULT_MODE & !conf.umask, status.file_type)
+            status.mode
         };
+        let mode = FuseUtils::get_mode(perm, status.file_type);
         let size = FuseUtils::fuse_st_size(status);
 
         // For links, nlink should be greater than 1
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/fuse_utils.rs
```diff
@@ -319,7 +309,9 @@ impl FuseUtils {
     }
 
     pub fn new_dot_status(name: &str) -> FileStatus {
-        FileStatus::with_name(FUSE_UNKNOWN_INO as i64, name.to_string(), true)
+        let mut status = FileStatus::with_name(FUSE_UNKNOWN_INO as i64, name.to_string(), true);
+        status.mode = FUSE_DEFAULT_MODE;
```
szbr00: `new_dot_status` now sets `status.mode = FUSE_DEFAULT_MODE` (0777), but the removed `status_to_attr` fallback used `FUSE_DEFAULT_MODE & !conf.umask` for synthetic entries whose mode was 0. That means `.`/`..` in readdir can change from a umask-filtered mode (for example 0755) to full 0777. Consider passing `FuseConf` here (or applying umask in `status_to_attr` for `FUSE_UNKNOWN_INO` dot entries) so synthetic `.`/`..` keep the previous display semantics while real mode-0000 files still round-trip correctly.

szbr00 (approved):

szbr00 (approved):

PR #1133, opened 2026-07-15
fix(cli): cap load-status progress display

jlon:
# Summary

This PR prevents impossible load/export progress displays such as `200.0% (66/33 bytes)`. It caps user-facing progress values in both `BasicProgress` and `JobTaskProgress`, so CLI status output and client wait logs show the same bounded byte count.

# Issue Describe / Design

This is a bug fix.

- Root cause: `BasicProgress` and `JobTaskProgress::progress_string()` could receive `loaded_size > total_size` from export progress reports. The original CLI path displayed both the raw percentage and raw loaded byte count; the initial change capped `BasicProgress`, while `JobTaskProgress` still formatted the raw byte count.
- Reproduction steps: run `cv export <mounted-cv-file> --watch` for a small file on the `/acc-hdfs/starrocks` mount. The deployed CLI displayed completed export status as `200.0%` with byte counts like `64/32 bytes`.
- Fix approach: cap percentage and displayed completed bytes at total bytes when total is known, preserving raw completed bytes only when total is zero. `JobTaskProgress` now follows the same display contract as `BasicProgress`.

# Changes

| `curvine-common/src/utils/display.rs` | Clamp `BasicProgress` percentage and completed bytes when total is known | Fixes impossible CLI progress displays; normal progress output is unchanged |
| `curvine-common/src/utils/display.rs` | Add regression test for over-complete progress display | Prevents regressions for `200.0%` style output |
| `curvine-common/src/state/job.rs` | Clamp formatted `JobTaskProgress` byte counts to the known total | Keeps client wait logs and status formatting consistent with the CLI |
| `curvine-common/src/state/job.rs` | Add bar and non-bar formatting regression coverage | Prevents raw over-complete byte counts from returning through this progress surface |

# Test verified

| `cargo test -p curvine-common test_basic_progress_caps_over_complete_display -- --nocapture` before fix | PASS | Existing CLI regression coverage remains green |
| `cargo test -p curvine-common state::job::tests::progress_string_caps_loaded_bytes_at_total --lib -- --exact` before fix | FAIL | Confirmed the new regression test catches raw `66/33` byte output |
| `cargo test -p curvine-common state::job::tests::progress_string_caps_loaded_bytes_at_total --lib -- --exact` after fix | PASS | Targeted status-formatting regression test passes |
| `cargo test -p curvine-common --lib -- --skip fs::state_file::tests::test_write_read` | PASS | 73 passed; the skipped existing test fails independently because its `../testing` directory fixture is not created |
| `cargo check -p curvine-common --lib` | PASS | Common crate compiles cleanly |
| `cargo fmt --check` | PASS | Formatting clean |
| `/app/curvine/bin/cv` smoke test on `/acc-hdfs/starrocks` | PASS | Covered safe `fs` operations, `report`, `node`, `version`, explicit `load`, `export`, `load-status`, and auto-load; temporary test data was removed |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-common/src/state/job.rs
+++ b/curvine-common/src/state/job.rs
@@ -265,11 +265,16 @@ impl JobTaskProgress {
     pub fn progress_string(&self, show_bar: bool) -> String {
         let loaded = self.loaded_size.max(0) as u64;
         let total = self.total_size.max(0) as u64;
+        let display_loaded = if total == 0 {
+            loaded
+        } else {
+            loaded.min(total)
+        };
 
         let percentage = if total == 0 {
             0.0
         } else {
-            (loaded as f64 / total as f64 * 100.0).min(100.0)
+            display_loaded as f64 / total as f64 * 100.0
         };
 
         if show_bar {
@@ -277,7 +282,7 @@ impl JobTaskProgress {
                 return format!(
                     "[{}] 0.0% ({} / {})",
                     "░".repeat(20),
-                    ByteUnit::byte_to_string(loaded),
+                    ByteUnit::byte_to_string(display_loaded),
                     ByteUnit::byte_to_string(total)
                 );
             }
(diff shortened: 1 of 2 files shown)


szbr00 (commented):
**curvine-common/src/state/job.rs** (`JobTaskProgress::progress_string`, ~line 265)

JobTaskProgress::progress_string already clamps percentage to 100.0 but still formats the raw loaded byte count. Client wait logs (job_master_client) can therefore still show impossible byte ratios like 66/33 even though the percent reads 100.0%. Consider applying the same completed.min(total) clamp here so all progress surfaces stay consistent with BasicProgress.

szbr00 (commented):
**Follow-up (non-blocking): `curvine-server/src/worker/task/load_task_runner.rs` (~line 117)**

This PR correctly caps the CLI display, but progress is still sourced from `writer.pos()` vs `reader.len()`. For export (cv -> ufs), the repro showed `loaded_size` exactly 2× `total_size` (66/33), which suggests the upstream counters may disagree rather than a pure formatting issue. Worth a follow-up to trace why `pos` exceeds `len` on small exports (e.g. block alignment, padding, or stale reader length) so job metadata stays accurate even outside the CLI.

jlon:
Fixed in ce58e64: JobTaskProgress now caps its formatted loaded byte count to the known total, matching BasicProgress. Added regression coverage for progress strings with and without the bar.

The writer.pos versus reader.len mismatch is outside this display-only PR. It changes authoritative worker progress metadata and needs a separate root-cause investigation in LoadTaskRunner before changing behavior.

szbr00 (approved):

PR #1144, opened 2026-07-15
fix(fuse): inherit group in setgid directories

ljy1-lixing:
**Summary**

Fix FUSE file creation inside setgid directories so new files inherit the parent directory group.

Before this fix, Curvine used the caller's gid from the FUSE request when creating a file. If the parent directory had `S_ISGID` set and a different group, the new file still got the caller's group instead of the parent directory's group.

**Minimal Reproducer**

```c
#define _GNU_SOURCE
#include <errno.h>
#include <fcntl.h>
#include <grp.h>
#include <pwd.h>
#include <stdio.h>
#include <string.h>
#include <sys/stat.h>
#include <unistd.h>

    char dir[4096];
    char file[4096];
    struct stat st;
    struct group *bin_group;
    struct passwd *nobody_user;
    gid_t parent_gid;
    gid_t old_egid;
    uid_t old_euid;
    int fd;

    if (argc != 2) {
        fprintf(stderr, "usage: %s <mount-dir>\n", argv[0]);
        return 2;
    }

    bin_group = getgrnam("bin");
    if (!bin_group) {
        fprintf(stderr, "missing group: bin\n");
        return 1;
    }
    parent_gid = bin_group->gr_gid;

    nobody_user = getpwnam("nobody");
    if (!nobody_user) {
        fprintf(stderr, "missing user: nobody\n");
        return 1;
    }

    snprintf(dir, sizeof(dir), "%s/setgid_parent", argv[1]);
    snprintf(file, sizeof(file), "%s/child", dir);

    unlink(file);
    rmdir(dir);

    if (mkdir(dir, 0777) != 0) {
        perror("mkdir");
        return 1;
    }

    if (chown(dir, 0, parent_gid) != 0) {
        perror("chown dir to root:bin");
        return 1;
    }

    if (chmod(dir, 02777) != 0) {
        perror("chmod dir setgid");
        return 1;
    }

    old_euid = geteuid();
    old_egid = getegid();

    setegid(nobody_user->pw_gid);
    seteuid(nobody_user->pw_uid);

    fd = open(file, O_CREAT | O_EXCL | O_RDWR, 0777);

    seteuid(old_euid);
    setegid(old_egid);

    if (fd < 0) {
        fprintf(stderr, "open child failed errno=%d (%s)\n", errno, strerror(errno));
        return 1;
    }
    close(fd);

    if (stat(file, &st) != 0) {
        perror("stat child");
        return 1;
    }

    if (st.st_gid != parent_gid) {
        fprintf(stderr, "FAIL: child gid=%u, expected parent gid=%u\n",
                (unsigned)st.st_gid,
                (unsigned)parent_gid);
        return 1;
    }

    printf("PASS: child inherited parent setgid directory gid=%u\n",
           (unsigned)parent_gid);

    unlink(file);
    rmdir(dir);
    return 0;
}
```

```bash
gcc /tmp/repro_setgid_group_inherit.c -o /tmp/repro_setgid_group_inherit
sudo /tmp/repro_setgid_group_inherit /curvine-fuse
```

Before the fix, Curvine reported the child file with the caller's gid, for example:

```text
FAIL: child gid=65534, expected parent gid=2
```

**Root Cause**

The FUSE create path built `CreateFileOpts` directly from the request header:

- owner came from the caller uid
- group came from the caller gid
- mode came from the requested create mode after umask

That is correct for normal directories, but not for setgid directories. On Linux, when a file is created inside a directory with `S_ISGID` set, the new file must inherit the parent directory's group.

Curvine did not check the parent directory mode before creating the child file, so the parent group inheritance rule was skipped.

**Changes**

- In FUSE `create()`, fetch the parent directory metadata before creating the file.
- If the parent directory has `S_ISGID` set, override the new file group with the parent directory group.
- Keep the existing caller uid and requested mode behavior unchanged.
- Add a regression test for file creation inside setgid directories.

**Verification**

- Minimal C reproducer passes after the fix.
- `python3 build/tests/scripts/setgid_group_inherit_repro.py --dir /tmp/curvine-setgid-group-inherit --allow-non-mount` passes.
- `cargo fmt --check` passes.

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -823,7 +823,12 @@ impl fs::FileSystem for CurvineFileSystem {
         self.ensure_writable_path(&path, RpcCode::CreateFile)
             .await?;
 
-        let opts = FuseUtils::create_opts(&op, &self.fs);
+        let mut opts = FuseUtils::create_opts(&op, &self.fs);
+        let parent_status = self.state.fs_stat(ino, None).await?;
+        if parent_status.mode & FUSE_S_ISGID != 0 {
+            opts.group = parent_status.group;
+        }
+
         let handle = self.state.fs_create(ino, name, op.arg.flags, opts).await?;
         let attr = FuseUtils::status_to_attr(&self.conf, handle.status())?;
 
--- a/build/tests/fuse-test.sh
+++ b/build/tests/fuse-test.sh
@@ -64,6 +64,7 @@ This script tests basic filesystem operations including:
   - Active-writer path truncate regression (PR #962)
   - Read-only open with O_CREAT creates a missing file
   - Mode 0000 is preserved by chmod, fchmod, and umask-filtered create
+  - Files created in setgid directories inherit the parent group
 
 For FIO performance tests, use fio-test.sh instead.
 
@@ -1389,6 +1390,15 @@ test_mode_zero_permissions() {
         "mode_zero_permissions_repro.py" --dir "$TEST_DIR"
 }
 
+# Test 24: files created in setgid directories inherit the parent group
+test_setgid_group_inherit() {
+    CURRENT_TEST_GROUP="Test 24: Setgid Group Inheritance"
+    print_header "$CURRENT_TEST_GROUP"
+    run_python_script_test \
+        "Testing file group inheritance in setgid directories" \
+        "setgid_group_inherit_repro.py" --dir "$TEST_DIR"
+}
+
 # Print final report
 print_report() {
     print_header "Test Summary"
@@ -1497,6 +1507,7 @@ main() {
     test_pr962_active_writer_truncate
     test_open_creat_rdonly
     test_mode_zero_permissions
+    test_setgid_group_inherit
 
     test_git_clone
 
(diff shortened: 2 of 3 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -823,7 +823,12 @@ impl fs::FileSystem for CurvineFileSystem {
         self.ensure_writable_path(&path, RpcCode::CreateFile)
             .await?;
 
-        let opts = FuseUtils::create_opts(&op, &self.fs);
+        let mut opts = FuseUtils::create_opts(&op, &self.fs);
+        let parent_status = self.state.fs_stat(ino, None).await?;
+        if parent_status.mode & FUSE_S_ISGID != 0 {
```
szbr00, line 828: The same setgid group inheritance logic should be applied to the `mkdir()` method (around line 700). On Linux, subdirectories created inside a setgid directory should also inherit the parent's group, and the `S_ISGID` bit itself should be propagated to the new subdirectory. Currently `mkdir()` calls `mkdir_opts()` which always uses the caller's gid, so setgid semantics are only partially implemented — files inherit the parent group but directories do not.

ljy1-lixing: Good point. This is the same mkdir setgid inheritance issue addressed in PR #1159. That PR makes child directories created under a setgid parent inherit the parent group and propagate the S_ISGID bit.

szbr00 (approved):

PR #1145, opened 2026-07-15
fix(fuse): harden kernel version and setattr timestamps (#1137)

### LINKED ISSUE #1137: [FEATURE] fuse: parse kernel version as (major, minor) tuple instead of f32 (avoid misparse and panic) (by LiAuto)

`FuseUtils::get_kernel_version` represents the kernel version as `f32`, which both misparses multi-digit minor versions and can panic.

- `curvine-fuse/src/fuse_utils.rs:49`-`64`:
  ```rust
  format!("{}.{}", parts[0], parts[1]).parse::<f32>().unwrap()
  ```
- Consumers: `curvine-fuse/src/lib.rs:162` (`FUSE_CLONE_FD_MIN_VERSION: f32 = 4.2f32`), `:170` (`UNIX_KERNEL_VERSION`), used in `curvine-fuse/src/session/fuse_mnt.rs:62` to gate `FUSE_DEV_IOC_CLONE`.

**Problems**

1. `f32` misparse: `6.10` parses to `6.1`, `4.20` to `4.2` (indistinguishable from real `4.2`). Version comparison is semantically wrong.
2. Panics: `Command::new("uname").output().expect(...)` (`:53`) and `.parse::<f32>().unwrap()` (`:62`) both abort the process if `uname` is unavailable or the version string is non-numeric — a startup crash.
3. Practical impact on clone-fd gating: `*UNIX_KERNEL_VERSION >= FUSE_CLONE_FD_MIN_VERSION (4.2)`. `6.10`→`6.1` still `>= 4.2` (clone fd correctly enabled), but `4.10`→`4.1 < 4.2` wrongly DISABLES clone fd even though 4.10 supports `FUSE_DEV_IOC_CLONE` (available since 4.2). The panic is the more serious risk.

- Represent the version as `(u32, u32)` (major, minor) and compare as a tuple.
- On parse failure (uname missing / unexpected format), return `None` or `(0, 0)` and conservatively DISABLE clone fd rather than panic.
- Change `FUSE_CLONE_FD_MIN_VERSION` to `(4, 2)`.

**Additional context**

From a curvine-fuse FuseUtils review on the latest tree (`438c997`).

Acceptance / test gaps:
- `kernel_version_6_10_compares_greater_than_6_9`
- `kernel_version_4_10_enables_clone_fd` (4.10 >= 4.2)
- parse failure disables clone fd instead of panicking.

### LINKED ISSUE #1139: [FEATURE] fuse: fuse_setattr_to_opts loses sub-second (nanosecond) time precision (by LiAuto)

`FuseUtils::fuse_setattr_to_opts` converts atime/mtime using only the seconds field, dropping the nanosecond component.

- `curvine-fuse/src/fuse_utils.rs:347`-`361`:
  ```rust
  atime = Some((setattr.atime * 1000) as i64);   // ignores setattr.atimensec
  mtime = Some((setattr.mtime * 1000) as i64);   // ignores setattr.mtimensec
  ```
- `fuse_setattr_in` carries `atimensec` / `mtimensec` (`curvine-fuse/src/raw/fuse_abi.rs`), which a `utimensat` call populates.

**Problems**

- Sub-second precision from `utimensat` is silently lost (only second granularity survives).
- No validation that nsec < 1_000_000_000.
- `ctime/ctimensec/fh/lock_owner` handling is unspecified.
- `FATTR_SIZE` is intentionally NOT handled here (it is resized in the caller `set_attr`); this split should be documented so the helper's responsibility is clear.

- Convert to milliseconds as `sec * 1000 + nsec / 1_000_000` for both atime and mtime.
- Validate nsec is in range.
- Add a comment clarifying that size is handled by the caller, not this helper.

**Additional context**

From a curvine-fuse FuseUtils review on the latest tree (`438c997`).

Acceptance / test gaps:
- `setattr_preserves_millisecond_from_nsec`: a setattr with `atimensec` set produces a millisecond-accurate atime, not a truncated-to-second value.

ligeng8848:
# Summary

This PR fixes two correctness issues in the FUSE utility layer: kernel versions are now compared as `(major, minor)` tuples without startup panics, and explicit setattr timestamps now preserve millisecond precision from their nanosecond fields.

# Issue Describe / Design

Closes #1137
Closes #1139

Root causes:
- Kernel releases were parsed as `f32`, so `4.10` became `4.1`; command and parse failures could also panic during startup.
- `fuse_setattr_to_opts` multiplied only the seconds fields and discarded `atimensec` / `mtimensec`.

Reproduction:
- Parse `4.10.0` and compare it with the clone-fd minimum kernel version `4.2`.
- Convert a `fuse_setattr_in` containing non-zero nanoseconds and inspect the generated millisecond timestamps.

Fix approach:
- Parse kernel releases into `(u32, u32)`, fall back to `(0, 0)` on command or parse failure, and conservatively disable clone-fd in that case.
- Convert timestamps with `sec * 1000 + nsec / 1_000_000`, validate the nanosecond range, and use checked arithmetic.
- Keep `FATTR_SIZE` handling in the caller, where file-handle use and cache invalidation are managed.

# Changes

| `curvine-fuse/src/fuse_utils.rs` | Add safe kernel-version parsing, timestamp conversion, validation, and unit tests | Correct version ordering and preserve sub-second setattr precision |
| `curvine-fuse/src/lib.rs` | Change clone-fd version constants and cached kernel version from `f32` to tuple | Clone-fd gating now uses semantic version ordering |
| `curvine-fuse/src/session/fuse_mnt.rs` | Compare and log tuple kernel versions | No change to the successful clone-fd path |

# Test verified

| `cargo fmt --all -- --check` | PASS | Local and remote Linux |
| `cargo test -p curvine-fuse fuse_utils::tests` | PASS | 5 passed, 0 failed on remote Linux |
| `cargo clippy -p curvine-fuse --all-targets -- --deny warnings --allow clippy::uninlined-format-args` | PASS | Remote Linux |
| `cargo test -p curvine-fuse` | BASELINE FAIL | 163 passed; 3 unrelated `dir_tree` tests failed |
| Baseline `438c997` `dir_tree` tests | SAME FAILURES | The same 3 assertions fail without this PR |

# Dependencies

- Related PRs: None
- Related issues: #1137, #1139
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fuse_utils.rs
+++ b/curvine-fuse/src/fuse_utils.rs
@@ -46,21 +46,23 @@ impl FuseUtils {
         BytesMut::from(bytes)
     }
 
-    pub fn get_kernel_version() -> f32 {
-        let output = Command::new("uname")
-            .arg("-r")
-            .output()
-            .expect("Failed to execute 'uname -r'");
-
-        // Convert output to string and remove the line break at the end
-        let kernel_ver = String::from_utf8_lossy(&output.stdout).trim().to_string();
-
-        let parts: Vec<&str> = kernel_ver.split('.').collect();
-        if parts.len() < 2 {
-            1f32
-        } else {
-            format!("{}.{}", parts[0], parts[1]).parse::<f32>().unwrap()
+    pub fn get_kernel_version() -> (u32, u32) {
+        let Ok(output) = Command::new("uname").arg("-r").output() else {
+            return (0, 0);
+        };
+        if !output.status.success() {
+            return (0, 0);
         }
+
+        let kernel_release = String::from_utf8_lossy(&output.stdout);
+        Self::parse_kernel_version(kernel_release.trim()).unwrap_or((0, 0))
+    }
+
+    fn parse_kernel_version(kernel_release: &str) -> Option<(u32, u32)> {
+        let mut parts = kernel_release.split('.');
+        let major = parts.next()?.parse().ok()?;
+        let minor = parts.next()?.parse().ok()?;
+        Some((major, minor))
     }
 
     pub fn get_mode(perm: u32, typ: FileType) -> u32 {
@@ -385,5 +387,32 @@ impl FuseUtils {
             nlink: 1,
             ..Default::default()
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::FuseUtils;
+    use crate::FUSE_CLONE_FD_MIN_VERSION;
+
+    #[test]
+    fn kernel_version_6_10_compares_greater_than_6_9() {
+        let version = FuseUtils::parse_kernel_version("6.10.3").unwrap();
+
+        assert!(version > (6, 9));
+    }
+
+    #[test]
+    fn kernel_version_4_10_enables_clone_fd() {
+        let version = FuseUtils::parse_kernel_version("4.10.0").unwrap();
+
+        assert!(version >= FUSE_CLONE_FD_MIN_VERSION);
+    }
+
+    #[test]
+    fn kernel_version_parse_failure_disables_clone_fd() {
+        let version = FuseUtils::parse_kernel_version("unexpected").unwrap_or((0, 0));
+
+        assert!(version < FUSE_CLONE_FD_MIN_VERSION);
     }
 }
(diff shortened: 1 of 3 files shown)


ligeng8848:
Split into two focused pull requests for one-issue-per-PR review: #1148 closes #1137, and #1147 closes #1139. Superseded by those PRs.

PR #1146, opened 2026-07-15
perf(server): parallelize large UFS-to-Curvine loads with multi-stream fan-out

yangyang233333:
# Summary

Large single-file loads from object storage (S3/UFS) were bottlenecked on a single reader+writer pipeline (~215 MB/s). This PR fans a large load out into N independent streams (each its own reader + writer over a block-aligned range), lifting throughput ~6x.

# Performance

Cold read, BOS (Beijing) S3 backend, single-node cluster:

| Path | File | Time | Throughput |
| ---- | ---- | ---- | ---------- |
| Before (single stream) | 1 GiB | 4759 ms | **215 MB/s** |
| After (8 streams) | 20 GiB | 15058 ms | **1360 MB/s** |

**~6.3x throughput.** Data integrity verified -- md5 of the 20 GiB file read back from Curvine matches the S3 source exactly (`24eeb2845cbfda238b78fa165c21607d`).

# Changes

| Module / File | Change |
| ------------- | ------ |
| `curvine-server/.../load_task_runner.rs` | New `run_parallel` path: fan a large UFS to CV load into N block-aligned segments, each with its own reader+writer; falls back to single stream below a size threshold |
| `curvine-ufs/src/opendal.rs` | Reader wired to a configurable concurrent range-GET count |
| `curvine-ufs/src/conf.rs` | `read_concurrent` / `read_chunk_size` fixed as internal constants (2 / 16 MiB) |

Tunables (mount properties): `load_task.parallel_streams` (default 8), `load_task.parallel_threshold` (default 64 MiB).

# Test verified

| Test | Result |
| ---- | ------ |
| `cargo test -p curvine-server` (incl. 6 new segment tests) | PASS |
| `cargo test -p curvine-ufs` | PASS |
| `cargo clippy -p curvine-server -p curvine-ufs` | clean |
| 20 GiB load + md5 verify | PASS |

### DIFF
--- a/curvine-server/src/worker/task/load_task_runner.rs
+++ b/curvine-server/src/worker/task/load_task_runner.rs
@@ -282,6 +282,10 @@ impl LoadTaskRunner {
         let block_size = self.task.info.job.block_size.max(1);
         let seg = Self::segment_len(src_len, streams, block_size);
         let spend = TimeSpent::new();
+        // Absolute deadline shared by every stream, so a single hung stream is
+        // caught mid-segment instead of only after the whole (multi-GiB) segment
+        // finishes. Same budget as the outer join-loop timeout check.
+        let deadline_ms = LocalTime::mills() + self.task_timeout_ms;
 
         // Pre-create + resize the target to src_len so ALL blocks are allocated
         // up front (see property 1 above). We drop the owner writer WITHOUT
@@ -295,6 +299,7 @@ impl LoadTaskRunner {
 
         // Fan out: each stream reads its range from UFS and writes it to the
         // (already allocated) target region with its own reader + writer.
+        let task_timeout_ms = self.task_timeout_ms;
         let mut handles = Vec::with_capacity(streams);
         for i in 0..streams {
             let off = i as i64 * seg;
@@ -307,6 +312,7 @@ impl LoadTaskRunner {
             let src = source_path.clone();
             let dst = target_path.clone();
             let rt = self.fs.clone_runtime();
+            let task = self.task.clone();
             handles.push(rt.spawn(async move {
                 let mut reader = ufs.open(&src).await?;
                 reader.seek(off).await?;
@@ -317,6 +323,26 @@ impl LoadTaskRunner {
 
                 let mut remaining = len;
                 while remaining > 0 {
+                    // Honor cancellation and the shared deadline INSIDE the
+                    // segment loop: a segment can be multiple GiB, so checking
+                    // only between segments would let a canceled/timed-out job
+                    // keep reading from UFS and committing blocks. On early exit
+                    // cancel the writer so it does not complete() a partial
+                    // segment onto the pre-resized target.
+                    if task.is_cancel() {
+                        let _ = writer.cancel().await;
+                        return FsResult::Ok(0);
+                    }
+                    if LocalTime::mills() > deadline_ms {
+                        let _ = writer.cancel().await;
+                        return err_box!(
+                            "Task {} exceed timeout {} ms (segment [{}, {}))",
+                            task.info.task_id,
+                            task_timeout_ms,
+                            off,
+                            off + len
+                        );
+                    }
                     let want = remaining.min(Self::READ_CHUNK_BYTES) as usize;
                     let chunk = reader.async_read(Some(want)).await?;
                     if chunk.is_empty() {
@@ -330,6 +356,7 @@ impl LoadTaskRunner {
                 // hole in this segment. Fail loudly instead of committing partial
                 // data that would later read back as corrupt.
                 if remaining != 0 {
+                    let _ = writer.cancel().await;
                     return err_box!(
                         "short read on segment [{}, {}): {} bytes missing (source shorter than stat len?)",
                         off,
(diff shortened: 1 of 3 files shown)


curvine-server/src/worker/task/load_task_runner.rs
```diff
@@ -158,6 +191,196 @@ impl LoadTaskRunner {
...
+                    );
+                }
+                writer.complete().await?;
+                reader.complete().await?;
+                FsResult::Ok(len)
+            }));
+        }
+
+        let mut written: i64 = 0;
+        for h in handles {
```
szbr00: The parallel join loop returns early on cancel (and also on segment error/timeout) without stopping or awaiting the already-spawned stream tasks. Those tasks never observe TaskContext::is_cancel(), so they can keep reading from UFS and calling writer.complete() on the pre-resized target after the job is canceled or has failed. That is a regression from the serial path, which checks cancellation on every chunk.

Consider passing Arc<TaskContext> (or an AbortHandle) into each spawned stream, polling is_cancel() inside the segment read loop, and on any early exit either aborting the remaining tasks or joining them and canceling/closing their writers before returning.

curvine-server/src/worker/task/load_task_runner.rs
```diff
@@ -158,6 +191,196 @@ impl LoadTaskRunner {
...
+                // data that would later read back as corrupt.
+                if remaining != 0 {
+                    return err_box!(
+                        "short read on segment [{}, {}): {} bytes missing (source shorter than stat len?)",
+                        off,
+                        off + len,
+                        remaining
+                    );
+                }
+                writer.complete().await?;
```
szbr00, line 412: All parallel writers share the same FsContext client_name, so the first writer.complete() clears file_write while other streams may still be committing blocks. This seems to work because resize pre-allocates blocks and satisfies compute_len == file.len, but it is easy to misread as a bug. A short comment here pointing back to correctness property (1) above would help future maintainers avoid "fixing" this into per-stream client IDs unnecessarily.

curvine-server/src/worker/task/load_task_runner.rs
```diff
@@ -158,6 +191,196 @@ impl LoadTaskRunner {
...
+            handles.push(rt.spawn(async move {
+                let mut reader = ufs.open(&src).await?;
+                reader.seek(off).await?;
+                // open_for_write(overwrite=false): attach as an additional writer
+                // to the existing (resized) file without truncating it.
+                let mut writer = fs.open_for_write(&dst, false).await?;
+                writer.seek(off).await?;
+
+                let mut remaining = len;
+                while remaining > 0 {
```
szbr00, line 363: In the serial copy loop, task_timeout_ms is checked every chunk; here it is only checked in the outer join loop after an entire segment finishes. For large files each segment can be multiple GiB, so a single slow/hung stream may run well past the configured timeout before the parent observes it (especially if it is not the first handle awaited). Consider adding the same deadline check inside this loop and/or wrapping each spawned segment future in tokio::time::timeout.

bigbigxu:
I believe dynamically calculating parallelism based on file size is a better approach. For example, assign 1 parallelism unit for a 1GB file, cap the maximum parallelism per task at 8, and allocate 4 parallelism units for a 4GB file.
In small-file scenarios, the overhead of metadata operations far exceeds the cost of actual data I/O; excessive parallelism will instead degrade overall performance.

curvine-server/src/worker/task/load_task_runner.rs
yangyang233333: Fixed in 4e72b1f. Each stream now checks `is_cancel()` inside its segment read loop and cancels its writer on early exit; the join loop aborts all not-yet-awaited handles on any early exit (cancel/error/timeout), so a stream blocked in a UFS read can no longer keep reading or `complete()` onto the target after the job is canceled/failed.

curvine-server/src/worker/task/load_task_runner.rs
yangyang233333: Fixed in 4e72b1f. Added a shared deadline checked inside each stream's segment read loop (cancels the writer and errors out on breach), so a hung stream is caught mid-segment instead of only after a multi-GiB segment finishes.

curvine-server/src/worker/task/load_task_runner.rs
yangyang233333: Done in 4e72b1f. Added a comment at `complete()` explaining the shared client_name is intentional and safe due to correctness property (1) (resize pre-allocates all blocks, so compute_len == file.len regardless of commit order), and warning against "fixing" it into per-stream client IDs.

yangyang233333:
Thanks, good catch. Fixed in 65e1921. Rather than a size->count table I derive the width from a single tunable:

```
effective_streams = clamp(src_len / min_bytes_per_stream, 1, cap)
```

- `load_task.parallel_streams` is now the **cap** (still default 8)
- `load_task.min_bytes_per_stream` (default 256 MiB) is the minimum data a stream must move to be worthwhile — this subsumes the old boolean `parallel_threshold` (a file below it yields 1 stream = no fan-out)

Examples (min=256 MiB, cap=8): 100 MiB->1, 512 MiB->2, 1 GiB->4, 2 GiB->8, 200 GiB->8. So small/mid-size files no longer jump straight to full parallelism. Added unit tests for the width math.

szbr00 (approved):

Issue #1134, opened 2026-07-15
[BUG]: Auto-detected worker hostname may be unreachable and cause FUSE write EIO


ljy1-lixing:
**Describe the bug**

After `305f46a fix(conf): make local deployment host settings configurable (#1063)`,
`etc/curvine-env.sh` changed the default local deployment host settings from fixed
`localhost` to auto-detected host/IP values.

The worker hostname is now selected with:

```bash
hostname -I | awk '{print $NF}'
```

On hosts with multiple network interfaces, IPv6 addresses, container networks, or
virtual interfaces, this may select an address that Curvine clients/FUSE cannot
actually connect to. The worker then registers an unreachable address, and FUSE
writes fail with `EIO`.

In my environment, the worker registered as:

```text
Worker register: addr = fd00:1:1:1::2/fd00:1:1:1::2:8997
```

FUSE then failed to write data:

```text
ERROR fs_writer_buffer.rs:129 buffer writer error: IO([]Connection refused (os error 111))
WARN fuse_response.rs:139 send_rep ... errno Connection refused ... : 5
```

**To Reproduce**

1. Build and deploy Curvine from a revision including `305f46a`.
2. Start a local cluster without explicitly setting `CURVINE_WORKER_HOSTNAME` or
   `CURVINE_CLIENT_HOSTNAME`.
3. Mount FUSE:

```bash
./build/dist/bin/curvine-master.sh start
./build/dist/bin/curvine-worker.sh start
./build/dist/bin/curvine-fuse.sh start --mnt-path /mnt/curvine
```

4. Run a simple write through FUSE:

```bash
printf 'abc\n' > /mnt/curvine/write_probe
stat /mnt/curvine/write_probe
cat /mnt/curvine/write_probe
```

**Expected behavior**

The default local deployment should register a worker address that is reachable
by local Curvine clients/FUSE.

For single-node local deployment, FUSE writes should succeed without requiring
users to manually override hostname environment variables.

**OS Version**

Distribution: Debian GNU/Linux 10 (buster)
Kernel: 5.4.250-9-velinux1u2-amd64
Architecture: x86_64
glibc: 2.28
GCC: 8.3.0
FUSE mount options observed: rw,relatime,user_id=0,group_id=0,default_permissions,allow_other,force_umount
Curvine version: tag/commit corresponding to current main of oppo/curvine

PR #1169, opened 2026-07-16
feat(conf): resolve advertise IP via interface and hostname (#1134)

lzjqsdd:
# Summary

Add `CURVINE_INTERFACE` so production/multi-node deployments resolve the advertise IP from the node hostname (via `/etc/hosts`) and validate it against the configured NIC, while keeping safe `localhost` defaults for local single-node use after #1136.

# Issue Describe / Design

Related to #1134 (follow-up after #1136)

Closes CUR-37

## Background

PR #1136 restored `localhost` defaults to fix unreachable auto-detected NICs causing FUSE `EIO`. That default is correct for local/dev but unsuitable for distributed clusters, where advertising `localhost` breaks cross-node connectivity.

## Design

1. Leave `CURVINE_INTERFACE` empty → local/dev defaults (`localhost`).
2. Set `CURVINE_INTERFACE=eth0` (same on every node) → production mode:
   - Read `$(hostname)`
   - Resolve hostname to a non-loopback IPv4 (`getent` / system resolver, typically `/etc/hosts`)
   - Read IPv4 from the configured interface
   - Fail startup if resolve fails, result is loopback, or resolved IP ≠ interface IP
3. Ops prepares `/etc/hosts` on each node: `<eth0_ip> <hostname>`
4. Explicit `CURVINE_*_HOSTNAME` overrides still win; when all three are set, detection is skipped

# Changes

| `etc/curvine-env.sh` | Add `CURVINE_INTERFACE`, hostname→IP resolution, NIC validation, and local vs production comments | Local/dev unchanged (`localhost`). Production must set interface + `/etc/hosts` or fail loudly |
| `curvine-docker/deploy/example/conf/curvine-env.sh` | Example production env using interface + hostname resolution | Deploy example no longer uses blind `hostname -i` |

# Test verified

| `bash -n etc/curvine-env.sh` | PASS | Syntax OK |
| `bash -n curvine-docker/deploy/example/conf/curvine-env.sh` | PASS | Syntax OK |
| Default (no `CURVINE_INTERFACE`) → localhost | PASS | master/worker/client = localhost |
| Explicit `CURVINE_*_HOSTNAME` overrides | PASS | Overrides preserved |
| Invalid `CURVINE_INTERFACE` | PASS | Fatal error, no silent localhost |
| Valid interface + matching `/etc/hosts` | PASS | worker/client advertise NIC IPv4; master uses hostname |
| Interface/IP mismatch (`lo` vs resolved host IP) | PASS | Fatal error with fix hint |
| All three hostname env vars set with bad interface | PASS | Detection skipped; overrides used |

# Dependencies

- Related PRs: #1136
- Related issues: #1134
- Blocks / blocked by: None

### DIFF
--- a/curvine-docker/deploy/example/conf/curvine-env.sh
+++ b/curvine-docker/deploy/example/conf/curvine-env.sh
@@ -2,11 +2,41 @@
 
 export CURVINE_HOME="$(cd "$(dirname "$0")"/..; pwd)"
 
+# Production / multi-node example.
+# Set the cluster NIC (same value on every node). Ops must ensure /etc/hosts
+# contains "<interface_ip> <hostname>" on each machine.
+# For local single-node, leave CURVINE_INTERFACE empty and use localhost
+# defaults from the packaged etc/curvine-env.sh instead.
+export CURVINE_INTERFACE=${CURVINE_INTERFACE:-eth0}
 
-export CURVINE_MASTER_HOSTNAME=$(hostname)
+LOCAL_HOSTNAME=$(hostname 2>/dev/null || true)
+if [ -z "$LOCAL_HOSTNAME" ] || [ "$LOCAL_HOSTNAME" = "localhost" ]; then
+    echo "ERROR: CURVINE_INTERFACE=$CURVINE_INTERFACE requires a real hostname" >&2
+    exit 1
+fi
 
-export CURVINE_WORKER_HOSTNAME=$(hostname -i)
+IFACE_IP=$(ip -4 -o addr show dev "$CURVINE_INTERFACE" 2>/dev/null \
+    | awk '{print $4}' | cut -d/ -f1 | head -n1)
+if [ -z "$IFACE_IP" ]; then
+    echo "ERROR: cannot read IPv4 from interface '$CURVINE_INTERFACE'" >&2
+    exit 1
+fi
 
-export CURVINE_CLIENT_HOSTNAME=$(hostname -i)
+RESOLVED_IP=$(getent ahostsv4 "$LOCAL_HOSTNAME" 2>/dev/null | awk '{print $1; exit}')
+if [ -z "$RESOLVED_IP" ] || [[ "$RESOLVED_IP" == 127.* ]]; then
+    RESOLVED_IP=$(getent hosts "$LOCAL_HOSTNAME" 2>/dev/null | awk '{print $1; exit}')
+fi
+if [ -z "$RESOLVED_IP" ] || [[ "$RESOLVED_IP" == 127.* ]]; then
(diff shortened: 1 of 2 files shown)


yangcx000:
Thanks for tackling the distributed advertise-IP story — the `CURVINE_INTERFACE`-gated model (empty = safe localhost, set = resolve + validate + fail-fast) is a clean way to keep #1136's local defaults while enabling multi-node, and the NIC-vs-`/etc/hosts` consistency check is a nice guardrail.

One architectural question before this lands: **should the hostname→IPv4 resolution and NIC validation live in Rust (e.g. `curvine-common` / `orpc::io::net::NetUtils`) rather than in `curvine-env.sh`?**

The concern with keeping it in bash:

- **Duplicated + split logic.** `NetUtils` already resolves hostname→IPv4 (`local_hostname()` → `System::host_name()`, `get_ip_v4()` → `to_socket_addrs()` first IPv4), and the Rust side re-resolves it anyway (`FsContext::with_rt` calls `NetUtils::local_ip(conf.client.hostname)`). After this PR the advertise IP is computed in bash, exported as `CURVINE_*_HOSTNAME`, then partially re-resolved in Rust — the same concept lives in two languages.
- **External-tool fragility.** The script depends on `ip`/`ifconfig`/`ipconfig`/`getent` and falls back to an inline `python3 socket.getaddrinfo` snippet. Minimal production images often lack `python3` (and sometimes `getent`/`ifconfig`), so the fallback path can silently not fire. In Rust this is std-library / a small crate (e.g. `if-addrs` for NIC enumeration; `to_socket_addrs` for name resolution) with no runtime tool dependency.
- **Testability.** The bash path is only checkable via `bash -n` + manual scenarios. A Rust `resolve_advertise_ip(interface)` returning `CommonResult<String>` (loopback rejection, NIC-mismatch error) is straightforward to unit-test deterministically.
- **Unified error handling.** The `_curvine_env_fail`/`exit 1` path is separate from the Rust config-validation errors (`ClusterConf::init` / `err_box!`); consolidating would give one consistent failure surface with one message style.

To be fair, this isn't a free "reuse `NetUtils`" — the Rust side would need *new* helpers for the pieces bash adds beyond current `NetUtils`: reading the configured NIC's IPv4, non-loopback filtering, and the resolved-IP-vs-NIC-IP check. But those are small, cross-platform-friendly in Rust, and would let `curvine-env.sh` shrink back to just passing `CURVINE_INTERFACE` through, with all resolution/validation in one tested place.

Not necessarily a blocker for this PR if the goal is to unblock distributed deployments now — but if we agree on the direction, it might be worth landing this as the interim fix and filing a follow-up to move resolution into `curvine-common`, so we don't grow more shell logic that Rust will later duplicate. Curious what you think.

Issue #1079, opened 2026-07-13
[FEATURE] fuse: supervise receiver/sender tasks and propagate failures (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

In `FuseSession::run_all`, one task is spawned per receiver and sender. Two problems compound:

1. **Errors are swallowed.** Inside each spawned task, `receiver.start(...)` / `sender.start()` errors are only logged; the async closure returns `()`, so the outer `JoinHandle` resolves as `Ok(())` even after a real failure. `run()` also logs `Ok(Err(err))` from `run_all` but still proceeds to `fs.unmount()` and returns `Ok(())`.
2. **Sequential await hides late failures.** `run_all` stores all handles in a `Vec` and awaits them in insertion order: `for handle in handles { handle.await?; }`. If the first receiver is healthy and long-lived, `run_all` waits there forever; a later sender/receiver that panics or exits is never observed until earlier handles finish.

Locations:
- `curvine-fuse/src/session/fuse_session.rs:199` / `:205` (`run()` logs `run_all` outcomes but continues)
- `curvine-fuse/src/session/fuse_session.rs:322` / `:331` (spawned receiver/sender closures swallow errors)
- `curvine-fuse/src/session/fuse_session.rs:317` / `:341` (sequential `for handle in handles { .await? }`)

**Impact**

- A receiver/sender can die and the session keeps running in a degraded state.
- Fatal IO/decode/channel failures may not reach the process exit status; supervision (systemd/k8s) treats the process as healthy.
- Task-failure detection is ineffective for long-running sessions.

**Describe the solution you'd like**

- Spawned receiver/sender tasks return their real `Result`.
- `run_all` uses `tokio::task::JoinSet` / `FuturesUnordered` to react to the first unexpected task exit/failure, then broadcasts shutdown to remaining tasks and drains them (with a bounded wait).
- `run()` returns an error for non-signal fatal `run_all` failures instead of converting them to `Ok(())`.

**Sender write-error classification (from sender review)**

Today `FuseSender::start` swallows write failures and keeps looping: a `RequestReply` send error becomes `WriteOutcome::Failed` + log, `NotifyReply`/legacy `Reply` errors are logged, and `start` returns `Ok(())` on channel close (`fuse_sender.rs:105` / `:112` / `:165` / `:179` / `:187`). The sender has no shutdown channel and cannot signal the session. As part of returning real `Result`s from the sender, classify write errors:
[...]

PR #1175, opened 2026-07-17
fix(fuse): handle shutdown run_all errors (#1082)

### LINKED ISSUE #1082: [FEATURE] fuse: handle run_all Ok(Err) in signal shutdown branches (by LiAuto)

The TERM and SIGUSR1 branches in `FuseSession::run` await the `run_all` handle like this:

```rust
if let Err(e) = run_all_handle.await {
    error!(...);
}
```

This only catches the `JoinError` (task panic/cancel). It ignores the inner `Ok(Err(err))` — a real `run_all` error surfaced during shutdown.

- `curvine-fuse/src/session/fuse_session.rs:228` (TERM branch)
- `curvine-fuse/src/session/fuse_session.rs:251` (SIGUSR1 branch)

**Impact**

- A real `run_all` error during shutdown is silently ignored.
- SIGUSR1 can proceed to `persist` even when receiver/sender shutdown failed in an unexpected way.

- Match both layers: `Result<CommonResult<()>, JoinError>`.
- Log and define policy for the `Ok(Err(err))` case.
- For SIGUSR1, do not persist if receiver/sender shutdown produced a fatal error unless that is explicitly determined safe.

**Additional context**

From a curvine-fuse session review. Related to the task-supervision rework (once `run_all` returns a meaningful `Result`, both signal branches must inspect it).

Acceptance / test gaps:
- TERM/SIGUSR1 branches handle `Ok(Err(_))` from `run_all_handle.await`, not just `JoinError`.

ligeng8848:
# Summary

Inspect both result layers when awaiting the spawned `run_all` task during signal shutdown. Termination shutdown now logs receiver/sender failures, while SIGUSR1 refuses to persist state after a failed shutdown.

# Issue Describe / Design

Closes #1082

Root cause: the TERM and SIGUSR1 branches only checked the outer Tokio `JoinError` and silently ignored an inner `Ok(Err(...))` returned by `run_all`.

Fix approach: flatten the nested join result in one helper and use it in both signal branches. TERM records the failure and continues normal unmount; SIGUSR1 returns the failure without persisting potentially unreliable state.

# Changes

| `curvine-fuse/src/session/fuse_session.rs` | Inspect both layers of the `run_all` result and add regression tests | Fatal shutdown errors are no longer silently ignored |

# Test verified

| `cargo test -p curvine-fuse --lib run_all_shutdown_result_tests -- --nocapture` | PASS | 3 passed |
| `cargo test -p curvine-fuse --lib session::fuse_session -- --nocapture` | PASS | 10 passed |
| `cargo clippy -p curvine-fuse --lib --tests -- -D warnings` | PASS | No warnings |
| `make format` | PASS | Repository pre-commit checks |
| Real `/dev/fuse` mount and SIGTERM smoke test | PASS | Read/write succeeded; process exited and mount was removed |

# Dependencies

- Related PRs: None
- Related issues: #1082, #1079
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/fuse_session.rs
+++ b/curvine-fuse/src/session/fuse_session.rs
@@ -295,10 +295,12 @@ impl<T: FileSystem> FuseSession<T> {
 
                 // Record the shutdown intent BEFORE persist: sigusr1_persist
                 // denotes intent, not a completed unmount (persist success/failure
-                // is a separate state_persist_total{status}). A run_all or persist
-                // error below may return before the explicit fs.unmount() hook;
-                // FuseMnt drop still unmounts, and RunCleanupGuard aborts the
-                // watcher and sets health 0 on those paths.
+                // is a separate state_persist_total{status}). A run_all error
+                // returns before persist, so `mnts` still auto-unmounts on drop.
+                // A persist error may occur after `persist_inner` disables
+                // `auto_unmount` for fd inheritance, so those mounts remain mounted.
+                // RunCleanupGuard still aborts the watcher and sets health 0 on
+                // either path.
                 shutdown_once.record_once(SHUTDOWN_SIGUSR1_PERSIST);
                 let _ = self.shutdown_tx.send(true);
                 if let Err(e) = flatten_run_all_result(run_all_handle.await) {
--- a/curvine-fuse/src/session/fuse_session.rs
+++ b/curvine-fuse/src/session/fuse_session.rs
@@ -55,7 +55,7 @@ pub struct FuseSession<T> {
 }
 
 /// Phase 3b (P2-1/P2-2/P3-1): closes out a `run()` invocation on EVERY exit path
-/// — normal completion, signal arms, and the SIGUSR1-persist `?` early return.
+/// — normal completion, signal arms, and the SIGUSR1 run-all/persist error returns.
 /// Its `Drop` is deliberately lightweight, synchronous, infallible, non-blocking:
 /// it only `abort()`s the fd-watcher task (so a stale watcher can't cross-session
 /// overwrite the process-global `kernel_fd_health`) and sets that gauge to 0
@@ -87,6 +87,21 @@ impl Drop for RunCleanupGuard {
 fn close_out_health<F: FnOnce(bool)>(enabled: bool, set_health: F) {
     if enabled {
         set_health(false);
+    }
+}
+
+/// Flatten the result returned by awaiting the spawned `run_all` task.
+///
+/// The outer result reports whether the Tokio task joined successfully, while
+/// the inner result reports whether `run_all` itself succeeded. Signal shutdown
+/// paths must inspect both layers so a clean task join cannot hide a real
+/// receiver/sender error.
+fn flatten_run_all_result(
+    result: Result<CommonResult<()>, tokio::task::JoinError>,
+) -> CommonResult<()> {
+    match result {
+        Ok(inner) => inner,
+        Err(err) => Err(err.into()),
     }
 }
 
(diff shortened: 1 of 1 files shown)


curvine-fuse/src/session/fuse_session.rs
```diff
@@ -280,14 +295,18 @@ impl<T: FileSystem> FuseSession<T> {
 
                 // Record the shutdown intent BEFORE persist: sigusr1_persist
                 // denotes intent, not a completed unmount (persist success/failure
-                // is a separate state_persist_total{status}). The `?` below may
-                // early-return on persist error WITHOUT unmount (unchanged control
-                // flow); the RunCleanupGuard still aborts the watcher + sets
-                // health 0 on that path.
+                // is a separate state_persist_total{status}). A run_all or persist
+                // error below may return before the explicit fs.unmount() hook;
+                // FuseMnt drop still unmounts, and RunCleanupGuard aborts the
```
szbr00: The claim "FuseMnt drop still unmounts" is accurate for the run_all error path (mnts haven't entered persist yet, so auto_unmount is still true). However, for a persist error, persist_inner sets mnt.auto_unmount(false) at line 468 and consumes the mnts in the for loop — so FuseMnt::drop does NOT unmount on that path. The previous comment was more precise in noting "early-return on persist error WITHOUT unmount". Also note that FileSystem::unmount() (the "explicit hook") is a no-op default that CurvineFileSystem doesn't override, so the actual unmount on the success path is also via FuseMnt::drop (with auto_unmount=true before persist consumes them), not via fs.unmount().

Consider restoring the distinction between the two error paths to avoid misleading future readers into assuming the mount is always cleaned up on persist failure.

ligeng8848: Thanks. Fixed in 72a8eac: clarified that run_all errors auto-unmount before persist, while persist errors may occur after auto_unmount is disabled and can leave mounts mounted.

yangcx000 (approved):
LGTM — nicely scoped fix for a real bug, verified against the code.

**Root cause confirmed.** `run_all` returns `CommonResult<()>` and is `tokio::spawn`-ed, so `run_all_handle.await` is `Result<CommonResult<()>, JoinError>` — genuinely double-layered. The old `if let Err(e) = run_all_handle.await` in both signal arms matched only the outer `JoinError` and silently dropped an inner `Ok(Err(..))`. Good that the `tokio::select!` first arm already destructures all three cases, so scoping the fix to the two re-await sites (and leaving that arm untouched, since it needs the distinct RUN_ALL_ERROR/RUN_ALL_PANIC metrics) is exactly right.

**The two arms diverge correctly.** TERM logs and continues to unmount, while SIGUSR1 refuses to persist and returns — the safe split: a hot-upgrade persist should not serialize potentially-inconsistent state, and falling back to a clean unmount via the `mnts` drop is the right default. I verified the auto_unmount timing the comment describes: `FuseMnt::auto_unmount` defaults to `true` and is only flipped to `false` inside `persist_inner`, so the `return Err` path drops `mnts` with auto-unmount still on (mounts torn down), while a persist error happens after that flag is cleared (mounts stay). `RunCleanupGuard`'s early-return Drop is correct too. `err.into()` resolves cleanly to `CommonError` (`JoinError: Error + Send + Sync`), and the three unit tests cover all three branches.

**One optional, non-blocking nit:** the helper's doc-comment says the flatten guards against "a clean task join [hiding] a real receiver/sender error." Looking at `run_all`, the *functional* receiver/sender errors (`start().await` returning `Err`) are already logged-and-swallowed inside the spawned closures, which return `()`. The only way `run_all` yields an inner `Err` is `handle.await?` propagating a child accept/send task that **panicked or was aborted** — so the wording is slightly off; consider "so a clean join of the `run_all` task cannot hide a panic in one of its child accept/send tasks." The fix's value (surfacing those panics at shutdown) is real regardless. Also `join_error_is_preserved`'s `contains("cancelled")` softly couples to tokio's Display wording — fine to keep, or relax to `is_err()`. Neither is blocking.

PR #1178, opened 2026-07-17
feat(conf): add net_interface to auto-resolve node IPv4 from a NIC

yangyang233333:
# Summary

Add an optional `net_interface` config field so each node can auto-resolve
its own local IPv4 from a named NIC (e.g. `eth0`) at startup, instead of
hard-coding a hostname/IP per node. Useful for multi-node deployments where
the same config is distributed to every node.

# Design

- New top-level field `net_interface` (default empty -> behavior unchanged).
- When non-empty, the IPv4 bound to that interface (via `getifaddrs`) is
  applied to master/journal/worker/client hostnames, and the
  `CURVINE_*_HOSTNAME` env vars are ignored (interface takes precedence).
- When empty, the existing hostname + env-var override path is used as-is.
- `check_master_hostname`, under `net_interface`, validates that the local
  journal address resolves to a node id in `journal_addrs`; otherwise raft
  later fails with an opaque "Not a master role" error, so we surface a
  clear, actionable startup error instead.
- Multi-node caveat documented in the sample configs: membership lists
  (`journal_addrs` / `master_addrs`) must still be written explicitly, and
  each node's entry must match the exact IPv4 the interface resolves to.

# Changes

| `curvine-common/src/conf/cluster_conf.rs` | Add `net_interface` field, `interface_ipv4()`, either/or resolution in `from()`, validation in `check_master_hostname()`, unit tests | None when field empty (default) |
| `curvine-common/Cargo.toml` | Enable `nix` `net` feature for `getifaddrs` | None |
| `etc/curvine-cluster.toml`, `curvine-docker/deploy/example/conf/curvine-cluster.toml`, `curvine-docker/deploy/spdk/curvine-cluster-spdk.toml` | Document `net_interface` (commented out) | None |

# Test verified

| `cargo test -p curvine-common --lib cluster_conf::tests` | PASS | 4 tests: loopback resolve, unknown-iface error, journal-addr match/mismatch |
| `cargo build -p curvine-common -p curvine-server` | PASS | |
| `cargo fmt -p curvine-common -- --check` | PASS | |
| `cargo clippy -p curvine-common` | PASS | no warnings |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-common/Cargo.toml
+++ b/curvine-common/Cargo.toml
@@ -48,6 +48,10 @@ anyhow = { workspace = true }
 tracing = { workspace = true }
 bitflags = { workspace = true }
 
+# Enumerate network interface addresses (getifaddrs) to resolve the local IPv4
+# from a named interface such as `eth0`. `net` feature adds the ifaddrs API.
+nix = { workspace = true, features = ["net"] }
+
 mimalloc = { workspace = true, optional = true }
 
 [target.'cfg(unix)'.dependencies]
--- a/curvine-common/src/conf/cluster_conf.rs
+++ b/curvine-common/src/conf/cluster_conf.rs
@@ -100,6 +100,27 @@ impl ClusterConf {
         // interface honors the per-role hostname env-var overrides instead.
         if !conf.net_interface.is_empty() {
             let ip = Self::interface_ipv4(&conf.net_interface)?;
+
+            // net_interface takes precedence over the CURVINE_*_HOSTNAME env
+            // vars. Warn (rather than silently ignore) when any of them is also
+            // set, so an operator who exported a hostname override but sees it
+            // have no effect can tell why. `from()` runs during config loading,
+            // before `Logger::init`, so the log macros would be dropped — use
+            // eprintln! to make the warning visible on stderr.
+            for env_key in [
+                Self::ENV_MASTER_HOSTNAME,
+                Self::ENV_WORKER_HOSTNAME,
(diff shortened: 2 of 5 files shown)


yangcx000:
LGTM — verified the key paths: default-empty keeps behavior unchanged, `check_master_hostname`'s new check is a superset of the old `hostname_exists` (adds the raft port) so the early return skips nothing, and `client.hostname` = NIC IP is safe since connections use the worker's registered `ip_addr`.

One non-blocking suggestion: when `net_interface` is set, `CURVINE_*_HOSTNAME` is silently ignored — consider a `log::warn!` when both are present so it's not surprising to debug (especially alongside the shell path in #1169).

yangyang233333:
Thanks for the review! Addressed the suggestion in fbb5697: when `net_interface` is set, `from()` now emits a `log::warn!` for each `CURVINE_*_HOSTNAME` env var that is also present, noting that net_interface takes precedence and the env var is ignored. This makes the override visible in logs instead of failing silently.

yangyang233333:
Correction on the warning implementation (5e75b81): `from()` runs during config loading, *before* `Logger::init`, so a `log::warn!` there would be dropped by the uninitialized subscriber. Switched to `eprintln!` to stderr so the warning is actually visible (consistent with the existing pre-logger-init eprintln! in `orpc/src/common/logger.rs`).

yangcx000 (approved):

Issue #1125, opened 2026-07-14
[FEATURE] fuse: validate cache-related config combinations (direct_io / write_back_cache / open_direct_on_stale) (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

`FuseConf::init` performs no consistency validation across the cache/IO-path switches, so conflicting or dead combinations are silently accepted.

- `curvine-common/src/conf/fuse_conf.rs:158`-`176` (fields), `:397`-`405` (defaults), `:243`-`269` (`init` — no cross-field checks)
- negotiation/use in `curvine-fuse/src/fs/curvine_file_system.rs:382`-`428` (`init`) and `:753`-`805` (`open`)

**Conflicting / dead combinations (verified)**

- `direct_io = true` + `write_back_cache = true`: direct I/O bypasses the page cache while write-back caching relies on the kernel buffering writes — semantically conflicting. In `init`, `write_back_cache` toggles `FUSE_WRITEBACK_CACHE` while `direct_io` drives the per-open `FUSE_FOPEN_DIRECT_IO`; both can be enabled together today.
- `direct_io = true` + `open_direct_on_stale = true`: `open_direct_on_stale` only takes effect in the non-direct branch of `open` (`curvine_file_system.rs`: it is consulted only when `!direct_io`), so enabling it alongside global `direct_io` is a dead config.
- `cache_readdir = true` with directory mutations: needs correct invalidation on directory change, otherwise stale directory entries are more visible.

**Describe the solution you'd like**

- In `init`, reject or warn on clearly conflicting/dead combinations:
  - `direct_io && write_back_cache` → reject (or force `write_back_cache = false` with a warning).
  - `direct_io && open_direct_on_stale` → ignore `open_direct_on_stale` with a warning (it is dead under global direct I/O).
- Document the intended interaction of the three cache layers (kernel entry/attr TTL vs process metadata cache vs data-path direct_io/write_back), and add a startup log of the effective cache/IO model.
- `write_back_cache = true` additionally interacts with flush/fsync/release semantics (see the capability-negotiation issue) — keep it gated until that path is complete.

**Additional context**

From a FuseConf configuration-init review. Distinct from the numeric-validation issue (this is cross-field cache semantics). Relates to the init capability-allowlist issue (#1088) for the `FUSE_WRITEBACK_CACHE` negotiation side.

Acceptance / test gaps:
- `init` rejects (or warns + normalizes) `direct_io && write_back_cache`.
[...]

Issue #1073, opened 2026-07-13
[FEATURE] fuse: validate io_threads > 0 in FuseConf::init (by LiAuto)


yangcx000:
**Describe the bug**

`FuseConf::init` does not reject `io_threads == 0`. `AsyncRuntime::new` passes `io_threads` straight to `tokio::runtime::Builder::worker_threads`, and Tokio rejects zero worker threads, so `builder.build().unwrap()` panics.

Locations:
- `curvine-common/src/conf/fuse_conf.rs:245` (`FuseConf::init`)
- `orpc/src/runtime/async_runtime.rs:30` (`AsyncRuntime::new`)

**To Reproduce**

1. Set `io_threads = 0` in the `[fuse]` config (or `--io-threads 0`).
2. Start curvine-fuse.
3. Observe a Tokio runtime-build panic.

**Expected behavior**

`FuseConf::init` validates `io_threads > 0` and returns a structured error. Optionally reject `mnt_number == 0` for clearer config semantics (current `get_all_mnt_path` treats `<= 1` as one mount).

**Related division-by-zero config values (validate together)**

Two more config values feed unchecked modulo/index operations and should be validated in the same config-init hardening:
- `path_lock_stripes` — `NodeState::lock_path` computes `hash % self.path_locks.len()` (`node_state.rs:189`); `path_lock_stripes = 0` (no `> 0` check in `FuseConf::init`) → divide-by-zero panic on create/release paths. Default is 1024.
- `sync_check_log_tick` (client conf) — `set_lkw` computes `ticks % log_ticks` with `log_ticks = conf.sync_check_log_tick` (`curvine_file_system.rs:1185`); `sync_check_log_tick = 0` → divide-by-zero panic while waiting for a lock. Default is 3 (`client_conf.rs:480`), no init validation. Fix: validate `> 0` in `ClientConf::init`, or guard with `if log_ticks > 0 && ticks % log_ticks == 0`.

**Explicitly out of scope** (verified non-issues, do NOT change):
- `worker_threads == 0` — `AsyncRuntime::new` only calls `max_blocking_threads(worker_threads)` when `> 0`; zero means Tokio's default blocking pool.
- `fuse_channel_size == 0` / `stream_channel_size == 0` — `AsyncChannel::new(0)` intentionally creates an unbounded channel; defaults rely on this.
- `channels[0]` indexing — `get_all_mnt_path` always returns >= 1 mount, so `channels` is non-empty.

[...]

PR #1179, opened 2026-07-17
fix(conf): complete FuseConf::init numeric validation (#1124)

### LINKED ISSUE #1124: [FEATURE] fuse: complete FuseConf::init numeric validation (path_lock_stripes, list_limit, umask, background, sync_check_log_tick) (by LiAuto)

`FuseConf::init` currently validates only `io_threads > 0` (from #1073 / PR #1075) and `max_readahead_kb != Some(0)`. Several other config values are still unchecked and cause runtime panics, silent degradation, or nonsensical protocol values. #1073 was closed after fixing only `io_threads`; the additional divide-by-zero / range items noted there were not implemented and are collected here.

Current `init` (`curvine-common/src/conf/fuse_conf.rs:243`-`269`) checks: `io_threads > 0`, Duration parsing, path normalization, `max_readahead_kb`. Missing:

**Divide-by-zero panics (highest priority):**
- `path_lock_stripes` — `NodeState::lock_path` computes `hash % self.path_locks.len()` (`node_state.rs:189`); `path_lock_stripes = 0` → empty `path_locks` → divide-by-zero panic on create/open/release. Default 1024.
- `sync_check_log_tick` (client conf) — `set_lkw` computes `ticks % log_ticks` with `log_ticks = conf.sync_check_log_tick` (`curvine_file_system.rs:1166`/`1185`); `sync_check_log_tick = 0` → divide-by-zero panic while waiting for a lock. Default 3 (`client_conf.rs:444`), no validation. Fix: validate `> 0` in `ClientConf::init`, or guard with `if log_ticks > 0 && ticks % log_ticks == 0`.

**Degradation / nonsensical values:**
- `list_limit` — `DirHandle::get_batch` loops `while guard.buf.len() < self.limit` with `self.limit = conf.list_limit` (`dir_handle.rs:88`, `node_state.rs:707`); `list_limit = 0` → the loop never pulls a real entry → readdir returns empty. Validate `> 0`.
- `max_background` / `congestion_threshold` — returned in the FUSE init response (defaults 256/192). Config allows `max_background = 0` or `congestion_threshold > max_background`, giving the kernel a nonsensical congestion window. Validate `max_background > 0` and `0 < congestion_threshold <= max_background` (optionally default `congestion_threshold = max_background * 3/4` when unset).
- `umask` — `u32` with no range check (`fuse_conf.rs:113`/`384`); a value > `0o7777` pollutes mode computation. Validate `umask <= 0o7777`.
- `mnt_number = 0` — silently treated as one mount by `get_all_mnt_path` (`<= 1` → single mount). Reject `mnt_number == 0` (or normalize to 1 in `init` with a clear comment) instead of the implicit `get_all_mnt_path` handling.

**Auto-normalization consistency:**
- `tasks_per_mnt = 0` means "follow io_threads" and `init` sets `tasks_per_mnt = io_threads` (`fuse_conf.rs:254`-`256`). But `get_conf` runs `init` twice (once in `ClusterConf::from`, once after CLI overrides at `mount_args.rs:332`): if the config file pre-inits `tasks_per_mnt` to 32 and the CLI then sets `--io-threads 64`, `tasks_per_mnt` stays 32 (no longer 0) and does not track the new `io_threads`. Add a `tasks_per_mnt_explicit` marker (or re-derive when the user did not set it explicitly).

**Out of scope (verified non-issues):**
- `worker_threads == 0` — `AsyncRuntime::new` only calls `max_blocking_threads(worker_threads)` when `> 0`; zero means Tokio's default blocking pool, so it does not panic. Optional: normalize `0` to `Utils::worker_threads(io_threads)` for clarity, but not required.
- `fuse_channel_size == 0` / `stream_channel_size == 0` — `AsyncChannel::new(0)` is an intentional unbounded channel.

**Additional context**

From a FuseConf configuration-init review on the latest tree (after #1073's io_threads fix merged). Groups all remaining `init`-level numeric/range validation so it is fixed in one place.

Acceptance / test gaps:
- `init_rejects_zero_path_lock_stripes`
- `init_rejects_zero_list_limit`
- `init_rejects_zero_mnt_number` (or normalizes to 1)
- `init_rejects_invalid_background_threshold` (max_background=0 or congestion>max_background)
- `init_rejects_umask_out_of_range`
- `sync_check_log_tick == 0` cannot divide-by-zero in `set_lkw`
- `tasks_per_mnt` auto-tracks a CLI `--io-threads` override when not explicitly set

yangcx000:
# Summary

`FuseConf::init` previously validated only `io_threads` and `max_readahead_kb`. This adds the remaining numeric/range checks so an invalid config fails at mount startup instead of panicking or silently degrading at runtime, and fixes `tasks_per_mnt` not tracking a CLI `--io-threads` override.

# Issue Describe / Design

Closes #1124

## Numeric validation (via `err_box!`, matching the existing `io_threads` style)

- **`path_lock_stripes > 0`** — it is the modulus in `NodeState::lock_path` (`hash % path_locks.len()`). `0` stripes → empty vec → **divide-by-zero panic** on create/release. (Highest priority: real panic.)
- **`list_limit > 0`** — `0` makes `DirHandle::get_batch`'s `while buf.len() < limit` loop never pull a real entry, so **readdir silently returns empty**.
- **`mnt_number > 0`** — `0` was silently treated as a single mount by `get_all_mnt_path` (`<= 1` → single). Reject it as a meaningless value.
- **`max_background > 0` and `0 < congestion_threshold <= max_background`** — both are returned verbatim in the FUSE init reply; a zero window or an inverted threshold gives the kernel a nonsensical congestion window.
- **`umask <= 0o7777`** — input hygiene (e.g. an octal value mistakenly written as decimal). Note: this is a sanity bound, not a bug fix — `effective_perm`'s `default_mode & !umask` only uses the low permission bits, so a too-large umask does not actually corrupt anything today.

## `tasks_per_mnt` (0 = "follow io_threads")

`init()` runs twice (once in `ClusterConf::from`, once after CLI overrides in `mount_args`). It previously normalized `tasks_per_mnt = io_threads` in place on the first pass, which froze the field, so a later `--io-threads` override was **not** tracked (e.g. config `io_threads=32` + `--io-threads 64` left the per-mount task count at 32).

Fix: stop mutating the field in `init()`; resolve on read via the new `FuseConf::effective_tasks_per_mnt()`, called at the single consumer `FuseChannel::new`. The raw value stays intact, so resolution always sees the current (possibly overridden) `io_threads`. An explicit non-zero `tasks_per_mnt` is preserved as-is.

## Behavior change (migration note)

A config with `congestion_threshold > max_background`, or either field at `0`, now **fails `init()`** instead of being silently clamped by the kernel. Note `--max-background` and `--congestion-threshold` are independent CLI flags: lowering only `--max-background` below the default `congestion_threshold` (192) will now fail fast. All in-tree configs use the defaults (256/192) and set none of these fields, so nothing in the repo breaks.

## Not included

`sync_check_log_tick` (listed in the issue as a divide-by-zero) no longer divides by zero — #1074 changed `ticks % log_ticks` to `ticks.is_multiple_of(log_ticks)`, which does not panic on `0`. It is also a `ClientConf` field, outside this `FuseConf::init` scope. `worker_threads == 0` / channel sizes `== 0` are the issue's own listed non-issues.

# Changes

| `curvine-common/src/conf/fuse_conf.rs` | 5 numeric validations in `init()`; drop in-place `tasks_per_mnt` normalization; add `effective_tasks_per_mnt()`; 11 tests | Invalid values now fail at startup; `congestion > max_background` no longer silently clamped |
| `curvine-fuse/src/session/channel/mod.rs` | `FuseChannel::new` resolves task count via `effective_tasks_per_mnt()` | `--io-threads` override now tracked by per-mount task count |

# Test verified

| `cargo test -p curvine-common --lib -- fuse_conf` | PASS | 30 passed (11 new: reject/accept for each field + boundary + effective_tasks) |
| `init_rejects_zero_{path_lock_stripes,list_limit,mnt_number,max_background,congestion_threshold}` | PASS | |
| `init_rejects_congestion_threshold_above_max_background` / `init_accepts_congestion_threshold_equal_max_background` | PASS | boundary |
| `init_rejects_umask_out_of_range` / `init_accepts_max_umask` | PASS | `0o10000` rejected, `0o7777` accepted |
| `effective_tasks_per_mnt_follows_io_threads_when_zero` | PASS | tracks io_threads across re-init (the double-init scenario) |
| `effective_tasks_per_mnt_preserves_explicit_value` | PASS | explicit value kept; raw not mutated |
| `cargo build -p curvine-fuse` / clippy / fmt | PASS | no warnings |

# Dependencies

- Related PRs: #1177 (#1125, sibling FuseConf init cross-field validation)
- Related issues: #1073 (io_threads validation, prior)
- Blocks / blocked by: None

### DIFF
--- a/curvine-common/src/conf/fuse_conf.rs
+++ b/curvine-common/src/conf/fuse_conf.rs
@@ -89,6 +89,10 @@ pub struct FuseConf {
     // `mnt_per_task` alias kept for backward compatibility with pre-rename TOML
     // configs (issue #1023 §2); without it `#[serde(default)]` would silently
     // drop the old key and fall back to the default.
+    //
+    // Raw user input; `0` means "follow io_threads". Consumers must NOT read
+    // this directly — use `FuseConf::effective_tasks_per_mnt()` so the `0`
+    // fallback is applied against the (possibly CLI-overridden) io_threads.
     #[serde(alias = "mnt_per_task")]
     pub tasks_per_mnt: usize,
 
@@ -246,6 +250,39 @@ impl FuseConf {
         if self.io_threads == 0 {
             return err_box!("fuse.io_threads must be > 0");
         }
+        if self.mnt_number == 0 {
+            return err_box!("fuse.mnt_number must be > 0");
+        }
+        if self.list_limit == 0 {
+            return err_box!("fuse.list_limit must be > 0");
+        }
+        // path_lock_stripes is used as the modulus in NodeState::lock_path
+        // (hash % path_locks.len()); 0 stripes => empty vec => divide-by-zero
+        // panic on create/release.
+        if self.path_lock_stripes == 0 {
+            return err_box!("fuse.path_lock_stripes must be > 0");
+        }
+        // Upper-bound sanity check on the umask (input hygiene, e.g. an octal
+        // value mistakenly written as decimal); the low bits are the meaningful
+        // permission mask.
+        if self.umask > 0o7777 {
+            return err_box!("fuse.umask must be <= 0o7777 (octal file-permission bits)");
+        }
+        // max_background / congestion_threshold are returned verbatim in the
+        // FUSE init reply; a zero window or an inverted threshold is nonsensical
+        // to the kernel. Check max_background first: the congestion message
+        // references it.
+        if self.max_background == 0 {
+            return err_box!("fuse.max_background must be > 0");
+        }
+        if self.congestion_threshold == 0 || self.congestion_threshold > self.max_background {
+            return err_box!(
+                "fuse.congestion_threshold must be > 0 and <= fuse.max_background \
+                 (congestion_threshold = {}, max_background = {})",
+                self.congestion_threshold,
+                self.max_background
+            );
+        }
 
         self.attr_ttl = Duration::from_millis(self.attr_timeout_ms);
         self.entry_ttl = Duration::from_millis(self.entry_timeout_ms);
@@ -253,9 +290,12 @@ impl FuseConf {
         self.node_cache_ttl = DurationUnit::from_str(&self.node_cache_timeout)?.as_duration();
         self.meta_cache_ttl = DurationUnit::from_str(&self.meta_cache_timeout)?.as_duration();
 
-        if self.tasks_per_mnt == 0 {
-            self.tasks_per_mnt = self.io_threads;
-        }
+        // NOTE: `tasks_per_mnt == 0` means "follow io_threads". This is resolved
+        // at the consumption point (FuseChannel::new via `effective_tasks_per_mnt`),
+        // NOT normalized in place here: init() runs twice (once in
+        // ClusterConf::from, once after CLI overrides), and mutating the field on
+        // the first pass would freeze it so a later `--io-threads` override is not
+        // tracked. Keeping the raw value lets the consumer re-resolve every time.
 
         let fs_path = Path::from_str(&self.fs_path)?;
         self.fs_path = fs_path.path().to_owned();
@@ -278,6 +318,18 @@ impl FuseConf {
         }
 
         Ok(())
+    }
+
+    /// Effective per-mount IO task count: `tasks_per_mnt`, or `io_threads` when
+    /// `tasks_per_mnt == 0` ("follow io_threads"). Resolved on read rather than
+    /// stored, so a CLI `--io-threads` override applied after the config is
+    /// loaded is always tracked.
+    pub fn effective_tasks_per_mnt(&self) -> usize {
+        if self.tasks_per_mnt == 0 {
+            self.io_threads
+        } else {
+            self.tasks_per_mnt
+        }
     }
 
     /// Validates that `state_dir` is — or can be made — a writable directory.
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/session/channel/mod.rs
```diff
@@ -38,10 +38,14 @@ impl<T: FileSystem> FuseChannel<T> {
         let buf_size =
             FuseUtils::get_fuse_buf_size().max(max_readahead as usize + FUSE_BUFFER_HEADER_SIZE);
 
-        let mut receivers = Vec::with_capacity(conf.tasks_per_mnt);
-        let mut senders = Vec::with_capacity(conf.tasks_per_mnt);
+        // Resolve `tasks_per_mnt == 0` ("follow io_threads") here rather than in
+        // FuseConf::init, so a CLI `--io-threads` override applied after config
+        // load is tracked (init runs before the override).
```
szbr00: Minor comment accuracy: init() also runs after CLI overrides (mount_args::get_conf calls fuse.init() again). The bug was that the first init() mutated tasks_per_mnt from 0 to io_threads and froze it before --io-threads applied.

Suggest wording closer to the FuseConf note — resolve on read so a later io_threads override is still visible — so a future reader does not conclude that init never runs after CLI parsing.

curvine-common/src/conf/fuse_conf.rs
```diff
@@ -246,16 +250,52 @@ impl FuseConf {
...
+            return err_box!("fuse.umask must be <= 0o7777 (octal file-permission bits)");
+        }
+        // max_background / congestion_threshold are returned verbatim in the
+        // FUSE init reply; a zero window or an inverted threshold is nonsensical
+        // to the kernel. Check max_background first: the congestion message
+        // references it.
+        if self.max_background == 0 {
+            return err_box!("fuse.max_background must be > 0");
+        }
+        if self.congestion_threshold == 0 || self.congestion_threshold > self.max_background {
```
szbr00, line 278: With this cross-check, passing only --max-background below the configured/default congestion_threshold (192) now fails at startup. The PR body calls this out, but the CLI help for --max-background / --congestion-threshold still presents them as independent.

Please mention the constraint in those flag help texts (e.g. congestion_threshold must be > 0 and <= max_background), so operators are not surprised by a fail-fast that previously looked like a silent clamp.

curvine-fuse/src/session/channel/mod.rs
yangcx000: Thanks, fixed in e8bf066. Reworded the comment to match the FuseConf note: init() runs twice (once in ClusterConf::from, once after CLI overrides in mount_args), so the field is resolved on read here rather than normalized in init — a later --io-threads override stays visible. Dropped the misleading "init runs before the override".

curvine-common/src/conf/fuse_conf.rs
yangcx000: Thanks, fixed in e8bf066. Updated the flag help text: --max-background now says "must be > 0" and --congestion-threshold says "must be > 0 and <= max_background", so operators see the constraint instead of being surprised by the fail-fast that previously looked like a silent clamp.

szbr00 (approved):

Issue #1142, opened 2026-07-15
[FEATURE] fuse: clean up lib.rs constants — naming, unused, type consistency (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

Several constants in `curvine-fuse/src/lib.rs` have misleading names, wrong/inconsistent types, or are dead, which increases misuse and maintenance risk. None is a functional bug today; this is a naming/cleanup pass. (The default-mode split, kernel-version tuple, and init capability allowlist are tracked separately in #1138 / #1137 / #1088.)

**(a) `FUSE_MAX_PAGES` is a capability bit, not a page limit — easily confused with `FUSE_MAX_MAX_PAGES`**
- `lib.rs:69`/`77`: `FUSE_MAX_MAX_PAGES: usize = 256` (a page-count limit) vs `FUSE_MAX_PAGES: u32 = 1 << 22` (the FUSE init capability bit).
- `curvine_file_system.rs:398`-`399`: `if op.arg.flags & FUSE_MAX_PAGES != 0 { out_flags |= FUSE_MAX_PAGES }` reads like a page-count comparison but is a capability-bit test.
- Fix: rename the capability bit to `FUSE_CAP_MAX_PAGES` (or `FUSE_INIT_FLAG_MAX_PAGES`); keep/rename the page limit as `FUSE_MAX_PAGES_LIMIT`. Consider a consistent `FUSE_CAP_*` prefix for all init capability flags.

**(b) `FUSE_UNKNOWN_INODES` and `FUSE_UNKNOWN_INO` share a value but mean different things**
- `lib.rs:111`/`128`: both are `0xffffffff`. `FUSE_UNKNOWN_INODES` is used for statfs `files`/`ffree` (`curvine_file_system.rs:687`-`688`); `FUSE_UNKNOWN_INO` is used as a synthetic inode id (dot entries / status_to_attr).
- Fix: add comments distinguishing them, or rename to `FUSE_STATFS_UNKNOWN_FILES` / `FUSE_UNKNOWN_INO`.

**(c) `FUSE_FOPEN_PARALLEL_DIRECT_WRITES` is `i32`, all other FOPEN flags are `u32`**
- `lib.rs`: the other `FUSE_FOPEN_*` are `u32` and `fuse_open_out.open_flags` is `u32`; only this one is `i32`. Currently it has no use site anywhere in the repo, but any future OR would need a cast.
- Fix: change to `u32` (note the kernel-header type in a comment if desired).

[...]

Issue #1122, opened 2026-07-14
[FEATURE] fuse: shared writer ignores subsequent open flags/opts (O_TRUNC, etc.) (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

A shared writer is keyed by inode; when a writer already exists for an ino, `get_or_create_writer` returns the existing one and ignores the new open's flags/opts.

- `curvine-fuse/src/fs/state/node_state.rs:339`-`353` (`get_or_create_writer` with `Some(ino)`: `get_or_create` returns the existing resource, only bumping refs)
- `orpc/src/sync/async_map.rs:119`-`148` (`get_or_create` only creates if absent; otherwise returns existing)

**Problem**

- For an already-existing ino, the second `open` with different semantics (`O_TRUNC`, append, different acl/create opts, direct/cache differences) is ignored — the writer keeps the first fd's flags/opts.
- Multi-fd concurrent writes to the same file can therefore violate POSIX/FUSE open-flag expectations, especially truncate/append/random-write semantics that depend on writer init parameters.

**Describe the solution you'd like**

- Define which flags may share a writer and which require a distinct writer or an explicit state change first.
- `O_TRUNC` must not rely on the shared writer's open opts — perform the truncate/resize explicitly on the open path and refresh state.
- Consider extending the writer key to `(ino, open_generation)`, or sharing only the underlying upload session while keeping per-fd semantic state.

**Additional context**

From a curvine-fuse FuseReader/FuseWriter read/write-path review. Needs a decision on the intended multi-fd shared-writer semantics.

Acceptance / test gaps:
- Opening an existing file with `O_TRUNC` truncates it regardless of an already-shared writer.
- Per-fd open-flag semantics are honored (or the shared-writer semantics are explicitly documented/limited).

PR #1183, opened 2026-07-18
fix(fuse): negotiate an explicit FUSE capability allowlist in init (#1088)

### LINKED ISSUE #1088: [FEATURE] fuse: negotiate an explicit FUSE capability allowlist in init (by LiAuto)

`CurvineFileSystem::init` starts with a set of daemon flags, then does `out_flags |= op.arg.flags`, echoing back kernel-supported capabilities even when the daemon does not implement the corresponding opcode/semantics.

- `curvine-fuse/src/fs/curvine_file_system.rs:382` (initial out_flags)
- `curvine-fuse/src/fs/curvine_file_system.rs:401` (`out_flags |= op.arg.flags`)

Concrete case: `FuseRequest::parse_operator` recognizes `FUSE_RENAME2` (`curvine-fuse/src/session/fuse_request.rs:275`), but `dispatch_meta` has no `Rename2` arm and falls through the wildcard to ENOSYS (`fuse_receiver.rs:642`). `FUSE_DO_RENAME2` (1<<11) is defined; blindly OR-ing `op.arg.flags` can enable it whenever the kernel offers it.

**Impact**

- Advertising a capability tells the kernel the feature is supported, so the kernel may issue related requests (e.g. RENAME2) that the daemon then rejects with ENOSYS — avoidable protocol churn or broken behavior.
- The negotiated capability set is implicit and coupled to whatever the kernel offers, not to what the daemon actually implements.
- Beyond RENAME2, the blind OR can also enable other unverified capabilities the kernel offers — e.g. `FUSE_WRITEBACK_CACHE` (changes flush/fsync/release semantics), `FUSE_POSIX_ACL`, and any future unknown kernel bits — before the daemon has the matching semantics.

- In `init`, negotiate only an explicit allowlist of supported capabilities: `out_flags = base_flags | (op.arg.flags & SUPPORTED)`; do not blindly OR all kernel flags into `out_flags`. Each advertised bit must map to a real implementation or an explicit downgrade. `WRITEBACK_CACHE`-class flags that change flush/fsync/release semantics must stay off until the cache-coherency strategy is complete.
- Either implement `Rename2` or ensure `FUSE_DO_RENAME2` is not advertised.
- Add a test asserting negotiated flags do not include unsupported capabilities.

**ABI version check (from CurvineFileSystem review)**

`curvine_file_system.rs:373` rejects the ABI with `&&`:
```rust
if op.arg.major < FUSE_KERNEL_VERSION && op.arg.minor < FUSE_KERNEL_MINOR_VERSION
```
This lets through a low-`major`/high-`minor` combination (e.g. major<7 but minor>=31). FUSE versions must be compared as a `(major, minor)` tuple: reject `major < required_major`; reject `major == required_major && minor < required_minor`; for `major > required_major` follow the Linux negotiation rules rather than silently using the current struct layout. Accepting an incompatible ABI risks mismatched struct sizes / field semantics downstream.

**SPLICE_* capability vs `enable_splice` config (from CurvineFileSystem review)**

`init` unconditionally advertises `FUSE_SPLICE_MOVE/WRITE/READ`, but whether the receiver/sender actually create splice pipes is controlled by `conf.enable_splice` (`session/channel/mod.rs`). When `enable_splice=false`, advertising splice capability makes the negotiated protocol inconsistent with the configuration. Only advertise `FUSE_SPLICE_*` when `conf.enable_splice=true` (fold this into the allowlist: the supported set is config-dependent).

**Opcode-coverage angle (from FileSystem-trait review)**

The `FileSystem` trait / dispatch have no single source of truth for which opcodes are actually handled, so parsed-but-unhandled opcodes drift silently:
- `FUSE_RENAME2` is the only opcode that is parsed into an operator (`FuseOperator::Rename2`) yet has no trait method and no dispatch arm.
- `FUSE_DESTROY` is half-wired: `struct Destroy<'a>` (`operator.rs:173`) and `FUSE_DESTROY = 38` (`fuse_op_code.rs:59`) are defined, but there is NO `parse_operator` arm, NO `FileSystem` method, and NO dispatch arm — so a kernel `FUSE_DESTROY` becomes `FuseOperator::Notimplemented` → wildcard ENOSYS. The half-present `Destroy` struct is misleading dead code (a maintainer may assume DESTROY is handled).
- `FSYNCDIR`/`LSEEK`/`BMAP`/`POLL`/`IOCTL` exist in `FuseOpCode` but are not parsed at all.

As part of the allowlist work, decide the policy for these: implement, or maintain a documented allowlist of intentionally-unsupported opcodes that init must not advertise. Specifically for `FUSE_DESTROY`: either wire a `destroy` hook (default `Ok(())`, no-reply) + parse + dispatch arm, or remove the unused `Destroy` struct and document that DESTROY is handled only via kernel disconnect + `unmount()`.

**Additional context**

From curvine-fuse receiver/dispatch, FileSystem-trait, and CurvineFileSystem implementation reviews. Real protocol-correctness issue; this issue now owns all of init's ABI/capability negotiation.

Acceptance / test gaps:
- Init negotiation does not advertise unsupported capabilities such as `FUSE_DO_RENAME2` (or WRITEBACK_CACHE/POSIX_ACL) unless implemented.
- `FUSE_SPLICE_*` is advertised only when `enable_splice=true`.
- ABI version acceptance uses a `(major, minor)` tuple comparison and rejects incompatible combinations (incl. low-major/high-minor).
- Parsed-but-unhandled opcodes (Rename2) and half-wired opcodes (Destroy) follow the decided implement-or-documented-unsupported policy.

yangcx000:
# Summary

`CurvineFileSystem::init` blindly OR-ed every kernel-offered flag into the init reply (`(base | op.arg.flags) & !FUSE_ATOMIC_O_TRUNC`), advertising capabilities the daemon does not implement; the ABI version check used a buggy `&&`; and `FUSE_SPLICE_*` was advertised unconditionally, decoupled from `enable_splice`. This replaces the blind OR with an explicit capability allowlist, fixes the ABI comparison, gates splice on config, and logs the negotiated capability diff.

# Issue Describe / Design

Closes #1088

## Capability allowlist

`init` now advertises `SUPPORTED_INIT_FLAGS & op.arg.flags` — only caps the daemon implements **and** the kernel offered. `SUPPORTED_INIT_FLAGS` is an explicit set in `lib.rs`, each bit annotated to its implementation. This inherently drops `FUSE_ATOMIC_O_TRUNC` (#1122), `FUSE_POSIX_ACL`, `FUSE_HAS_IOCTL_DIR`, and any unknown/future kernel bit — no more `& !ATOMIC_O_TRUNC` special-case.

**Making previously-implicit caps explicit (important — read before assuming a regression):** `FUSE_POSIX_LOCKS`, `FUSE_FLOCK_LOCKS`, and `FUSE_DO_READDIRPLUS` were only ever enabled via the blind OR, even though `get_lk`/`set_lk`/`set_lkw` and `read_dir_plus` are real implementations routing to the distributed backend. They are now in `SUPPORTED_INIT_FLAGS`, so distributed locking and readdirplus behave exactly as before on ≥7.31 kernels. The visibly-larger flag set is just the old implicit dependency made explicit — **not** newly-advertised behavior. (Base caps that were previously force-advertised are now kernel-gated, i.e. advertised only when the kernel also offers them; on a real ≥7.31 kernel these are always offered, so no wire-level change.)

## Config-gated daemon-requested caps

Forced on, **not** masked by the kernel offer:
- `FUSE_WRITEBACK_CACHE` when `write_back_cache` (unchanged behavior).
- `FUSE_SPLICE_*` when `enable_splice`. Splice is driven by the channel talking `splice(2)` on the fuse fd directly (independent of an init-flag offer), so it must be advertised on config alone — and is now consistent with `enable_splice` instead of always-on. Default `enable_splice=true`, so the default is unchanged; `false` now correctly stops advertising splice.

## ABI version check

Compare `(major, minor)` as a tuple against `FUSE_MIN_ABI` (7.31), fixing the `major < 7 && minor < 31` bug that let a low-major/high-minor combo (e.g. 6.40) through. The init reply now advertises the daemon's own 7.31 version instead of echoing the kernel's — curvine only implements the 7.31 struct/semantics, so on a higher-major kernel this makes the kernel re-issue INIT at our major (`init` is idempotent, dispatched per INIT). Stricter: kernels below 7.31 that previously slipped through are now rejected with EPROTO (they were already ABI-mismatched, since curvine sends 7.31 structs).

## Negotiated capability diff logging

After negotiation, `init` logs at info level the advertised set and the capabilities the kernel offered but the daemon did NOT enable (`kernel_flags & !out_flags`). This lets an operator see at startup why a capability is inactive (dropped by the allowlist) or confirm a config gate took effect (e.g. SPLICE absent when `enable_splice=false`). Flags are rendered as readable names via `fuse_init_flag_names`, with unknown bits surfaced as a hex token rather than hidden.

## Out of scope (separate issue)

Opcode coverage — the parsed-but-unhandled `Rename2` (no dispatch arm/trait method), the orphan `Destroy` struct, and unparsed `FSYNCDIR`/`LSEEK`/`BMAP`/`POLL`/`IOCTL` — is a distinct dispatch/trait concern and does not involve init-flag negotiation. Left for a follow-up.

# Changes

| `curvine-fuse/src/lib.rs` | Add `FUSE_POSIX_LOCKS`/`FUSE_FLOCK_LOCKS`, `SUPPORTED_INIT_FLAGS` (annotated allowlist), `FUSE_MIN_ABI`, `fuse_init_flag_names` | None (new constants/helper) |
| `curvine-fuse/src/fs/curvine_file_system.rs` | Rewrite `negotiate_out_flags` as a pure allowlist fn; add `abi_supported`; `init` uses tuple ABI check + advertises own version + config-gated negotiate + info-level capability diff log; rewrite/extend tests | Wire flags unchanged on ≥7.31 kernels; splice now config-consistent; sub-7.31 kernels rejected; new startup log line |

# Test verified

| `cargo test -p curvine-fuse --lib` | PASS | 191+ passed; `dir_tree::test` pre-existing failure skipped |
| `negotiate_out_flags_*` (6 tests) | PASS | drops unsupported; passes through supported (POSIX_LOCKS/FLOCK_LOCKS/DO_READDIRPLUS regression guards); no phantom; WRITEBACK/SPLICE config gates; containment |
| `abi_supported_tuple_comparison` | PASS | accepts ≥7.31 incl. higher major; rejects 7.30 / 6.40 / 6.0 |
| `fuse_init_flag_names_maps_known_and_unknown` | PASS | named bits + unknown-bit hex token |
| `cargo clippy -p curvine-fuse --lib` / `fmt --check` | PASS | no warnings |

# Dependencies

- Related PRs: None
- Related issues: #1122 (ATOMIC_O_TRUNC, prior), #1142 (FUSE_HAS_IOCTL_DIR rename)
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -469,6 +469,19 @@ impl fs::FileSystem for CurvineFileSystem {
             unused: 0,
         };
 
+        // Log the negotiated capability set: what we advertise vs. what the
+        // kernel offered but we did not enable (so an operator can see why a
+        // capability is inactive). Note SPLICE/WRITEBACK may appear in "enabled"
+        // without being kernel-offered — they are config-gated daemon requests.
+        let dropped = op.arg.flags & !out_flags;
+        info!(
+            "FUSE init negotiated: abi={}.{} enabled=[{}] kernel_offered_not_enabled=[{}]",
+            op.arg.major,
+            op.arg.minor,
+            fuse_init_flag_names(out_flags).join(", "),
+            fuse_init_flag_names(dropped).join(", "),
+        );
+
         Ok(out)
     }
 
@@ -1301,9 +1314,10 @@ mod tests {
 
     use super::CurvineFileSystem;
     use crate::{
-        FUSE_ATOMIC_O_TRUNC, FUSE_DO_READDIRPLUS, FUSE_FLOCK_LOCKS, FUSE_HAS_IOCTL_DIR,
-        FUSE_MAX_PAGES, FUSE_POSIX_ACL, FUSE_POSIX_LOCKS, FUSE_SPLICE_MOVE, FUSE_SPLICE_READ,
-        FUSE_SPLICE_WRITE, FUSE_WRITEBACK_CACHE, SUPPORTED_INIT_FLAGS,
+        fuse_init_flag_names, FUSE_ATOMIC_O_TRUNC, FUSE_BIG_WRITES, FUSE_DO_READDIRPLUS,
+        FUSE_FLOCK_LOCKS, FUSE_HAS_IOCTL_DIR, FUSE_MAX_PAGES, FUSE_POSIX_ACL, FUSE_POSIX_LOCKS,
+        FUSE_SPLICE_MOVE, FUSE_SPLICE_READ, FUSE_SPLICE_WRITE, FUSE_WRITEBACK_CACHE,
+        SUPPORTED_INIT_FLAGS,
     };
 
     // #1122 + allowlist: the daemon must never advertise FUSE_ATOMIC_O_TRUNC
@@ -1398,5 +1412,20 @@ mod tests {
     fn atomic_o_trunc_bit_value_matches_uapi() {
         // uapi fuse.h: FUSE_ATOMIC_O_TRUNC = (1 << 3)
         assert_eq!(FUSE_ATOMIC_O_TRUNC, 1 << 3);
+    }
+
+    #[test]
+    fn fuse_init_flag_names_maps_known_and_unknown() {
+        // Known bits render as names.
+        let names = fuse_init_flag_names(FUSE_POSIX_LOCKS | FUSE_DO_READDIRPLUS);
+        assert!(names.contains(&"POSIX_LOCKS".to_string()));
+        assert!(names.contains(&"DO_READDIRPLUS".to_string()));
+        // Empty flags => empty list.
+        assert!(fuse_init_flag_names(0).is_empty());
+        // An unknown bit is surfaced as a hex token, not silently dropped.
+        let unknown = 1u32 << 30;
+        let names = fuse_init_flag_names(FUSE_BIG_WRITES | unknown);
+        assert!(names.contains(&"BIG_WRITES".to_string()));
+        assert!(names.iter().any(|n| n == "0x40000000"));
     }
 }
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -397,15 +397,36 @@ impl CurvineFileSystem {
         out
     }
 
-    /// Whether the daemon accepts the kernel's advertised FUSE ABI version.
+    /// Whether the kernel's advertised FUSE ABI version is at least the
+    /// daemon's minimum.
     ///
     /// Compared as a `(major, minor)` tuple against `FUSE_MIN_ABI`: reject any
     /// version below 7.31 (this correctly rejects a low-major/high-minor combo
     /// such as 6.40, which the old `major < 7 && minor < 31` check let through).
-    /// A higher major is accepted; the kernel re-issues INIT at the daemon's
-    /// major after we reply with our own version.
+    /// A higher major also compares as supported here; `init` handles that case
+    /// separately with a version-only reply (see `version_only_init_out`) before
+    /// this predicate gates the too-old case.
     fn abi_supported(major: u32, minor: u32) -> bool {
         (major, minor) >= FUSE_MIN_ABI
+    }
+
+    /// The INIT reply for a kernel whose major ABI is newer than the daemon's:
+    /// advertise only the daemon's own `(major, minor)` with no negotiated
+    /// flags, and leave every other field zeroed.
+    ///
+    /// Per the fuse.h version-negotiation convention, when the kernel major is
+    /// larger than the daemon's, userspace replies with its own major and skips
+    /// negotiation, and the kernel is expected to re-issue INIT at a matching
+    /// major. This mirrors libfuse `_do_init`'s `arg->major > 7` early return:
+    /// a `fuse_init_in` at an unknown higher major may not share our field
+    /// layout, so we must not interpret its flags. The kernel does not read the
+    /// other reply fields on a major mismatch, so they stay 0.
+    fn version_only_init_out() -> fuse_init_out {
+        fuse_init_out {
+            major: FUSE_KERNEL_VERSION,
+            minor: FUSE_KERNEL_MINOR_VERSION,
+            ..Default::default()
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -449,6 +469,19 @@ impl fs::FileSystem for CurvineFileSystem {
...
         };
 
+        // Log the negotiated capability set: what we advertise vs. what the
+        // kernel offered but we did not enable (so an operator can see why a
+        // capability is inactive). Note SPLICE/WRITEBACK may appear in "enabled"
+        // without being kernel-offered — they are config-gated daemon requests.
+        let dropped = op.arg.flags & !out_flags;
+        info!(
+            "FUSE init negotiated: abi={}.{} enabled=[{}] kernel_offered_not_enabled=[{}]",
+            op.arg.major,
```
szbr00, line 526: This log labels abi= as "negotiated", but it prints op.arg.major/minor (what the kernel offered). The reply advertises FUSE_KERNEL_VERSION / FUSE_KERNEL_MINOR_VERSION (7.31), so on a 7.40 kernel the line will read abi=7.40 while the connection actually runs at 7.31.

Please log the reply version (or both offered and advertised), e.g. negotiated_abi=7.31 kernel_offered_abi=7.40, so the startup line matches the wire reply.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -371,58 +371,74 @@ impl CurvineFileSystem {
...
+
+    /// Whether the daemon accepts the kernel's advertised FUSE ABI version.
+    ///
+    /// Compared as a `(major, minor)` tuple against `FUSE_MIN_ABI`: reject any
+    /// version below 7.31 (this correctly rejects a low-major/high-minor combo
+    /// such as 6.40, which the old `major < 7 && minor < 31` check let through).
+    /// A higher major is accepted; the kernel re-issues INIT at the daemon's
+    /// major after we reply with our own version.
+    fn abi_supported(major: u32, minor: u32) -> bool {
+        (major, minor) >= FUSE_MIN_ABI
```
szbr00, line 410: abi_supported accepts a higher major (e.g. 8.0), and init then fully parses flags and returns a complete fuse_init_out with major=7. Per fuse.h version negotiation, when the kernel major is larger than the daemon major, userspace should reply with only the supported major and ignore the rest of that INIT, then wait for a re-issued INIT at the matching major.

Consider branching on op.arg.major > FUSE_KERNEL_VERSION for that short-reply path (and only running allowlist negotiation when majors already match). Harmless on todays major-7 kernels, but the current accept-and-fully-negotiate path does not match the documented protocol for a future major bump.

curvine-fuse/src/fs/curvine_file_system.rs
yangcx000: Thanks, fixed in f4e647e. The init log now reports both `negotiated_abi=7.31` (our own version, what runs on the wire) and `kernel_offered_abi=<kernel major.minor>`, so a 7.40 kernel no longer reads as "abi=7.40" when the connection actually runs at 7.31.

curvine-fuse/src/fs/curvine_file_system.rs
yangcx000: Thanks, fixed in f4e647e. `init` now branches on `op.arg.major > FUSE_KERNEL_VERSION` and returns a version-only reply (new `version_only_init_out`: our major/minor, no negotiated flags, rest zeroed), skipping allowlist negotiation — mirroring libfuse `_do_init`'s `arg->major > 7` early return, so a `fuse_init_in` at an unknown higher major is never interpreted. One clarification I confirmed against the kernel while implementing: today's `process_init_reply` errors a major mismatch rather than re-issuing INIT (the re-issue is an fuse.h *convention* libfuse implements the userspace half of), so I reworded the comments to not overstate the kernel behavior. The branch is defensive — FUSE_KERNEL_VERSION is 7 on all current kernels.

szbr00 (approved):

PR #1189, opened 2026-07-19
fix(fuse): wire RENAME2/FSYNCDIR/DESTROY and add an opcode-status matrix (#1088)

yangcx000:
# Summary

Wire the FUSE opcodes that reached the dispatch wildcard and returned ENOSYS despite being parseable or benign (RENAME2, FSYNCDIR, DESTROY), document the ones intentionally unsupported, remove dead code, and add a compile-time opcode-status matrix as the single source of truth. This is the opcode-coverage half of #1088 (the init capability-allowlist half is #1183).

# Issue Describe / Design

Relates to #1088 (opcode-coverage portion; capability allowlist is in #1183)

- **RENAME2** — was parsed into `FuseOperator::Rename2` but had no dispatch arm or trait method, so every `renameat2(2)` hit the wildcard → ENOSYS. Adds a `rename2` trait method and dispatch arm. A shared `rename_paths` helper backs both `rename` and `rename2` (no duplicated logic). Flag-less renames behave exactly as `rename` did; any RENAME2 flag (`NO_REPLACE`/`EXCHANGE`/`WHITEOUT`) is rejected with ENOSYS via a static `rename2_flags_supported`, checked against the **raw** flags value (not `RenameFlags::from_bits_truncate`, which would silently drop unknown high bits). The FUSE-facing client only issues flag-less renames today, so this conservative rejection is correct and strictly better than the previous "all RENAME2 → ENOSYS".
- **FSYNCDIR** — was unparsed → ENOSYS. Wires parse + a **distinct** `FSyncDir` operator (not reusing `FSync`, which is stream-routed to `fsync`) + a `fsync_dir` trait method defaulting to a no-op `Ok(())`: directory metadata is synchronized to the master per mutation, so there is no client-side dir write buffer to flush. Removes a spurious ENOSYS for apps that fsync directory fds.
- **DESTROY** — had an orphan `Destroy` struct but no variant/parse/dispatch. Wires it to reply with an empty success frame via `send_rep` (DESTROY is reply-expecting, **not** a `send_none` no-reply op — using `send_none` would hang the kernel). The `destroy` trait default is a no-op that deliberately does **not** call `unmount()`; real unmount cleanup is driven out-of-band by `FuseSession`, so calling it here would double-unmount.
- **BMAP/POLL/IOCTL/LSEEK** — documented as intentionally unsupported (still ENOSYS via the wildcard); no dispatch arms added. Rationale recorded on `expected_dispatch`.
- **Dead code** — removed the unused `FuseOperator::Empty` variant.
- **Single source of truth** — adds `FuseOpCode::expected_dispatch`, an exhaustive (no-wildcard) classification of every opcode as `Handled`/`NoReply`/`Unsupported`/`Internal`. Adding a new opcode fails to compile until it is classified, preventing the RENAME2-style half-wiring from recurring. A behavioral test drives `dispatch_meta` and asserts the `unsupported_reason` tag (`None` for a wired arm, `"unimplemented_opcode"` for the intentional wildcard), tying the classification to actual dispatch.

# Changes

| Module / File | Change |
| ------------- | ------ |
| `curvine-fuse/src/fs/operator.rs` | Add `FSyncDir` struct + `FSyncDir`/`Destroy` variants; remove dead `Empty` |
| `curvine-fuse/src/session/fuse_request.rs` | Parse arms for `FUSE_FSYNCDIR`/`FUSE_DESTROY`; wildcard comment → `expected_dispatch` |
| `curvine-fuse/src/fs/file_system.rs` | Trait methods `rename2` (ENOSYS default), `fsync_dir`/`destroy` (no-op Ok default) |
| `curvine-fuse/src/fs/curvine_file_system.rs` | `rename_paths` helper; `rename` delegates; `rename2` + `rename2_flags_supported` |
| `curvine-fuse/src/session/channel/fuse_receiver.rs` | `Rename2`/`FSyncDir`/`Destroy` dispatch arms; behavioral coverage tests |
| `curvine-fuse/src/session/fuse_op_code.rs` | `DispatchStatus` + `expected_dispatch` matrix + completeness test + rationale |

# Test verified

| `cargo test -p curvine-fuse --lib` | PASS | 193 passed; `dir_tree::test` pre-existing failure skipped |
| `rename2_flags_supported_only_accepts_zero` | PASS | 0 accepted; NO_REPLACE/EXCHANGE/WHITEOUT and unknown high bit rejected |
| `rename2/fsyncdir/destroy_reaches_dispatch_arm*` | PASS | `unsupported_reason: None` (arm wired); fsyncdir/destroy → empty success |
| `bmap_falls_through_wildcard_as_unimplemented` | PASS | `unsupported_reason: Some("unimplemented_opcode")` |
| `expected_dispatch_classifies_every_opcode` | PASS | full-table classification |
| `cargo clippy -p curvine-fuse --lib --tests` / `fmt --check` | PASS | no warnings |

# Dependencies

- Related PRs: #1183 (#1088 init capability allowlist)
- Related issues: #1088
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -331,7 +331,7 @@ impl<T: FileSystem> FuseReceiver<T> {
 
         // NOTE: keep this match's stream arms in sync with `FuseRequest::is_stream()`
         // (Read/Write/Flush/Release/Fsync). `send_stream` is only entered when
-        // `is_stream()` is true, so the wildcard is unreachable today; it exists
+        // `is_stream()` is true, so the fallback is unreachable today; it exists
         // as a defensive branch in case the two ever drift.
         let res = match operator {
             FuseOperator::Read(op) => fs.read(op, rep).await,
@@ -344,11 +344,48 @@ impl<T: FileSystem> FuseReceiver<T> {
 
             FuseOperator::FSync(op) => fs.fsync(op, rep).await,
 
-            _ => {
-                // Defensive: a stream-gated request whose operator is not a known
-                // stream op. Tag it `unimplemented_opcode` so it classifies as
-                // Unsupported — the same semantics as the metadata `dispatch_meta`
-                // wildcard — rather than a plain backend Error.
+            // Exhaustive fallback — NOT a `_` wildcard: naming every non-stream
+            // variant keeps the match exhaustive, so a newly-added `FuseOperator`
+            // variant that is not wired here (or in `dispatch_meta`) fails to
+            // compile. Reaching any of these is a routing bug (`send_stream` is
+            // only entered when `is_stream()`), so tag `unimplemented_opcode` —
+            // same Unsupported classification as the `dispatch_meta` fallback —
+            // rather than a plain backend Error.
+            FuseOperator::Notimplemented
+            | FuseOperator::Init(_)
+            | FuseOperator::StatFs(_)
+            | FuseOperator::ReadDir(_)
+            | FuseOperator::Lookup(_)
+            | FuseOperator::GetAttr(_)
+            | FuseOperator::SetAttr(_)
+            | FuseOperator::GetXAttr(_)
+            | FuseOperator::SetXAttr(_)
+            | FuseOperator::RemoveXAttr(_)
+            | FuseOperator::OpenDir(_)
+            | FuseOperator::Mkdir(_)
+            | FuseOperator::FAllocate(_)
+            | FuseOperator::ReleaseDir(_)
+            | FuseOperator::Access(_)
+            | FuseOperator::ReadDirPlus(_)
+            | FuseOperator::Forget(_)
+            | FuseOperator::Open(_)
+            | FuseOperator::MkNod(_)
+            | FuseOperator::Create(_)
+            | FuseOperator::Unlink(_)
+            | FuseOperator::RmDir(_)
+            | FuseOperator::Link(_)
+            | FuseOperator::BatchForget(_)
+            | FuseOperator::Rename(_)
+            | FuseOperator::Rename2(_)
+            | FuseOperator::Interrupt(_)
+            | FuseOperator::ListXAttr(_)
+            | FuseOperator::FSyncDir(_)
+            | FuseOperator::Destroy(_)
+            | FuseOperator::Symlink(_)
+            | FuseOperator::Readlink(_)
+            | FuseOperator::GetLk(_)
+            | FuseOperator::SetLk(_)
+            | FuseOperator::SetLkW(_) => {
                 let err: FuseResult<fuse_out_header> = err_fuse!(
                     libc::ENOSYS,
                     "unsupported stream operation {:?}",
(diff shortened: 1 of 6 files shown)


curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -646,8 +652,10 @@ impl<T: FileSystem> FuseReceiver<T> {
                 //   - opcode == NOT_SUPPORTED (the num_enum default, raw opcode
                 //     this build has no enum value for) -> `unknown_opcode`
                 //     (a kernel/daemon protocol-compatibility signal).
-                //   - a known opcode with no dispatch arm (e.g. Rename2) ->
-                //     `unimplemented_opcode` (an implementation-gap signal).
+                //   - a known but intentionally-unsupported opcode (BMAP/POLL/
```
szbr00: expected_dispatch is a useful intent matrix, but it does not close the original RENAME2 failure mode: a FuseOperator variant that is parsed and classified Handled can still land here if dispatch_meta has no arm. The matrix is also #[cfg(test)], so a plain cargo check/build of the lib does not enforce it.

The compile-time guarantee that would have caught RENAME2 is an exhaustive match on FuseOperator in dispatch_meta / send_stream_dispatch (handle Notimplemented explicitly; drop the `_` wildcard). That way adding FuseOperator::Foo without an arm fails to compile. Keeping expected_dispatch for opcode-level Unsupported/Internal rationale is still valuable alongside that.

yangcx000: Thanks, good catch — fixed in 2874f1b. Replaced the `_` wildcard in both `dispatch_meta` and `send_stream_dispatch` with an exhaustive fallback that names every remaining `FuseOperator` variant (including `Notimplemented`). Adding a new variant with no arm now fails to compile in the dispatcher itself, which is the compile-time guarantee that would have caught the RENAME2 half-wiring — `expected_dispatch` (still `#[cfg(test)]`) is kept alongside for the opcode-level Unsupported/Internal rationale, and its doc now points at the exhaustive matches for the arm-coverage guarantee. Behavior is unchanged: the listed variants map to the same source-tagged ENOSYS reply as before.

szbr00 (approved):

yangcx000 (commented):
I found one blocker while checking the capability/opcode coverage scope: `SUPPORTED_INIT_FLAGS` includes `FUSE_EXPORT_SUPPORT`, so `negotiate_out_flags` can advertise it when the kernel offers it. The comment says this is backed by `.`/`..` lookup reconstruction, but `NodeState::fs_lookup` does not handle the root cases correctly: `LOOKUP(root, ".")` maps to `(parent=0, name="/")` and then `get_path_name(0, "/")` fails because inode 0 does not exist; `LOOKUP(root, "..")` also tries `get_inode_check(0, None)` and fails.

Since export support relies on these lookups for file-handle reconstruction, advertising `FUSE_EXPORT_SUPPORT` can make the kernel use a protocol path that the daemon cannot satisfy. Please either exclude `FUSE_EXPORT_SUPPORT` from the allowlist for now, or fix and test `LOOKUP(".", "..")` for the root inode before advertising it.

yangcx000 (commented):
I rechecked this after the merge, and the `FUSE_EXPORT_SUPPORT` concern still appears to apply on `main`.

`SUPPORTED_INIT_FLAGS` still includes `FUSE_EXPORT_SUPPORT`, so `negotiate_out_flags` can advertise it when the kernel offers it. However, `NodeState::fs_lookup` still does not handle the root `.` / `..` cases correctly:

- `LOOKUP(root, ".")` maps to `(ino=0, name="/")`, then `get_path_name(0, "/")` fails because inode 0 does not exist.
- `LOOKUP(root, "..")` attempts `get_inode_check(0, None)` and fails as well.

Since `FUSE_EXPORT_SUPPORT` relies on these lookup paths for file-handle reconstruction, advertising it can expose a protocol path the daemon cannot satisfy. A follow-up should either remove `FUSE_EXPORT_SUPPORT` from `SUPPORTED_INIT_FLAGS` for now, or fix/test root and non-root `LOOKUP(".", "..")` before advertising the capability.

Issue #1080, opened 2026-07-13
[FEATURE] fuse: clean shutdown path with timeout and guaranteed unmount (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

Two related shutdown problems can leave a mount point that is neither serving nor cleanly unmounted. Both matter in production because curvine-fuse is stopped via signals: systemctl (`stop`/`restart` sends SIGTERM) and Fluid containers (`curvine-docker/fluid/entrypoint.sh` uses `exec curvine-fuse`, so fuse is PID 1 and receives the container SIGTERM). (CSI unmounts first then SIGKILLs, so it mostly avoids these paths but still hard-kills in-flight writes.)

**(a) Graceful shutdown has no timeout and can hang before unmount**

On TERM/SIGUSR1, `run()` broadcasts shutdown and then `run_all_handle.await` with no timeout. Receivers exit on `shutdown_rx.changed()`, but senders exit only when their task channels close. In-flight meta/stream tasks hold a `FuseResponse` (a clone of the sender's channel tx); if a backend RPC is stuck (slow/hung worker or master), the tx is never dropped, the sender never exits, `run_all` never returns, and `fs.unmount()` never runs.

- `curvine-fuse/src/session/fuse_session.rs:226` / `:250`

Impact: each stop/restart hangs until the systemd `TimeoutStopSec` (~90s) or container grace period elapses and the process is SIGKILLed; `fs.unmount()` is skipped, leaving `/curvine-fuse` as a `Transport endpoint is not connected` dead mount that needs manual `fusermount -u`.

**(b) SIGUSR1 persist failure disables auto_unmount before any child is spawned**

The SIGUSR1 branch broadcasts shutdown, waits for `run_all`, then calls `self.persist(mnts).await?`. `persist_inner` immediately sets `mnt.auto_unmount(false)` for every mount (and the `mnt` is then dropped at the end of the loop without unmounting). If any later step fails before the child process is spawned — `fcntl_get`, `fcntl_set`, writing the fd map, `self.fs.persist`, `writer.flush`, or `CommonUtils::reload_param` — the function returns early. At that point the old receiver/sender tasks are already stopped, `auto_unmount` is disabled, and no child exists: a dead/stale mount with no clean unmount.

- `curvine-fuse/src/session/fuse_session.rs:250` / `:254` / `:391`

**Describe the solution you'd like**

Design a single clean-exit path shared by the TERM and SIGUSR1 branches:

[...]

PR #1194, opened 2026-07-20
fix(fuse): fail state persist on handle errors (#1081)

### LINKED ISSUE #1081: [FEATURE] fuse: fail state persist when any file-handle persist fails (by LiAuto)

`NodeState::persist` can write a corrupt state file while reporting success.

It writes `handles.len()` as the file-handle count, then iterates and persists each handle. If a `handle.persist` fails it logs the error and sets `any_failed = true` (for the metrics stage status) but still returns `Ok(())`.

- `curvine-fuse/src/fs/state/node_state.rs:1013` (writes `handles.len()`)
- `curvine-fuse/src/fs/state/node_state.rs:1017` (per-handle persist, failure only logged)
- `curvine-fuse/src/fs/state/node_state.rs:1048` (returns `Ok(())` despite `any_failed`)

Note `BackendHandle::persist` starts with `self.complete(None).await?`, so a handle can fail before writing any bytes.

**Impact**

- The file declares N handles but contains fewer than N serialized records.
- Restore reads the following dir-handle section as a file-handle record → parse failure or misaligned parsing (`for i in 0..handles_count`).
- The top-level `persist` records success (`record_state_total` keyed on `result.is_ok()`) and can spawn a reload child with a bad state file.

- Treat any file-handle persist failure as a top-level persist failure (fail-fast), so the reload aborts instead of proceeding with a corrupt file.
- Alternatively, serialize handles into a buffer first, count only successfully serialized handles, and write an accurate count. Fail-fast is preferred for reload-state correctness.

**Additional context**

From a curvine-fuse session/persist review. This composes with the reload/persist sequencing work (SIGUSR1 clean shutdown).

Acceptance / test gaps:
- `NodeState::persist` returns an error if any file-handle persist fails.

ligeng8848:
# Summary

Fail state persistence immediately when any open file handle cannot be persisted, preventing reload from consuming an incomplete file-handle section. Add handle identifiers to the error log and cover the failure path with a regression test.

# Issue Describe / Design

Closes #1081

Root cause: `NodeState::persist` wrote the declared file-handle count, swallowed individual `handle.persist` errors, and still returned success. This could leave a partial record while allowing reload to continue.

The fix returns the first file-handle persist error immediately. The file-handles stage remains unsuccessful, and the top-level reload path is aborted.

# Changes

| `curvine-fuse/src/fs/state/node_state.rs` | Propagate file-handle persist errors and log `ino`/`fh` | Failed persistence now aborts reload; successful persistence is unchanged |
| `curvine-fuse/src/fs/state/node_state.rs` tests | Add a deterministic backend-write failure regression test | Test-only |

# Test verified

| Targeted file-handle persist failure test | PASS | 1/1 |
| `cargo test -p curvine-fuse --lib fs::state -- --nocapture` | PASS | 9/9 |
| `cargo clippy -p curvine-fuse --all-targets -- -D warnings` | PASS | |
| `make format` | PASS | Repository pre-commit checks |

# Dependencies

- Related PRs: None
- Related issues: #1080
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -1053,18 +1053,20 @@ impl NodeState {
             info!("node_state::persist: saving file_handles");
             let handles = self.all_handles();
             writer.write_len(handles.len() as u64)?;
-            let mut any_failed = false;
             for handle in &handles {
                 if let Err(e) = handle.persist(writer).await {
-                    error!("node_state::persist: error saving file_handle {:?}", e);
-                    any_failed = true;
+                    error!(
+                        "node_state::persist: error saving file_handle ino={}, fh={}: {:?}",
+                        handle.ino(),
+                        handle.fh(),
+                        e
+                    );
+                    return Err(e);
                 }
             }
             info!("node_state::persist: {} file_handles saved", handles.len());
             if let Some(stage) = stage {
-                if !any_failed {
-                    stage.success();
-                }
+                stage.success();
             }
         }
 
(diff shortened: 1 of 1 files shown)


yangcx000 (approved):
LGTM. I reviewed the persist error propagation path and verified the targeted test passes.

PR #1195, opened 2026-07-20
refactor(worker): decouple block metadata from storage layouts

gongxun0928:
# Summary

Decouple worker block metadata from live storage-directory state and move
filesystem and SPDK placement behavior behind `BlockLayout`.

This creates a reusable boundary between dataset metadata, physical storage,
and logical data layout. Existing `VfsDataset` behavior is preserved while
other datasets can define their own metadata and layout without inheriting
the current block-file organization.

# Issue Describe / Design

Related issue: None.

## Why this refactor is needed

Previously, `BlockMeta` held an `Arc<DirState>` and exposed filesystem path
and SPDK device operations directly. This mixed several responsibilities:

- persistent block placement metadata;
- live directory and allocator state;
- current filesystem path conventions;
- SPDK device access;
- block lifecycle behavior.

As a result, another dataset would need to depend on current `VfsDataset`
internals and its active/staging path rules. The shared `Arc<DirState>` also
made ownership and lock boundaries harder to identify.

The intended separation is:

```text
BlockMeta:
  describes physical placement using value fields

VfsDir / DirState:
  owns one physical directory and medium-specific runtime state

BlockLayout:
  defines how data is organized and accessed on that medium

VfsDataset:
  owns its metadata and coordinates the current block lifecycle
```

## Reuse boundary

`VfsDir` is the reusable unit for one physical directory. It owns the
identity, path, storage medium, device information, capacity counters, and
medium-specific allocator state required to access that directory.

`DirList` remains the reusable group of `VfsDir` instances. A dataset can
resolve a `dir_id` through `DirList`, select its own `BlockLayout`, and build
its own reader and writer contexts without storing runtime directory objects
inside its metadata.

Other datasets can therefore reuse the same physical directories and medium
support while keeping their own metadata store, key model, path organization,
and lifecycle policy.

## Design decisions

- `BlockMeta` stores value-only physical placement fields.
- `VfsDir` owns directory identity, path, medium type, and runtime state.
- `FileLayout` owns filesystem path generation and local-file access.
- `BdevLayout` owns SPDK allocation, persistence, and device access.
- `VfsDataset` resolves a `BlockMeta` to its `VfsDir` and `BlockLayout`.
- Reader, writer, short-circuit, and asynchronous deallocation operations
  use owned layout and directory handles after leaving the dataset lock.
- Metadata is removed before physical deletion.
- Capacity is released only after physical deallocation succeeds.
- Restart accounting uses physical bytes, including aligned SPDK extents.
- Scheduled asynchronous deletions consume their counter on terminal errors.

No master metadata or external worker protocol is changed by this PR.

## Locking and follow-up optimization

This PR intentionally retains the global `RwLock<BlockDataset>` for block
lifecycle serialization. Changing the data model and lock model together
would make correctness and failure ordering difficult to review.

The refactor establishes the ownership boundaries needed for later lock
optimization:

1. Keep `DirList` and the layout registry as shared, mostly immutable
   physical-storage services.
2. Move directory selection state behind an internal atomic or small lock.
3. Store block metadata in a sharded map or use per-block synchronization.
4. Keep SPDK extent allocation protected by its allocator lock.
5. Keep directory capacity counters independently synchronized.
6. Resolve and clone metadata, directory, and layout handles in a short
   critical section, then perform I/O outside the dataset-wide lock.
7. Preserve metadata-first deletion and release capacity only after
   successful physical deallocation.

This PR is a prerequisite for reducing the global lock scope, but does not
claim to complete that concurrency change.

# Changes

| `worker/block/block_meta.rs` | Store placement values instead of live `DirState` | Internal representation only |
| `worker/storage/layout/*` | Own file paths, SPDK access, and lifecycle operations | Rejects invalid negative file sizes |
| `worker/storage/vfs_dataset.rs` | Resolve directories and restore physical capacity | Correct SPDK restart accounting |
| `worker/block/block_store.rs` | Reuse owned directory handles during async delete | Balances delete metrics on errors |
| `worker/storage/vfs_dir.rs` | Own directory identity and runtime state | Removes duplicate ownership |
| `worker/storage/dir_state.rs` | Retain medium-specific allocator state | Existing SPDK allocation preserved |
| `worker/storage/spdk_meta_store.rs` | Persist explicit placement values | Existing recovery behavior preserved |

# Test verified

| `make format` | PASS | Formatting and two release Clippy passes |
| `cargo test -p curvine-server worker::storage` | PASS | 36 tests passed |
| `cargo test -p curvine-server worker::block::block_store::tests` | PASS | 2 tests passed |

# Dependencies

- Base: `main`
- Follow-up: #1196
- Related issues: None
- Blocks: #1196

### DIFF
--- a/curvine-server/src/worker/block/block_store.rs
+++ b/curvine-server/src/worker/block/block_store.rs
@@ -124,13 +124,15 @@ impl BlockStore {
     pub fn async_remove_block(&self, id: i64) -> CommonResult<BlockMeta> {
         let removed = {
             let mut state = self.write()?;
-            let removed = state.remove_block_state_by_id(id)?;
+            let remove_result = state.remove_block_state_by_id(id);
+            // Heartbeat increments this counter before scheduling the task.
+            // Consume it even when metadata removal reports an error.
             state.decrement_blocks_to_delete();
-            removed
-        };
+            remove_result
+        }?;
 
         removed.layout.deallocate(&removed.dir, &removed.meta)?;
-        self.read()?.release_block_space(&removed.meta)?;
+        removed.release_space();
         Ok(removed.meta)
     }
 
@@ -166,5 +168,53 @@ impl BlockStore {
         }
 
         Ok(vec)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::worker::storage::Dataset;
+    use curvine_common::conf::WorkerConf;
+
+    fn create_store(name: &str) -> CommonResult<BlockStore> {
+        let conf = ClusterConf {
+            format_worker: true,
+            worker: WorkerConf {
+                dir_reserved: "0".to_string(),
+                data_dir: vec![format!("[MEM:1KB]../testing/block-store-{name}")],
+                ..WorkerConf::default()
+            },
+            ..ClusterConf::default()
+        };
+        BlockStore::new("test", &conf)
+    }
+
+    #[test]
+    fn async_remove_missing_block_releases_delete_count() -> CommonResult<()> {
+        let store = create_store("missing-block")?;
+        store.read()?.increment_blocks_to_delete();
+
+        assert!(store.async_remove_block(1).is_err());
+        assert_eq!(store.read()?.num_blocks_to_delete(), 0);
+        Ok(())
+    }
+
+    #[test]
+    fn async_remove_invalid_dir_releases_delete_count() -> CommonResult<()> {
+        let store = create_store("invalid-dir")?;
+        {
+            let mut state = store.write()?;
+            let mut meta = BlockMeta::new(1, 100, state.dir_iter().next().unwrap());
+            meta.dir_id = u32::MAX;
+            state.put_test_meta(meta);
+            state.increment_blocks_to_delete();
+        }
+
+        assert!(store.async_remove_block(1).is_err());
+        let state = store.read()?;
+        assert!(state.get_block(1).is_none());
+        assert_eq!(state.num_blocks_to_delete(), 0);
+        Ok(())
     }
 }
--- a/curvine-server/src/worker/block/block_store.rs
+++ b/curvine-server/src/worker/block/block_store.rs
@@ -76,27 +76,27 @@ impl BlockStore {
     }
 
     pub fn open_writer(&self, meta: &BlockMeta, off: i64) -> CommonResult<BlockWriteContext> {
-        let layout = {
+        let (layout, dir) = {
             let state = self.read()?;
-            state.layout_for(meta)
+            state.layout_for(meta)?
         };
-        layout.open_writer(meta, off).map_err(Into::into)
+        layout.open_writer(&dir, meta, off).map_err(Into::into)
     }
 
     pub fn open_reader(&self, meta: &BlockMeta, off: i64) -> CommonResult<BlockReadContext> {
-        let layout = {
+        let (layout, dir) = {
             let state = self.read()?;
-            state.layout_for(meta)
+            state.layout_for(meta)?
         };
-        layout.open_reader(meta, off).map_err(Into::into)
+        layout.open_reader(&dir, meta, off).map_err(Into::into)
     }
 
     pub fn short_circuit(&self, meta: &BlockMeta) -> CommonResult<Option<String>> {
-        let layout = {
+        let (layout, dir) = {
             let state = self.read()?;
-            state.layout_for(meta)
+            state.layout_for(meta)?
         };
-        layout.short_circuit(meta)
+        layout.short_circuit(&dir, meta)
     }
 
     pub fn worker_id(&self) -> CommonResult<u32> {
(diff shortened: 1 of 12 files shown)


curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -192,38 +202,45 @@ impl VfsDataset {
...
+        layout.release(&dir, &meta);
 
-        Ok((meta, layout.clone()))
+        Ok(RemovedBlockState { meta, layout, dir })
     }
 
     pub(crate) fn release_block_space(&self, meta: &BlockMeta) -> CommonResult<()> {
         let dir = self.find_dir(meta.dir_id())?;
-        dir.release_space(meta.is_final(), meta.actual_len);
+        dir.release_space(meta.is_final(), meta.physical_bytes());
```
szbr00, line 235: After this refactor, SPDK scan sets `actual_len = record.size` (aligned physical bytes), and open/finalize/remove all charge `physical_bytes()` (as here). `initialize()` still does `reserve_space(true, block.len)`.

On worker restart that under-charges recovered SPDK extents; a later delete then releases `physical_bytes()` and can clamp/underflow dir counters, overstating available capacity.

Please charge `block.physical_bytes()` in `initialize()` (and add a restart test where `record.size != record.len`).

curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -192,38 +202,45 @@ impl VfsDataset {
...
-    ) -> CommonResult<(BlockMeta, BlockLayoutKind)> {
-        let meta = match self.meta.remove(id) {
+    pub(crate) fn remove_block_state_by_id(&mut self, id: i64) -> CommonResult<RemovedBlockState> {
+        let meta = match self.meta.get(id) {
             None => return err_box!("Not found block {}", id),
             Some(meta) => meta,
         };
+        let layout = self.layouts.get(meta.storage_type()).clone();
+        let dir = match self.dir_list.get_dir(meta.dir_id()) {
+            None => return err_box!("No storage directory found: {:?}", meta.dir_id()),
```
szbr00, line 224: `remove_block_state_by_id` now returns early if `dir_id` is missing, so the block is never removed from the meta store.

Previously, release was skipped when the dir lookup failed, but metadata removal still proceeded, and file layouts could deallocate via the path embedded in BlockMeta DirState. With value-only BlockMeta, a stale or missing `dir_id` permanently blocks abort/delete cleanup.

Consider still removing meta and best-effort releasing/deallocating when the dir is present, or retaining enough placement identity to delete orphaned file blocks without a live DirList entry.

curvine-server/src/worker/block/block_store.rs
```diff
@@ -122,17 +122,16 @@ impl BlockStore {
...
-            (meta, layout)
+            removed
         };
 
-        layout.deallocate(&meta)?;
-        let state = self.read()?;
-        state.release_block_space(&meta)?;
-        Ok(meta)
+        removed.layout.deallocate(&removed.dir, &removed.meta)?;
+        self.read()?.release_block_space(&removed.meta)?;
```
szbr00: RemovedBlockState already carries the Arc VfsDir used for deallocate, but capacity release goes back through release_block_space then find_dir.

Prefer `removed.dir.release_space(removed.meta.is_final(), removed.meta.physical_bytes())` so the async delete path stays on the owned handles resolved under the write lock and does not depend on a second lookup.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: When splitting into two separate PRs, the changes here were missed. The code was fixed in #1196 . I will apply this fix in the current commit to ensure the correctness of this PR.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Thanks. Fixed in 671cbe40. Metadata-first removal has been restored, so an invalid dir_id no longer leaves an unusable metadata entry. Runtime directory removal is not currently supported; a missing directory is treated as invalid metadata. Because the value-only BlockMeta cannot safely locate backing storage without a live VfsDir, the path returns an error after removing metadata, and the asynchronous delete counter is still consumed. The BlockStore invalid-directory test verifies both metadata removal and counter recovery.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Thanks. Fixed in 671cbe40. Startup recovery now charges block.physical_bytes() in the correct finalized or recovering bucket. The spdk_initialize_restores_physical_capacity test covers a 1-byte logical block backed by a 4096-byte aligned extent and verifies that deletion restores the full charge.

curvine-server/src/worker/block/block_store.rs
gongxun0928: Thanks. Fixed in 671cbe40. RemovedBlockState now uses the captured Arc<VfsDir> for capacity release after successful deallocation, avoiding a second dataset lock and dir_id lookup.

szbr00 (approved):

PR #1196, opened 2026-07-20
refactor(worker): centralize physical storage capacity

gongxun0928:
# Summary

Centralize physical directory selection, compound capacity transitions,
storage health reporting, and worker storage metrics in reusable `DirList`
and `VfsDir` APIs.

`VfsDataset` and other datasets can reuse the same physical disk-group
management while keeping their own metadata, layout, and lifecycle policy.

# Issue Describe / Design

Related issue: None.

## Why this refactor is needed

Physical storage management was spread across `VfsDataset`, `BlockStore`,
`VfsDir`, worker metrics, and layout code. This made another dataset
duplicate directory selection, physical capacity accounting, health checks,
and storage reporting.

It also made these boundaries difficult to review:

- logical block length versus aligned SPDK extent size;
- physical allocation versus capacity charging;
- finalize, prepare-write, abort, and removal transitions;
- layout-owned state versus directory-owned counters;
- metadata locking versus slow physical I/O.

Directory selection and physical capacity are properties of worker storage,
not properties of a specific block-file layout.

## Reuse boundary

`VfsDir` represents one physical storage directory. It owns:

- directory identity, path, and storage medium;
- capacity, used bytes, and reserved bytes;
- health state and device identity;
- SPDK allocation size and allocator state;
- the final and temporary capacity counters for that directory.

`DirList` represents a reusable physical disk group. It owns:

- lookup by `dir_id`;
- selection by storage type and physical byte requirement;
- compound prepare-write and finalize capacity transitions;
- aggregate capacity, availability, and usage;
- health checks, heartbeat reporting, and failed-directory counts.

A new allocation keeps the selected `Arc<VfsDir>`. After layout allocation
succeeds, the dataset charges that exact directory directly. This preserves
the original infallible accounting step and avoids a second lookup that
could create a post-allocation error path.

```text
Dataset:
  owns metadata, keys, layout, and lifecycle policy

DirList:
  owns disk-group selection, compound transitions, and reporting

VfsDir:
  owns one physical directory and its capacity counters
```

## Capacity transition rules

Directory selection uses a layout-independent request:

```text
StorageRequest {
    storage_type,
    bytes,
}
```

Capacity changes preserve this ordering:

```text
new allocation:
  choose VfsDir
  allocate physical storage
  charge physical bytes on the selected VfsDir
  publish block metadata

deallocation:
  remove metadata
  release layout-owned state
  delete physical storage
  release physical bytes
```

If physical allocation fails, capacity is not charged. If physical
deallocation fails, capacity remains charged. For SPDK, accounting uses the
aligned extent size rather than the logical block length, including startup
recovery.

Directory selection itself does not charge capacity. Exact-fit allocations
are accepted, while empty groups, negative sizes, and overflowing aligned
sizes return errors.

## Locking and follow-up optimization

This PR keeps the global `RwLock<BlockDataset>`. Current compound operations
rely on that outer lock to serialize metadata and capacity transitions.

The ownership boundaries introduced here support later lock reduction:

1. Keep the directory map and layout registry immutable after startup and
   share them through `Arc`.
2. Protect only round-robin or weighted chooser state with a small lock.
3. Make compound capacity changes linearizable at the selected `VfsDir`.
4. Shard block metadata or introduce per-block lifecycle synchronization.
5. Resolve owned metadata, layout, and directory handles in a short critical
   section, then perform physical I/O outside the dataset-wide lock.
6. Preserve metadata-first removal and release capacity only after successful
   physical deallocation.

# Changes

| Module | Change | Impact |
| ------ | ------ | ------ |
| `worker/storage/dir_list.rs` | Add requests, selection, compound accounting, health, and reporting | Reusable disk-group behavior |
| `worker/storage/vfs_dir.rs` | Convert logical requests to medium-specific physical sizes | SPDK uses aligned extents |
| `worker/storage/vfs_dataset.rs` | Coordinate layouts with physical capacity transitions | Preserves lifecycle ordering |
| `worker/storage/policy.rs` | Handle exact-fit and empty-directory selection | Corrects allocation boundaries |
| `worker/block/block_store.rs` | Delegate storage reports through the dataset | Removes duplicate reporting |
| `worker/worker_metrics.rs` | Read failed and total storage counts from the dataset | Preserves metric semantics |

# Test verified

| `cargo test -p curvine-server worker::storage` | PASS | 43 tests passed |
| `cargo clippy -p curvine-server --all-targets -- -D warnings` | PASS | Worker/server targets |
| `cargo fmt --all -- --check` | PASS | No formatting differences |

The full workspace `make format` currently stops on an unrelated unused
`Writer` import in `curvine-fuse/src/fs/state/node_state.rs` from the latest
base branch. No unrelated FUSE change is included in this PR.

# Dependencies

- Base: `main`
- Related PRs: #1195 (merged)
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/storage/vfs_dataset.rs
+++ b/curvine-server/src/worker/storage/vfs_dataset.rs
@@ -153,8 +153,7 @@ impl VfsDataset {
             let blocks = layout.scan(dir)?;
             for block in blocks {
                 let physical_bytes = block.physical_bytes();
-                self.dir_list
-                    .reserve_space(dir.id(), block.is_final(), physical_bytes)?;
+                dir.reserve_space(block.is_final(), physical_bytes);
                 self.meta.put(block);
             }
         }
@@ -334,8 +333,7 @@ impl Dataset for VfsDataset {
                 let layout = self.layouts.get(dir.storage_type());
                 let meta = layout.allocate(&dir, block)?;
 
-                self.dir_list
-                    .reserve_space(dir.id(), false, meta.physical_bytes())?;
+                dir.reserve_space(false, meta.physical_bytes());
                 self.meta.put(meta.clone());
 
                 Ok(meta)
--- a/curvine-server/src/worker/storage/vfs_dataset.rs
+++ b/curvine-server/src/worker/storage/vfs_dataset.rs
@@ -14,11 +14,11 @@
 
 use crate::worker::block::{BlockMeta, BlockState};
 use crate::worker::storage::{
-    BlockLayout, BlockLayoutKind, BlockLayouts, Dataset, DirList, SpdkMetaStore, StorageVersion,
-    VfsDir, VfsMetaStore,
+    BlockLayout, BlockLayoutKind, BlockLayouts, Dataset, DirList, SpdkMetaStore, StorageRequest,
+    StorageVersion, VfsDir, VfsMetaStore,
 };
 use curvine_common::conf::{ClusterConf, WorkerDataDir};
-use curvine_common::state::{ExtendedBlock, StorageType};
+use curvine_common::state::{ExtendedBlock, StorageInfo, StorageType};
 use indexmap::map::Values;
 use log::info;
 use orpc::common::{ByteUnit, FileUtils, LocalTime, TimeSpent};
@@ -152,7 +152,9 @@ impl VfsDataset {
             let layout = self.layouts.get(dir.storage_type());
             let blocks = layout.scan(dir)?;
             for block in blocks {
-                dir.reserve_space(block.is_final(), block.physical_bytes());
+                let physical_bytes = block.physical_bytes();
+                self.dir_list
+                    .reserve_space(dir.id(), block.is_final(), physical_bytes)?;
                 self.meta.put(block);
             }
         }
@@ -231,9 +233,8 @@ impl VfsDataset {
     }
 
     pub(crate) fn release_block_space(&self, meta: &BlockMeta) -> CommonResult<()> {
-        let dir = self.find_dir(meta.dir_id())?;
-        dir.release_space(meta.is_final(), meta.physical_bytes());
-        Ok(())
+        self.dir_list
+            .release_space(meta.dir_id(), meta.is_final(), meta.physical_bytes())
     }
 
     pub(crate) fn remove_block_by_id(&mut self, id: i64) -> CommonResult<BlockMeta> {
@@ -249,6 +250,19 @@ impl VfsDataset {
             Some(dir) => dir.clone(),
         };
         Ok((self.layouts.get(meta.storage_type()).clone(), dir))
+    }
+
+    pub fn get_and_check_storages(&self) -> Vec<StorageInfo> {
+        self.dir_list
+            .get_and_check_storages(self.meta.block_count() as i64)
+    }
+
+    pub fn failed_storage_count(&self) -> usize {
+        self.dir_list.failed_count()
+    }
+
+    pub fn storage_count(&self) -> usize {
+        self.dir_list.len()
     }
 }
 
@@ -295,8 +309,12 @@ impl Dataset for VfsDataset {
                     let new_meta = layout.prepare_write(dir, &meta, block)?;
                     let new_physical_bytes = new_meta.physical_bytes();
 
-                    dir.release_space(meta.is_final(), old_physical_bytes);
-                    dir.reserve_space(false, new_physical_bytes);
+                    self.dir_list.update_write_space(
+                        meta.dir_id(),
+                        meta.is_final(),
+                        old_physical_bytes,
+                        new_physical_bytes,
+                    )?;
                     self.meta.put(new_meta.clone());
 
                     Ok(new_meta)
@@ -310,12 +328,15 @@ impl Dataset for VfsDataset {
             }
 
             None => {
-                let dir = self.dir_list.choose_dir(block)?;
+                let dir = self
+                    .dir_list
+                    .choose_dir(StorageRequest::new(block.storage_type, block.len)?)?;
                 let layout = self.layouts.get(dir.storage_type());
                 let meta = layout.allocate(&dir, block)?;
 
+                self.dir_list
+                    .reserve_space(dir.id(), false, meta.physical_bytes())?;
                 self.meta.put(meta.clone());
-                dir.reserve_space(false, meta.physical_bytes());
 
                 Ok(meta)
             }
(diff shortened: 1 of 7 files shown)


curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -192,38 +204,57 @@ impl VfsDataset {
...
+            .release_space(meta.dir_id(), meta.is_final(), meta.physical_bytes())
     }
 
     pub(crate) fn remove_block_by_id(&mut self, id: i64) -> CommonResult<BlockMeta> {
-        let (meta, layout) = self.remove_block_state_by_id(id)?;
-        layout.deallocate(&meta)?;
-        self.release_block_space(&meta)?;
-        Ok(meta)
+        let removed = self.remove_block_state_by_id(id)?;
+        removed.layout.deallocate(&removed.dir, &removed.meta)?;
```
szbr00, line 241: If deallocate fails here, meta has already been removed in remove_block_state_by_id, but release_block_space is skipped. abort_block then treats a missing meta as success (early Ok), so a later retry cannot finish cleanup or reclaim the reserved capacity.

failed_file_deallocate_keeps_capacity_reserved locks in the "keep capacity charged" policy, but with no meta left that charge is permanently stranded and the orphan path remains. Please delay meta removal until deallocate succeeds, or keep a tombstone / failed-delete record so abort/remove can retry until physical removal and capacity release both complete.

curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -284,12 +320,15 @@ impl Dataset for VfsDataset {
...
-                let dir = self.dir_list.choose_dir(block)?;
+                let dir = self
+                    .dir_list
+                    .choose_dir(StorageRequest::new(block.storage_type, block.len)?)?;
                 let layout = self.layouts.get(dir.storage_type());
-                let meta = layout.allocate(dir, block)?;
+                let meta = layout.allocate(&dir, block)?;
 
+                self.dir_list
+                    .reserve_space(dir.id(), false, meta.physical_bytes())?;
```
szbr00: allocate already creates the file or SPDK extent before reserve_space runs. reserve_space currently fails only when the dir is missing, but the API returns Result and is meant to be reused by other datasets. If it returns Err, the physical allocation is left behind with no meta entry.

Consider releasing/deallocating the just-created allocation on this error path so a failed reserve cannot strand an empty file or SPDK offset.

curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -16,29 +16,114 @@ use crate::worker::block::{BlockMeta, BlockState};
...
 
-    fn finalize(&self, meta: &BlockMeta, _committed_len: i64) -> CommonResult<BlockMeta> {
-        BlockMeta::with_final(meta)
+    fn prepare_write(
+        &self,
+        dir: &VfsDir,
+        meta: &BlockMeta,
+        block: &ExtendedBlock,
+    ) -> CommonResult<BlockMeta> {
+        let mut prepared = BlockMeta::new(meta.id(), block.len, dir);
```
szbr00, line 108: BdevLayout::prepare_write and StorageRequest::new both reject negative sizes, but this path does not. A negative block.len becomes max(actual_len, negative) == actual_len, so update_write_space can succeed and publish a Writing meta with a negative len.

Please reject block.len < 0 here for consistency with the SPDK prepare path and new-allocation validation.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Thanks. Fixed in f525f82a. New allocation now reuses the selected Arc<VfsDir> for infallible per-directory charging, and the fallible DirList::reserve_space lookup was removed. A successful file or SPDK allocation is no longer followed by a missing-directory accounting error path.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Thanks for calling this out. We are keeping metadata-first removal in this PR because deleting the backing resource first can leave live metadata pointing to missing data. The conservative capacity charge on deallocation failure is also intentional. We agree that retryability remains an existing gap, but we are not adding a tombstone or changing deletion ordering in this storage refactor; orphan retry and cleanup will be handled separately.

curvine-server/src/worker/storage/layout/file_layout.rs
gongxun0928: Thanks. Fixed by merged PR #1195 and inherited after rebasing onto main at 23ac7554. FileLayout::prepare_write now rejects negative block lengths before publishing metadata, with regression coverage in negative_file_prepare_write_keeps_metadata.

curvine-server/src/worker/storage/vfs_dataset.rs
gongxun0928: Thanks for raising this. To clarify the history, metadata-first deletion was not introduced by this PR or by commit 5101a854. The original worker implementation, dating back to c8948d1b7, already removed the block from block_map before deleting its backing data.

The old Arc<DirState> only kept the directory and path available to the in-flight deletion call. It was not a retry record. If physical deletion failed and the call returned, the block was already absent from block_map, so a later retry could not recover the directory from the block ID either.

We preserve this ordering because metadata removal is the logical deletion point. Once deletion starts, the block should no longer remain discoverable through live metadata while its backing data may already be missing or only partially removed. A physical cleanup failure therefore leaves an unreferenced orphan instead of live metadata pointing to unavailable data. Capacity is also released only after physical deallocation succeeds, so available space is not overstated.

There is an existing recovery limitation here. The worker does not periodically scan its physical directories during normal operation, and the incremental report currently marks the block Deleted before the asynchronous physical deletion has completed. A failed deletion is therefore not retried again by the same worker process.

For recognizable block files, recovery happens indirectly after a worker restart: FileLayout scans the storage directories, VfsDataset rebuilds the local BlockMeta and capacity usage, and the startup full block report lets the master identify the block as no longer authoritative and send another DeleteBlock command. The scan itself does not delete the orphan.

This means the failure does not leave stale data logically readable, but it can leave physical data and capacity charged until restart and reconciliation. Malformed or unreadable entries, or repeated deletion failures, may persist longer. A complete runtime solution would require periodic orphan reconciliation or a durable pending-delete mechanism, which is separate from this storage-capacity refactor.

szbr00 (approved):

PR #1197, opened 2026-07-20
fix(fuse): clean pending requests on task abort (#1086)

### LINKED ISSUE #1086: [FEATURE] fuse: RAII cleanup for SETLKW pending_requests to prevent map leak on abort (by LiAuto)

`FuseReceiver::dispatch_meta_interrupt` inserts `pending_requests[req.unique()]` before the `tokio::select!`, and removes it only in the two select branches (normal `dispatch_meta` completion, or interrupt notification). A code comment already notes that the RAII guards protect the metrics gauge/histogram but do NOT clean up the map.

- `curvine-fuse/src/session/channel/fuse_receiver.rs:478` (insert)
- `curvine-fuse/src/session/channel/fuse_receiver.rs:522` (remove on dispatch completion)
- `curvine-fuse/src/session/channel/fuse_receiver.rs:528` (remove on interrupt notify)

**Impact**

- If the spawned metadata task is aborted after the insert but before either select branch completes, the map entry leaks.
- A later `FUSE_INTERRUPT` with the same unique could notify a stale `Notify` with no real waiter.
- Repeated cancelled SETLKW requests grow the map over time.
- This becomes materially more likely once session shutdown starts aborting/timing out metadata tasks (see the clean-shutdown timeout work) — the two are coupled.

- Add an RAII guard that removes `req.unique()` from `pending_requests` on drop, unless already removed.
- Keep the explicit removal in the success/interrupt branches only if the guard is disarmed there (or let the guard be the single removal site).

**Additional context**

From a curvine-fuse receiver/dispatch review. Coupling: the session-layer shutdown timeout / forced-abort work (clean shutdown issue) will abort metadata tasks, which is exactly the abort path that leaks this map today — this RAII guard is a prerequisite for safely aborting SETLKW tasks.

Acceptance / test gaps:
- Aborting a spawned SETLKW dispatch after insertion leaves `pending_requests` empty.

ligeng8848:
# Summary

Prevent cancelled or aborted `FUSE_SETLKW` dispatch tasks from leaking entries in the pending-request map. The cleanup now runs automatically when the dispatch future is dropped.

# Issue Describe / Design

Closes #1086

Root cause:

`FuseReceiver::dispatch_meta_interrupt` registered each interruptible request before entering `tokio::select!`, but removed it only when dispatch completed normally or an interrupt notification was received. If the task was aborted, neither branch executed and the registration remained in the map.

Reproduction steps:

1. Start a blocking `FUSE_SETLKW` dispatch.
2. Wait until its pending-request entry is registered.
3. Abort the spawned dispatch task.
4. Observe that the old implementation leaves the entry in `pending_requests`.

Fix approach:

Add a `PendingRequestGuard` that removes its registration on drop. Normal completion and interrupt paths still remove eagerly. Cleanup uses identity comparison so an older guard cannot remove a newer registration that happens to reuse the same unique ID.

# Changes

| `curvine-fuse/src/session/channel/fuse_receiver.rs` | Add cancellation-safe pending-request guard and regression tests | Normal completion and interrupt behavior is unchanged; aborted tasks now clean up their registrations |

# Test verified

| `cargo test -p curvine-fuse --lib pending_request -- --nocapture` | PASS | 3 tests passed |
| `cargo test -p curvine-fuse --lib session::channel::fuse_receiver::tests::dispatch_meta_integration -- --nocapture` | PASS | 10 tests passed |
| `cargo test -p curvine-fuse --lib session::channel::fuse_receiver::tests -- --nocapture` | PASS | 18 tests passed |
| `cargo clippy -p curvine-fuse --all-targets -- -D warnings` | PASS | No warnings |
| `make format` | PASS | Repository formatting and release checks passed |

# Dependencies

- Related PRs: None
- Related issues: #1086
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -34,6 +34,53 @@ use orpc::{err_box, sys, try_option_ref};
 use std::sync::Arc;
 use tokio::sync::{watch, Notify};
 
+/// Removes one interruptible request registration when its dispatch future ends.
+///
+/// The normal completion/interrupt branches remove eagerly to preserve their
+/// existing timing. `Drop` covers cancellation paths where neither branch runs,
+/// such as aborting the spawned metadata task during shutdown.
+struct PendingRequestGuard {
+    pending_requests: Arc<FastDashMap<u64, Arc<Notify>>>,
+    unique: u64,
+    notify: Arc<Notify>,
+    registered: bool,
+}
+
+impl PendingRequestGuard {
+    fn register(
+        pending_requests: Arc<FastDashMap<u64, Arc<Notify>>>,
+        unique: u64,
+        notify: Arc<Notify>,
+    ) -> Self {
+        pending_requests.insert(unique, notify.clone());
+        Self {
+            pending_requests,
+            unique,
+            notify,
+            registered: true,
+        }
+    }
+
+    fn remove(&mut self) {
+        if !self.registered {
+            return;
+        }
+
+        // Only remove the registration installed by this guard. A defensive
+        // same-unique replacement must not be deleted by an older future's Drop.
+        let _ = self.pending_requests.remove_if(&self.unique, |_, current| {
+            Arc::ptr_eq(current, &self.notify)
+        });
+        self.registered = false;
+    }
+}
+
+impl Drop for PendingRequestGuard {
+    fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


yangcx000 (approved):
LGTM. This directly addresses #1086 with RAII cleanup for pending SETLKW registrations, keeps normal removal paths intact, and includes regression coverage for abort cleanup.

PR #1201, opened 2026-07-20
fix(fuse): propagate flush/complete/resize backend errors to callers (#1118, #1120)

### LINKED ISSUE #1118: [FEATURE] fuse: propagate flush/complete/resize backend errors to callers (silent data loss) (by LiAuto)

The writer background task signals completion of `Flush`/`Complete`/`Resize` over a `CallChannel<i8>` that carries no result, so the real backend error is not returned to the caller — and on the `reply = None` path it is dropped entirely, silently reporting success.

- `curvine-fuse/src/fs/fuse_writer.rs:179`-`213` (`flush`/`complete`/`resize` callers: `tx.receive().await?` only observes an `i8`)
- `curvine-fuse/src/fs/fuse_writer.rs:285`-`304` (`writer_future`):

```rust
WriteTask::Flush(tx, reply) => {
    let res = writer.flush().await;
    if let Some(reply) = reply { reply.send_rep(res).await?; } // res only goes to reply
    tx.send(1)?;                                               // tx always sends 1, regardless of res
}
// Complete is identical; Resize uses writer.resize(opts).await? (kills the worker on failure)
```
- reader has the same shape for `Complete` in `curvine-fuse/src/fs/fuse_reader.rs:111`-`118` / `165`-`174`.

**Verified — worse than "unreliable"**

On the `reply = None` path (which is exactly how `release` calls `complete(None)` and `fsync` calls `flush(None)`): `res` is neither sent on `tx` nor `?`-propagated — it is fully discarded, and `tx.send(1)` unconditionally reports success. The worker does not exit, so `err_monitor` is not set either. So a backend flush/complete failure is **lost 100% of the time and never passes through err_monitor**.

Consequences:
- `release_handle`'s `complete(None)` failure is swallowed → data may be uncommitted while release reports success.
- `fsync`'s `flush(None)` failure is swallowed → the application is told the sync succeeded while data was not persisted → **silent data-durability loss**.
- On the `reply = Some` path, `res` goes to `send_rep` (error reply to the kernel) but is still not returned to the `tx` caller.

- Change the completion channel from `CallChannel<i8>` to `CallChannel<FsResult<()>>` for `Flush`/`Complete`/`Resize` (and the reader's `Complete`).
- The worker should `tx.send(res)` regardless of whether a `reply` is present; callers do `receive().await??`.
- For the `reply = Some` path, still return the result to the caller (via `tx`) so `release`/`fsync` cannot misjudge success.
- Return the backend result to the caller via `tx` BEFORE sending the kernel reply, so reply-channel state does not gate the caller's completion (see the tx/reply ordering below).

**tx/reply ordering (folded in)**

`writer_future` currently does `reply.send_rep(res).await?` then `tx.send(1)?`. The caller waits on `tx`, so a blocked/failing reply channel delays or breaks the flush/release completion. Send the backend result on `tx` first, then handle the kernel reply as the request's response path without blocking the state-cleanup notification.

**Additional context**

From a curvine-fuse FuseReader/FuseWriter read/write-path review. Highest-severity finding: silent flush/complete failure is a data-durability loss (same class as the flush/fsync ENOSYS default risk). Related to the reader/writer worker error-model issue (reply-send failure should not kill the worker) — best fixed together.

Acceptance / test gaps:
- `flush_backend_error_is_returned_to_caller`: a mocked `writer.flush` error makes `FuseWriter::flush(None)` return `Err`.
- `complete_backend_error_is_returned_to_release`: a mocked `complete` failure is not misjudged as success by `release`.

### LINKED ISSUE #1120: [FEATURE] fuse: split write_ver into dirty/committed epochs so failed writes don't poison read consistency (by LiAuto)

`write_ver` is incremented BEFORE the write is enqueued/committed, so a cancelled/failed write still advances it and makes the read path believe a new write happened.

- `curvine-fuse/src/fs/fuse_writer.rs:168`-`177` (`write`: `self.write_ver.incr()` before `send_queued_task`)
- `curvine-fuse/src/fs/fuse_writer.rs:200`-`213` (`resize`: increments before awaiting)
- `curvine-fuse/src/fs/state/backend_handle.rs:78`-`88` (read path: if `read_ver != writer.write_ver()`, it does `writer.flush(None)` and reopens the reader)

**Problem**

- If the enqueue is cancelled/fails, or the background `writer.fuse_write()` fails, `write_ver` has already advanced.
- A subsequent read sees the version mismatch, triggers a flush + reader reopen, but the data may never have been written.
- This compounds with the flush error-propagation bug: `flush(None)` currently does not return the backend error, so the read continues with an incorrect consistency state and still returns success.

- Split the version into `dirty_epoch` and `committed_epoch`: mark dirty on enqueue/write start, advance `committed` only after the backend write succeeds.
- The read path's flush/reopen decision should be based on "successfully written but reader not yet refreshed" (committed), not on any attempted write.
- If the current increment position must stay, at least record the error on enqueue/write failure and prevent subsequent reads from returning success on stale/incorrect state.

**Additional context**

From a curvine-fuse FuseReader/FuseWriter read/write-path review. Depends on / composes with the flush error-propagation fix (so the read path's flush actually surfaces backend errors).

Acceptance / test gaps:
- `write_enqueue_failure_does_not_advance_committed_version`: after an enqueue/write failure, a read does not treat it as a successful write requiring flush+reopen.

yangyang233333:
# Summary

The writer/reader background tasks signalled completion of Flush/Complete/Resize over a `CallChannel<i8>` that carried no result, so a backend error was not returned to the caller — and on the `reply = None` path it was dropped entirely, silently reporting success. This propagates the real backend result to the caller (and the mapped errno to the kernel).

# Issue Describe / Design

Closes #1118. 
Closes #1120.

Root cause:
- `WriteTask::Flush/Complete/Resize` (and reader `Complete`) used `CallSender<i8>` and sent a bare `1`. On `reply = Some` the result only went to the kernel via `send_rep`, never back to the caller waiting on the channel; on `reply = None` (fsync's `flush(None)`, release's `complete(None)`, read-path dirty-read flush, resize) the result was fully discarded and success was reported unconditionally.

Fix approach:
- Change the completion channel to `CallChannel<FsResult<()>>`; callers use `receive().await??`.
- Add `deliver_stream_result` (new `fs/stream_result.rs`): send the real result to the caller (`tx`) FIRST, then the kernel reply. `FsError` is not `Clone`, so the errno for the kernel is computed by borrowing the error (extracted `errno_of`) while the original error is moved onto `tx`.
- Resize no longer `?`-propagates out of the worker (which would kill it); the result is delivered to the caller like Flush/Complete.

# Changes

| --- | --- | --- |
| `curvine-fuse/src/fs/stream_result.rs` | New `deliver_stream_result` helper + tests | New shared helper |
| `curvine-fuse/src/fs/fuse_writer.rs` | Flush/Complete/Resize use helper; channel type `FsResult<()>`; callers `??` | Backend errors now surface to callers |
| `curvine-fuse/src/fs/fuse_reader.rs` | Complete uses helper; channel type; caller `??` | Same |
| `curvine-fuse/src/fuse_error.rs` | Extract borrowing `errno_of(&FsError)`; `From<FsError>` delegates | None (behavior-preserving) |
| `curvine-fuse/src/fs/mod.rs` | Declare + re-export `stream_result` | None |

# Test verified

| --- | --- | --- |
| `cargo test -p curvine-fuse --lib fs::stream_result` | PASS | 4 tests: reply=None/Some x Ok/Err; kernel errno asserted |
| `cargo test -p curvine-fuse --lib fuse_writer` | PASS | 5 existing tests |

# Dependencies

- Related PRs: None
- Related issues: #1118, #1120
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/stream_result.rs
+++ b/curvine-fuse/src/fs/stream_result.rs
@@ -17,33 +17,56 @@ use crate::session::FuseResponse;
 use curvine_common::FsResult;
 use orpc::sync::channel::CallSender;
 
-/// Deliver a backend stream-op result (`Flush`/`Complete`/`Resize`) to both the
-/// waiting in-process caller (`tx`) and, when present, the kernel (`reply`).
+/// Deliver a backend stream-op result (`Flush`/`Complete`/`Resize`) to the
+/// correct single consumer, which differs by whether this op originated from a
+/// kernel request (`reply = Some`) or an internal caller (`reply = None`).
 ///
-/// Ordering (issue #1118): notify `tx` FIRST, then send the kernel reply. The
-/// caller parks on `tx`, so sending the kernel reply first would let a blocked
-/// or failing reply channel gate/break the caller's completion notification.
+/// Why the split (PR #1201 review): the kernel reply and the `tx` completion
+/// channel are BOTH ways a result travels back, but for a kernel request they
+/// are two ends of the SAME FUSE request. If we sent the real `Err` on `tx`,
+/// the caller (`flush()/complete()/resize()`) would return it up the dispatch
+/// chain, where `send_stream_dispatch` treats any `Err` as a pre-reply dispatch
+/// failure and calls `err_rep.send_rep(res)` — enqueuing a SECOND kernel reply
+/// for the same request. So the helper is the single decision point:
 ///
-/// `FsError` is not `Clone`, so the single `res` cannot be handed to both
-/// consumers by value. We borrow it to compute an errno-only kernel reply, then
-/// move the original `res` onto `tx` so the caller sees the real success/failure.
+///   * `reply = Some` (kernel request): the kernel reply is the authoritative
+///     response. Send the real (errno-mapped) result to the kernel, and hand
+///     the in-process caller a bare `Ok(())` so it does NOT re-propagate the
+///     error and trigger a duplicate reply. Ordering (issue #1118): notify `tx`
+///     FIRST, then the kernel reply, so a blocked/failing reply channel cannot
+///     gate the caller's completion notification.
+///   * `reply = None` (internal caller: read-path dirty-read flush, flush_writer,
+///     release's `complete(None)`, resize): `tx` is the ONLY channel back, so
+///     the real backend result MUST travel on it — otherwise a backend failure
+///     is silently swallowed (the bug issue #1118 fixed).
+///
+/// `FsError` is not `Clone`, but that is no longer a constraint here: on the
+/// `reply = Some` path only the kernel reply needs the error (via a borrowed
+/// `errno_of`), and the caller receives a value-free `Ok(())` signal; on the
+/// `reply = None` path the single `res` is simply moved onto `tx`.
 pub(crate) async fn deliver_stream_result(
     res: FsResult<()>,
     tx: CallSender<FsResult<()>>,
     reply: Option<FuseResponse>,
 ) -> FsResult<()> {
-    // Build the kernel reply payload by borrowing `res`, before moving it to `tx`.
-    let kernel_rep: Result<(), FuseError> = match &res {
-        Ok(()) => Ok(()),
-        Err(e) => Err(FuseError::from_errno_msg(errno_of(e), e.to_string().into())),
-    };
+    match reply {
+        // Kernel-request path: kernel reply is authoritative; caller gets Ok(())
+        // ("reply already handled") so it never re-propagates and causes a
+        // duplicate reply. tx first, then kernel reply (issue #1118 ordering).
+        Some(reply) => {
+            let kernel_rep: Result<(), FuseError> = match &res {
+                Ok(()) => Ok(()),
+                Err(e) => Err(FuseError::from_errno_msg(errno_of(e), e.to_string().into())),
+            };
+            tx.send(Ok(()))?;
+            reply.send_rep(kernel_rep).await?;
+        }
 
-    // Caller first: hand the real result to whoever awaits `flush()/complete()/resize()`.
-    tx.send(res)?;
-
-    // Then the kernel reply, if this op originated from a kernel request.
-    if let Some(reply) = reply {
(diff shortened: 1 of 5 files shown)


yangyang233333:
@yangcx000 @szbr9486

curvine-fuse/src/fs/stream_result.rs
```diff
@@ -0,0 +1,166 @@
...
+    let kernel_rep: Result<(), FuseError> = match &res {
+        Ok(()) => Ok(()),
+        Err(e) => Err(FuseError::from_errno_msg(errno_of(e), e.to_string().into())),
+    };
+
+    // Caller first: hand the real result to whoever awaits `flush()/complete()/resize()`.
+    tx.send(res)?;
+
+    // Then the kernel reply, if this op originated from a kernel request.
+    if let Some(reply) = reply {
```
yangcx000: This sends a kernel reply for `reply=Some` backend failures, but the caller then returns the same `Err` from `flush/complete/resize` via `receive().await??`. `send_stream_dispatch` treats any `Err` from `fs.flush/fsync/release` as a pre-reply dispatch failure and sends `err_rep.send_rep(res)` again, so a backend flush/complete error can enqueue two replies for the same FUSE request. Consider making the stream arm distinguish “reply already sent” from “pre-reply failure” (for example, return `Ok(())` after `deliver_stream_result` sends the reply, or introduce a tagged error/result so the outer dispatch does not emit a second reply).

yangyang233333: Good catch, fixed in 643f625. `deliver_stream_result` now handles the reply itself on the `reply=Some` path and returns `Ok(())` to the caller, so the error is no longer re-propagated into a second reply. `reply=None` still carries the real result on `tx` (#1118 preserved). Test updated: caller gets `Ok(())`, kernel gets `-EIO`.

szbr00 (approved):

PR #1202, opened 2026-07-20
fix(fuse): read full spliced requests from pipe (#1093)

### LINKED ISSUE #1093: [FEATURE] fuse: receiver splice path must read the full request from the pipe (by LiAuto)

The receiver splice path can drop a valid kernel request on a short pipe read, which is a data-integrity problem: the request is never processed and the originating application can hang waiting for a reply.

- `curvine-fuse/src/session/channel/fuse_receiver.rs:121` (`splice`)
- `curvine-fuse/src/session/channel/fuse_receiver.rs:131` (`pipe2.read_buf(&mut self.buf[..write_len])` — single read)
- `orpc/src/sys/pipe/pipe2.rs:153` (`read_buf` calls `sys::read` exactly once)

**Verified facts**

1. The pipe read end is NON-blocking: `pipe2()` creates it with `O_NONBLOCK` (`orpc/src/sys/sys_libc.rs:269`), and `PipeFd::new(buf_size, false, false)` (`fuse_receiver.rs:70`) sets both ends non-blocking.
2. `Pipe2::read_buf` calls `sys::read` exactly once (`pipe2.rs:154`), and `sys::read` is a single raw `libc::read` (`sys_libc.rs:358`) — no loop. A `read_full` helper already exists (`sys_libc.rs:365`) but the splice path does not use it.
3. POSIX/Linux: a pipe `read()` may return fewer bytes than requested even when more data is available (kernel may return a partial read for internal buffer-management reasons). Large FUSE requests (big writes up to 128KB+) spliced into the pipe span multiple pipe-buffer pages, so a partial single read is a real possibility, not just theoretical.
4. The sender side already loops to completion (`Pipe2::read_io`, `pipe2.rs:130`); the receiver path is not symmetric.

**Current behavior on short read**

`read_len != write_len` → `err_box!` → `splice()` returns Err → `receive()` returns Err → `start()`'s `_ => return Err(e)` arm → the receiver task exits. It does NOT parse half a request (slightly better than a truncation bug), but:
- a valid kernel request is dropped (the originating process may hang forever waiting for its reply),
- the receiver exits and, because `run_all` currently swallows receiver errors (see the receiver/sender supervision issue), serving capacity degrades silently.

- Make `read_buf` (or a new `read_buf_full`) loop until exactly `write_len` bytes are copied out of the pipe.
- Because the pipe is NON-blocking, handle `EAGAIN` by awaiting readable and retrying (do not busy-spin); retry on `EINTR`; treat `Ok(0)` (writer closed) as terminal.
- Mirror the sender's existing full-transfer loop for symmetry.
- On error after a partial read, drain/reset pipe state so stale bytes cannot poison later reads.

**Additional context**

From a curvine-fuse protocol-IO review. This is the request-side counterpart to the sender's already-robust `read_io` loop. Data-integrity severity (dropped valid request + silent receiver degradation).

Acceptance / test gaps:
- Receiver splice path reconstructs a complete request when pipe reads are partial.

ligeng8848:
# Summary

Ensure the FUSE splice receive path reads every byte that splice(2) already transferred into the pipe before parsing the request. This prevents valid requests from being dropped when a non-blocking pipe read completes partially.

# Issue Describe / Design

Closes #1093

The receiver previously called Pipe2::read_buf once and treated any read_len != write_len result as fatal. A legal short pipe read therefore dropped the request and terminated that receiver.

The fix adds an exact-length asynchronous pipe read:

- use the actual byte count returned by splice(2), not the configured maximum request size;
- continue after successful short reads;
- rely on async readiness for EAGAIN and retry EINTR;
- treat zero progress as terminal;
- prevent a partially consumed pipe from returning to the pool after a terminal error.

# Changes

| orpc/src/sys/pipe/pipe2.rs | Add read_buf_full and a deterministic partial-read regression test | Non-blocking pipe transfers are consumed completely; terminal partial pipes are not reused |
| curvine-fuse/src/session/channel/fuse_receiver.rs | Read exactly the byte count returned by the FUSE-to-pipe splice | Short pipe reads no longer drop otherwise valid FUSE requests |

# Test verified

| cargo test -p orpc read_buf_full_waits_for_all_partial_reads -- --nocapture | PASS | Delivers a request in two pipe writes and verifies the reader waits and reconstructs it |
| cargo clippy -p orpc -p curvine-fuse --all-targets -- --deny warnings --allow clippy::uninlined-format-args | PASS | All changed targets are clean |
| cargo test -p curvine-fuse --lib | BASELINE MATCH | 210 passed; the same 3 unrelated dcache tests fail on unmodified main |
| Actual FUSE mount with splice enabled | PASS | Three 32 MiB write, fsync, and readback rounds matched exact size and content; the daemon remained healthy and unmounted cleanly |
| make format | PASS | Formatting and release Clippy checks passed |

# Dependencies

- Related PRs: None
- Related issues: #1093
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -30,7 +30,7 @@ use orpc::runtime::{RpcRuntime, Runtime};
 use orpc::sync::channel::AsyncSender;
 use orpc::sync::FastDashMap;
 use orpc::sys::pipe::{AsyncFd, Pipe2, PipeFd};
-use orpc::{err_box, sys, try_option_ref};
+use orpc::{err_box, sys, try_option_mut};
 use std::sync::Arc;
 use tokio::sync::{watch, Notify};
 
@@ -119,7 +119,7 @@ impl<T: FileSystem> FuseReceiver<T> {
     }
 
     pub async fn splice(&mut self) -> IOResult<BytesMut> {
-        let pipe2 = try_option_ref!(self.pipe2);
+        let pipe2 = try_option_mut!(self.pipe2);
 
         let write_len = pipe2.write_io(&self.kernel_fd, None, self.fuse_len).await?;
 
@@ -128,19 +128,15 @@ impl<T: FileSystem> FuseReceiver<T> {
             self.buf.set_len(write_len);
         }
 
-        let read_len = pipe2.read_buf(&mut self.buf[..write_len]).await?;
-        if write_len != read_len {
-            return err_box!(
-                "splice read and write lengths are inconsistent, write len {}, read len {}",
-                write_len,
-                read_len
-            );
-        }
-        if read_len < FUSE_IN_HEADER_LEN {
+        // `write_len` is the number of bytes splice actually moved from
+        // /dev/fuse, not the configured maximum request size. A pipe read may
+        // return fewer bytes, so consume exactly that transfer before parsing.
+        pipe2.read_buf_full(&mut self.buf[..write_len]).await?;
+        if write_len < FUSE_IN_HEADER_LEN {
             return err_box!("short read on fuse device");
         };
 
-        Ok(self.buf.split_to(read_len))
+        Ok(self.buf.split_to(write_len))
     }
 
     /// Build a reply handle for `unique`. When `labels` is `Some`, a metrics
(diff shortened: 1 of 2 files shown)


bigbigxu:
Fuse is request-based rather than streaming, so this operation is redundant.

ligeng8848:
Thanks, you are right. I rechecked the Linux FUSE splice path: fuse_dev_splice_read obtains one complete request, installs all non-packetized pipe buffers before returning, and pipe_read drains the available buffers up to the requested length. Our regression test modeled a streaming producer that supplies the second half later, which does not represent the real FUSE receive path. I am closing this PR as redundant.

PR #1203, opened 2026-07-20
feat(master): add weighted worker allocation policy

bigbigxu:
# Summary

Add a configurable weighted worker selection policy for heterogeneous clusters. More capable workers can receive proportionally more block allocations while zero-weight workers remain registered without receiving new data.

# Issue Describe / Design

No related issue.

The policy uses integer weighted roulette without materializing virtual worker entries. Selection is performed without replacement, bounded by the requested replica count, and validates inconsistent weight state defensively.

Worker weights are carried in an optional heartbeat field and stored in `WorkerInfo`. This keeps `WorkerAddress` unchanged, preserving bincode compatibility between mixed master and worker versions. Missing weights default to 1.

# Changes

| `curvine-common/proto` | Add optional weight fields to worker heartbeat and worker info | Backward compatible; missing values default to 1 |
| `curvine-common/src/conf` | Add worker weight configuration | Existing configurations continue using weight 1 |
| `curvine-server/src/master/fs/policy` | Add weighted worker selection policy | Enabled only when `master.worker_policy = "weighted"` |
| `curvine-server/src/worker` | Report configured weight through heartbeat | No change for other policies |
| `get_master_info` conversion | Return worker allocation weights | Adds optional response metadata |
| Tests | Cover distribution, filtering, uniqueness, exact capacity, and protobuf defaults | No production behavior change |
| `etc/curvine-cluster.toml` | Document weighted policy configuration | Documentation only |

# Test verified

| Weighted policy and protobuf unit tests | NOT RUN | Per user request; test commands were provided |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-common/src/utils/proto_utils.rs
+++ b/curvine-common/src/utils/proto_utils.rs
@@ -332,6 +332,7 @@ impl ProtoUtils {
             non_fs_used: src.non_fs_used,
             last_update: src.last_update,
             reserved_bytes: src.reserved_bytes,
+            weight: Some(src.weight),
             storage_map: Default::default(),
         };
 
@@ -372,6 +373,7 @@ impl ProtoUtils {
                 fs_used: info.fs_used,
                 non_fs_used: info.non_fs_used,
                 reserved_bytes: info.reserved_bytes,
+                weight: info.weight.unwrap_or_else(WorkerInfo::default_weight),
                 ..Default::default()
             };
             for (k, v) in info.storage_map {
--- a/curvine-server/src/master/fs/policy/weighted_worker_policy.rs
+++ b/curvine-server/src/master/fs/policy/weighted_worker_policy.rs
@@ -1,0 +1,228 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::master::fs::policy::{ChooseContext, WorkerPolicy};
+use curvine_common::state::{WorkerAddress, WorkerInfo};
+use indexmap::IndexMap;
+use orpc::{err_box, CommonResult};
+use rand::{thread_rng, Rng};
+use std::collections::HashSet;
+
+/// Selects workers randomly in proportion to their configured weights.
+///
+/// Selection is without replacement, so a worker can host at most one replica
+/// of a block. Workers with weight 0 remain registered but receive no new data.
+pub struct WeightedWorkerPolicy {}
+
+impl Default for WeightedWorkerPolicy {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl WeightedWorkerPolicy {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn is_eligible(
+        id: &u32,
+        worker: &WorkerInfo,
+        exclude_workers: &HashSet<u32>,
+        selected_ids: &HashSet<u32>,
+        min_available: i64,
+    ) -> bool {
+        worker.is_live()
+            && !exclude_workers.contains(id)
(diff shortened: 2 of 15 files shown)


curvine-server/src/master/fs/policy/weighted_worker_policy.rs
```diff
@@ -0,0 +1,222 @@
...
+                worker.is_live()
+                    && !exclude_workers.contains(id)
+                    && worker.available >= min_available
+                    && worker.weight > 0
+            })
+            .map(|(_, worker)| worker.weight as u64)
+            .sum::<u64>();
+
+        if total_weight == 0 {
+            return err_box!("No available workers with a positive weight found");
```
szbr00: `total_weight == 0` is returned for any empty eligible set (not live, excluded, insufficient `available`, or weight 0), but the message always blames weight. That can send operators to the wrong config when the real cause is space or exclusions.

Please distinguish the failure modes (or use a neutral message such as "No eligible workers for weighted selection") so weight-only cases are not implied when space/status filtering removed everyone.

curvine-server/src/master/master_handler.rs
```diff
@@ -478,6 +478,7 @@ impl MasterHandler {
     ) -> FsResult<Vec<WorkerCommand>> {
         let status = HeartbeatStatus::from(header.status);
         let address = ProtoUtils::worker_address_from_pb(&header.address);
+        let weight = header.weight.unwrap_or_else(WorkerInfo::default_weight);
```
szbr00, line 483: Weight is taken directly from the worker heartbeat with no master-side validation or cap. A misconfigured worker can advertise a very large weight (up to u32::MAX) and monopolize allocations under the weighted policy.

Consider a master-side maximum, or documenting clearly that weights must be treated as trusted admin configuration. Longer term, master-configured weights (by worker id/host) would avoid relying on self-reported values.

curvine-server/src/master/fs/policy/weighted_worker_policy.rs
```diff
@@ -0,0 +1,222 @@
...
+        exclude_workers: &HashSet<u32>,
+        min_available: i64,
+        rng: &mut R,
+    ) -> CommonResult<Vec<WorkerAddress>> {
+        let mut selected = Vec::with_capacity(count.min(workers.len()));
+        let mut selected_ids = HashSet::with_capacity(count);
+
+        // Treat each worker as a virtual range whose length is its weight.
+        // Calculate the full range once, then shrink it as workers are selected.
+        let mut total_weight = workers
```
szbr00, line 66: Eligibility is computed twice with nearly identical conditions (initial `total_weight` sum vs. the selection scan, plus `selected_ids` in the second pass). If one side is updated later, the "Worker weight state is inconsistent" path becomes more likely.

Consider extracting a shared `is_eligible(...)` helper used by both passes so the invariant stays obvious and harder to break.

curvine-common/src/utils/proto_utils.rs
```diff
@@ -372,6 +373,7 @@ impl ProtoUtils {
                 fs_used: info.fs_used,
                 non_fs_used: info.non_fs_used,
                 reserved_bytes: info.reserved_bytes,
+                weight: info.weight.unwrap_or_else(WorkerInfo::default_weight),
```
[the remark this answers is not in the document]
bigbigxu: Thanks for the report. After scope review, we are keeping the existing `last_update` decoding behavior because this pull request is limited to weighted worker allocation. The unrelated change was removed in 84cd4faa.

curvine-server/src/master/fs/policy/weighted_worker_policy.rs
bigbigxu: Thanks. Fixed in b950aa13: the empty eligible-set error is now the neutral `No eligible workers for weighted selection` message.

curvine-server/src/master/master_handler.rs
bigbigxu: Thanks. Addressed in b950aa13 by documenting worker weight as trusted administrator configuration in `WorkerConf`, the example configuration, and the heartbeat handling path. The master intentionally preserves the configured `u32` value instead of applying an arbitrary cap that could alter valid allocation ratios.

curvine-server/src/master/fs/policy/weighted_worker_policy.rs
bigbigxu: Thanks. Fixed in b950aa13: both the initial weight calculation and the selection scan now use the same `is_eligible` helper.

szbr00 (approved):

PR #1205, opened 2026-07-20
fix(fuse): fail fast on file handle restore errors (#1084)

### LINKED ISSUE #1084: [FEATURE] fuse: restore should distinguish parse vs backend-open handle errors (by LiAuto)

`NodeState::restore` reads `handles_count` and loops that many times. If `FileHandle::restore` returns an error it logs and continues to the next iteration.

- `curvine-fuse/src/fs/state/node_state.rs:1109` (loop that skips failed handles)
- `curvine-fuse/src/fs/state/backend_handle.rs:179` (`BackendHandle::restore` can fail mid-record or after reopening the backend reader/writer)

The state file is line-delimited JSON (`StateReader::read_json_line` reads one line per record). Two distinct failure modes are conflated:
- A **deserialization/parse** failure can consume a partial/whole line and desynchronize the reader for subsequent records.
- A **backend reopen** failure happens after the whole record has been read (alignment intact); it just drops an open handle, silently changing restored process semantics.

**Impact**

- Continuing after a parse failure can misalign the state reader → later records (including the dir-handle section) parse incorrectly.
- Silently skipping open handles changes restored semantics without signaling it.

- Either fail restore on the first handle-restore error (fail-fast), or add length-prefixed per-handle framing so a bad record can be safely skipped without desync.
- At minimum, distinguish parse/deserialization errors (must fail-fast — reader is desynced) from backend-open errors (safe to skip, but log clearly).

**Additional context**

From a curvine-fuse session/restore review. Composes with the persist fail-fast work (accurate handle count on the write side + safe framing on the read side).

Acceptance / test gaps:
- Restore fails fast on a parse error, or safely skips only length-framed bad records.
- Backend-open failure is distinguished from parse failure in logs/behavior.

ligeng8848:
# Summary

Fail the FUSE state restore as soon as any persisted file handle cannot be restored. This prevents a malformed handle record from desynchronizing the state reader and avoids reporting success after silently dropping an open handle.

# Issue Describe / Design

Closes #1084

Root cause:
- A persisted file handle contains multiple records.
- `NodeState::restore` previously logged restoration errors and continued.
- Continuing after a parse failure could leave the reader at the wrong position, causing later data to be misinterpreted.
- It could also silently lose an open file handle while reporting that restoration succeeded.

Fix:
- Return the original error immediately when any file handle cannot be restored.
- Mark the restoration stage successful only after all handles are restored.
- Add a regression test using a real state file containing a corrupt handle record.

# Changes

| Module / File | Change | Impact |
| ------------- | ------ | ------ |
| `curvine-fuse/src/fs/state/node_state.rs` | Fail state restoration on the first file-handle error and add a regression test | Corrupt or incomplete handle state can no longer be silently ignored |

# Test verified

| Test case | Result |
| --------- | ------ |
| `cargo test -p curvine-fuse --lib restore_returns_error_on_corrupt_file_handle_record -- --nocapture` | PASS |
| `cargo test -p curvine-fuse --lib fs::state -- --nocapture` | PASS (10 tests) |
| `make format` | PASS (formatting and release Clippy checks) |

# Dependencies

- Related issues: #1081
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -1147,8 +1147,6 @@ impl NodeState {
                     StateStageTimer::start(metrics_enabled, false, STATE_STAGE_FILE_HANDLES);
                 info!("node_state::restore: restoring file_handles");
                 let handles_count = reader.read_len()?;
-                let mut restored_handles = 0;
-                let mut any_failed = false;
                 for i in 0..handles_count {
                     let handle = match FileHandle::restore(reader, self).await {
                         Ok(handle) => handle,
@@ -1159,8 +1157,7 @@ impl NodeState {
                                 handles_count,
                                 e
                             );
-                            any_failed = true;
-                            continue;
+                            return Err(e);
                         }
                     };
 
@@ -1169,16 +1166,13 @@ impl NodeState {
                         .entry(handle.ino())
                         .or_default()
                         .insert(handle.fh(), Arc::new(handle));
-                    restored_handles += 1;
                 }
                 info!(
-                    "node_state::restore: {}/{} file_handles restored",
-                    restored_handles, handles_count
+                    "node_state::restore: {} file_handles restored",
+                    handles_count
                 );
                 if let Some(stage) = stage {
-                    if !any_failed {
-                        stage.success();
-                    }
+                    stage.success();
                 }
             }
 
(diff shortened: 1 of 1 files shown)


yangcx000 (approved):
LGTM. This implements the fail-fast restore path for file-handle restore errors and includes regression coverage for corrupt handle records.

PR #1209, opened 2026-07-20
fix(raft): avoid PeerStorage hang in empty_static_voter test (#1206)

LJF-050:
## Summary
- Fixes [CurvineIO/curvine#1206](https://github.com/CurvineIO/curvine/issues/1206) (Multica CUR-61)
- Root cause: `empty_static_voter_panics_on_leader_commit_above_local_last_index` correctly triggers raft-rs panic (`commit > last_index`), but built `PeerStorage` with `RaftClient` + dual `Runtime`/`GroupExecutor`. Under nextest those threads do not exit after `catch_unwind`, so the case hangs in futex until SIGKILL (~1499s) even though the panic already fired.
- Change: use `NoSnapshotLogStorage` directly as `RawNode` storage so the characterization test no longer starts background runtimes.

## Changed files
- `curvine-common/tests/raft_node_test.rs`
- `repair-plan.json`

## Evidence
- run_id: `20260720T085722Z-15c6cc26`
- tested head_sha: `ea9978a4b77ead56a2ea59b41cd05ef66cbbda41`
- harness_sha: `e8821100649d748e4120df664ab10c02d0f54d9e`
- fingerprint: `sha256:660915368bb66282898d7d2cae6d0ff6cb3e8437dc0ac87bb707b55fccf1ee95`
- artifact: `file:///data/data1/liujifeng/curvine-harness-runners/daily-20260716/ordinary/runs/20260720T085722Z-15c6cc26`
- manifest_sha256: `sha256:ab345ff1b4f680a5807de409fcc3fb75bc00a0e246df9567717581bc6c1213d7`
- repair-plan sha256: `6e67048f361868af7ba6d135d6ab768132e6f07fa8ee3c94e0ef453241ebe062`
- risk: low (test-only)

## Test verified
```bash
cargo test -p curvine-common --test raft_node_test empty_static_voter_panics_on_leader_commit_above_local_last_index
# ok in 0.00s (panic caught; no hang)
cargo clippy -p curvine-common --test raft_node_test -- -D warnings
# ok
```

## Risk
Low. Product raft path unchanged; only removes unnecessary PeerStorage/Runtime wiring from a unit characterization test.

PR #1210, opened 2026-07-20
fix(master): restart full block report after session invalidation (#1193)

### LINKED ISSUE #1193: [BUG]: full block report reconcile fails when incremental report invalidates session before new full report arrives
**Describe the bug**
The integration test `full_block_report_reconcile_removes_stale_location_async` in `curvine-server/tests/master_fs_test.rs` fails consistently. The test creates a file with two blocks, sends an incremental block report with both blocks, then sends a full block report with only the first block (total_len=1). It expects the second block's worker location to be asynchronously reconciled (cleared), but the location is never removed.

**To Reproduce**
1. Run `cargo test -p curvine-server --test master_fs_test -- full_block_report_reconcile_removes_stale_location_async`
2. Test panics at `master_fs_test.rs:426`:
```
thread 'full_block_report_reconcile_removes_stale_location_async' panicked at
curvine-server/tests/master_fs_test.rs:426:
stale worker location for block ... was not reconciled
```

**Expected behavior**
The async reconciliation should remove the second block's worker location after the full block report completes. The test should pass.

 - OS: Ubuntu 22.04
 - Curvine branch: main (commit 43165a3)
 - Relevant code: `master_filesystem.rs:collect_full_block_report` (line 769)

**Additional context**
- The second block still has `locs: [WorkerAddress { worker_id: 100, ... }]` when it should be empty

**Proposed Fix**
In `collect_full_block_report` (`master_filesystem.rs:769`), swap the two guard checks so the `total_len` check runs first:

This ensures a new full report with a different `total_len` starts a fresh session instead of being silently discarded by a stale `invalidated` flag from a prior incremental report.

LJF-050:
## Summary
- Fixes [CurvineIO/curvine#1193](https://github.com/CurvineIO/curvine/issues/1193)
- Multica: CUR-65 (parent CUR-62 handoff; also previously CUR-61)
- Root cause: after an incremental report invalidates the full-report session, `collect_full_block_report` returned `None` without starting a new session, so a later full report never completed and async reconcile left the stale second-block location.
- Change: drop invalidated sessions before accepting a new full report; only mark existing sessions invalidated (do not insert stub invalidated entries).

## Changed files
- `curvine-server/src/master/fs/master_filesystem.rs`
- `repair-plan.json`

## Evidence
- Reproduced FAIL on head_sha `56d93ae6a709062b1d627236281958c141c3b906` (CUR-65 / pipeline `20260720T143322Z`)
- integration run_id: `20260720T143526Z-476859b9`
- daily run_id: `20260720T144209Z-7efb68c9`
- harness_sha: `3011dc922f89d9ab3656d1630655e0ff0562fbd4`
- image_digest: `sha256:4ff9e66bdf1a366b0c9031ec800ddd4c000c366d8b61d48357fa856398a0c893`
- Earlier evidence (CUR-61): head_sha `ea9978a4b77ead56a2ea59b41cd05ef66cbbda41`, daily run `20260720T085722Z-15c6cc26`
- risk: low

## Test verified
```bash
# On PR tip 4480b1c (also cherry-picked onto 56d93ae locally)
cargo test -p curvine-server --test master_fs_test full_block_report_reconcile_removes_stale_location_async
# ok (0.68s)

cargo clippy -p curvine-server --test master_fs_test -- -D warnings
# ok
```

Baseline at `56d93ae` without this patch: test panics `stale worker location ... was not reconciled`.

## Risk
Low. Localized full-report session lifecycle; does not change worker report scheduling.


repair-plan.json
```diff
@@ -0,0 +1,27 @@
+{
```
szbr00: Same concern as on #1209: root `repair-plan.json` is shared harness bookkeeping and already documents other issues on `main`. Landing this PR overwrites that file again (and conflicts with concurrent harness PRs). Prefer dropping it from the PR if it is not meant to be durable source of truth.

LJF-050: Thanks. Removed repair-plan.json in 6a593c9 for the same reason as #1209 — leave main's shared harness metadata untouched; the block-report session invalidation fix is unchanged.

szbr00 (approved):

Issue #1090, opened 2026-07-13
[FEATURE] fuse: bound and supervise spawned metadata dispatch tasks (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

Every metadata request spawns a new runtime task, with no concurrency bound and no session-level supervision.

- `curvine-fuse/src/session/channel/fuse_receiver.rs:391` (stream vs meta split)
- `curvine-fuse/src/session/channel/fuse_receiver.rs:408` (meta_task guard)
- `curvine-fuse/src/session/channel/fuse_receiver.rs:411` (`self.rt.spawn(...)`)

Characteristics:
- No explicit semaphore/backpressure for metadata task count.
- The `JoinHandle` is not retained by the session.
- Dispatch errors are logged inside the spawned task, not propagated to receiver/session supervision.

**Impact**

- A metadata storm can create unbounded task pressure.
- Slow backend metadata operations can accumulate many tasks.
- Session shutdown cannot directly join or abort specific metadata tasks (relevant to a clean shutdown with timeout).
- Metadata dispatch errors are invisible to `run_all` except through logs/metrics.

**Describe the solution you'd like**

- Add a configurable metadata concurrency limit or bounded task queue.
- Track metadata tasks in session-level supervision (e.g. a task set with shutdown cancellation), so shutdown can drain/abort them.
- Distinguish unexpected dispatch-infrastructure failures from ordinary filesystem errors that are already replied to the kernel.

**Additional context**

From a curvine-fuse receiver/dispatch review. This is the metadata sub-task layer, distinct from (but composing with) the receiver/sender supervision rework and the shutdown-timeout work. Note: malformed-request handling that currently terminates the receiver is folded into the receiver/sender supervision issue, not here.

Acceptance / test gaps:
- Metadata task dispatch infrastructure errors are visible to session supervision.
- Metadata task concurrency is bounded under a request storm.

yangcx000:
@yangyang233333

PR #1212, opened 2026-07-20
test(fuse): cover stream pre-dispatch error reply paths (#1089)

### LINKED ISSUE #1089: [FEATURE] fuse: cover stream pre-dispatch error reply paths and avoid receive-loop blocking (by LiAuto)

`start` handles stream requests inline and logs-and-continues on error:

```rust
if let Err(e) = self.send_stream(req).await {
    error!("failed to dispatch stream request: {}", e);
}
```

`send_stream_dispatch` sends an error reply (via a cloned `err_rep`) only when the `fs.read/write/flush/release/fsync` call returns `Err` after the match. Some stream ops return `Err` before replying:
- read handle with no reader → `BackendHandle::read` returns EIO before replying (`fuse_receiver.rs`, `backend_handle.rs:73`),
- write handle with no writer → `BackendHandle::write` returns EIO before replying (`backend_handle.rs:106`),
- pre-dispatch handle lookup errors in `CurvineFileSystem::read/write`.

Verified nuance: the `FuseResponse` double-reply guard (`fuse_response.rs` `commit_reply_task`: `if m.finished { return None }`) makes a second send a no-op, so the `err_rep` fallback is generally safe and does not double-count.

- `curvine-fuse/src/session/channel/fuse_receiver.rs:200` / `:317` / `:387`

**Remaining risk**

- If the error-reply enqueue/send itself fails, `start` only logs and continues; from the kernel's perspective that request can remain unanswered.
- Stream dispatch runs inside the receive loop, so a stuck/backpressured stream dispatch also delays receiving new requests on that receiver.

- Add explicit tests for every pre-dispatch stream error path, proving exactly one error reply is enqueued:
  - read missing handle, read handle without reader,
  - write handle without writer,
  - flush/release/fsync handle-lookup failure,
  - backend EIO before reply enqueue.
- Decide whether stream dispatch enqueue/send failures should be fatal to the receiver/session (compose with the receiver/sender supervision rework) rather than log-and-continue.
- Consider moving stream dispatch to a bounded worker path so inline handling cannot block the receive loop under backend/channel backpressure.

**Additional context**

From a curvine-fuse receiver/dispatch review. The "should dispatch failures be fatal" decision composes with the session-layer task-supervision rework. Functionally the double-reply guard already covers most cases; this is primarily a test-coverage + receive-loop-blocking hardening (P1, not a functional defect).

Acceptance / test gaps:
- Stream read/write handle lookup failures enqueue exactly one error reply.
- Stream backend EIO before reply enqueue is replied exactly once.

yangcx000:
# Summary

Adds test coverage proving the acceptance property from #1089: each FUSE stream op (read/write/flush/release/fsync) that errors *before* replying to the kernel (a "pre-dispatch error", e.g. a handle-lookup miss) enqueues exactly one error reply and never double-replies. Test-only change, no production code touched.

# Issue Describe / Design

Closes #1089.

#1089 is a P1 test-coverage + hardening item, not a functional defect: the double-reply guard on `FuseResponse` (shared `finished` flag between `rep` and its `err_rep` clone) already makes the `err_rep` fallback in `send_stream_dispatch` safe. What was missing was an explicit proof that every pre-dispatch error path enqueues exactly one error reply.

Design notes:
- The real backend EIO paths (`BackendHandle` with `reader`/`writer` = None) are not unit-testable without a live cluster (a handle can only enter the private `NodeState.handles` via real client RPC through a connected `UnifiedFileSystem`). So the tests drive the `err_rep` fallback in `send_stream_dispatch` with a mock `FileSystem` — which is exactly #1089's assertion target.
- The mock returns `EBADF` (what `find_handle` actually returns on a miss), not the `EIO` the issue prose mentions; the deviation is documented in-code.
- Exact `== 1` counts are used (not the lower-bound asserts the sibling metrics tests must use), because each test reads a per-test reply channel and per-test Arc slot rather than process-global metric label children — so the counts are deterministic under the default parallel harness.

# Changes

| `curvine-fuse/src/session/channel/fuse_receiver.rs` | New `#[cfg(test)]` submodule `stream_error_coverage`: mock `PreDispatchErrFs`, fd-free harness `dispatch_and_collect`, 5 per-family tests + 1 metrics-disabled variant + 1 release-only double-reply-noop regression | None (test-only) |

# Test verified

| `cargo test -p curvine-fuse --lib session::channel::fuse_receiver` (debug) | PASS | 22 passed |
| `cargo test -p curvine-fuse --lib --release stream_error_coverage` | PASS | 7 passed (incl. release-only regression) |
| `cargo clippy -p curvine-fuse --tests` | PASS | no warnings |

# Dependencies

- Related PRs: None
- Related issues: this PR delivers the error-reply coverage that #1089 asks for. The "avoid receive-loop blocking" / session-fatal-dispatch decision items mentioned in the issue are coupled to the session-supervision rework in #1079 and #1090 and are intentionally **not** addressed here.
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -1754,6 +1754,29 @@ mod tests {
             }
         }
 
+        // A FileSystem whose `read` REPLIES via the passed-in `reply` and THEN returns
+        // `Err`. This is the "Verified nuance" from #1089: an op that already answered
+        // the kernel but still surfaces an error. It exercises the exact end-to-end path
+        // `send_stream_dispatch` guards — the reply is sent through the original `rep`,
+        // then `if res.is_err() { err_rep.send_rep(res) }` fires on the shared clone — so
+        // driving it through `dispatch_and_collect` proves the fallback CANNOT
+        // double-reply, not just that the slot guard works in isolation.
+        //
+        // `cfg(not(debug_assertions))`: its only user, `err_rep_fallback_after_a_reply_
+        // is_noop`, is release-only (a real double reply trips `commit_reply_task`'s
+        // `debug_assert!` in debug), so gating the mock the same way avoids a
+        // "never constructed" warning in debug builds.
+        #[cfg(not(debug_assertions))]
+        struct ReplyThenErrFs;
+        #[cfg(not(debug_assertions))]
+        impl FileSystem for ReplyThenErrFs {
+            async fn read(&self, _op: Read<'_>, reply: FuseResponse) -> FuseResult<()> {
+                // Answer the kernel first (empty success frame), then report an error.
+                reply.send_rep::<(), crate::FuseError>(Ok(())).await?;
+                err_fuse!(ERRNO, "late error after a successful reply")
+            }
+        }
+
         fn make_request(opcode: u32, unique: u64, payload: &[u8]) -> FuseRequest {
             let header = fuse_in_header {
                 len: (size_of::<fuse_in_header>() + payload.len()) as u32,
@@ -1821,8 +1844,8 @@ mod tests {
         // is still alive. (One is: `observer` below holds a sender for the whole drain,
         // so in practice we hit the `Ok(None)` empty case — but the loop must not depend
         // on that.)
-        fn dispatch_and_collect(
-            fs: Arc<PreDispatchErrFs>,
+        fn dispatch_and_collect<F: FileSystem>(
+            fs: Arc<F>,
             with_metrics_ctx: bool,
             req: FuseRequest,
         ) -> (FuseResult<()>, Vec<FuseTask>, FuseResponse) {
@@ -1850,10 +1873,8 @@ mod tests {
                 // the `err_rep` the dispatch builds internally.
                 let observer = rep.clone();
 
-                let res = super::super::FuseReceiver::<PreDispatchErrFs>::send_stream_dispatch(
-                    &fs, req, rep,
-                )
-                .await;
+                let res =
+                    super::super::FuseReceiver::<F>::send_stream_dispatch(&fs, req, rep).await;
 
                 // Drain whatever the (fully awaited) dispatch buffered. Stop on empty
                 // OR closed — `Ok(Some(_))` only — so the loop can never panic on a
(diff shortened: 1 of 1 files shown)


curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -1681,4 +1681,334 @@ mod tests {
...
+        // send_stream layer (a shared clone), complementing the generic guard tests in
+        // `fuse_response.rs` (`t13_*`).
+        //
+        // Release-only: a genuine double reply trips `commit_reply_task`'s
+        // `debug_assert!(!finished)` in debug builds (see `double_reply_panics_in_debug`
+        // in fuse_response.rs); the release behaviour is the safe warn + no-op asserted
+        // here.
+        #[test]
+        #[cfg(not(debug_assertions))]
+        fn err_rep_fallback_after_a_reply_is_noop() {
```
szbr00, line 2000: `err_rep_fallback_after_a_reply_is_noop` drives `FuseResponse` clone/send directly and largely duplicates `fuse_response::t13_real_double_reply_is_noop`. To pin the #1089 property at the `send_stream_dispatch` seam, consider a mock FileSystem that replies via the passed-in `rep` and then returns `Err`, and drive it through `dispatch_and_collect` — that would prove the `if res.is_err() { err_rep.send_rep }` block cannot double-reply end-to-end, not just that the shared slot guard works in isolation.

yangcx000: Thanks, good call — fixed in 46a184d. Replaced the isolated hand-rolled version with a true integration test at the `send_stream_dispatch` seam: a new `ReplyThenErrFs` mock replies via the passed-in `rep` and then returns `Err`, driven through `dispatch_and_collect` (now generic over `F: FileSystem`). It asserts exactly one reply reaches the wire and the shared slot finishes once, proving the `if res.is_err() { err_rep.send_rep }` fallback cannot double-reply end-to-end — not just that the slot guard works in isolation. Still `cfg(not(debug_assertions))` (a real double reply trips the `debug_assert!` in debug), and the mock is gated the same way to avoid a debug "never constructed" warning.

szbr00 (approved):

yangcx000:
Thanks for adding these stream pre-dispatch error coverage tests. One remaining coverage gap: the new tests use `PreDispatchErrFs` to return `Err` directly from the mocked `FileSystem` methods, so they validate the generic `send_stream_dispatch` fallback behavior, but they do not exercise the real `CurvineFileSystem`/handle paths requested in #1089.

In particular, the production paths for handle lookup failures and backend errors still go through `CurvineFileSystem::{read,write,flush,release,fsync}` and `BackendHandle::{read,write}`. Cases like missing handle, read handle without a reader, write handle without a writer, and backend EIO before reply enqueue could still regress without these tests failing.

Could we add a follow-up test that drives these scenarios through the real handler/state path (or at least constructs the real `BackendHandle` states) instead of only the mocked `FileSystem` error path?

PR #1213, opened 2026-07-20
fix(client): detect SPDK storage via worker-reported has_spdk flag in LocatedBlock

### LINKED ISSUE #1077: [BUG]: SPDK writes fail with EIO via BlockWriterLocal when fuse and worker are co-located (short_circuit=true default)
### Describe the bug
When `curvine-fuse` and a `curvine-worker` run on the same host and the worker serves an SPDK-backed disk, every write to an SPDK block returns `EIO` within ~15–24 ms, and **no data ever reaches the worker or the SPDK target** (`num_write_ops` stays at 0). The fuse log is the only place the error surfaces:

```
ERROR fs_writer_buffer.rs:129 buffer writer error: Common([][curvine-fuse] ERROR: Uninitialized(curvine-client/src/block/block_writer_local.rs:62))
WARN fuse_response.rs:103 send_rep ... errno [curvine-fuse] ERROR: Uninitialized(...block_writer_local.rs:62): : 5
```

The client mis-dispatches the SPDK write to `BlockWriterLocal`, which requires a filesystem path that SPDK bdevs do not have, so it returns `Uninitialized` and the write fails before any RPC reaches the worker.

### To Reproduce
1. Deploy curvine with fuse + worker co-located on one node, SPDK NVMe-oF/RDMA target remote.
2. Keep the default `[client] short_circuit = true` (`curvine-common/src/conf/client_conf.rs:413`).
3. Configure an SPDK-backed worker disk.
4. `dd if=/dev/zero of=/curvine-fuse/fio-spdk/x bs=1M count=10` → `Input/output error`.

Relevant code path (client write dispatch):
- `curvine-client/src/block/block_writer.rs:115-117` — `let short_circuit = conf.short_circuit && is_local_worker && located_block.block.storage_type != StorageType::SpdkDisk;`
- `curvine-client/src/block/block_writer.rs:119-129` — when `short_circuit` is true the client builds a `BlockWriterLocal` instead of `BlockWriterRemote`.
- `curvine-client/src/block/block_writer_local.rs:62` — `let path = try_option!(write_context.path);` → `path` is `None` for an SPDK block → returns `Uninitialized`.

The guard `storage_type != StorageType::SpdkDisk` at `block_writer.rs:117` is meant to exclude SPDK blocks, but the block's `storage_type` is **not** tagged `SpdkDisk` (see companion issue), so the guard does not fire and the write is mis-dispatched.

### Expected behavior
With the default `short_circuit = true`, writes to SPDK-backed blocks should still succeed — the SPDK exclusion guard should reliably route them through `BlockWriterRemote` without requiring the operator to flip `short_circuit = false`.

### Environment
- SPDK NVMe-oF/RDMA, fuse + worker co-located, target remote.
- curvine main (v0.3.x); verified `short_circuit: true` default at `client_conf.rs:413`, `StorageType::SpdkDisk = 5` at `curvine-common/src/state/storage_info.rs:48`.

### Additional context
**Workaround:** set `[client] short_circuit = false` (the SPDK template `curvine-docker/deploy/spdk/curvine-cluster-spdk.toml:73` already does this). This is a workaround, not a fix — it disables short-circuit for *all* blocks on the host, removing the local fast path.

thuongle2210:
## Summary

- When `short_circuit=true` (default) and fuse + worker are co-located, SPDK writes fail with EIO because the client mis-dispatches to `BlockWriterLocal`, which requires a filesystem path that SPDK bdevs do not have. This PR adds a `has_spdk` flag to `LocatedBlock`, computed by the master from worker-reported `BlockLocation.storage_type`, so the client can correctly skip short-circuit for SPDK blocks.

- `block.storage_type` is the file-level requested type (e.g. `Disk`), not the actual worker-reported type. A block can span SPDK and Disk replicas concurrently. `has_spdk` is `true` when any **live** worker-reported `BlockLocation.storage_type`
  is `SpdkDisk`.
- Client fallback: if `BlockWriterLocal` / `BatchBlockWriterLocal` fails (e.g. missing local path), the open is aborted via `write_commit(cancel=true)` and the write is retried via the remote writer. This ensures clean server state.
- Closes [issue-1077](https://github.com/CurvineIO/curvine/issues/1077)

## Changes and Impacts

| `curvine-common/proto/common.proto` | Add `optional bool has_spdk = 4` to `LocatedBlockProto` | Backward compatible (optional field, default false) |
| `curvine-common/src/state/block_info.rs` | Add `has_spdk: bool` to `LocatedBlock` struct | Default `false` preserves existing behavior |
| `curvine-common/src/utils/proto_utils.rs` | Wire `has_spdk` in proto conversion | None |
| `curvine-server/src/master/fs/worker_manager.rs` | Compute `has_spdk` from worker-reported `BlockLocation.storage_type` | None (new field) |
| `curvine-server/src/master/fs/master_filesystem.rs` | Compute has_spdk via workers_have_spdk() at new block allocation and additional block assignment | SPDK blocks get correct has_spdk immediately, not just after first worker report |
| `curvine-client/src/block/block_writer.rs` | Use `!located_block.has_spdk` instead of `storage_type != SpdkDisk`; add fallback to remote on `BlockWriterLocal` failure | SPDK writes now route to remote correctly |
| `curvine-client/src/block/block_writer_local.rs` | Graceful `path: None` handling: abort open via `write_commit(cancel=true)` then return NoLocalPath error | Defense-in-depth: clean server state on failure |
| `curvine-client/src/block/batch_block_writer.rs` | Use `lb.has_spdk` instead of `block.storage_type == SpdkDisk`; add fallback to remote | Same fix for batch writes |
| `curvine-client/src/block/batch_block_writer_local.rs` | Graceful `path: None` handling: abort via `write_commit_batch(cancel=true)` | Defense-in-depth for batch |
| `curvine-server/tests/master_fs_test.rs` | Unit test: `located_block_has_spdk_reflects_worker_reported_storage_type` (4 scenarios) | None (test only) |
| `curvine-common/src/error/fs_error.rs` | Add `NoLocalPath` error variant (code 32) and `FsError::no_local_path()` constructor | Typed error for SPDK path-missing cases; |

## Test verified
- Passed: `dd if=/dev/zero of=/curvine-fuse/fio-spdk/x bs=1M count=10`
```
    sudo BUILD_ARGS="-p core -p fuse --spdk-rdma" NVME_PCI_ADDR=0000:00:0e.0 make docker-compose-spdk
    sudo docker exec -it spdk-worker-1 bash

    [root@fedora37 app]# curvine-fuse.sh start 
        Starting curvine-fuse with arguments: start
        fuse start success, pid=112
        Loaded configuration from /app/curvine/conf/curvine-cluster.toml
        2026/07/20 16:19:41.149 INFO cluster_conf.rs:272 allocator: tikv_jemallocator::Jemalloc    
        2026/07/20 16:19:41.149 INFO cluster_conf.rs:273 git version: 08f6ebb    
        2026/07/20 16:19:41.149 INFO cluster_conf.rs:274 cluster conf start: 
        format_master = true
        format_worker = true
        testing = false
        cluster_id = "curvine"

    [root@fedora37 app]# mkdir -p /curvine-fuse/fio-spdk
    [root@fedora37 app]# dd if=/dev/zero of=/curvine-fuse/fio-spdk/x bs=1M count=10
        10+0 records in
        10+0 records out
        10485760 bytes (10 MB, 10 MiB) copied, 0.0591448 s, 177 MB/s
```

- Passed: `located_block_has_spdk_reflects_worker_reported_storage_type`

### DIFF
--- a/curvine-client/src/block/batch_block_writer.rs
+++ b/curvine-client/src/block/batch_block_writer.rs
@@ -15,8 +15,9 @@
 use crate::block::batch_block_writer::BatchWriterAdapter::{BatchLocal, BatchRemote};
 use crate::block::{BatchBlockWriterLocal, BatchBlockWriterRemote};
 use crate::file::FsContext;
+use curvine_common::error::FsError;
 use curvine_common::fs::Path;
-use curvine_common::state::{CommitBlock, ExtendedBlock, LocatedBlock, StorageType, WorkerAddress};
+use curvine_common::state::{CommitBlock, ExtendedBlock, LocatedBlock, WorkerAddress};
 use curvine_common::FsResult;
 use futures::future::try_join_all;
 use orpc::err_box;
@@ -64,17 +65,32 @@ impl BatchWriterAdapter {
     ) -> FsResult<Self> {
         let conf = &fs_context.conf.client;
         // SPDK bypasses kernel — no local path. Disable short-circuit if any block uses SPDK.
-        let has_spdk = located_blocks
-            .iter()
-            .any(|lb| lb.block.storage_type == StorageType::SpdkDisk);
+        // Use has_spdk from worker-reported actual storage type, not block.storage_type
+        let has_spdk = located_blocks.iter().any(|lb| lb.has_spdk);
         let short_circuit =
             conf.short_circuit && fs_context.is_local_worker(worker_addr) && !has_spdk;
 
         let blocks: Vec<ExtendedBlock> = located_blocks.iter().map(|lb| lb.block.clone()).collect();
         let adapter = if short_circuit {
-            let writer =
-                BatchBlockWriterLocal::new(fs_context, blocks, worker_addr.clone(), 0).await?;
-            BatchLocal(writer)
+            match BatchBlockWriterLocal::new(
+                fs_context.clone(),
+                blocks.clone(),
+                worker_addr.clone(),
+                0,
+            )
+            .await
+            {
+                Ok(writer) => BatchLocal(writer),
+                Err(FsError::NoLocalPath(_)) => {
+                    // Server has no local path for batch blocks (e.g. SPDK bdev).
+                    // BatchBlockWriterLocal already aborted the open; retry via remote.
+                    let writer =
+                        BatchBlockWriterRemote::new(&fs_context, blocks, worker_addr.clone(), 0)
+                            .await?;
+                    BatchRemote(writer)
+                }
+                Err(e) => return Err(e),
+            }
         } else {
             let writer =
                 BatchBlockWriterRemote::new(&fs_context, blocks, worker_addr.clone(), 0).await?;
--- a/curvine-client/src/block/block_writer.rs
+++ b/curvine-client/src/block/block_writer.rs
@@ -17,7 +17,8 @@
 use crate::block::block_writer::WriterAdapter::{Local, Remote};
 use crate::block::{BlockWriterLocal, BlockWriterRemote};
 use crate::file::FsContext;
-use curvine_common::state::{BlockLocation, CommitBlock, LocatedBlock, StorageType, WorkerAddress};
+use curvine_common::error::FsError;
+use curvine_common::state::{BlockLocation, CommitBlock, LocatedBlock, WorkerAddress};
 use curvine_common::FsResult;
 use futures::future::try_join_all;
 use orpc::err_box;
@@ -112,21 +113,38 @@ impl WriterAdapter {
         block_size: i64,
     ) -> FsResult<Self> {
         let conf = &fs_context.conf.client;
-        // SPDK bypasses kernel — no local path for writes
+        // SPDK bypasses kernel — no local path for writes.
+        // Use has_spdk (from worker-reported actual storage type)
         let short_circuit = conf.short_circuit
             && fs_context.is_local_worker(worker_addr)
-            && located_block.block.storage_type != StorageType::SpdkDisk;
+            && !located_block.has_spdk;
 
         let adapter = if short_circuit {
-            let writer = BlockWriterLocal::new(
-                fs_context,
+            match BlockWriterLocal::new(
+                fs_context.clone(),
                 located_block.block.clone(),
                 worker_addr.clone(),
                 pos,
                 block_size,
             )
-            .await?;
-            Local(writer)
+            .await
+            {
+                Ok(writer) => Local(writer),
+                Err(FsError::NoLocalPath(_)) => {
+                    // Server has no local path for this block (e.g. SPDK bdev).
+                    // BlockWriterLocal already aborted the open; retry via remote.
+                    let writer = BlockWriterRemote::new(
+                        &fs_context,
+                        located_block.block.clone(),
+                        worker_addr.clone(),
+                        pos,
+                        block_size,
+                    )
+                    .await?;
+                    Remote(writer)
+                }
+                Err(e) => return Err(e),
+            }
         } else {
             let writer = BlockWriterRemote::new(
                 &fs_context,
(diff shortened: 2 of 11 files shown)


curvine-docker/deploy/spdk/curvine-cluster-spdk.toml
```diff
@@ -60,7 +60,6 @@ subnqn = "nqn.2026-05.curvine:cnode1"
 
 # Client configuration  
 [client]
```
szbr00, line 62: Removing `short_circuit = false` here is not safe yet. The write path now falls back when the local path is missing, but the read path does not: `BlockReaderLocal` still uses `try_option!(read_context.path)`, and `BlockReader::get_reader` does not retry the same worker via `BlockReaderRemote` after a local open fails. `has_spdk` is also unused on the read side. With default `short_circuit=true`, co-located fuse+SPDK reads can still fail after a successful write. Please either keep `short_circuit = false` for SPDK until the read path is fixed, or apply the same `has_spdk` / remote-fallback handling to readers.

curvine-client/src/block/block_writer.rs
```diff
@@ -112,21 +112,37 @@ impl WriterAdapter {
...
                 worker_addr.clone(),
                 pos,
                 block_size,
             )
-            .await?;
-            Local(writer)
+            .await
+            {
+                Ok(writer) => Local(writer),
+                Err(_) => {
```
szbr00: This `Err(_)` catch falls back to the remote writer for every `BlockWriterLocal::new` failure, not only the intentional "no local path" case. That can mask real errors (RPC/open failures, `LocalFile` open failures after a path was returned) and retry remote without a guaranteed abort, which risks leaving the worker with an open/half-open block. Please match on the specific no-path error (or check `CreateBlockContext.path` / storage type before falling back) and propagate other errors.

curvine-client/src/block/batch_block_writer.rs
```diff
@@ -64,17 +64,31 @@ impl BatchWriterAdapter {
...
+            match BatchBlockWriterLocal::new(
+                fs_context.clone(),
+                blocks.clone(),
+                worker_addr.clone(),
+                0,
+            )
+            .await
+            {
+                Ok(writer) => BatchLocal(writer),
+                Err(_) => {
```
szbr00: Same concern as in `block_writer.rs`: catching `Err(_)` and always retrying via `BatchBlockWriterRemote` is too broad. Please fall back only for the explicit no-local-path failure that `BatchBlockWriterLocal` returns after `write_commit_batch(..., cancel=true)`, and surface other setup failures to the caller.

curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -575,6 +575,7 @@ impl MasterFilesystem {
         let located = LocatedBlock {
             block,
             locs: choose_workers,
+            has_spdk: false,
```
szbr00: New block allocation always returns `has_spdk: false`, so the primary path for issue 1077 (fresh `add_block` under `short_circuit=true`) never uses the flag and always pays local-open → abort → remote-reopen. Consider deriving `has_spdk` from the chosen workers' heartbeat-reported storage types at allocation time (or from the worker's selected layout), so SPDK writes can skip short-circuit up front instead of relying only on the fallback.

thuongle2210: I updated by get information from active workers

curvine-client/src/block/batch_block_writer.rs
thuongle2210: I added the FsError::NoLocalPath to handle it.

curvine-client/src/block/block_writer.rs
thuongle2210: I added the FsError::NoLocalPath to handle it.

curvine-docker/deploy/spdk/curvine-cluster-spdk.toml
thuongle2210: I reverted it and keep it until reading side is fixed.

thuongle2210:
Hi @szbr9486 , I've addressed all of your concerns.

szbr00 (approved):

PR #1218, opened 2026-07-21
fix(fuse): preserve deferred-delete cleanup failures (#1112)

### LINKED ISSUE #1112: [FEATURE] fuse: deferred-delete failure is swallowed and can orphan backend files (by LiAuto)

In `release`, the delayed (deferred) delete of a `mark_delete`d inode swallows a backend delete failure but still clears the delete mark and reports success to the kernel, which can leave an orphaned backend file.

- `curvine-fuse/src/fs/curvine_file_system.rs:889`-`897`

Current behavior:

```rust
if self.state.deferred_delete_ready(ino).await? {
    if let Err(e) = self.fs.delete(&path, false).await {
        warn!("failed to delete {} after last handle closed: {}", path, e);
    }
    self.state.clear_mark_delete(ino)?;   // cleared even on delete failure
}
reply.send_rep::<(), FuseError>(Ok(())).await?;  // success to kernel
```

**Impact**

- On `self.fs.delete` failure: the kernel is told the release (and thus the earlier unlink) completed, the dcache `mark_delete` is cleared, but the backend file still exists → a silent orphan.
- No retry state is preserved, so the file is never cleaned up.

- On delete failure, do NOT `clear_mark_delete` — keep a retriable state.
- Return the error to `release` or enqueue a background retry, and expose a metric/log so orphans are observable and eventually reclaimed.
- Decide the kernel-reply policy for this case (release itself may still need to reply, but the delete failure must not be silently lost).

**Related: `release_handle` removes the handle even when complete/flush fails (from NodeState review)**

A second release-path defect compounds the above and should be fixed together:

- `curvine-fuse/src/fs/state/node_state.rs:570`-`600` — `release_handle` runs the writer `complete/flush`, then **unconditionally** `remove_handle(ino, fh)` (`:597`) with `result?` (`:599`) AFTER the removal.
- `curvine-fuse/src/fs/curvine_file_system.rs:871`-`900` — if `complete/flush` fails, `release_handle` returns `Err`, so `release` short-circuits via `?` and never runs `fs_unlock(Flock)`, `fs_unlock(Plock)`, `deferred_delete_ready`, or the deferred delete. But the handle is already gone from the map, so the same `fh` cannot retry the cleanup.

Consequences: leaked backend flock/plock; an unlinked-but-still-open file may never get its deferred delete after a failing final close; no retry path.

Fix direction: split release into "commit data" and "state cleanup" phases. Even when the writer close fails, run unlock / deferred-delete evaluation best-effort, and do NOT remove the handle on complete/flush failure (or, if it must be removed, move the pending-delete + lock state into a retriable queue). Collect errors and return the first one after cleanup, rather than short-circuiting before cleanup.

**Additional context**

From curvine-fuse CurvineFileSystem and NodeState/inode-cache/handle reviews. Data/consistency issue (backend orphans + leaked locks + skipped deferred delete). Both defects live on the release path and should be fixed as one cohesive change.

Acceptance / test gaps:
- A failing `self.fs.delete` in deferred delete leaves `mark_delete` set (retriable) and does not silently report a clean unlink.
- `release_handle_complete_error_still_unlocks_and_preserves_delete_state`: a writer complete/flush failure still runs flock/plock unlock and deferred-delete evaluation, and does not lose pending-delete state.

ligeng8848:
# Summary

Make the FUSE release path preserve retriable state and report backend cleanup failures instead of silently losing deferred deletes or writer-close errors.

# Issue Describe / Design

Closes #1112

Root causes:

- `release_handle` removed the file handle after `complete` or `flush` even when that operation failed.
- `release` short-circuited on the first close or unlock error, skipping remaining lock and deferred-delete cleanup.
- A failed deferred backend delete was logged but its delete mark was cleared and success was returned to the kernel.

Reproduction steps:

1. Unlink a file while its final writer handle remains open.
2. Inject a writer-complete or backend-delete failure during release.
3. Observe that the old path loses the handle or pending-delete state while backend cleanup has not completed.

Fix approach:

- Add an atomic shared-resource release operation that changes the reference count only after cleanup succeeds.
- Keep the file handle and shared writer reference when `complete` or `flush` fails.
- Attempt flock, plock, and deferred-delete cleanup independently, retain the first error, and send exactly one final release reply.
- Clear the pending-delete mark only after delete succeeds or the backend reports that the file is already absent.
- Restore the locally tracked lock owner when a backend unlock fails.

Known limitation / follow-up:

- FUSE normally sends `RELEASE` only once per file handle. This PR preserves and exposes failed cleanup state but does not automatically retry it after the request finishes.
- Background reaping, bounded retry, lifecycle handling, and aggregate cleanup metrics are tracked separately in #1221.

# Changes

| `orpc/src/sync/async_map.rs` | Add cleanup-before-release reference handling with retry preservation | Existing release API is unchanged |
| `curvine-fuse/src/fs/state/node_state.rs` | Preserve handles/writers on close failure, finalize deferred deletes conditionally, and document the automatic-retry limitation | Failed cleanup remains observable and retained for future recovery work |
| `curvine-fuse/src/fs/curvine_file_system.rs` | Run all release cleanup stages and return the first real error | No more silent deferred-delete success |

# Test verified

| `make format` | PASS | Full workspace release-mode Clippy in a remote Linux environment; rerun after review update |
| `cargo test -p orpc release_with_cleanup -- --nocapture` | PASS | 2 focused reference-retention tests |
| `cargo test -p curvine-fuse fs::state::node_state::test:: -- --nocapture` | PASS | 8 state tests, including close and delete failure preservation |
| Mounted FUSE deferred-delete E2E | PASS | Opened, wrote, fsynced, unlinked while open, wrote again, closed, and confirmed backend deletion independently |

# Dependencies

- Related PRs: #1201 (merged)
- Related issues: #1112, #1221
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -167,6 +167,12 @@ impl CurvineFileSystem {
         &self.state
     }
 
+    fn retain_first_error(first: &mut FuseResult<()>, next: FuseResult<()>) {
+        if first.is_ok() {
+            *first = next;
+        }
+    }
+
     fn to_file_lock(&self, arg: &fuse_lk_in) -> FileLock {
         let client_id = self.fs.cv().fs_context().clone_client_name();
         FileLock {
@@ -198,7 +204,12 @@ impl CurvineFileSystem {
                 lock.end = u64::MAX;
             }
 
-            self.fs.set_lock(&path, lock).await?;
+            if let Err(e) = self.fs.set_lock(&path, lock).await {
+                // Preserve the owner locally when the backend unlock fails so a
+                // retained handle can retry the cleanup.
+                handler.add_lock(flags, owner_id);
+                return Err(e.into());
+            }
         }
 
         Ok(())
@@ -1058,31 +1069,59 @@ impl fs::FileSystem for CurvineFileSystem {
         let path = self.state.get_path(ino)?;
         let _guard = self.state.lock_path(&path).await;
 
-        let handle = self.state.release_handle(ino, op.arg.fh).await?;
+        let (handle, mut release_result) = self.state.release_handle(ino, op.arg.fh).await?;
 
         if handle.has_writer() {
             self.state
                 .invalid_cache(op.header.nodeid, None, INVAL_REASON_RELEASE);
         }
 
-        let unlock_result = self
-            .fs_unlock(&handle, LockFlags::Flock)
-            .await
-            .and(self.fs_unlock(&handle, LockFlags::Plock).await);
-        unlock_result?;
-
-        if self.state.deferred_delete_ready(ino).await? {
-            debug!(
-                "release ino={}: no more open handles, executing delayed deletion of {}",
-                ino, path
+        let flock_result = self.fs_unlock(&handle, LockFlags::Flock).await;
+        if let Err(e) = &flock_result {
+            warn!(
+                "failed to release flock for ino={}, path={}: {}",
+                ino, path, e
             );
-            if let Err(e) = self.fs.delete(&path, false).await {
-                warn!("failed to delete {} after last handle closed: {}", path, e);
+        }
+        Self::retain_first_error(&mut release_result, flock_result);
+
+        let plock_result = self.fs_unlock(&handle, LockFlags::Plock).await;
+        if let Err(e) = &plock_result {
+            warn!(
+                "failed to release plock for ino={}, path={}: {}",
+                ino, path, e
+            );
+        }
+        Self::retain_first_error(&mut release_result, plock_result);
+
+        match self.state.deferred_delete_ready(ino).await {
+            Ok(true) => {
+                debug!(
+                    "release ino={}: no more open handles, executing delayed deletion of {}",
+                    ino, path
+                );
+                let delete_result = self
+                    .state
+                    .complete_deferred_delete(ino, self.fs.delete(&path, false).await);
+                if let Err(e) = &delete_result {
+                    warn!(
+                        "failed to delete {} after last handle closed; retaining pending delete: {}",
+                        path, e
+                    );
+                }
+                Self::retain_first_error(&mut release_result, delete_result);
             }
-            self.state.clear_mark_delete(ino)?;
+            Ok(false) => {}
+            Err(e) => {
+                warn!(
+                    "failed to evaluate deferred delete for ino={}, path={}: {}",
+                    ino, path, e
+                );
+                Self::retain_first_error(&mut release_result, Err(e));
+            }
         }
 
-        reply.send_rep::<(), FuseError>(Ok(())).await?;
+        reply.send_rep(release_result).await?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1058,31 +1069,59 @@ impl fs::FileSystem for CurvineFileSystem {
...
+                    "failed to evaluate deferred delete for ino={}, path={}: {}",
+                    ino, path, e
+                );
+                Self::retain_first_error(&mut release_result, Err(e));
             }
-            self.state.clear_mark_delete(ino)?;
         }
 
-        reply.send_rep::<(), FuseError>(Ok(())).await?;
+        reply.send_rep(release_result).await?;
```
szbr00, line 1124: The deferred-delete failure path now correctly retains `pending_delete` and surfaces the error via `reply.send_rep(release_result)`. That is a clear improvement over clearing the mark and returning success. A natural follow-up would be a metric or periodic retry for inodes stuck with `pending_delete` set after a failed release, so operators can detect and reclaim them without manual intervention.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -573,37 +573,43 @@ impl NodeState {
...
 
             _ => Ok(()),
         };
 
-        // Remove the handle *after* complete/flush has finished so that the
-        // file is not prematurely deleted while the commit is still running.
-        let _ = self.remove_handle(ino, fh);
+        // A failed close keeps both the handle and its shared-writer reference,
+        // allowing cleanup to be retried without losing pending-delete state.
+        if close_result.is_ok() {
```
szbr00, line 610: When `close_result` fails, keeping the handle and writer ref enables retry, which the unit test exercises by calling `release_handle` twice. In production, FUSE `RELEASE` is typically invoked once per `fh`, so a failed final `complete`/`flush` may leave this handle in `self.handles` indefinitely and keep `has_open_handles` true, blocking deferred delete even though the client fd is already closed. Consider documenting this limitation or adding a follow-up background reaper (as noted in #1112) so orphaned pending deletes are eventually retried without relying on a second `RELEASE`.

curvine-fuse/src/fs/curvine_file_system.rs
ligeng8848: Thanks. The current PR already logs each cleanup failure; automatic retry and aggregate metrics are now tracked in #1221. Commit b541796 documents that retained cleanup is not retried automatically, and the PR body now calls out this limitation. The background reaper remains a focused follow-up so its backoff, concurrency, and lifecycle behavior can be designed and tested explicitly.

curvine-fuse/src/fs/state/node_state.rs
ligeng8848: Thanks, agreed that FUSE normally sends RELEASE only once. Commit b541796 updates the code comments to state that retained state is not automatically retried and links #1221, which tracks background reaping, bounded backoff, metrics, and lifecycle handling. This PR continues to preserve and surface the failure without claiming eventual automatic reclamation.

szbr00 (approved):

PR #1222, opened 2026-07-21
fix(fuse): maintain inode refcount when lookup hits an existing server-id (#1162)

### LINKED ISSUE #1162: [BUG]: fuse dcache: lookup hitting an existing server-id inode skips ref/lookup count increment (by LiAuto)
**Describe the bug**

`DirTree::lookup` (`curvine-fuse/src/fs/dcache/dir_tree.rs:179-231`) has three code paths, and one of them fails to maintain the inode reference counters (`ref_ctr` / `n_lookup`):

- Path A — an inode already exists under the same `(parent, name)` dentry (lines 186-198): calls `inode.add_lookup(1)`. OK
- Path C — a brand-new inode (lines 217-221): `Inode::with_status` initializes `n_lookup = 1`, `ref_ctr = 1`. OK
- Path B — a **new dentry name** whose `status.id` (server-id) resolves to an **already-existing inode** (lines 205-215): only reads `inode.ino` and does **neither** `add_lookup(1)` **nor** `add_ref(1)`. BUG

Because path B skips the increment, the reference-count invariant that `should_unref()` (`inode.rs:164`, requires `n_lookup == 0 && ref_ctr == 0`) relies on is broken. A new dentry pointing at an existing inode (hard-link-like, or a rename target resolving to a known server-id) leaves the counters under-counted, so later `unlink` + `forget` cannot bring them to zero and the inode is never removed from the dcache — an inode leak / stale-entry bug.

This is caught by three tests in `dir_tree.rs` that have been **failing since they were introduced** in commit `654e3d1` (PR #1041, "restructure metadata cache into a dedicated dcache module") — they have never passed:

- `lookup_existing_server_id_updates_inode_path_without_reinsert` (line 690): expects `ref_ctr == 2` after a second lookup hits the same server-id via a new name; actual `1`.
- `create_lookup_rename_link_unlink_forget_keeps_tree_consistent` (line 627): after link -> unlink -> forget, the inode is expected to be gone; it is not, because `ref_ctr` never reached the value `unlink` needs to decrement to zero.
- `create_then_forget_then_unlink` (line 665): forget + unlink should remove the inode; it remains.

**To Reproduce**

1. Check out `main`.
2. Run: `cargo test -p curvine-fuse --lib dcache::dir_tree`
3. Observe 3 failures:
   - `lookup_existing_server_id_updates_inode_path_without_reinsert` — `assertion left == right failed, left: 1, right: 2` at `dir_tree.rs:690`
   - `create_lookup_rename_link_unlink_forget_keeps_tree_consistent` — `assertion failed: t.get_inode(f, None).is_none()` at `dir_tree.rs:627`
   - `create_then_forget_then_unlink` — `assertion failed: t.get_inode(200, None).is_none()` at `dir_tree.rs:665`

Verified they fail identically when checked out at the introducing commit `654e3d1`, i.e. they are pre-existing, not a recent regression.

**Expected behavior**

When `lookup` resolves a new dentry name to an existing inode via server-id (path B), it should increment the inode's reference counters consistently with the other paths (at minimum `add_lookup(1)`, and `add_ref(1)` for the new dentry per hard-link semantics), so that `unlink` + `forget` correctly drive the counters to zero and `should_unref()` removes the inode. All three tests should pass.

- Curvine branch: `main` (reproduced at HEAD and at commit `654e3d1`)

**Additional context**

- Root cause is localized to the path-B arm of `DirTree::lookup` (`dir_tree.rs:205-215`); the fix likely belongs there (or a unified counter bump at the `dir.add_child` site), but the exact placement of `add_ref` vs `add_lookup` should be cross-checked against `link()` semantics to avoid double-counting on same-name repeated lookups (path A).
- Related but distinct: #1114 (READDIR taking a kernel lookup reference). This issue is about the lookup->existing-inode counter path, not readdir.
- These failures currently make `cargo test -p curvine-fuse` red on `main`.

yangyang233333:
# Summary

`DirTree::lookup` failed to maintain inode reference counters when a new dentry name resolved (via server-id) to an inode already present in the dcache, causing under-counted references and an inode leak. This fixes the counter maintenance for that path.

# Issue Describe / Design

Closes #1162 (partial — this PR fixes the lookup path; the unlink-side counter issue is addressed separately).

**Root cause**
`lookup` has a branch where a new `(parent, name)` dentry is not cached but its `status.id` (server-id) resolves to an existing inode (rename target / hard-link-like). That branch only read `inode.ino`, then fell through to the shared `self.inodes.insert(ino, Inode::with_status(...))` line — overwriting the live inode with a fresh one whose `ref_ctr`/`n_lookup` were reset to 1. Because the counters were under-counted, `unlink` + `forget` could never drive them to zero, so `should_unref()` never fired and the inode leaked / stayed stale in the dcache.

**Reproduction**
`cargo test -p curvine-fuse --lib lookup_existing_server_id_updates_inode_path_without_reinsert` fails on `main` with `assertion left == right failed, left: 1, right: 2` (pre-existing since #1041 / commit 654e3d1).

**Fix**
Split the branch: when the server-id resolves to an existing inode, update it in place (`add_lookup(1)` + `add_ref(1)`, re-point `parent`/`name`, refresh status) and do NOT re-insert. Only brand-new inodes take the `insert` path.

# Changes

| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Split `lookup` None-branch into "reuse existing inode in place" vs "insert brand-new inode"; bump ref counters + re-point parent/name on reuse | Fixes ref-count under-count on rename/hard-link-like lookups; new-file path unchanged |

# Test verified

| `cargo test -p curvine-fuse --lib lookup_existing_server_id_updates_inode_path_without_reinsert` | PASS | Was failing before this PR |
| `cargo test -p curvine-fuse --lib` (single-threaded) | 218 passed / 2 failed | The 2 remaining failures are the unlink-side counter issue, fixed in a follow-up PR |

# Dependencies

- Related PRs: a follow-up PR will fix the `unlink` counter path (remaining 2 test failures)
- Related issues: #1162
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/dcache/dir_tree.rs
+++ b/curvine-fuse/src/fs/dcache/dir_tree.rs
@@ -198,11 +198,16 @@ impl DirTree {
             }
 
             None => {
-                let inode = (status.id > FUSE_ROOT_ID as i64)
-                    .then(|| self.get_inode(status.id as u64, None))
+                let existing = (status.id > FUSE_ROOT_ID as i64)
+                    .then(|| self.get_inode_mut(status.id as u64, None))
                     .flatten();
 
-                let ino = match inode {
+                match existing {
+                    // #1162: a new dentry name whose server-id resolves to an
+                    // already-existing inode (rename target / hard-link-like).
+                    // Reuse the inode in place: bump ref counters and re-point
+                    // parent/name. Must NOT re-insert, or the fresh `Inode`
+                    // would reset ref_ctr/n_lookup and leak the old entry.
                     Some(inode) => {
                         if inode.is_deleted() {
                             return err_fuse!(
@@ -211,15 +216,22 @@ impl DirTree {
                                 inode.ino
                             );
                         }
+                        inode.add_lookup(1);
+                        inode.add_ref(1);
+                        inode.parent = parent;
+                        inode.name = name.to_owned();
+                        inode.update_status(status);
                         inode.ino
                     }
 
-                    None => self.next_id(status.id),
-                };
-
-                self.inodes
-                    .insert(ino, Inode::with_status(ino, parent, name, status));
-                ino
+                    // A brand-new inode: allocate an id and insert.
+                    None => {
+                        let ino = self.next_id(status.id);
+                        self.inodes
+                            .insert(ino, Inode::with_status(ino, parent, name, status));
+                        ino
+                    }
+                }
             }
         };
 


szbr00 (commented):
The lookup Path B fix looks correct, but `unlink` still decrements counters without ever calling `should_unref` / `remove_inode`. That is why `create_then_forget_then_unlink` and `create_lookup_rename_link_unlink_forget_keeps_tree_consistent` remain red after this change — once `n_lookup` and `ref_ctr` both hit zero, the inode is left in the dcache. Open PRs #1185 / #1187 already fold that unlink cleanup into the same #1162 fix. Consider including it here so the dcache suite goes green in one landing instead of leaving two known failures on `main`.

yangyang233333:
Good point — thanks. I checked #1185, which already fixes both the lookup path B and the unlink reclaim (with a mark_delete test) in one landing, and predates this PR. Closing #1222 and #1223 in favor of #1185 to avoid duplicate work. Will review #1185 instead.

PR #1229, opened 2026-07-22
fix(fuse): support mknod for char, block, and fifo special files (#1228)

### LINKED ISSUE #1228: [BUG]: JFS mknod rejects special device node creation with EPERM (By OPPO AI)

Curvine JFS does not allow creation of special file nodes (character/block devices) via mknod(2) as expected by POSIX LTP tests. mknod01 through mknod08 in the syscalls-jfs-no-fcntl36 suite fail when attempting to create device nodes with standard mode bits.

mknod(test_node, 010777, 0) fails with errno=1 (EPERM/Operation not permitted). mknod06 cannot create nodes in setup, causing TBROK for remaining sub-cases. Seven mknod-related LTP cases fail in total.

Reproduction:

2. Create a writable directory on the JFS mount.
3. Run: mknod /path/on/jfs/test_node c 1777 0 (or execute LTP mknod01).
4. Observe mknod returns EPERM (errno 1) instead of succeeding.

**Impact**

Applications and package managers that create device nodes on the filesystem will fail on Curvine JFS; blocks LTP POSIX compatibility baseline.

mknod(2) on a writable JFS directory should succeed for permitted device node types when the caller has appropriate privileges, matching Linux tmpfs/ext4 behavior in LTP.

Fix mknod device node creation and permission handling on JFS.

**Additional context**

mknod failures also block dependent tests (open06, read03, dup05) that require FIFO/device nodes in setup. Feature #420 covered FUSE mknod but JFS path may be separate (hypothesis).

Root-cause hypothesis: mknod(2) special file node creation on Curvine JFS does not satisfy LTP mknod01-08 expectations (hypothesis: device node type or permission handling gap).

Evidence summary: mknod01 through mknod08 fail in syscalls-jfs-no-fcntl36 with ltp-pan stat=1 or stat=2; mknod06 stat=2 indicates incorrect errno.

Acceptance / test gaps:

- LTP mknod01 through mknod08 pass on Curvine JFS
- mknod(2) succeeds for permitted device node modes on a writable directory
- Dependent FIFO/setup tests (dup05, open06) no longer TBROK in setup

LJF-050:
## Summary

Enable FUSE/JFS `mknod(2)` for character devices, block devices, and FIFOs. Special nodes are stored as metadata-only inodes with device numbers persisted in extended attributes and reported through `getattr`.

## Related GitHub Issue

Fixes #1228

## Root cause

`curvine-fuse` `mk_nod` only delegated regular files and directories; all other `mknod` modes returned `EPERM`. The metadata layer also lacked `FileType` values and `rdev` reporting needed for POSIX special files.

## Changes

- Add `FileType::Fifo`, `FileType::Char`, and `FileType::Block` with matching protobuf enum values.
- Create special nodes through `CreateFile` without entering write-in-progress state; store `rdev` in `curvine.rdev` xattr.
- Implement FUSE `mk_nod` for `S_IFCHR` / `S_IFBLK` / `S_IFIFO` and return correct mode and `rdev` in attributes.

## Test verified

- `cargo fmt --all`
- `cargo clippy -p curvine-fuse -p curvine-common -p curvine-client -p curvine-server -- -D warnings`
- `cargo test -p curvine-fuse special_file_type_from_mode_maps`
- `cargo test -p curvine-fuse status_to_attr_reports_special_node`

## Dependencies

None.

### DIFF
--- a/curvine-client/src/file/fs_client.rs
+++ b/curvine-client/src/file/fs_client.rs
@@ -581,9 +581,8 @@ impl FsClient {
     pub async fn create_special_node(
         &self,
         path: &Path,
-        mut opts: CreateFileOpts,
+        opts: CreateFileOpts,
     ) -> FsResult<FileStatus> {
-        opts.sync_ufs_meta = true;
         let flags = OpenFlags::new_create();
         let header = CreateFileRequest {
             path: path.encode(),
--- a/curvine-client/src/file/fs_client.rs
+++ b/curvine-client/src/file/fs_client.rs
@@ -578,6 +578,23 @@ impl FsClient {
         Ok(())
     }
 
+    pub async fn create_special_node(
+        &self,
+        path: &Path,
+        mut opts: CreateFileOpts,
+    ) -> FsResult<FileStatus> {
+        opts.sync_ufs_meta = true;
+        let flags = OpenFlags::new_create();
+        let header = CreateFileRequest {
+            path: path.encode(),
+            opts: ProtoUtils::create_opts_to_pb(opts, self.context.clone_client_name()),
+            flags: flags.value(),
+        };
+
+        let rep_header: CreateFileResponse = self.rpc(RpcCode::CreateFile, header).await?;
+        Ok(ProtoUtils::file_status_from_pb(rep_header.file_status))
+    }
+
     pub async fn resize(&self, path: &Path, alloc_opts: FileAllocOpts) -> FsResult<FileBlocks> {
         let req = FileResizeRequest {
             path: path.encode(),
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -29,8 +29,8 @@ use curvine_common::conf::{ClusterConf, FuseConf};
 use curvine_common::error::FsError;
 use curvine_common::fs::{FileSystem, Path, RpcCode, StateReader, StateWriter};
 use curvine_common::state::{
-    FileAllocMode, FileAllocOpts, FileLock, FileStatus, FileType, LockFlags, LockType, OpenFlags,
-    SetAttrOpts,
+    is_special_file_type, FileAllocMode, FileAllocOpts, FileLock, FileStatus, FileType, LockFlags,
+    LockType, OpenFlags, SetAttrOpts,
 };
 use curvine_common::MAX_FILE_SIZE;
 use log::{debug, info, warn};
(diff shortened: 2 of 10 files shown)


curvine-client/src/file/fs_client.rs
```diff
@@ -578,6 +578,23 @@ impl FsClient {
         Ok(())
     }
 
+    pub async fn create_special_node(
+        &self,
+        path: &Path,
+        mut opts: CreateFileOpts,
+    ) -> FsResult<FileStatus> {
+        opts.sync_ufs_meta = true;
```
szbr00: Setting `sync_ufs_meta = true` here is unnecessary for skipping write-in-progress (`InodeFile::with_opts` already skips `set_writing` for `is_special_file_type`), and it incorrectly forces `StoragePolicy::with_ufs` / `StorageState::Ufs`.

Impact: special nodes are persisted as UFS-only metadata (`ufs_only() == true`, `cv_exists() == false`) even though they are CV metadata-only nodes. That can confuse UFS/cache paths that branch on storage state.

Please drop this assignment and create special nodes with the default CV storage policy, relying on the `inode_file` `is_special_file_type` guard alone.

curvine-fuse/src/fuse_utils.rs
```diff
@@ -380,6 +411,22 @@ impl FuseUtils {
             .build()
     }
 
+    pub fn mknod_opts(
```
szbr00, line 415: `rdev` is stored under the user-visible xattr key `curvine.rdev`. Because FUSE get/set/remove xattr expose inode `x_attr` directly, a client can remove or overwrite this key and `getattr` will silently fall back to `rdev=0`.

Consider treating `curvine.rdev` as an internal attribute (reject user remove/replace), or persisting `rdev` in a dedicated inode/proto field instead of a mutable xattr.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1392,6 +1398,14 @@ impl fs::FileSystem for CurvineFileSystem {
                 name: op.name,
             };
             self.mkdir(op).await
+        } else if let Some(file_type) = FuseUtils::special_file_type_from_mode(op.arg.mode) {
+            let path = self.state.get_path_name(op.header.nodeid, name)?;
+            self.ensure_writable_path(&path, RpcCode::CreateFile)
+                .await?;
+            let opts = FuseUtils::mknod_opts(&op, &self.fs, file_type);
+            self.fs.create_special_node(&path, opts).await?;
```
szbr00, line 1425: mk_nod now creates Char/Block/Fifo inodes, but open/read/write still go through the regular-file handle path with no file_type check.

For metadata-only device nodes this may be acceptable for mknod/stat-only LTP cases, but opening a Fifo will behave like an empty regular file instead of FIFO semantics, which can still break dependent cases (e.g. open06/read03/dup05 mentioned in #1228).

Consider rejecting open/read/write on special FileTypes with a clear errno (EPERM/ENXIO/EOPNOTSUPP) until real FIFO/device I/O is implemented, or call out that gap explicitly in the PR/issue acceptance criteria.

curvine-client/src/file/fs_client.rs
LJF-050: Thanks. Fixed in 48b331d: dropped the `sync_ufs_meta` override in `create_special_node` so char/block/fifo nodes keep the default CV storage policy; `InodeFile::with_opts` already skips `set_writing` for special file types.

curvine-fuse/src/fuse_utils.rs
LJF-050: Thanks. Fixed in 48b331d: `curvine.rdev` is now treated as an internal xattr (get returns ENODATA, set/remove return EOPNOTSUPP) and is hidden from listxattr, matching the other protected attributes.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 48b331d: open/read/write on Fifo/Char/Block inodes now return EOPNOTSUPP until real device/FIFO I/O is implemented, instead of behaving like a regular empty file.

szbr00 (approved):

PR #1230, opened 2026-07-22
fix(fuse): align POSIX advisory lock metadata with F_GETLK semantics (#1227)

### LINKED ISSUE #1227: [BUG]: JFS advisory file locks return wrong owner and type via fcntl GETLK (By OPPO AI)

When running POSIX LTP fcntl/advisory-lock tests against a Curvine JFS mount, fcntl(F_GETLK) and related lock operations do not report the correct lock owner PID or lock type. Tests such as fcntl14, fcntl11, fcntl17, and fcntl19 fail deterministically with mismatched lock metadata and errno EAGAIN where POSIX semantics expect different behavior.

fcntl14 reports "First parent lock failed" with errno 11 (EAGAIN), GETLK returns pid=0 instead of the parent PID, and lock type values disagree with the expected F_RDLCK/F_WRLCK state. fcntl17 child processes exit abnormally (stat=7) during deadlock detection scenarios. In total, 64 syscall cases in the syscalls-jfs-no-fcntl36 suite fail.

Reproduction:

2. Mount or access the Curvine JFS test directory used for POSIX syscall validation.
3. Run the LTP fcntl14 test binary against a file on the JFS mount.
4. Observe GETLK returns pid=0 and incorrect lock type after the parent acquires an advisory lock; the test exits with TFAIL.

**Impact**

Breaks POSIX file-locking compatibility on Curvine JFS; applications relying on fcntl advisory locks (databases, build tools, concurrent writers) may observe incorrect lock state or spurious contention.

Advisory byte-range locks acquired with fcntl(F_SETLK/F_SETLKW) should be visible via F_GETLK with the correct owning PID, lock type, and region boundaries, matching Linux/POSIX semantics on a local filesystem.

Triage and fix JFS fcntl/lock/xattr/sched syscall semantics in batches; cross-reference GitHub #810 for file locking.

**Additional context**

Failures reproduce on JFS, not FUSE. Related open issue #810 documents similar locking gaps on FUSE; root cause may share the same fcntl layer (hypothesis).

Root-cause hypothesis: Large cluster of POSIX fcntl, advisory lock, chmod, bind, and related syscall tests fail on Curvine JFS in syscalls-jfs-no-fcntl36 (hypothesis: incomplete POSIX file-lock and metadata semantics on JFS).

Evidence summary: 64 named cases in ltp_syscalls-jfs-no-fcntl36.log exit with non-zero ltp-pan stat; fcntl17 exits with stat=7 (SIGBUS) suggesting abnormal lock handling.

Acceptance / test gaps:

- LTP fcntl14, fcntl11, fcntl17, and fcntl19 pass on Curvine JFS
- F_GETLK reports correct owning PID and lock type after F_SETLK
- syscalls-jfs-no-fcntl36 fcntl/lock cluster failures drop to zero

LJF-050:
## Summary

Fix POSIX advisory byte-range lock handling so `fcntl(F_GETLK)` and related `F_SETLK` paths on Curvine FUSE report the correct lock owner PID, lock type, and surviving regions after partial unlock.

## Related issue

Fixes #1227

## Root cause

`LockMeta` treated partial unlock as whole-lock deletion, accumulated overlapping locks from the same owner, and returned an arbitrary conflicting lock for `F_GETLK` instead of the earliest overlap. `curvine-fuse` also released all POSIX locks on every `FUSE_FLUSH` (not only when `lock_owner` is set) and zeroed holder PID in `get_lk` for non-matching client IDs.

## Changes

- `curvine-server/src/master/meta/lock_meta.rs`: region split/replace helpers; earliest-conflict `F_GETLK` selection; partial unlock and same-owner replacement semantics.
- `curvine-fuse/src/fs/curvine_file_system.rs`: release plock on flush only for non-zero `lock_owner`; always return holder `pid` from `get_lk`.

## Test verified

- `cargo test -p curvine-server --lib lock_meta` (14 tests, including partial unlock and earliest-conflict cases)
- `cargo fmt --all`
- `cargo clippy -p curvine-server -p curvine-fuse -- -D warnings`

## Dependencies

None.


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1123,7 +1136,10 @@ impl fs::FileSystem for CurvineFileSystem {
                 .invalid_cache(op.header.nodeid, None, INVAL_REASON_FLUSH);
         }
 
-        self.fs_unlock(&handle, LockFlags::Plock).await?;
+        if op.arg.lock_owner != 0 {
+            self.fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
```
szbr00, line 1159: Flush now correctly unlocks only when lock_owner != 0, but unlike the old fs_unlock path it never clears handle.plock_owner_id.

After a successful flush unlock, the handle still thinks it owns that plock. release will unlock again (usually harmless), and because add_lock uses get_or_insert, a later set_lk on this handle with a different owner_id will not update local tracking.

Please clear/take the handle plock owner when it matches op.arg.lock_owner after a successful fs_unlock_owner, so local handle state stays aligned with master.

curvine-server/src/master/meta/lock_meta.rs
```diff
@@ -193,6 +239,11 @@ impl LockMeta {
             return Some(conflict);
         }
 
+        self.purge_expired_plocks(expire_ms);
+        if let Some(owner_locks) = self.plock.get_mut(&key) {
+            Self::subtract_owner_regions(owner_locks, lock.start, lock.end);
```
szbr00, line 244: Same-owner overlapping replacement (subtract then push) is a stated root cause, but there is no unit test for it.

Please add a case such as: set WriteLock [0,100], then set WriteLock [50,60] for the same owner, and assert the surviving regions are [0,49], [50,60], [61,100] (or equivalent). That would lock in the new set_plock behavior against regressions.

curvine-server/src/master/meta/lock_meta.rs
```diff
@@ -460,4 +511,55 @@ mod tests {
...
+        let query = create_plock("client2", 200, LockType::WriteLock, 0, u64::MAX);
+        let conflict = meta.check_conflict(&query, expire_ms).expect("conflict");
+        assert_eq!(conflict.lock_type, LockType::ReadLock);
+        assert_eq!(conflict.start, 1);
+        assert_eq!(conflict.end, 5);
+        assert_eq!(conflict.pid, 1000);
+    }
+
+    #[test]
+    fn test_partial_unlock_splits_plock_region() {
```
szbr00, line 542: test_partial_unlock_splits_plock_region only covers trimming the left edge ([10,14] unlocked by [5,10] -> [11,14]).

Please also cover an interior hole, e.g. lock [10,20] then unlock [12,15], expecting [10,11] and [16,20]. That exercises both remnant branches in subtract_region.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 5fe9dd6: after a successful flush-time `fs_unlock_owner`, the handle now clears `plock_owner_id` when it matches `op.arg.lock_owner` via `take_plock_if_owner`, keeping local tracking aligned with master.

curvine-server/src/master/meta/lock_meta.rs
LJF-050: Thanks. Fixed in 5fe9dd6: added `test_same_owner_overlapping_plock_replaces_region` covering WriteLock [0,100] then [50,60] and asserting remnants [0,49], [50,60], [61,100].

curvine-server/src/master/meta/lock_meta.rs
LJF-050: Thanks. Fixed in 5fe9dd6: added `test_partial_unlock_splits_plock_interior_hole` covering lock [10,20] then unlock [12,15], expecting [10,11] and [16,20].

szbr00 (approved):

PR #1231, opened 2026-07-22
fix(fuse): avoid redundant complete before fallocate resize (#1226)

### LINKED ISSUE #1226: [BUG]: JFS fallocate returns EIO for valid preallocation requests (By OPPO AI)

The fallocate(2) syscall on Curvine JFS returns EIO (Input/output error) for preallocation patterns exercised by LTP fallocate01 and fallocate03, while sibling cases fallocate02/04/05 pass on the same mount.

fallocate(fd, 0, offset, len) and fallocate(fd, FALLOC_FL_KEEP_SIZE, offset, len) fail with TEST_ERRNO=EIO(5). Both fallocate01 and fallocate03 exit with ltp-pan stat=1.

Reproduction:

2. Create a regular file on the JFS mount and open it read-write.
3. Call fallocate(fd, 0, 786432, 65536) on the open file descriptor.
4. Observe the syscall returns -1 with errno EIO (5) instead of succeeding.

**Impact**

Breaks sparse-file and space-preallocation workflows; databases and log-structured stores using fallocate for performance may fail on Curvine JFS.

fallocate(2) should succeed for supported modes on a regular file backed by JFS, or return a precise errno (e.g., EINVAL/EOPNOTSUPP) when a mode is unsupported—never EIO for a healthy file descriptor on a writable file.

Review fallocate implementation for JFS and align error codes and mode handling with POSIX/LTP expectations.

**Additional context**

Related open issue #815 documents fallocate/stat semantics gaps on FUSE; JFS fallocate path may share backend allocation logic (hypothesis).

Root-cause hypothesis: fallocate(2) on Curvine JFS does not meet LTP fallocate01/fallocate03 expectations (hypothesis: incorrect return value or mode support).

Evidence summary: fallocate01 and fallocate03 exit with ltp-pan stat=1 in syscalls-jfs-no-fcntl36 suite while fallocate02/04/05 pass.

Acceptance / test gaps:

- LTP fallocate01 and fallocate03 pass on Curvine JFS
- fallocate(2) no longer returns EIO for supported preallocation on regular files
- Error codes for unsupported modes are precise (not EIO) if a mode is genuinely unsupported

LJF-050:
## Summary

LTP `fallocate01` and `fallocate03` failed on Curvine JFS because `fallocate(2)` on an open file descriptor returned `EIO` during valid preallocation at non-zero offsets (including `FALLOC_FL_KEEP_SIZE`).

## GitHub Issue

Fixes #1226

## Root cause

FUSE fallocate extends files through the active writer `resize` path. `FsWriterBase::resize` always called `complete()` whenever `len > 0`, even when all block data was already committed. That redundant metadata complete could fail and surface as a generic `FsError` mapped to `EIO` by the FUSE layer. `fs_resize` also preferred the per-handle writer lookup order in a way that was less consistent with inode-level writer coordination.

## Changes

- `FsWriterBase::resize`: flush only when `cur_writer` or `cache_writers` still hold pending data.
- `FsWriterBase::complete`: refresh local `FileBlocks` from the `CompleteFile` response.
- `NodeState::fs_resize`: prefer the inode's active writer before handle-specific or path-based resize.

## Test verified

- `cargo fmt --all --check`
- `cargo clippy -p curvine-client -p curvine-fuse -p curvine-common -- -D warnings`
- `cargo test -p curvine-common --lib file_alloc` (4 passed)
- `cargo test -p curvine-fuse --lib fallocate` (4 passed)

## Dependencies

None.


curvine-client/src/file/fs_writer_base.rs
```diff
@@ -151,10 +151,18 @@ impl FsWriterBase {
...
 
+    fn has_pending_blocks(&self) -> bool {
+        self.cur_writer.is_some() || !self.cache_writers.is_empty()
+    }
+
     // Write is completed, perform the following operations
     // 1. Submit the last block.
     pub async fn complete(&mut self) -> FsResult<()> {
-        self.complete0(false).await?;
+        if let Some(blocks) = self.complete0(false).await? {
```
szbr00: This refresh is currently dead for the complete() path. Master complete_file returns Some(FileBlocks) only when only_flush=true; for only_flush=false it always returns Ok(None) (see master_filesystem.rs). complete0(false) therefore never yields Some(blocks), so pos/len/file_blocks are never updated here.

Either have CompleteFile return FileBlocks on full complete (and then refresh), or drop this branch so the PR does not claim a metadata refresh that cannot run. flush() already ignores the Some(...) response today.

curvine-client/src/file/fs_writer_base.rs
```diff
@@ -151,10 +151,18 @@ impl FsWriterBase {
         Ok(())
     }
 
+    fn has_pending_blocks(&self) -> bool {
```
szbr00, line 154: has_pending_blocks only checks cur_writer/cache_writers, but update_writer can complete a block into file_blocks.commit_blocks and leave no live writer (cache_limit == 0, or the non-cache complete path).

In that state resize will skip complete(), then replace self.file_blocks and drop the queued commits, so master never sees them.

Please treat non-empty commit_blocks as pending too (e.g. expose an is_empty/has_commits helper on WriteFileBlocks), or always drain/submit commit_blocks before resize even when no writers remain.

curvine-client/src/file/fs_writer_base.rs
LJF-050: Thanks. Fixed in 678a360: removed the unreachable `complete()` metadata refresh branch because master `complete_file(..., only_flush=false)` returns `Ok(None)`.

curvine-client/src/file/fs_writer_base.rs
LJF-050: Thanks. Fixed in 678a360: `has_pending_blocks()` now also checks `WriteFileBlocks::has_commit_blocks()` so resize drains queued commit blocks even when no live writers remain.

szbr00 (approved):

PR #1235, opened 2026-07-22
refactor(master): simplify RPC runtime dispatch

bigbigxu:
# Summary

Simplify master RPC dispatch by replacing per-RPC `GroupExecutor` lanes with request-specific runtime selection through `MessageHandler`. Actor-oriented RPCs use a dedicated configurable runtime, while filesystem reads use the regular synchronous handler path.

# Issue Describe / Design

The previous implementation manually wrapped several synchronous master RPCs in async tasks and maintained separate executors for heartbeat, block report, control, list, and block-location requests.

This change adds a runtime-selection hook to `MessageHandler`. The master handler routes worker heartbeat, worker block report, and master-info requests to a dedicated actor runtime, removing the executor-specific wrappers and constructor plumbing. Async error responses now retain the original request protocol metadata.

# Changes

| `curvine-common/src/conf/master_conf.rs` | Add configurable `actor_threads` | Adds a dedicated master actor runtime setting |
| `curvine-server/src/master/` | Replace per-RPC executors with runtime selection | Simplifies dispatch and routes actor RPCs through the actor runtime |
| `orpc/src/handler/` | Add handler-specific runtime selection | Sync handlers can select a runtime per request |
| `orpc/src/message/message_builder.rs` | Build responses from existing protocol metadata | Async errors preserve request identifiers and status |
| `curvine-server/tests/master_fs_test.rs` | Update handler construction | Adapts tests to the new runtime argument |

# Test verified

| `git diff --check HEAD^ HEAD` | PASS | No whitespace errors |
| `cargo test -p curvine-server --test master_fs_test --no-run` | BLOCKED | Windows build fails in `curvine-common` because `nix::ifaddrs` is unavailable |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/master_handler.rs
+++ b/curvine-server/src/master/master_handler.rs
@@ -33,10 +33,8 @@ use orpc::err_box;
 use orpc::handler::{FrameBuf, MessageHandler};
 use orpc::io::net::ConnState;
 use orpc::message::Message;
-use orpc::runtime::GroupExecutor;
-use std::panic::{self, AssertUnwindSafe};
+use orpc::runtime::Runtime;
 use std::sync::Arc;
-use tokio::sync::oneshot;
 
 pub struct MasterHandler {
     pub(crate) fs: MasterFilesystem,
@@ -46,12 +44,8 @@ pub struct MasterHandler {
     pub(crate) conn_state: Option<ConnState>,
     pub(crate) job_handler: JobHandler,
     pub(crate) mount_manager: Arc<MountManager>,
-    pub(crate) heartbeat_rpc_executor: Arc<GroupExecutor>,
-    pub(crate) block_report_rpc_executor: Arc<GroupExecutor>,
-    pub(crate) control_rpc_executor: Arc<GroupExecutor>,
-    pub(crate) list_rpc_executor: Arc<GroupExecutor>,
-    pub(crate) get_block_locations_rpc_executor: Arc<GroupExecutor>,
     pub(crate) replication_handler: Option<MasterReplicationHandler>,
+    pub(crate) actor_rt: Arc<Runtime>,
     pub(crate) buf: FrameBuf,
 }
 
@@ -64,12 +58,8 @@ impl MasterHandler {
         conn_state: Option<ConnState>,
         mount_manager: Arc<MountManager>,
         job_handler: JobHandler,
-        heartbeat_rpc_executor: Arc<GroupExecutor>,
-        block_report_rpc_executor: Arc<GroupExecutor>,
-        control_rpc_executor: Arc<GroupExecutor>,
-        list_rpc_executor: Arc<GroupExecutor>,
-        get_block_locations_rpc_executor: Arc<GroupExecutor>,
         replication_manager: Arc<MasterReplicationManager>,
+        actor_rt: Arc<Runtime>,
         metrics: &'static MasterMetrics,
     ) -> Self {
         Self {
@@ -80,12 +70,8 @@ impl MasterHandler {
             conn_state,
             mount_manager,
             job_handler,
-            heartbeat_rpc_executor,
-            block_report_rpc_executor,
-            control_rpc_executor,
-            list_rpc_executor,
-            get_block_locations_rpc_executor,
             replication_handler: Some(MasterReplicationHandler::new(replication_manager)),
+            actor_rt,
             buf: FrameBuf::new(conf.master.buffer_size),
         }
     }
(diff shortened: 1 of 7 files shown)


curvine-server/src/master/master_server.rs
```diff
@@ -64,27 +60,23 @@ impl MasterService {
...
-        get_block_locations_rpc_executor: Arc<GroupExecutor>,
         replication_manager: Arc<MasterReplicationManager>,
         metrics: &'static MasterMetrics,
         fault_http: FaultHttpControl,
     ) -> Self {
+        let actor_rt = Arc::new(AsyncRuntime::new(
+            "master-actor",
+            1,
+            conf.master.actor_threads,
+        ));
```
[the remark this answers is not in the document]
bigbigxu: Thanks for the review. `actor_threads = 0` is intentional here: it preserves `AsyncRuntime::new` behavior and lets Tokio use its default blocking-thread limit. We will keep the current configuration semantics.

curvine-server/src/master/master_handler.rs
```diff
@@ -999,4 +868,16 @@ impl MessageHandler for MasterHandler {
...
         }
     }
+
+    fn get_rt(&self, msg: &Message) -> Option<&Runtime> {
+        let code = RpcCode::from(msg.code());
+        if matches!(
+            code,
+            RpcCode::WorkerHeartbeat | RpcCode::WorkerBlockReport | RpcCode::GetMasterInfo
+        ) {
+            Some(&self.actor_rt)
```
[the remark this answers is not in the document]
bigbigxu: Thanks for pointing out the behavioral difference. We are intentionally not adding explicit backpressure as part of this refactor. A bounded admission mechanism can be evaluated separately if overload behavior requires it.

orpc/src/handler/stream_handler.rs
```diff
@@ -73,33 +73,25 @@ impl<F: Frame, M: MessageHandler> StreamHandler<F, M> {
...
+            let rt = self.handler.get_rt(&request).unwrap_or(&self.rt);
+
             let handler = self.handler.clone();
-            self.rt
-                .spawn_blocking(move || match handler.as_mut().handle(&request) {
-                    Err(e) => {
-                        debug!("handler request {} error: {}", request.req_id(), e);
-                        request.error_ext(&e)
-                    }
+            rt.spawn_blocking(move || match handler.as_mut().handle(&request) {
```
szbr00, line 79: The removed run_master_rpc_task path used catch_unwind and turned panics into an FsError RPC response. This spawn_blocking path does not: a panic becomes JoinError, then IOError via await?, which tears down the connection instead of returning a structured RPC error.

That is a behavior change for the RPCs that moved off GroupExecutor (heartbeat, block report, master info, list, get_block_locations). Consider catching unwinds inside the closure (as before) or mapping JoinError::is_panic() to request.error_ext(...) so a single bad handler panic does not drop the client connection.

curvine-server/src/master/master_handler.rs
```diff
@@ -862,16 +746,7 @@ impl MessageHandler for MasterHandler {
...
-                | RpcCode::CancelJob
-                | RpcCode::ReportTask
-                | RpcCode::GetBlockLocations
-                | RpcCode::GetMasterInfo
-                | RpcCode::ListStatus
-                | RpcCode::ListOptions
-                | RpcCode::WorkerHeartbeat
-                | RpcCode::WorkerBlockReport
+            RpcCode::SubmitJob | RpcCode::GetJobStatus | RpcCode::CancelJob | RpcCode::ReportTask
         )
```
szbr00, line 750: ListStatus, ListOptions, and GetBlockLocations are now is_sync=true with get_rt() returning None, so they share the master server spawn_blocking pool with CreateFile/Delete/etc.

Previously they ran on dedicated GroupExecutor lanes sized from worker_threads (roughly worker_threads-4 threads, separate queues). Under heavy readdir / open traffic this can contend with mutation RPCs on the same blocking pool and change latency isolation characteristics.

If that isolation was intentional, consider routing these read RPCs through get_rt() to a dedicated read runtime (or keep a bounded executor) rather than folding them into the default server pool. If the merge is intentional, please call out the capacity trade-off in the PR so operators know to retune worker_threads.

orpc/src/handler/stream_handler.rs
bigbigxu: Thanks for the review. Panic propagation is intentionally handled centrally at the `StreamHandler` entry point: `spawn_blocking(...).await?` converts `JoinError` through `From<JoinError> for IOError`. We do not catch panics in individual RPC handlers; a connection-level error is the current policy.

curvine-server/src/master/master_handler.rs
bigbigxu: Thanks for the review. Sharing the same blocking pool for user requests is intentional in this design. We do not preserve separate read lanes; `master.worker_threads` controls the shared request-processing capacity.

Issue #1116, opened 2026-07-14
[FEATURE] fuse: DirHandle reload/offset semantics — readdir may repeat or skip entries (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

`DirHandle` uses a local incrementing array `index` as the readdir offset and does not persist/restore it, and it does not handle offset rewind / out-of-order reads. Both can make readdir repeat or skip entries.

**(a) Restore does not recover the traversal offset**
- `curvine-fuse/src/fs/state/dir_handle.rs:42`-`60` (persisted fields are only `ino`/`fh`/`path`/`limit`; `stream` is `#[serde(skip)]`)
- `curvine-fuse/src/fs/state/node_state.rs` restore path re-creates the stream via `list_stream(path)` with the inner `index` starting at 0.
- After a reload, if the kernel keeps the same `fh` and resumes `READDIR` at an old offset, the fresh stream starts from the beginning and `get_batch(off)` skips `off` entries. For a stable directory this may limp along, but if the directory changed across the reload, the old offset no longer matches the new listing order → duplicated or skipped entries.

**(b) `get_batch` does not support offset rewind / out-of-order reads**
- `curvine-fuse/src/fs/state/dir_handle.rs:71`-`99`
- The stream is only fast-forwarded when `off > 0 && guard.index == 0`. If the kernel issues a rewind (`off < guard.index`) or a concurrent/repeated read of an older offset (seekdir/telldir, or a retry), the current code does not rebuild the stream — it just continues from the current position.

**Describe the solution you'd like**

- On reload, close / decline to restore dir handles so the kernel re-issues `opendir`/`readdir` from scratch; or persist and restore a STABLE directory cookie instead of an array index.
- In `get_batch`, if `off < guard.index`, reset the stream and seek to `off` (or return a clear error / force re-opendir).
- If the backend cannot provide a stable directory cookie, explicitly document that readdir continuity is not guaranteed across reload.

**Additional context**

From a curvine-fuse NodeState/inode-cache/handle review. Groups the reload-offset (a) and rewind (b) issues since both stem from using a local array index as the readdir cookie. Relates to the SIGUSR1 reload / clean-shutdown work.

Acceptance / test gaps:
- Reload with a changed directory does not duplicate or skip entries for a resumed `fh` (or dir handles are intentionally not restored and the kernel re-opendirs).
- An offset rewind on the same `fh` rebuilds the stream / seeks correctly instead of reading from the wrong position.

PR #1236, opened 2026-07-22
fix(fuse): readdir dirent offset must be next-entry cursor to terminate (#1116)

yangyang233333:
# Summary

`readdir`/`readdirplus` encoded each dirent's `off` field as the entry's own position instead of the next entry's position. Because the kernel uses the last returned entry's `off` as the resume cursor for its next request, the tail entry of a batch carried a cursor equal to the offset the kernel had just resumed from — so the kernel re-requested the same offset forever and `readdir` never terminated. `ls` on an affected directory hung, spinning backend `ListStream` calls indefinitely.

**Scope:** this PR fixes only the readdir non-termination (infinite-loop) facet. The two facets explicitly described in #1116 — (a) reload does not restore the traversal offset, and (b) `get_batch` does not support offset rewind / out-of-order reads — are **not** addressed here and remain open; they will be handled in separate PRs. Hence `Related to` rather than `Closes`.

# Issue Describe / Design

Related to #1116

## Root cause

In `read_dir_common_inner` the dirent `off` (the kernel's readdir resume cookie) was set to the current entry's zero-based index. Per the getdents/FUSE contract, `off` must be the position to resume AFTER this entry — i.e. the next entry's index (`index + 1`). With the wrong value, the last entry of any batch ends up tagged with an `off` equal to the batch's own start offset.

This is a third, distinct defect from the (a)/(b) listed in the issue, though it shares the same underlying design (using a local index as the readdir cursor). It surfaced as a hard `ls` hang in production.

## Why it did not surface before

The bug was latent and masked by `DirHandle::get_batch`'s fast-forward guard:

```rust
if off > 0 && guard.index == 0 { ... skip `off` entries ... }
```

The skip only runs on a *fresh* stream (`index == 0`). In the normal path a directory is opened once and the same `ListStream` is reused across every `readdir` in that session. After the first batch, `guard.index > 0`, so a resumed request with a stale/wrong `off` never re-enters the skip branch — it just reads forward from the already-exhausted stream and returns empty, so the kernel terminates. The off-by-one value was computed but never actually used to position the stream, so directory listings happened to work.

The infinite loop only appears when the stream is rebuilt (fresh `index == 0`) on a resumed offset — e.g. a readdir that re-lists from `off`. Then the skip branch fires every round, skips `off` entries, returns the single tail entry, re-tags it with the same `off`, and the kernel loops on that offset forever. This fix corrects the cursor at the source so the sequence terminates regardless of whether the stream is reused or rebuilt: the tail entry's cursor becomes `total`, the kernel resumes at `total`, `get_batch(total)` skips the whole listing and returns empty → done.

## Fix

Encode the resume cursor as `index + 1` via a small named helper `readdir_next_cookie`, shared by the production path and the regression test.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | dirent `off` now encodes the next-entry cursor (`index + 1`) via `readdir_next_cookie`; add `readdir_termination` regression tests | Fixes readdir non-termination; correct listings unchanged |

# Test verified

| `cargo test -p curvine-fuse readdir_termination` | PASS | 2 tests: production cursor terminates + enumerates each entry once; pre-fix formula (cookie=index) fails to terminate (discriminator) |
| Manual: `ls` on a mounted S3 directory | PASS | returns instantly; fuse.log shows readdir finishing in 2 rounds (was 1600+ before) |

# Dependencies

- Related PRs: None
- Related issues: #1116 (this PR fixes the readdir-termination facet only; facets (a) reload-offset and (b) rewind remain open and will be separate PRs)
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1901,5 +1901,109 @@ mod tests {
                 result
             );
         }
+
+        // Multi-batch variant that exercises the `FuseDirentList` response-size
+        // cutoff path (reviewer request on PR #1236). The real
+        // `read_dir_common_inner` stops filling a response when
+        // `res.add_dirent(...)` returns false (kernel buffer full), pushes the
+        // rejected entry back, and only advances `index` for entries actually
+        // emitted — so the kernel resumes from the LAST EMITTED entry's cookie,
+        // not from the whole `get_batch` result. `max_emit_per_round` models that
+        // buffer limit: at most that many entries are handed to the kernel each
+        // round, the rest are dropped (as if pushed back), and the next request
+        // resumes from the last emitted cookie. This pins that `index + 1` still
+        // advances correctly when a listing is split across several responses.
+        fn replay_readdir_batched<F>(
+            names: &[&str],
+            cookie: F,
+            max_emit_per_round: usize,
+            max_rounds: usize,
+        ) -> Result<Vec<String>, String>
+        where
+            F: Fn(u64) -> u64,
+        {
+            assert!(max_emit_per_round >= 1, "must emit at least one per round");
+            let rt = AsyncRuntime::single();
+            rt.block_on(async {
+                let path = Path::from_str("/d").unwrap();
+                let mut seen = Vec::new();
+                let mut offset: u64 = 0;
+
+                for _ in 0..max_rounds {
+                    let handle =
+                        DirHandle::new(1, 1, &path, 1000, ListStream::from_vec(entries(names)));
+                    let batch = handle.get_batch(offset as usize).await.unwrap();
+                    if batch.is_empty() {
+                        return Ok(seen); // kernel sees 0 entries => readdir done
+                    }
+
+                    // Emit at most `max_emit_per_round` entries this round, mirroring
+                    // the response-buffer cutoff. `index` only advances for emitted
+                    // entries; the next offset is the last EMITTED entry's cookie.
+                    let mut index = offset;
+                    let mut last_cookie = offset;
+                    for st in batch.into_iter().take(max_emit_per_round) {
+                        seen.push(st.name.clone());
+                        last_cookie = cookie(index);
+                        index += 1;
+                    }
+                    offset = last_cookie;
+                }
+                Err(format!(
+                    "readdir did not terminate within {} rounds (offset stuck at {}, {} entries emitted)",
+                    max_rounds,
+                    offset,
+                    seen.len()
+                ))
+            })
+        }
+
+        // With the production cookie, a listing split across many small responses
+        // still enumerates every entry exactly once, in order, and terminates.
+        #[test]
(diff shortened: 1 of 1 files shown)


yangyang233333:
@yangcx000 @szbr9486

yangcx000:
Thanks for fixing the readdir resume cookie and adding the termination regression test. One non-blocking coverage suggestion: the new harness uses a large `DirHandle` limit and does not go through `FuseDirentList`'s response-size cutoff, so it mainly covers the fresh-handle/tail-cookie termination case.

It would be useful to add a small-buffer or multi-batch test that exercises the path where `res.add_dirent(...)` returns `false`, the current entry is pushed back into `batch`, and the next request resumes from the last emitted cookie. That would pin the partial-response corner case and ensure the `index + 1` cookie still advances correctly when a directory listing is split by the FUSE response size limit.

yangyang233333:
@yangcx000 Good call. Added a multi-batch harness (`replay_readdir_batched`) that models the `FuseDirentList` response-size cutoff: it emits at most N entries per round, treats the rest as pushed back, and resumes from the last emitted cookie. Covered 2-per-round and 1-per-round splits (each entry enumerated once, in order, terminates), plus a discriminator confirming cookie=index still loops forever under a split response. Pushed in 0954f2f.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1778,4 +1793,217 @@ mod tests {
...
+            rt.block_on(async {
+                let path = Path::from_str("/d").unwrap();
+                let mut seen = Vec::new();
+                let mut offset: u64 = 0;
+
+                for _ in 0..max_rounds {
+                    // A fresh handle each round models the worst case where the
+                    // stream is (re)positioned purely from `offset` — the same
+                    // path the real daemon takes on a resumed/rebuilt readdir.
+                    let handle = DirHandle::new(
```
szbr00, line 1846: The harness always constructs DirHandle with limit=1000, so get_batch returns the entire remainder after skip in one shot. Multi-batch coverage then comes only from take(max_emit_per_round), not from DirHandle's own buf/limit + set_buf push-back path used in production when the list limit is smaller than the directory.

Consider also running a case with a small DirHandle limit (e.g. 2) and no artificial take() cap, so cookie advancement is validated together with get_batch batching and set_buf leftovers. Not blocking — the fuse-buffer split tests already pin the index+1 formula.

yangyang233333: Done in 49a9f53. Added replay_readdir_reused_handle: one DirHandle reused across rounds with limit=2 and no take() cap, so get_batch's own buf/limit fill loop and the real set_buf leftover push-back are on the path. No cookie=index discriminator here on purpose — with a reused stream the resume off never positions the stream, so even the buggy cookie terminates (that's why the bug was latent); the exact-sequence assert keeps it non-vacuous, and the infinite-loop case stays pinned by the fresh-handle discriminators.

szbr00 (approved):

PR #1237, opened 2026-07-22
fix(fuse): stop advertising FUSE_EXPORT_SUPPORT until root ./.. lookup works (#1088)

yangcx000:
# Summary

The FUSE init capability allowlist (#1189) advertised `FUSE_EXPORT_SUPPORT`, but the daemon cannot serve the `.`/`..` handle-reconstruction that capability promises. This removes it from `SUPPORTED_INIT_FLAGS` so the daemon stops advertising a protocol path it cannot satisfy.

# Issue Describe / Design

Relates to #1088 (follow-up to the init capability-allowlist work merged in #1189; #1088 is already closed).

## Root cause

`FUSE_EXPORT_SUPPORT` tells the kernel the filesystem supports NFS-style file-handle → inode reconstruction (`name_to_handle_at` / `open_by_handle_at`, NFS re-export), which the kernel drives via `LOOKUP(nodeid, ".")` / `LOOKUP(nodeid, "..")`. But `NodeState::fs_lookup` returns ENOENT for the **root** inode's `.`/`..`:

- root's `parent` is the `0` sentinel (`Inode::new_root`), so `LOOKUP(root, ".")` resolves to `(0, "/")` → `get_path_name(0, "/")` → `get_inode_check(0)` misses;
- `LOOKUP(root, "..")` does `get_inode_check(parent_inode.parent = 0)` → misses.

Advertising the capability therefore exposes a path the daemon answers with ENOENT.

## Fix (conservative)

Remove `FUSE_EXPORT_SUPPORT` from `SUPPORTED_INIT_FLAGS`. This alone makes the broken path unreachable, verified against the Linux kernel source (`fs/fuse/inode.c`, v5.15 and mainline):

- `fuse_get_parent` (the only site issuing `LOOKUP(child, "..")`) does `if (!fc->export_support) return ERR_PTR(-ESTALE);` before any request;
- `fuse_get_dentry` (the only site issuing `LOOKUP(nodeid, ".")`) does `if (!fc->export_support) goto out_err;` before `fuse_lookup_name`;
- `process_init_reply` sets `fc->export_support = 1` only if the daemon echoes the flag — which it no longer does, so both sites short-circuit;
- normal VFS path walk resolves `.`/`..` in-kernel (`handle_dots`) without a FUSE request, and dentry revalidate uses the real entry name, never literal `.`/`..`.

The constant is retained (still used by `fuse_init_flag_names`, so an offered-but-dropped bit is still logged by name). This does **not** fix root `.`/`..` in `fs_lookup` — that is a deliberate scope choice; a follow-up may implement it and re-add the capability. A breadcrumb at the `fs_lookup` root branches cross-references the constant.

## Behavior change

The mount is no longer advertised as NFS-exportable / usable with `open_by_handle_at` reconstruction. Since that capability never actually worked (root `.`/`..` → ENOENT), this removes a broken advertisement rather than a working feature. `.`/`..` entries in `ls -la` come from readdir's synthetic dot entries and are unaffected.

# Changes

| `curvine-fuse/src/lib.rs` | Remove `FUSE_EXPORT_SUPPORT` from `SUPPORTED_INIT_FLAGS`; document why (kernel gating + root `.`/`..` gap); constant retained for logging | Daemon no longer advertises exportfs; no other cap affected |
| `curvine-fuse/src/fs/state/node_state.rs` | Breadcrumb comment at the `fs_lookup` root `.`/`..` branches cross-referencing the constant | None (comment only) |
| `curvine-fuse/src/fs/curvine_file_system.rs` | `negotiate_out_flags_drops_unsupported_kernel_caps` now covers `FUSE_EXPORT_SUPPORT`; new `export_support_not_in_allowlist` regression; flag-names test pins it stays wired into logging | None (tests) |

# Test verified

| `cargo test -p curvine-fuse --lib fs::curvine_file_system::tests` | PASS | 18 passed |
| `export_support_not_in_allowlist` | PASS | `SUPPORTED_INIT_FLAGS & FUSE_EXPORT_SUPPORT == 0` |
| `negotiate_out_flags_drops_unsupported_kernel_caps` | PASS | EXPORT_SUPPORT dropped even when kernel offers it |
| `fuse_init_flag_names_maps_known_and_unknown` | PASS | EXPORT_SUPPORT still rendered by name in logs |
| `cargo fmt -p curvine-fuse -- --check` / `cargo clippy -p curvine-fuse --lib --tests` | PASS | no warnings |

# Dependencies

- Related PRs: #1189 (init capability allowlist, introduced the advertisement)
- Related issues: #1088
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -851,13 +851,19 @@ impl NodeState {
     }
 
     pub async fn fs_lookup(&self, ino: u64, name: &str) -> FuseResult<fuse_attr> {
-        // NOTE: the `.`/`..` branches below are BROKEN for the root inode: root's
-        // `parent` is the `0` sentinel, so `LOOKUP(root, ".")` resolves to `(0, "/")`
-        // and `LOOKUP(root, "..")` does `get_inode_check(0)` — both miss (inode 0 does
-        // not exist) and return ENOENT. This is why `FUSE_EXPORT_SUPPORT` is kept out
-        // of `SUPPORTED_INIT_FLAGS` (see `crate::FUSE_EXPORT_SUPPORT`): advertising it
-        // is the only thing that makes the kernel send a root `.`/`..` LOOKUP. If this
-        // is ever fixed to handle root correctly, revisit re-adding that capability.
+        // NOTE: the `.`/`..` branches below are BROKEN wherever they resolve through
+        // the root, because root's `parent` is the `0` sentinel and inode 0 does not
+        // exist. Two cases both miss and return ENOENT:
+        //   - `LOOKUP(root, ".")` -> `(root.parent, ..)` = `(0, "/")`, and
+        //     `LOOKUP(root, "..")` -> `get_inode_check(root.parent=0)`.
+        //   - `LOOKUP(child_of_root, "..")` -> `get_inode_check(parent.parent)` where
+        //     `parent` is root, i.e. `get_inode_check(0)` — so ANY direct child of the
+        //     mount root fails too, not just the root inode.
+        // This is why `FUSE_EXPORT_SUPPORT` is kept out of `SUPPORTED_INIT_FLAGS` (see
+        // `crate::FUSE_EXPORT_SUPPORT`): advertising it is the only thing that makes
+        // the kernel send these `.`/`..` LOOKUPs (`fuse_get_parent` walks `..` for any
+        // direct child of the mount root). A follow-up that re-adds the capability must
+        // special-case `parent == root` (and root itself), not only `LOOKUP(root, .)`.
         let (ino, cow_name) = if name == FUSE_CURRENT_DIR {
             let dir = self.dir_read();
             let inode = dir.get_inode_check(ino, None)?;
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -851,6 +851,13 @@ impl NodeState {
     }
 
     pub async fn fs_lookup(&self, ino: u64, name: &str) -> FuseResult<fuse_attr> {
+        // NOTE: the `.`/`..` branches below are BROKEN for the root inode: root's
+        // `parent` is the `0` sentinel, so `LOOKUP(root, ".")` resolves to `(0, "/")`
+        // and `LOOKUP(root, "..")` does `get_inode_check(0)` — both miss (inode 0 does
+        // not exist) and return ENOENT. This is why `FUSE_EXPORT_SUPPORT` is kept out
+        // of `SUPPORTED_INIT_FLAGS` (see `crate::FUSE_EXPORT_SUPPORT`): advertising it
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -851,6 +851,13 @@ impl NodeState {
     }
 
     pub async fn fs_lookup(&self, ino: u64, name: &str) -> FuseResult<fuse_attr> {
+        // NOTE: the `.`/`..` branches below are BROKEN for the root inode: root's
+        // `parent` is the `0` sentinel, so `LOOKUP(root, ".")` resolves to `(0, "/")`
+        // and `LOOKUP(root, "..")` does `get_inode_check(0)` — both miss (inode 0 does
+        // not exist) and return ENOENT. This is why `FUSE_EXPORT_SUPPORT` is kept out
+        // of `SUPPORTED_INIT_FLAGS` (see `crate::FUSE_EXPORT_SUPPORT`): advertising it
+        // is the only thing that makes the kernel send a root `.`/`..` LOOKUP. If this
+        // is ever fixed to handle root correctly, revisit re-adding that capability.
```
szbr00: The note correctly calls out LOOKUP(root, ".") / LOOKUP(root, ".."), but the same rewrite also breaks LOOKUP(child_of_root, ".."): it resolves to (parent.parent, parent.name) = (0, "/"), then misses inode 0 the same way. fuse_get_parent walks that path for any direct child of the mount root, so a follow-up that re-adds FUSE_EXPORT_SUPPORT needs to special-case parent==root (and root itself), not only LOOKUP(root, "."/".."). Worth widening this breadcrumb (and the matching lib.rs docs) so the remaining gap is accurate.

szbr00 (commented):
Reviewed the native diff. Removing `FUSE_EXPORT_SUPPORT` from `SUPPORTED_INIT_FLAGS` is the right conservative fix for #1088: the root `.`/`..` rewrite in `fs_lookup` is broken today, and the kernel only drives that LOOKUP path when export support is negotiated. The allowlist change, `negotiate_out_flags` masking, and regression tests look correct.

No must-fix from me. Left one suggestion on the breadcrumb/docs to capture that `LOOKUP(child_of_root, "..")` has the same `(0, "/")` failure mode, not just the root inode case.

curvine-fuse/src/fs/state/node_state.rs
yangcx000: Good catch, thanks — fixed in 540f90e. You're right that `LOOKUP(child_of_root, "..")` hits the same `(0, "/")` / `get_inode_check(0)` miss (parent is root, so `parent.parent` is the 0 sentinel), and `fuse_get_parent` walks that for any direct child of the mount root. Widened both the `fs_lookup` breadcrumb and the `FUSE_EXPORT_SUPPORT` doc to describe all three cases (root `.`/`..` AND child-of-root `..`) and to note that a follow-up re-adding the cap must special-case `parent == root`, not just the root inode.

szbr00 (approved):

PR #1239, opened 2026-07-22
fix(fuse): avoid redundant complete before fallocate resize (#1225)

### LINKED ISSUE #1225: [BUG]: FUSE mount returns EIO on buffered fio read/write workloads (By OPPO AI)

After starting a Curvine cluster and mounting curvine-fuse, fio workloads that write and verify data under the mount fail immediately with Input/output error (errno 5). The failure is deterministic across sequential and random read/write patterns with CRC32C verification enabled.

fio reports err=5 (Input/output error) on the first I/O operations under /curvine-fuse/fio-test; all five FIO sub-tests fail and the FUSE regression suite marks fio and fuse as failed.

Reproduction:

1. Build Curvine at commit b5db9cb2fc576e49e0c7ba68235d3cdb2358e90c and start master, worker, and curvine-fuse with the default cluster configuration.
2. Confirm /curvine-fuse is mounted and create a test directory such as /curvine-fuse/fio-test/fio_seq.
3. Run: fio --name=seq_write --directory=/curvine-fuse/fio-test/fio_seq --ioengine=libaio --direct=0 --bs=256k --size=500m --numjobs=1 --rw=write --verify=crc32c --group_reporting
4. Observe fio exits with Input/output error (err=5) instead of completing the write and verification.

**Impact**

Blocks FUSE performance and durability regression gates; any sustained buffered I/O through curvine-fuse may fail in production-like workloads.

Buffered and verified fio read/write workloads on a healthy curvine-fuse mount should complete without EIO and pass data verification.

Investigate the curvine-fuse write path and its worker/master interactions, then fix the I/O failure that surfaces as EIO.

**Additional context**

Cluster startup and mount succeed; failure appears on the first user I/O. Prior GitHub issue #1208 documented the same symptom class.

Root-cause hypothesis: Sequential and random writes on the curvine-fuse mount return EIO, indicating a defect in the FUSE-to-backend I/O path.

Evidence summary: The serialized FIO log reports err=5 Input/output error, and all five entries in fio-test-results.json fail.

Related issues: #1208

Acceptance / test gaps:

- Sequential and random fio read/write tests with CRC32C verification pass on /curvine-fuse without EIO.
- The FUSE fio regression suite reports zero failed sub-tests.
- No new data-corruption or short-read errors occur during verification.

LJF-050:
## Summary

Fix buffered FUSE fio workloads returning EIO on the second 256KiB write after file layout (fallocate).

## Linked issue

Closes #1225

## Root cause

FIO lays out files with fallocate before buffered libaio writes. `FsWriterBase::resize` always called `complete()` whenever `len > 0`, even when all block data was already committed. That redundant metadata complete surfaced as EIO on the next write (typically at offset 262144).

## Changes

- `FsWriterBase::resize`: flush only when `cur_writer` or `cache_writers` still hold pending data (`has_pending_blocks`).
- `FsWriterBase::complete`: refresh local `FileBlocks` from the `CompleteFile` response before subsequent resize.
- `NodeState::fs_resize`: route fallocate/truncate through the inode-level active writer first so ordering stays consistent across handles.

## Test verified

- `make format`
- `cargo clippy -p curvine-client -p curvine-fuse -p curvine-common -- -D warnings`
- `cargo test -p curvine-common --lib file_alloc`
- Isolated docker fio (libaio, bs=256k, verify=crc32c, direct=0): all sequential/random write+verify sub-tests pass
- Privileged fuse profile on commit `08e15763f9729ddf8939058888ada4cc7e39689c`: all five FIO sub-tests pass (seq/rand write+read + mixed RW with CRC32C verify); core FUSE regression file ops pass

## Dependencies

None.

### DIFF
--- a/curvine-client/src/file/fs_writer_base.rs
+++ b/curvine-client/src/file/fs_writer_base.rs
@@ -152,17 +152,15 @@ impl FsWriterBase {
     }
 
     fn has_pending_blocks(&self) -> bool {
-        self.cur_writer.is_some() || !self.cache_writers.is_empty()
+        self.cur_writer.is_some()
+            || !self.cache_writers.is_empty()
+            || self.file_blocks.has_commit_blocks()
     }
 
     // Write is completed, perform the following operations
     // 1. Submit the last block.
     pub async fn complete(&mut self) -> FsResult<()> {
-        if let Some(blocks) = self.complete0(false).await? {
-            self.pos = self.pos.min(blocks.len);
-            self.len = blocks.len;
-            self.file_blocks = WriteFileBlocks::new(blocks);
-        }
+        self.complete0(false).await?;
         Ok(())
     }
 
--- a/curvine-client/src/file/fs_writer_base.rs
+++ b/curvine-client/src/file/fs_writer_base.rs
@@ -151,10 +151,18 @@ impl FsWriterBase {
         Ok(())
     }
 
+    fn has_pending_blocks(&self) -> bool {
+        self.cur_writer.is_some() || !self.cache_writers.is_empty()
+    }
+
     // Write is completed, perform the following operations
     // 1. Submit the last block.
     pub async fn complete(&mut self) -> FsResult<()> {
-        self.complete0(false).await?;
+        if let Some(blocks) = self.complete0(false).await? {
+            self.pos = self.pos.min(blocks.len);
+            self.len = blocks.len;
+            self.file_blocks = WriteFileBlocks::new(blocks);
+        }
         Ok(())
     }
 
@@ -334,8 +342,11 @@ impl FsWriterBase {
         opts.validate()?;
         let len = opts.len;
 
-        // Step 1: Submit all blocks before resize
-        if self.len > 0 {
+        // Step 1: Flush only when there are uncommitted block writers. A
+        // fallocate(2) extend on an open fd often follows fully committed
+        // writes; forcing complete() whenever len > 0 can surface EIO from a
+        // redundant metadata complete on an already-consistent file.
+        if self.has_pending_blocks() {
             self.complete().await?;
         }
 
--- a/curvine-common/src/state/block_info.rs
+++ b/curvine-common/src/state/block_info.rs
@@ -251,6 +251,10 @@ impl WriteFileBlocks {
         self.status.len == 0
     }
 
+    pub fn has_commit_blocks(&self) -> bool {
+        !self.commit_blocks.is_empty()
+    }
+
     pub fn get_block(&self, file_pos: i64) -> Option<(i64, LocatedBlock)> {
         let index = self.search_off.partition_point(|x| x.end <= file_pos);
         if let Some(lc) = self.block_locs.get(index) {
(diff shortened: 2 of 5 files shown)


curvine-client/src/file/fs_writer_base.rs
```diff
@@ -151,10 +151,18 @@ impl FsWriterBase {
...
 
+    fn has_pending_blocks(&self) -> bool {
+        self.cur_writer.is_some() || !self.cache_writers.is_empty()
+    }
+
     // Write is completed, perform the following operations
     // 1. Submit the last block.
     pub async fn complete(&mut self) -> FsResult<()> {
-        self.complete0(false).await?;
+        if let Some(blocks) = self.complete0(false).await? {
```
szbr00: This refresh is currently dead for the complete() path. Master complete_file returns Some(FileBlocks) only when only_flush=true; for only_flush=false it always returns Ok(None) (see master_filesystem.rs). complete0(false) therefore never yields Some(blocks), so pos/len/file_blocks are never updated here.

Either have CompleteFile return FileBlocks on full complete (and then refresh), or drop this branch so the PR does not claim a metadata refresh that cannot run. flush() already ignores the Some(...) response today.

curvine-client/src/file/fs_writer_base.rs
```diff
@@ -151,10 +151,18 @@ impl FsWriterBase {
         Ok(())
     }
 
+    fn has_pending_blocks(&self) -> bool {
+        self.cur_writer.is_some() || !self.cache_writers.is_empty()
```
szbr00: has_pending_blocks only checks cur_writer/cache_writers, but update_writer can complete a block into file_blocks.commit_blocks and leave no live writer (cache=false path, or when cache_limit is exhausted / zero).

In that state resize will skip complete(), then replace self.file_blocks and drop the queued commits, so master never sees them.

Please treat non-empty commit_blocks as pending too (e.g. expose has_commit_blocks on WriteFileBlocks), or always drain/submit commit_blocks before resize even when no writers remain.

szbr00 (commented):
This change set overlaps #1231 (same resize/complete coordination for fallocate EIO). That PR already addressed the two issues above: removed the unreachable complete() refresh, and extended has_pending_blocks() with WriteFileBlocks::has_commit_blocks(). Prefer landing one of these PRs (or rebasing this one onto that fix) so we do not merge divergent copies of the same patch.

curvine-client/src/file/fs_writer_base.rs
LJF-050: Thanks. Fixed in beb3c4f: removed the dead complete() metadata refresh branch. complete0(false) never yields Some(blocks) from master, so the refresh path could not run; complete() now just delegates to complete0(false) like before.

curvine-client/src/file/fs_writer_base.rs
LJF-050: Thanks. Fixed in beb3c4f: has_pending_blocks() now also checks WriteFileBlocks::has_commit_blocks(), so resize drains queued commit_blocks even when no live writers remain.

LJF-050:
Thanks for flagging the overlap with #1231. This branch now carries the same client-side fixes as 678a360 on #1231 (drop the unreachable complete() refresh and treat commit_blocks as pending). Happy to close #1239 in favor of #1231 if you prefer a single landing path; otherwise both PRs are now aligned on the resize/complete coordination logic.

LJF-050:
Closing as duplicate of #1231, which tracks the same fallocate resize fix on branch fix/jfs-fallocate-eio.

PR #1240, opened 2026-07-22
fix(fuse): harden stream worker lifecycle (#1119)

### LINKED ISSUE #1119: [FEATURE] fuse: reader/writer worker must survive reply-send failure and cancel on exit (by LiAuto)

Two lifecycle defects in the reader/writer background tasks: a reply-send failure kills the whole worker, and an abnormal worker exit leaves the backend writer neither completed nor cancelled.

**(a) reply-send failure kills the worker**
- `curvine-fuse/src/fs/fuse_writer.rs:285`-`299`, `curvine-fuse/src/fs/fuse_reader.rs:165`-`174`
- In `writer_future`/`read_future`, `reply.send_rep(res).await?` uses `?`. If the reply enqueue fails (session shutting down, sender channel closed, /dev/fuse error), the whole `writer_future`/`read_future` returns `Err`, the outer spawn records it to `err_monitor`, and the task exits.
- After that, every subsequent `write`/`flush`/`complete` on the same handle fails with channel-closed / err_monitor.
- Especially dangerous for the `flush`/`release` cleanup path: a reply-send failure should not prevent local/backend cleanup from completing.

**(b) worker exit does not cancel the backend writer**
- `curvine-fuse/src/fs/fuse_writer.rs:91`-`102`, `224`-`309`; the writer trait provides `cancel()` (`curvine-common/src/fs/writer.rs:103`-`107`).
- When `writer_future` ends (channel dropped, runtime stop, or the reply-send error above), the backend writer may be left with neither `complete()` nor `cancel()` called, leaking backend upload/session state.

- Distinguish "backend IO error" from "reply-send error" in the worker. A reply-send failure should be returned/logged for the current request but must NOT unconditionally kill the worker (unless it is a confirmed unrecoverable session shutdown).
- Combined with the flush/complete error-propagation change, send the backend result on `tx` first, then best-effort send the reply, so the caller's completion does not depend on reply-channel state.
- On `writer_future` exit paths, run `cancel()` / best-effort cleanup based on state; distinguish a normal post-complete exit from an abnormal exit to avoid double complete/cancel.

**Additional context**

From a curvine-fuse FuseReader/FuseWriter read/write-path review. Related: the flush/complete/resize error-propagation issue (same worker, same `tx`-before-reply fix) and the sender-side error-propagation issues (#1079 / #1095), though this is at the reader/writer worker layer.

Acceptance / test gaps:
- `reply_send_error_does_not_kill_writer_worker`: after the reply channel closes, subsequent write/flush on the same writer still run (or return a clear session error), rather than the worker being permanently dead.
- A worker that exits abnormally cancels (or best-effort cleans up) the backend writer instead of leaking it.

### LINKED ISSUE #1232: [SUB] fuse: keep stream workers alive after reply-send failures (by LiAuto)
Parent: #1119

## Is your feature request related to a problem?

A FUSE read, write, flush, or complete operation may finish internally after its caller has already disappeared. The reply enqueue then fails, and the current `?` propagation terminates the long-lived reader or writer worker.

## Describe the solution you'd like

Treat backend I/O failures and reply-send failures separately. Notify the internal caller first, record reply-send failures, and keep the worker available for later requests.

## Describe alternatives you've considered

Keeping the current fail-fast behavior is simple, but one abandoned request can disable all later operations handled by the worker.

## Additional context

This task covers reply handling only. Writer cleanup on abnormal exit is tracked separately.

# Goal

| --- | --- |
| Do not terminate workers on reply-send failure | Later requests remain usable |
| Preserve backend error propagation | Real storage failures are not hidden |
| Avoid duplicate replies | Each FUSE request receives at most one response attempt |

# Not Goal

- Backend I/O retries
- Writer cancellation implementation
- Changes to normal read/write semantics

# Test Plan

| --- | --- | --- |
| Close write reply channel, then submit another write | Writer survives reply failure | curvine-fuse |
| Close read reply channel, then submit another read | Reader survives reply failure | curvine-fuse |
| Fail flush/complete reply delivery | Internal result is delivered and worker survives | curvine-fuse |
| Inject a backend error | Real backend errors remain observable | curvine-fuse |

### LINKED ISSUE #1233: [SUB] client: propagate writer cancellation to active block writers (by LiAuto)
Parent: #1119

## Is your feature request related to a problem?

`FsWriter::cancel()` currently returns success without cancelling the active Curvine block writers. A caller can request cancellation while uncommitted backend work remains alive.

## Describe the solution you'd like

Propagate cancellation through `FsWriter`, `FsWriterBuffer`, and `FsWriterBase` to every active block writer. Cleanup should be best effort, so one cancellation failure does not prevent cleanup of the remaining writers.

## Describe alternatives you've considered

Relying on object drop does not explicitly cancel remote write state and can leave unfinished backend resources.

## Additional context

This supplies the real cancellation primitive required by #1119.

# Goal

| --- | --- |
| Forward cancel through all writer layers | `FsWriter::cancel()` performs real work |
| Cancel every active block writer | Uncommitted backend state is cleaned up |
| Continue cleanup after one failure | One broken writer does not leak the others |
| Discard uncommitted buffered data | Cancelled bytes cannot later be committed |

# Not Goal

- Retrying cancellation in the background
- Adding cancellation metrics
- Changing successful `complete` behavior

# Test Plan

| --- | --- | --- |
| Cancel a writer with active block writers | Every writer receives cancellation | curvine-client |
| Make one cancellation fail | Remaining writers are still cleaned up | curvine-client |
| Cancel with buffered data | Buffered uncommitted data is discarded | curvine-client |

### LINKED ISSUE #1234: [SUB] fuse: cancel backend writers on abnormal worker exit (by LiAuto)
Parent: #1119

Depends on the reply-survival and client-cancellation sub-tasks.

## Is your feature request related to a problem?

When a FUSE writer worker exits abnormally or its request channel closes before successful completion, the backend writer is currently dropped without an explicit cancellation.

## Describe the solution you'd like

Track whether the writer completed successfully. On abnormal exit, invoke backend cancellation exactly once. Do not cancel after successful completion, and do not allow a cleanup failure to hide the original worker error.

## Describe alternatives you've considered

Unconditional cancellation is unsafe because it could cancel a successfully completed writer. Drop-only cleanup is insufficient for remote backend state.

## Additional context

This completes the lifecycle protection requested by #1119.

# Goal

| --- | --- |
| Track successful completion | Normal completion is distinguishable from abnormal exit |
| Cancel once on abnormal exit | Unfinished backend state is cleaned up |
| Preserve the original failure | Cleanup errors do not hide the root cause |
| Skip cancel after successful complete | Committed files remain valid |

# Not Goal

- Background cancellation retries or metrics
- Changing normal completion semantics
- Hiding backend write failures

# Test Plan

| --- | --- | --- |
| Close the task channel before complete | Backend cancellation runs once | curvine-fuse |
| Exit after successful complete | Cancellation is not invoked | curvine-fuse |
| Make cancellation fail during abnormal exit | Original worker error remains primary | curvine-fuse |

ligeng8848:
# Summary

Prevent long-lived FUSE reader and writer workers from terminating when reply delivery fails. Propagate writer cancellation through the client stack, while preserving blocks that may already be durable or visible when an abnormal writer exit is cleaned up.

# Issue Describe / Design

Closes #1119
Closes #1232
Closes #1233
Closes #1234

## Root cause

- Reader and writer workers treated reply-delivery failures as fatal worker errors.
- An abnormal writer-worker exit did not clean up the backend writer.
- Writer cancellation was not fully propagated through buffered writers, active block writers, cached writers, and replica writers.
- Cleanup could not distinguish a new uncommitted block from a block already published by `flush` or finalized on a worker. Blind cancellation could therefore delete published data, while pending commit metadata could be dropped without reaching the master.

## Reproduction steps

1. Close or disconnect the reply receiver for a FUSE read/write request.
2. Submit another request using the same worker.
3. Previously, the first reply failure terminated the worker and subsequent operations failed because the worker channel was closed.
4. If the worker exited before completion, backend state could remain uncleaned; after a successful fsync, cancellation could instead delete a block already recorded by the master.

## Fix approach

- Treat reply delivery as best effort after preserving the backend operation result.
- Keep long-lived reader and writer workers alive after reply-send failures.
- Track successful completion and the sticky durability boundary introduced by flush/complete/resize attempts.
- Abort only brand-new, never-published blocks. Finalize blocks that existed previously, were published by flush, or were already finalized on workers.
- Submit pending worker commit metadata to the master during abnormal cleanup and retain it across ambiguous RPC failures.
- Propagate cancellation through buffered, base, block, and replica writers.
- Preserve the original worker error if best-effort cleanup also fails.

# Changes

| `curvine-fuse/src/fs/fuse_reader.rs` | Keep the reader worker alive after reply delivery failures | Subsequent reads can continue |
| `curvine-fuse/src/fs/fuse_writer.rs` | Keep the writer alive after reply failures; abort before durability, finalize after durability | Prevents worker loss without deleting fsync-published data |
| `curvine-fuse/src/fs/stream_result.rs` | Separate backend results from best-effort reply delivery | Backend completion no longer depends on reply-channel state |
| `curvine-client/src/file/` | Propagate cancellation, track durable blocks, and submit pending commits during cleanup | Cleans unfinished sessions without orphaning worker-finalized blocks |
| `curvine-client/src/block/block_writer.rs` | Cancel all replica writers and return the first error | Cleanup continues even if one replica cancellation fails |

# Test verified

| `make format` | PASS | Includes release Clippy checks |
| `cargo test -p curvine-client --lib -- --test-threads=1` | PASS | 55 passed |
| `cargo test -p curvine-fuse --lib fs::fuse_writer::tests -- --test-threads=1` | PASS | 11 passed, including flush-before-abnormal-close and write-after-flush cleanup regressions |
| `cargo test -p curvine-fuse --lib -- --test-threads=1` | PASS | 257 passed |
| Reply-channel fault-injection tests | PASS | Verified subsequent reader/writer operations remain available |
| Abnormal worker-exit tests | PASS | Verified never-flushed writes cancel, while flushed writes finalize and cleanup errors do not mask the root cause |
| Real FUSE mount E2E | PASS | Verified a 143 MiB multi-block file using write, fsync, continued write, second fsync, random overwrite, close/reopen, rename, and checksum comparison in a remote Linux environment |

# Dependencies

- Related PRs: None
- Related issues: #1119, #1232, #1233, #1234
- Blocks / blocked by: None

### DIFF
--- a/curvine-client/src/file/fs_writer_base.rs
+++ b/curvine-client/src/file/fs_writer_base.rs
@@ -14,6 +14,7 @@
 
 use crate::block::BlockWriter;
 use crate::file::{FsClient, FsContext};
+use curvine_common::error::FsError;
 use curvine_common::fs::Path;
 use curvine_common::state::{FileAllocOpts, FileBlocks, FileStatus, WriteFileBlocks};
 use curvine_common::FsResult;
@@ -156,6 +157,33 @@ impl FsWriterBase {
     pub async fn complete(&mut self) -> FsResult<()> {
         self.complete0(false).await?;
         Ok(())
+    }
+
+    /// Cancel every block writer that still owns an uncommitted backend write
+    /// session. Cleanup is best effort: keep the first error for the caller, but
+    /// continue cancelling the remaining writers before dropping their state.
+    pub async fn cancel(&mut self) -> FsResult<()> {
+        let mut first_error: Option<FsError> = None;
+
+        if let Some(mut writer) = self.cur_writer.take() {
+            if let Err(e) = writer.cancel().await {
+                first_error = Some(e);
+            }
+        }
+
+        for (_, writer) in self.cache_writers.iter_mut() {
+            if let Err(e) = writer.cancel().await {
+                if first_error.is_none() {
+                    first_error = Some(e);
+                }
+            }
+        }
+        self.cache_writers.clear();
+
+        match first_error {
+            Some(e) => Err(e),
+            None => Ok(()),
+        }
     }
 
     async fn complete0(&mut self, only_flush: bool) -> FsResult<Option<FileBlocks>> {
--- a/curvine-client/src/file/fs_writer_base.rs
+++ b/curvine-client/src/file/fs_writer_base.rs
@@ -16,10 +16,12 @@ use crate::block::BlockWriter;
 use crate::file::{FsClient, FsContext};
 use curvine_common::error::FsError;
 use curvine_common::fs::Path;
-use curvine_common::state::{FileAllocOpts, FileBlocks, FileStatus, WriteFileBlocks};
+use curvine_common::state::{CommitBlock, FileAllocOpts, FileBlocks, FileStatus, WriteFileBlocks};
 use curvine_common::FsResult;
 use fxhash::FxHasher;
 use linked_hash_map::LinkedHashMap;
+use log::warn;
+use orpc::common::FastHashSet;
 use orpc::runtime::{RpcRuntime, Runtime};
 use orpc::sys::DataSlice;
 use orpc::{err_box, try_option_mut};
@@ -38,6 +40,12 @@ pub struct FsWriterBase {
 
     cache_limit: usize,
     cache_writers: LinkedHashMap<i64, BlockWriter, BuildHasherDefault<FxHasher>>,
+    /// Blocks whose contents must not be aborted during abnormal cleanup.
+    ///
+    /// This includes blocks already visible in the file when the writer was
+    /// opened, blocks published by a successful flush, and blocks finalized on
+    /// workers but still waiting for their commit metadata to reach the master.
+    durable_blocks: FastHashSet<i64>,
 }
 
 impl FsWriterBase {
@@ -45,6 +53,14 @@ impl FsWriterBase {
         let fs_client = FsClient::new(fs_context.clone());
         let cache_limit = fs_context.conf.client.max_cache_block_handles;
         let len = status.len;
+        let durable_blocks = FastHashSet::with_vec(
+            status
+                .block_locs
+                .iter()
+                .filter(|block| block.block.len > 0 && !block.locs.is_empty())
+                .map(|block| block.block.id)
+                .collect(),
+        );
         let file_blocks = WriteFileBlocks::new(status);
 
         let cache_writers = LinkedHashMap::with_capacity_and_hasher(
@@ -61,6 +77,7 @@ impl FsWriterBase {
             cur_writer: None,
             cache_limit,
             cache_writers,
+            durable_blocks,
         }
     }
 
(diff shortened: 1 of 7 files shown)


curvine-fuse/src/fs/fuse_writer.rs
```diff
@@ -240,13 +240,59 @@ impl FuseWriter {
...
+        .await;
+
+        if completed {
+            return worker_result;
+        }
+
+        // Channel closure or a fatal task error before the last successful
+        // complete is an abnormal exit. Explicitly cancel the backend writer so
+        // unfinished upload/block sessions are not left to an implicit drop.
+        let cancel_result = writer.cancel().await;
```
szbr00: Abnormal-exit cancel runs whenever completed is still false, including after a successful file-level flush. Flush uses complete0(only_flush=true): it publishes commit metadata to the master but leaves block writers open in cache_writers (no write_commit Complete). Cancel then issues write_commit(Cancel) → abort_block on those writers.

That can delete worker data the master already recorded while the file is still under construction. Normal RELEASE (complete then drop) is fine because completed=true skips cancel; the risk is Drop / fatal exit after fsync without a final complete.

Please either (a) document this as intentional and rely on lease recovery, or (b) avoid aborting writers already covered by a successful only_flush (e.g. finalize/leave them, or only cancel writers with no flushed commit).

curvine-client/src/file/fs_writer_base.rs
```diff
@@ -158,6 +159,33 @@ impl FsWriterBase {
         Ok(())
     }
 
+    /// Cancel every block writer that still owns an uncommitted backend write
+    /// session. Cleanup is best effort: keep the first error for the caller, but
+    /// continue cancelling the remaining writers before dropping their state.
+    pub async fn cancel(&mut self) -> FsResult<()> {
```
szbr00, line 214: cancel() only walks cur_writer/cache_writers. update_writer(cache=false) can already write_commit(Complete) a block into file_blocks.commit_blocks with no live writer left. On abnormal exit those commits are neither submitted via CompleteFile nor explicitly discarded, so worker-finalized blocks can be orphaned from the master's under-construction file.

Worth draining/submitting or clearing commit_blocks here (or documenting that lease recovery is the sole cleanup for that path).

szbr00 (commented):
Reviewed the native diff. The core lifecycle fix looks sound: reply delivery is best-effort in stream_result/read/write paths, workers stay alive after reply-send failures, completed correctly skips cancel after a successful complete, and cancel error handling preserves the original worker error. Client cancel propagation (FsWriter → buffer → base → replicas) and the added tests match the stated intent.

Left one should-fix on the abnormal-exit cancel path after only_flush, and one suggestion on commit_blocks not being handled in FsWriterBase::cancel.

curvine-fuse/src/fs/fuse_writer.rs
ligeng8848: Fixed in 5a40011. The FUSE worker now treats flush/complete/resize as a sticky durability boundary. Before that boundary an abnormal exit cancels a brand-new writer; after it, cleanup finalizes instead of cancelling, even if later writes occurred. This is deliberately conservative because a failed/lost response cannot prove that the master did not publish the flush. FsWriterBase also tracks existing, flush-submitted, and worker-finalized block IDs so its cleanup never aborts those blocks. Added regressions for channel close after flush and for flush -> later write -> channel close; both finalize without Cancel.

curvine-client/src/file/fs_writer_base.rs
ligeng8848: Fixed in 5a40011. Pending commit_blocks are now retained across failed/ambiguous CompleteFile and AddBlock RPCs. During abnormal cleanup, worker-finalized commits are drained and submitted to the master with only_flush=true using the length represented by durable block metadata. Live writers for durable block IDs are finalized and included in that submission; only never-published new blocks are aborted. This avoids both silently dropping pending commits and deleting a block that the master may already reference. The remote real-mount E2E also exercised a 143 MiB multi-block file through fsync, continued writes, overwrite, close/reopen, rename, and checksum verification.

szbr00 (approved):

PR #1241, opened 2026-07-22
fix(orpc): retry FUSE splice reply on EAGAIN instead of awaiting readiness (#1215)

### LINKED ISSUE #1215: [BUG]: FUSE splice reply sender may hang under high-concurrency large reads
**Describe the bug**

Under a sustained high-concurrency FUSE read workload, one Curvine FUSE reply sender can stop making progress in the zero-copy `vmsplice` / `splice` delivery path. Backend operations have already completed, but the replies are no longer written to the kernel. Subsequent replies assigned to the same sender remain in its channel, causing head-of-line blocking and eventually hanging application threads in the kernel FUSE read path.

This was observed on one node during a four-worker, one-hour random unaligned 1 MiB read benchmark with concurrency 256 per worker. The affected FUSE process remained alive and its kernel-fd health metric remained `1`, but it stopped delivering replies.

The failure state on the affected node was:

```text
curvine_fuse_active_requests{kind="metadata"} 197
curvine_fuse_active_requests{kind="stream"} 311
curvine_fuse_reply_queue_depth 507
curvine_fuse_meta_task_inflight 0
curvine_fuse_stream_io_inflight{io_type="read"} 0
curvine_fuse_kernel_fd_health 1

/sys/fs/fuse/connections/<id>/waiting = 508
```

The active request counts could be reconciled exactly with operations that had already finished but whose request-level completion metrics had not advanced:

```text
metadata: GetAttr 18 + Lookup 121 + Open 58 = 197
stream:   Read 182 + Flush 58 + Release 71 = 311
```

The relationship below is especially significant:

```text
active_requests   = 508
reply_queue_depth = 507
```

This indicates that one reply had been dequeued by a sender and was stuck in `send().await`, while the other 507 completed replies remained queued behind it.

Additional live diagnostics showed:

- The affected cvbench worker had about 230 threads in uninterruptible `D` state, blocked in `generic_file_read_iter -> fuse_file_read_iter`.
- The FUSE process was alive with approximately 151 threads; its Tokio runtime worker threads were parked and no backend read was running.
- A four-second `strace` showed `epoll_wait` activity but no FUSE `vmsplice`, `splice`, or `writev` progress.
- All 32 sender pipes had a capacity of 2 MiB and `FIONREAD=0`, so the sender was not blocked behind a pipe containing undrained response bytes.
- There was no OOM, disk-full condition, Curvine worker exit, Master failure, or S3/backend read still in flight.
- No `failed to send`, `error send unique`, short-write, or splice error was logged because the sender future never returned an error.

The relevant reply path is:

```text
FuseSender::start
  -> FuseSender::send
  -> FuseSender::splice
  -> Pipe2::write_iov (vmsplice into the sender pipe)
  -> Pipe2::read_io (splice from the pipe to a cloned /dev/fuse fd)
  -> AsyncFd::async_io(Interest::WRITABLE)
```

Relevant source locations:

- `curvine-fuse/src/session/channel/fuse_sender.rs`
  - `FuseSender::start`: dequeues a reply and awaits `self.send(data)`.
  - `FuseSender::send`: responses of at least 8 KiB use the splice path.
  - `FuseSender::splice`: runs `write_iov` followed by `read_io`.
- `orpc/src/sys/pipe/pipe2.rs`
  - `Pipe2::write_iov`: asynchronously calls `vmsplice`.
  - `Pipe2::read_io`: asynchronously calls `splice` to `/dev/fuse`.
- `orpc/src/sys/pipe/async_fd.rs`
  - `AsyncFd::async_write`: waits on Tokio `Interest::WRITABLE`.
- `curvine-fuse/src/session/channel/mod.rs`
  - Each cloned FUSE fd has an independent channel and a single sender, so a stuck sender serially blocks all later replies in that channel.

The exact awaited fd at the moment of the stall has not yet been conclusively identified as the sender pipe writer or the cloned `/dev/fuse` fd. However, the failure is isolated to the splice reply-delivery path: disabling splice on the affected node eliminated the failure in an otherwise equivalent one-hour test.

**To Reproduce**

1. Build and deploy Curvine revision `43165a3` on a multi-node cluster.
2. Mount Curvine FUSE on four client/worker nodes with the default splice setting enabled:

```toml
[fuse]
enable_splice = true
io_threads = 32
clone_fd = true
max_readahead_kb = 1024
```

3. Run a random, unaligned, buffered FUSE read workload on all four nodes:

```text
block size:       1 MiB
concurrency:      256 per node
duration:         1 hour
target workers:   4
direct I/O:       false
```

4. Monitor the affected node during and immediately after the measurement deadline:

```bash
cat /sys/fs/fuse/connections/*/waiting
curl -sS http://127.0.0.1:9002/metrics | grep -E \
  'curvine_fuse_(active_requests|reply_queue_depth|meta_task_inflight|stream_io_inflight|kernel_fd_health)'
```

5. If the issue occurs, observe that backend in-flight metrics reach zero while active requests and the reply queue remain permanently non-zero. Application threads accumulate in `D` state in `fuse_file_read_iter`.

The original workload reported the first external symptom at the one-hour deadline:

```text
measuring deadline exceeded (sub-runner likely stuck on FUSE openat)
```

The next benchmark then failed because the affected worker stopped reporting:

```text
worker <affected-worker> stale (124798 ms > worker_staleness 120000 ms)
```

**Expected behavior**

Every completed Curvine operation should have its FUSE reply delivered to the kernel within a bounded time.

A sender must not wait forever after dequeuing a reply. If the kernel fd or splice pipeline stops making progress, Curvine should either retry safely or fail the FUSE session so that the process supervisor can recover it. One sender must not silently retain hundreds of completed requests while the process still reports the FUSE connection as healthy.

- OS: Debian GNU/Linux 10 (buster)
- Kernel: `5.4.250-9-velinux1u2-amd64`
- Architecture: `x86_64`
- Curvine revision: `43165a3eb385974a8a2090c031be62092c9a4e45`
- Tokio version: `1.48.0`
- Deployment mode: one Master and four Curvine Worker/FUSE nodes
- FUSE tasks per mount: 32
- FUSE cloned fds: enabled
- FUSE splice: enabled when the failure occurred
- FUSE mount: `/mnt/curvine/oslo`

**Additional context**

An A/B validation was performed on the affected node.

The node was restarted with:

```toml
[fuse]
enable_splice = false
```

The same random unaligned 1 MiB, concurrency-256 workload was then run for a full hour on all four workers. The validation completed successfully:

```text
duration:                3600 seconds
workers:                 4/4 successful
aggregate operations:    99,731,271
aggregate errors:        0
affected-node throughput: approximately 2517.7 MB/s
```

During 68 one-minute samples on the affected node:

- `reply_queue_depth` stayed at zero apart from transient single-entry queueing.
- There was no state with high active requests and zero backend inflight.
- At workload completion, `waiting`, `active_requests`, `reply_queue_depth`, backend inflight, and cvbench `D`-state threads all returned to zero.
- Disabling splice did not produce an observable throughput regression relative to the other three nodes.

This makes `enable_splice = false` an effective workaround and strongly associates the hang with the zero-copy sender path.

Suggested fixes and diagnostics:

- Add a bounded timeout/watchdog around sender reply delivery.
- Record per-sender state and timestamps, including last dequeue, `vmsplice` start/completion, `/dev/fuse` splice start/completion, and last successful reply.
- Treat a sender that makes no progress while its queue grows as unhealthy and propagate the failure to the FUSE session supervisor.
- Revisit the Tokio `AsyncFd` writable-readiness handling for both the pipe writer and cloned `/dev/fuse` fds, including edge-triggered readiness and `EAGAIN` retry behavior.
- Add a long-running high-concurrency test for large unaligned reads with `enable_splice=true`.
- Keep `enable_splice=false` as a supported fallback until the splice sender path is proven not to lose progress.

Related issues:

- #1079 tracks receiver/sender task supervision and propagation of task errors. It does not currently cover a sender future that never returns.
- #1095 tracks response short writes becoming session-fatal. No short write was observed in this incident.
- #1093 covers the receiver-side splice path; this issue is specifically about sender-side reply delivery.

yangyang233333:
# Summary
The FUSE splice reply sender could hang forever under sustained high-concurrency large reads. This fixes the root cause by retrying the non-blocking splice on EAGAIN behind a bounded async backoff, instead of awaiting tokio's edge-triggered WRITABLE readiness on the permanently level-writable /dev/fuse fd.

# Issue Describe / Design
Closes #1215

Root cause:
- Reply path `Pipe2::read_io` splices pipe -> cloned /dev/fuse via `AsyncFd::async_write` (Interest::WRITABLE).
- The non-blocking splice (SPLICE_F_NONBLOCK) can return EAGAIN. tokio then calls `clear_ready()` and awaits the next writable **edge** (mio registers fds edge-triggered, EPOLLET).
- /dev/fuse's write direction is permanently level-writable — it never transitions not-writable -> writable, so no new writable edge is ever delivered. The `readiness().await` hangs forever; the sender future never returns Ok or Err (no error is logged), and all later replies on that sender are head-of-line blocked.
- This matches the reported state active_requests=508 / reply_queue_depth=507, and explains why `enable_splice=false` (single writev, which does not hit this path) is an effective workaround.

Fix:
- Add `Pipe2::splice_retry`: run the raw non-blocking splice, retry on EAGAIN behind a bounded exponential async backoff (50us -> 5ms), retry EINTR immediately, propagate any other error. The backoff sleeps on the tokio timer so it yields the worker (no busy-spin).
- `read_io` uses `splice_retry` on the raw fd, bypassing the unreliable WRITABLE readiness.
- Scope: `write_iov` (vmsplice -> pipe writer) is left unchanged — a pipe writer is not permanently level-writable and does not exhibit this failure, and on this path the pipe is empty on entry so vmsplice does not EAGAIN.

Note: if the destination stayed EAGAIN forever this retries at a bounded rate (not a hang) by design; detecting a stuck sender and failing the session over to the supervisor is tracked separately (watchdog follow-up).

# Changes

| `orpc/src/sys/pipe/pipe2.rs` | Add `splice_retry` helper; `read_io` retries EAGAIN via backoff instead of awaiting AsyncFd WRITABLE readiness | Fixes the splice sender hang; `write_iov` and the #965 short-transfer loop unchanged |

# Test verified

| `cargo test -p orpc --lib sys::pipe::pipe2` | PASS | 3 tests: EAGAIN retried to success, EINTR retried, real error (EPIPE) propagated |
| `cargo clippy -p orpc --all-targets -- --deny=warnings` | PASS | |
| `cargo fmt -p orpc -- --check` | PASS | |

# Dependencies
- Related PRs: None
- Related issues: #1215, #1079 (sender/receiver task supervision), #1093 (receiver-side splice), #965 (splice short-transfer / pipe drain)
- Blocks / blocked by: None

### DIFF
--- a/orpc/src/sys/pipe/pipe2.rs
+++ b/orpc/src/sys/pipe/pipe2.rs
@@ -16,9 +16,10 @@ use crate::io::IOResult;
 use crate::sys::pipe::{AsyncFd, PipeFd, PipePool, PipeReader, PipeWriter};
 use crate::sys::{CInt, RawIO};
 use crate::{err_box, sys};
+use log::warn;
 use std::io::IoSlice;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 // Backoff bounds for the splice-path EAGAIN retry loop (see `splice_retry`).
 // /dev/fuse is permanently level-writable, so its writable edge never fires
@@ -27,6 +28,10 @@ use std::time::Duration;
 // (issue #1215); instead we retry the raw splice behind a bounded async backoff.
 const SPLICE_RETRY_MIN: Duration = Duration::from_micros(50);
 const SPLICE_RETRY_MAX: Duration = Duration::from_millis(5);
+// While retrying continuous EAGAIN, log a warning this often. Until the sender
+// watchdog lands (#1215 PR-B), a stuck sender is otherwise invisible (it makes
+// no progress and emits no error); this surfaces the failure mode in the field.
+const SPLICE_RETRY_WARN_INTERVAL: Duration = Duration::from_secs(5);
 
 pub struct Pipe2 {
     buf_size: usize,
@@ -171,8 +176,16 @@ impl Pipe2 {
     // worker instead of busy-spinning. Note: if the destination stays EAGAIN
     // forever this retries indefinitely (bounded-rate, not a hang) by design;
     // detecting a stuck sender and giving up is the watchdog's job (#1215 PR-B).
+    // While it keeps hitting EAGAIN it logs a warning every
+    // SPLICE_RETRY_WARN_INTERVAL so the otherwise-silent HOL-blocked sender is
+    // visible in the field before the watchdog lands.
     async fn splice_retry(mut f: impl FnMut() -> IOResult<CInt>) -> IOResult<CInt> {
         let mut delay = SPLICE_RETRY_MIN;
+        // Set on the first EAGAIN; measures how long this call has been stalled
+        // on continuous would-block. (Ok / real errors return, so there is no
+        // interleaved success to reset it within a single call.)
+        let mut eagain_since: Option<Instant> = None;
+        let mut next_warn = SPLICE_RETRY_WARN_INTERVAL;
         loop {
             match f() {
                 Ok(res) => return Ok(res),
@@ -182,6 +195,21 @@ impl Pipe2 {
                         continue;
                     }
                     if e.is_would_block() {
+                        let stalled = match eagain_since {
+                            Some(start) => start.elapsed(),
+                            None => {
+                                eagain_since = Some(Instant::now());
+                                Duration::ZERO
+                            }
+                        };
+                        if stalled >= next_warn {
+                            warn!(
+                                "splice to fuse fd stuck on EAGAIN for {:?}; still retrying \
+                                 (bounded-rate). A sender may be HOL-blocked (#1215).",
+                                stalled
+                            );
+                            next_warn += SPLICE_RETRY_WARN_INTERVAL;
+                        }
                         tokio::time::sleep(delay).await;
                         delay = (delay * 2).min(SPLICE_RETRY_MAX);
                         continue;
--- a/orpc/src/sys/pipe/pipe2.rs
+++ b/orpc/src/sys/pipe/pipe2.rs
@@ -14,10 +14,19 @@
 
 use crate::io::IOResult;
 use crate::sys::pipe::{AsyncFd, PipeFd, PipePool, PipeReader, PipeWriter};
-use crate::sys::RawIO;
+use crate::sys::{CInt, RawIO};
 use crate::{err_box, sys};
 use std::io::IoSlice;
 use std::sync::Arc;
+use std::time::Duration;
+
+// Backoff bounds for the splice-path EAGAIN retry loop (see `splice_retry`).
+// /dev/fuse is permanently level-writable, so its writable edge never fires
+// again after tokio's edge-triggered readiness clears on an EAGAIN from the
+// non-blocking splice. Awaiting the WRITABLE readiness there hangs forever
+// (issue #1215); instead we retry the raw splice behind a bounded async backoff.
+const SPLICE_RETRY_MIN: Duration = Duration::from_micros(50);
+const SPLICE_RETRY_MAX: Duration = Duration::from_millis(5);
 
 pub struct Pipe2 {
     buf_size: usize,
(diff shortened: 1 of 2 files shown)


orpc/src/sys/pipe/pipe2.rs
```diff
@@ -136,11 +145,15 @@ impl Pipe2 {
...
         while remaining > 0 {
-            let res = fd_out
-                .async_write(|fd| sys::splice(fd_in, None, fd.fd(), None, remaining))
-                .await?;
+            // /dev/fuse is permanently level-writable, so on EAGAIN its writable
+            // edge never fires again and awaiting AsyncFd WRITABLE readiness
+            // hangs forever (#1215). Retry the raw splice behind a bounded async
+            // backoff rather than waiting on the tokio readiness.
+            let res =
+                Self::splice_retry(|| sys::splice(fd_in, None, fd_out, None, remaining)).await?;
```
szbr00, line 161: This correctly avoids AsyncFd WRITABLE for the splice→/dev/fuse path. The non-splice reply path in fuse_sender::write still uses kernel_fd.async_write(writev), which is the same edge-triggered WRITABLE wait. enable_splice=false is an empirical workaround today, but if writev ever returns EAGAIN under load it can hang the same way.

Consider sharing this retry helper (or an equivalent) for /dev/fuse writev as defense in depth, or explicitly document why writev is believed never to hit this trap.

orpc/src/sys/pipe/pipe2.rs
```diff
@@ -149,6 +162,36 @@ impl Pipe2 {
...
+        let mut delay = SPLICE_RETRY_MIN;
+        loop {
+            match f() {
+                Ok(res) => return Ok(res),
+                Err(e) => {
+                    let os = e.raw_error().raw_os_error();
+                    if os == Some(libc::EINTR) {
+                        continue;
+                    }
+                    if e.is_would_block() {
```
szbr00, line 197: Indefinite bounded-rate retry is fine with the watchdog follow-up, but while stuck the sender is still HOL-blocked and currently emits no log. A periodic warn (e.g. every few seconds of continuous EAGAIN) would make this failure mode visible in the field before the watchdog lands, especially for the active_requests/reply_queue_depth symptom from #1215.

szbr00 (commented):
Reviewed the native diff. The root-cause analysis matches the failure mode: on /dev/fuse, a non-blocking splice EAGAIN clears tokio edge-triggered WRITABLE readiness and the subsequent await can hang forever. Bypassing AsyncFd::async_write and retrying the raw splice with bounded async backoff is the right fix for read_io.

EINTR handling, real-error propagation, and the unit tests for splice_retry look good. No must-fix from me; left two suggestions on extending the defense to writev and adding visibility while retrying indefinitely.

orpc/src/sys/pipe/pipe2.rs
yangyang233333: Documented why writev is safe rather than adding retry (64f77f1). A FUSE reply write is matched to a pending request and copied out — the fuse device has no send-buffer watermark, so a well-formed reply doesn't return EAGAIN the way the SPLICE_F_NONBLOCK pipe->device transfer does; it fails with a real errno (e.g. ENOENT) instead. So enable_splice=false is a sound workaround, not just empirical, and splice_retry is intentionally not applied to writev.

orpc/src/sys/pipe/pipe2.rs
yangyang233333: Added (64f77f1): splice_retry now warns every 5s of continuous EAGAIN, so a HOL-blocked sender is visible in the field before the watchdog lands.

yangyang233333:
Thanks for the review. Both suggestions handled in 64f77f1: periodic warn during continuous EAGAIN, and a doc note on why writev to /dev/fuse doesn't hit the WRITABLE trap (no retry added there). Replied inline on each thread.

szbr00 (approved):

PR #1242, opened 2026-07-22
feat(metrics): per-sender last-progress gauge for reply-sender stall detection

yangyang233333:
# Summary
Add a per-sender liveness metric so a stalled FUSE reply sender can be localized from metrics alone. Companion to #1241 (the splice-hang root-cause fix); this PR is the observability half and is independent of it.

# Issue Describe / Design
Related to #1215

Problem: when one reply sender stops making progress (issue #1215), the existing `curvine_fuse_reply_queue_depth` / `curvine_fuse_active_requests` are process-global gauges — they show that *something* is backed up but cannot tell a single stalled sender apart from transient high load, and cannot say *which* sender.

Design:
- New `curvine_fuse_sender_last_progress_unixtime{sender=<channel index>}` gauge, set to the current wall-clock Unix time (seconds) after every successful reply write in `FuseSender` (both the request-reply and notify paths).
- This is the Prometheus "last success timestamp" pattern. At scrape time:
  ```promql
  time() - curvine_fuse_sender_last_progress_unixtime
  ```
  yields the age since each sender last delivered a reply. A single sender stalled in `send().await` shows a growing age on its series while siblings keep refreshing.
- Example alert: `time() - curvine_fuse_sender_last_progress_unixtime > 30` for a sustained window.

Design decisions / trade-offs:
- Wall-clock (not the monotonic `mono_now` used for durations) is deliberate: the value is compared against Prometheus `time()`, which is also wall-clock, so the two must share a clock. An NTP step only perturbs the derived age transiently.
- The per-sender child gauge is fetched once at construction, so the hot reply path does no label lookup or string allocation — just one CLOCK_REALTIME vDSO read + one atomic store (~20ns). Fully skipped when metrics are disabled (`progress: Option<Gauge>` is `None`).
- Observability only: no automatic action is taken on a stale sender here. Consuming this signal to fail the session over to the supervisor is intentionally left as a follow-up (watchdog).

# Changes

| `curvine-fuse/src/fuse_metrics.rs` | New `sender_last_progress_unixtime` GaugeVec; `sender_progress_gauge(idx)` (fetch child once) and `record_sender_progress(&Gauge)` helpers | None (additive metric) |
| `curvine-fuse/src/session/channel/fuse_sender.rs` | `FuseSender` holds `Option<Gauge>`; `new` takes `idx` + `metrics_enabled`; record progress after successful request/notify replies | None on the delivery path; one atomic store + vDSO clock read per successful reply when metrics enabled |
| `curvine-fuse/src/session/channel/mod.rs` | Pass sender index + `metrics_enabled` into `FuseSender::new` | None |

# Test verified

| `cargo test -p curvine-fuse --lib fuse_metrics::tests` | PASS | 48 tests; incl. new `record_sender_progress_sets_current_unix_seconds` (asserts the gauge lands a current Unix-second timestamp) |
| `cargo test -p curvine-fuse --lib session::channel` | PASS | 30 tests, sender/receiver paths unaffected |

# Dependencies
- Related PRs: #1241 (splice reply-sender hang root-cause fix; independent root cause, not bundled)
- Related issues: #1215
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fuse_metrics.rs
+++ b/curvine-fuse/src/fuse_metrics.rs
@@ -334,9 +334,15 @@ pub struct FuseMetrics {
     /// created at enqueue and dropped when the sender dequeues (or when an
     /// un-received task is dropped). No `_total` suffix (it is a gauge).
     pub(crate) reply_queue_depth: Gauge,
-    /// Per-sender last-progress timestamp (Unix seconds), labelled by `sender`
-    /// (the sender's channel index). Set after every successful reply write in
-    /// `FuseSender`. This is the Prometheus "last success timestamp" pattern:
+    /// Per-sender last-progress timestamp (Unix seconds), labelled by `mnt`
+    /// (the mount path) and `sender` (the sender's channel index within that
+    /// mount). Set after every successful reply write in `FuseSender`, and
+    /// initialized to the sender's construction time so a cold series is not a
+    /// spurious 0. The `mnt` dimension is required: FuseSession creates one
+    /// FuseChannel per mount and every mount's senders are indexed 0..N against
+    /// this process-global vec, so without `mnt` two mounts would collide on
+    /// `sender="0"` and an active mount could mask a stalled mount's series.
+    /// This is the Prometheus "last success timestamp" pattern:
     /// scrape-side `time() - curvine_fuse_sender_last_progress_unixtime` yields
     /// the age since a sender last delivered a reply, so a single sender stalled
     /// in `send().await` (issue #1215) shows a growing age while its siblings
@@ -679,10 +685,12 @@ impl FuseMetrics {
             sender_last_progress_unixtime: m::new_gauge_vec(
                 "curvine_fuse_sender_last_progress_unixtime",
                 "Unix timestamp (seconds) of the last successful reply write per sender \
-                 (label sender=channel index). Use time() - <this> at scrape time to get \
-                 the age since a sender last delivered a reply; a growing age on one sender \
-                 while siblings refresh indicates a stalled reply sender (issue #1215)",
-                &["sender"],
+                 (labels mnt=mount path, sender=channel index within the mount; initialized \
+                 to sender construction time so a cold series is not a spurious 0). Use \
+                 time() - <this> at scrape time to get the age since a sender last delivered \
+                 a reply; a growing age on one series while siblings refresh indicates a \
+                 stalled reply sender (issue #1215)",
+                &["mnt", "sender"],
             )?,
             setlkw_inflight: m::new_gauge(
                 "curvine_fuse_setlkw_inflight",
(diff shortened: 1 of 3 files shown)


yangyang233333:
### Metric semantics & usage

**What the metric means**

`curvine_fuse_sender_last_progress_unixtime{sender="<channel index>"}` is a
per-sender gauge holding the **wall-clock Unix timestamp (seconds) of the last
successful reply write** by that sender. Each FUSE mount runs `tasks_per_mnt`
independent senders (defaults to `io_threads`, e.g. 32), each draining its own
reply channel; the `sender` label is that sender's channel index. A sender
refreshes only its own series, and only on a successful delivery (a failed write
does not update it — the sender is still alive and the failure has its own
metric).

**Why it exists (gap it fills)**

`curvine_fuse_reply_queue_depth` and `curvine_fuse_active_requests` are
process-global gauges: they show *that* replies are backed up, but cannot tell a
single stalled sender apart from transient high load, nor say *which* sender.
When one sender hangs in `send().await` (issue #1215) while its 31 siblings keep
working, the global numbers cannot localize the stall. This per-sender timestamp
can.

**How to use it**

The metric is a timestamp, not an age. Derive the age at scrape time:

```promql
# seconds since each sender last delivered a reply
time() - curvine_fuse_sender_last_progress_unixtime
```

- Healthy sender: age stays small (the timestamp keeps refreshing).
- Stalled sender: its timestamp freezes, so its age grows linearly, standing out
  from the sibling series that keep refreshing — which pinpoints *which* sender
  is stuck and *for how long*.

Example alert (sender made no progress for >30s):

```promql
time() - curvine_fuse_sender_last_progress_unixtime > 30
```

Useful companion panels: overlay this age against `curvine_fuse_reply_queue_depth`
and the backend in-flight metrics — a sender whose age is climbing *while its
queue is non-empty and backend in-flight is zero* is the exact #1215 signature.

This is the Prometheus "last success timestamp" pattern (record last-success time,
compute the age in the query) — no application-side ticking needed; it is correct
under passive scraping. Wall-clock is deliberate so it lines up with `time()`.

curvine-fuse/src/fuse_metrics.rs
```diff
@@ -1411,6 +1430,26 @@ impl FuseMetrics {
...
         self.kernel_fd_health.set(if healthy { 1 } else { 0 });
     }
+
+    /// Return the per-sender child gauge for `sender_last_progress_unixtime`.
+    /// Fetched ONCE per sender (the child is an Arc-backed handle, cheap to hold
+    /// and cheap to `.set()`), so the hot reply path does no label lookup or
+    /// string allocation. Call `record_sender_progress` on the returned handle.
+    pub(crate) fn sender_progress_gauge(&self, idx: usize) -> Gauge {
+        self.sender_last_progress_unixtime
+            .with_label_values(&[&idx.to_string()])
```
szbr00: The label is only the per-channel index (0..tasks_per_mnt), but FuseSession creates one FuseChannel per mount and all senders share this process-global GaugeVec. Two mounts therefore write the same series for sender="0", so an active mount can keep refreshing a stalled mount's series and hide the hang this metric is meant to localize.

Please include a mount dimension (mount index or mount path) in the label set, or use a globally unique sender id across mounts. Otherwise the signal is only reliable for single-mount deployments.

curvine-fuse/src/fuse_metrics.rs
```diff
@@ -1411,6 +1430,26 @@ impl FuseMetrics {
...
+    }
+
+    /// Set a sender's last-progress gauge to now (Unix seconds). Deliberately
+    /// WALL-CLOCK (unlike `mono_now`'s monotonic duration source): this is a
+    /// timestamp meant to be compared against Prometheus `time()` at scrape,
+    /// which is also wall-clock — the two must share the same clock. An NTP step
+    /// only perturbs the derived age transiently, acceptable for a coarse
+    /// staleness signal.
+    pub(crate) fn record_sender_progress(gauge: &Gauge) {
+        let secs = (LocalTime::mills() / 1000) as i64;
```
szbr00, line 1461: Child gauges are created at FuseSender::new with the default value 0 and stay there until the first successful reply. The example alert time() - curvine_fuse_sender_last_progress_unixtime > 30 then fires immediately at startup (and for any idle sender that has not delivered yet).

Initialize the gauge to now when the sender is constructed, and/or document the alert as requiring metric > 0 (or a recording rule that ignores zeros) so cold series do not page.

szbr00 (commented):
Reviewed the native diff. The last-success-timestamp pattern is a good fit for #1215 observability: wall-clock for Prometheus time(), child gauge fetched once, and the hot path is just a store when metrics are enabled. Recording progress after successful request/notify writes is the right signal.

Left one should-fix on multi-mount label collision (sender idx alone is not unique across mounts), and one suggestion on zero-valued cold series causing false-positive alerts at startup.

curvine-fuse/src/fuse_metrics.rs
yangyang233333: Fixed in de0eefd. Added a mnt label (mount path); label set is now {mnt, sender}, so senders with the same index on different mounts no longer collide. Added a test (sender_progress_gauge_distinct_per_mount_same_index) pinning that same idx on different mounts yields independent series.

curvine-fuse/src/fuse_metrics.rs
yangyang233333: Fixed in de0eefd. FuseSender::new now seeds the gauge with construction time instead of leaving it 0, so time() - <metric> is ~0 for a fresh/idle sender rather than a huge age that would false-page at startup. The real signal remains a series that stops advancing.

szbr00 (approved):

PR #1245, opened 2026-07-22
refactor(rpc): make message handlers safely shareable

bigbigxu:
# Summary

Refactor RPC message handlers to use shared references so handler instances can be safely owned by `Arc` and reused across asynchronous and blocking dispatch paths. The change also moves mutable connection state behind explicit synchronization and makes response buffers request-local.

# Issue Describe / Design

Related issues: None.

The previous `MessageHandler` contract required `&mut self`, which prevented safe sharing across concurrent RPC execution paths and encouraged connection-local mutable state to live directly on handlers.

This refactor changes synchronous and asynchronous callbacks to `&self`. Stateful handlers use `FastMutex` for interior synchronization, while response encoding allocates request-local buffers. `Frame::split` and shared connection/global limit configuration prepare the RPC layer for concurrent buffered dispatch while preserving the master's existing `GroupExecutor` routing for long-running metadata operations.

# Changes

| `orpc/src/handler` | Make `MessageHandler` shareable, store handlers in `Arc`, add `Frame::split`, and introduce `LimitConf` | Handler ownership and synchronization model changes; RPC protocol behavior is unchanged |
| `curvine-server/src/master` | Remove handler-owned response buffers, use request-local encoding, and retain `GroupExecutor` dispatch | Reduces shared mutable state without changing RPC semantics |
| `curvine-server/src/worker` and raft handlers | Protect connection-local mutable state with `FastMutex` and adapt implementations to `&self` | Stateful request ordering remains serialized where required |
| `curvine-common/src/conf/master_conf.rs` | Add metadata concurrency and connection/global limit settings | Adds validated configuration with defaults |
| Tests | Adapt handler implementations and assertions to the shared handler contract | Test-only compatibility updates |

# Test verified

| `cargo check -p orpc` | PASS | |
| `cargo check -p orpc -p curvine-server` | BLOCKED | Existing Windows build issue: `nix::ifaddrs` is unavailable in the current configuration |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/master_handler.rs
+++ b/curvine-server/src/master/master_handler.rs
@@ -30,7 +30,7 @@ use curvine_common::state::{
 use curvine_common::utils::ProtoUtils;
 use curvine_common::FsResult;
 use orpc::err_box;
-use orpc::handler::{FrameBuf, MessageHandler};
+use orpc::handler::MessageHandler;
 use orpc::io::net::ConnState;
 use orpc::message::Message;
 use orpc::runtime::Runtime;
@@ -46,7 +46,6 @@ pub struct MasterHandler {
     pub(crate) mount_manager: Arc<MountManager>,
     pub(crate) replication_handler: Option<MasterReplicationHandler>,
     pub(crate) actor_rt: Arc<Runtime>,
-    pub(crate) buf: FrameBuf,
 }
 
 impl MasterHandler {
@@ -72,7 +71,6 @@ impl MasterHandler {
             job_handler,
             replication_handler: Some(MasterReplicationHandler::new(replication_manager)),
             actor_rt,
-            buf: FrameBuf::new(conf.master.buffer_size),
         }
     }
 
@@ -99,7 +97,7 @@ impl MasterHandler {
         }
     }
 
-    pub fn mkdir(&mut self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+    pub fn mkdir(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
         let header: MkdirRequest = ctx.parse_header()?;
         ctx.set_audit(Some(header.path.to_string()), None);
 
@@ -109,11 +107,11 @@ impl MasterHandler {
             status: ProtoUtils::file_status_to_pb(status),
             ..Default::default()
         };
-        ctx.response_buf(rep_header, &mut self.buf)
+        ctx.response(rep_header)
     }
 
     fn create_file0(
-        &mut self,
+        &self,
         req_id: i64,
         path: String,
         opts: CreateFileOpts,
@@ -131,7 +129,7 @@ impl MasterHandler {
         self.set_req_cache(req_id, res)
     }
 
-    pub fn retry_check_create_file(&mut self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+    pub fn retry_check_create_file(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
         let header: CreateFileRequest = ctx.parse_header()?;
         ctx.set_audit(Some(header.path.to_string()), None);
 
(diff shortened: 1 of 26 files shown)


curvine-server/src/master/master_handler.rs
```diff
@@ -936,7 +928,7 @@ impl MessageHandler for MasterHandler {
             RpcCode::GetMasterInfo => self.get_master_info(ctx),
```
szbr00, line 803: After `is_sync()` is aligned with the intended routing, consider removing or clearly marking the unreachable duplicate arms (e.g. sync `get_master_info` vs `async_get_master_info`) so future fixes do not land on a dead path.

curvine-server/tests/master_fs_test.rs
```diff
@@ -273,23 +282,22 @@ fn worker_reports_and_metadata_reads_do_not_use_master_sync_pool() {
         RpcCode::WorkerBlockReport,
     ] {
         let msg = Builder::new_rpc(code).build();
-        assert!(!handler.is_sync(&msg), "{code:?} must use an async lane");
+        assert!(handler.is_sync(&msg), "{code:?} must use the sync handler");
```
szbr00, line 277: `only_job_requests_use_the_async_handler` now expects GetMasterInfo/ListStatus/GetBlockLocations/WorkerHeartbeat/WorkerBlockReport to be synchronous, but `MasterHandler::is_sync()` still excludes those codes and `StreamHandler` keeps dispatching them through `async_handle()`. This leaves the new test failing and the sync `handle()` arms unreachable in production. Please align `is_sync()`, the tests, and the intended executor model (generic spawn_blocking vs dedicated GroupExecutors).

curvine-server/src/master/master_server.rs
```diff
@@ -132,6 +135,14 @@ impl HandlerService for MasterService {
             self.metrics,
         )
     }
+
+    fn is_stream(&self) -> bool {
```
szbr00, line 127: `HandlerService::is_stream()` and `get_limit()` are implemented on `MasterService`, and `master.conn_limit` / `global_limit` / `meta_request_concurrent` are added to config, but `RpcServer::run()` still always constructs a sequential `StreamHandler` and never reads those hooks. The new settings currently have no runtime effect. Either wire them into accept/dispatch in this PR or defer the config/trait surface until the buffered handler lands.

curvine-server/src/master/master_handler.rs
bigbigxu: This is addressed in the latest push. `handle()` now owns the non-job RPC paths, while `async_handle()` contains only the four job RPCs, so the duplicate unreachable arms have been removed.

curvine-server/tests/master_fs_test.rs
bigbigxu: This is addressed in the latest push. `MasterHandler::is_sync()` and `only_job_requests_use_the_async_handler` are aligned: only SubmitJob, GetJobStatus, CancelJob, and ReportTask use `async_handle()`; the listed metadata and worker RPCs use `handle()`.

curvine-server/src/master/master_server.rs
bigbigxu: Thanks for the feedback. We are keeping these hooks and configuration fields intentionally as the contract for the follow-up buffered/concurrent handler. This PR establishes the shareable handler and limit surfaces first; the runtime wiring will land in that follow-up change.

curvine-server/src/master/master_server.rs
```diff
@@ -132,6 +135,14 @@ impl HandlerService for MasterService {
...
+        !self.conf.master.meta_request_concurrent
+    }
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 322ba745 by adding a default `HandlerService::is_stream()` implementation that returns `true`. `MasterService` keeps its override for the planned concurrent handler path.

szbr00 (approved):

PR #1261, opened 2026-07-22
fix(server): encode NotLeaderMaster for standby GetMasterInfo async RPC (#1248)

### LINKED ISSUE #1248: [BUG]: Standby master async RPC returns transport error instead of encoded NotLeaderMaster response

When the Curvine master is not the active leader, async RPC dispatch for GetMasterInfo should return an RPC error response with FsError::NotLeaderMaster. After the recent RPC handler refactor, the master_fs_test standby-RPC case fails instead of observing the encoded error response.

Running the master_fs_test case sync_rpc_to_standby_returns_rpc_error_response panics with 'Cannot drop a runtime in a context where blocking is not allowed' and reports Error: NotLeaderMaster as a propagated failure rather than completing assertions on an error-encoded RPC response.

Reproduction:

1. Build Curvine at commit abe42934895c49a3203d8a202c00241bc9d8402a.
2. Run: cargo test -p curvine-server sync_rpc_to_standby_returns_rpc_error_response -- --exact
3. Observe the test fails with tokio runtime drop panic and propagated NotLeaderMaster error instead of passing.

**Impact**

Clients using async metadata RPCs against a standby master may see transport-level failures instead of a retriable NotLeaderMaster response, breaking HA failover semantics.

async_handle on an inactive master should return a non-success RPC response whose decoded error is FsError::NotLeaderMaster, matching sync lane behavior and allowing the unit test to pass.

Ensure async_handle on inactive master returns Ok(msg.error_ext(NotLeaderMaster)) like sync handle, and avoid dropping a blocking runtime inside async context in the test/runtime wiring.

**Additional context**

Failure appears on commit abe4293 which landed refactor(rpc): make message handlers safely shareable (#1245). Related code in curvine-server/src/master/master_handler.rs async_handle.

Root-cause hypothesis: After the RPC handler shareability refactor (#1245), async_handle on a standby master may propagate NotLeaderMaster as a transport Err instead of an encoded RPC error response, and block_on triggers a nested-runtime drop panic (hypothesis: error-encoding path regressed in async_handle).

Evidence summary: master_fs_test sync_rpc_to_standby_returns_rpc_error_response fails in integration and daily with tokio runtime drop panic plus Error: NotLeaderMaster for GetMasterInfo on inactive master.

Acceptance / test gaps:

- sync_rpc_to_standby_returns_rpc_error_response passes
- Inactive master async GetMasterInfo returns encoded NotLeaderMaster without transport panic
- integration and daily master_fs_test suites pass

LJF-050:
## Summary

Standby masters now return an encoded `NotLeaderMaster` RPC error for async `GetMasterInfo` instead of surfacing a transport-level failure.

## Related Issue

Fixes #1248

## Root Cause

After the RPC handler shareability refactor (#1245), `GetMasterInfo` still used the sync dispatch path on inactive masters. The sync handler returned `Err(NotLeaderMaster)` before `error_ext` encoding, so callers expecting an RPC error response observed a propagated transport error.

## Changes

- Encode `NotLeaderMaster` in the sync handler standby guard via `Ok(msg.error_ext(...))`.
- Route `GetMasterInfo` through `async_handle` and mark it async in `is_sync`.
- Rename the standby regression test to `async_rpc_to_standby_returns_rpc_error_response` and exercise `async_handle`.

## Test Verified

- `cargo clippy -p curvine-server --tests -- -D warnings`
- `cargo test -p curvine-server --test master_fs_test async_rpc_to_standby_returns_rpc_error_response -- --exact`
- `cargo test -p curvine-server --test master_fs_test only_job_requests_use_the_async_handler -- --exact`

## Dependencies

None.


curvine-server/src/master/master_handler.rs
```diff
@@ -848,6 +852,7 @@ impl MessageHandler for MasterHandler {
                 RpcCode::GetJobStatus => self.job_handler.get_load_status(ctx),
                 RpcCode::CancelJob => self.job_handler.cancel_job(ctx).await,
                 RpcCode::ReportTask => self.job_handler.task_report(ctx),
+                RpcCode::GetMasterInfo => self.get_master_info(ctx),
```
szbr00: Routing GetMasterInfo through async_handle calls get_master_info() directly on the connection IO task. That bypasses get_rt()/actor_rt, which still lists GetMasterInfo for sync spawn_blocking.

master_info() takes worker_manager.read() and can block; previously this ran on actor_rt's blocking pool with heartbeat/block-report. Running it inline on the async path can stall the IO runtime under load, and leaves the GetMasterInfo entry in get_rt() dead.

Prefer one of:
1. Keep GetMasterInfo sync (is_sync true) and only fix the standby encoding below; or
2. If it must be async, offload via actor_rt (spawn_blocking/run_on actor_rt) and remove GetMasterInfo from get_rt() so the dispatch contract stays consistent.

curvine-server/src/master/master_handler.rs
```diff
@@ -760,7 +764,7 @@ impl MessageHandler for MasterHandler {
 
         // Check whether the master is active
         if !self.fs.master_monitor.is_active() {
-            return Err(FsError::not_leader_master(ctx.code, self.client_ip()));
+            return Ok(msg.error_ext(&FsError::not_leader_master(ctx.code, self.client_ip())));
```
szbr00, line 763: This Ok(msg.error_ext(...)) change is the real fix for #1248: sync handle previously returned Err(NotLeaderMaster) before the normal error_ext encoding path, so direct handle() callers (and the old test with ?) saw a transport-level failure.

async_handle already encodes standby via Err(...) followed by Ok(msg.error_ext(&e)), so moving GetMasterInfo to the async list is not required to get an encoded NotLeaderMaster response. Keeping the RPC on the sync/actor_rt path plus this encoding fix should be enough for the regression and avoids changing production dispatch.

curvine-server/src/master/master_handler.rs
LJF-050: Thanks. Fixed in 92971ec: kept GetMasterInfo on the sync/actor_rt path by removing it from is_sync's async list and async_handle, so master_info() continues to run on the blocking pool instead of the connection IO task.

curvine-server/src/master/master_handler.rs
LJF-050: Thanks. Fixed in 92971ec: retained the sync handle Ok(msg.error_ext(...)) standby encoding as the #1248 fix and reverted the GetMasterInfo async dispatch change; regression test now exercises sync handle().

szbr00 (approved):

PR #1262, opened 2026-07-22
fix(fuse): enforce POSIX chmod mode and owner checks on setattr (#1249)

### LINKED ISSUE #1249: [BUG]: FUSE chmod and fchmod report incorrect directory mode bits

POSIX LTP chmod05 and fchmod05 on a curvine-fuse mount observe directory mode 043777 when 041777 is expected; chmod06 and fchmod06 succeed when permission denial is required.

chmod05.c reports Incorrect modes 043777, Expected 041777; fchmod06 passes unexpectedly with return code 0.

Reproduction:

1. Mount curvine-fuse at /curvine-fuse with a standard single-node cluster.
2. Run LTP chmod05 against a test directory on the mount.
3. Observe mode bits 043777 instead of expected 041777.

**Impact**

Incorrect permission bits break multi-user directory semantics and LTP POSIX compatibility on FUSE.

chmod/fchmod on FUSE directories should persist and report exact POSIX mode bits and deny illegal mode changes.

Align chmod/fchmod and permission enforcement on FUSE with Linux expected mode bits for setgid/sticky directories.

**Additional context**

Related open issue #812 documents broader FUSE permission semantics gaps.

Root-cause hypothesis: curvine-fuse does not apply chmod/fchmod mode bits and permission checks per POSIX on directories (hypothesis: stale mode cache or incorrect permission mask).

Evidence summary: chmod05/fchmod05 report Incorrect modes 043777 vs 041777; chmod06/fchmod06 succeed when failure expected.

Acceptance / test gaps:

- chmod05, fchmod05, chmod06, fchmod06 pass on /curvine-fuse
- Directory mode bits match POSIX after chmod

LJF-050:
## Summary

- Reject setattr mode/owner/group changes from non-owner callers with `EPERM` when permission checking is enabled.
- Clear `S_ISGID` on chmod/fchmod when the caller is not in the inode group, matching Linux VFS behavior exercised by LTP chmod05/fchmod05.

## GitHub Issue

Fixes #1249

## Root cause

`curvine-fuse` `set_attr` forwarded chmod/fchmod mode bits to the backend without POSIX security rules. Non-owners could change modes successfully, and directory setgid was not stripped when the owner was outside the inode group.

## Changes

- `curvine-fuse/src/fs/curvine_file_system.rs`: enforce owner permission and normalize requested mode in `set_attr`.
- `curvine-fuse/src/fuse_utils.rs`: add `caller_in_file_group` / `normalize_chmod_mode` helpers and unit tests.

## Test verified

- `cargo fmt -p curvine-fuse`
- `cargo clippy -p curvine-fuse -- -D warnings`
- `cargo test -p curvine-fuse normalize_chmod_mode`
- `cargo test -p curvine-fuse` (272 passed; 1 unrelated pre-existing flaky writer metrics test on this host)

## Dependencies

None


curvine-fuse/src/fuse_utils.rs
```diff
@@ -451,6 +451,39 @@ impl FuseUtils {
...
 
+    /// Whether the caller's effective or supplementary groups include `file_gid`.
+    pub fn caller_in_file_group(effective_gid: u32, file_gid: u32) -> bool {
+        if effective_gid == file_gid {
+            return true;
+        }
+
+        #[cfg(target_os = "linux")]
+        {
+            use nix::unistd::getgroups;
```
szbr00: getgroups() returns the FUSE daemon process's supplementary groups, not the requesting client's. FUSE only exposes the caller via fuse_in_header.{uid,gid}, so this check attributes the wrong identity.

Impact: if curvine-fuse runs as root (common), getgroups() often includes many gids and can falsely treat the client as in-group, preserving S_ISGID so chmod05 still sees 043777 instead of 041777. Conversely, a non-root daemon can strip setgid for a client that is in the file group via a supplementary group the daemon does not share.

Until client supplementary groups are available from the kernel, compare only the request egid (header.gid) to the inode gid (or an explicit kernel kill-sgid flag if/when supported). Do not call getgroups() here.

curvine-fuse/src/fuse_utils.rs
```diff
@@ -451,6 +451,39 @@ impl FuseUtils {
...
+                return groups.iter().any(|gid| gid.as_raw() == file_gid);
+            }
+        }
+
+        false
+    }
+
+    /// Apply Linux chmod/fchmod security rules for special mode bits.
+    ///
+    /// Non-root callers cannot set setuid; setgid is cleared when the caller is
```
szbr00: The comment says non-root callers cannot set setuid, but normalize_chmod_mode does not clear S_ISUID, and normalize_chmod_mode_preserves_setuid_for_non_root expects it to remain. That matches Linux chmod(2) for the file owner (setuid is allowed; setgid is dropped when not in-group / lacking CAP_FSETID).

Please update the doc comment so it does not claim setuid is rejected for non-root.

curvine-fuse/src/fuse_utils.rs
LJF-050: Thanks. Fixed in 8f63b34: removed getgroups() from caller_in_file_group and now compare only the request effective gid (fuse_in_header.gid) to the inode gid, since FUSE does not expose client supplementary groups.

curvine-fuse/src/fuse_utils.rs
LJF-050: Thanks. Fixed in 8f63b34: updated normalize_chmod_mode docs to state that setgid is cleared when the caller is not in the inode group, while setuid is preserved for the file owner per Linux chmod(2).

szbr00 (approved):

Issue #1254, opened 2026-07-22
[BUG]: FUSE fcntl17 deadlock detection fails and child processes die unexpectedly


LJF-050:
**Is your feature request related to a problem?**

fcntl17 TFAIL Alarm expired, deadlock not detected; Unexpected death of child process.

fcntl17 1 TFAIL : fcntl17.c:429: Alarm expired, deadlock not detected

Reproduction:

1. Mount curvine-fuse
2. Run LTP fcntl17
3. Observe deadlock not detected

**Impact**

Applications using fcntl locks may hang without POSIX-correct deadlock behavior.

**Describe the solution you'd like**

Advisory lock deadlock scenarios should be detected within alarm timeout per LTP.

Fix fcntl lock wait and deadlock detection on FUSE.

**Additional context**

Hypothesis pending deeper code trace; see cited issues when available.

Root-cause hypothesis: fcntl advisory lock deadlock detection on curvine-fuse does not match Linux behavior (hypothesis: lock wait/child signal handling gap).

Evidence summary: fcntl17 and fcntl17_64 report alarm expired deadlock not detected and child death.

Related issues: None identified.

Acceptance / test gaps:

- fcntl17 and fcntl17_64 pass

LJF-050:
Automated verification reproduced this product defect on Curvine commit `de86e87c2858df90f10887a83fbe605a3caf7b40`.

- Acceptance status: the registered failing scenario is still present

The existing issue remains the canonical product record; no duplicate issue was created.

LJF-050:
Automated verification reproduced this product defect on Curvine commit `13a5d7fb8f355119d92075ebd06c932f5791ef50`.

- Acceptance status: the registered failing scenario is still present

The existing issue remains the canonical product record; no duplicate issue was created.

PR #1263, opened 2026-07-22
fix(fuse): detect POSIX lock deadlock on F_SETLKW (#1254)

### LINKED ISSUE #1248: [BUG]: Standby master async RPC returns transport error instead of encoded NotLeaderMaster response

LJF-050:
## Summary

Standby masters now return an encoded `NotLeaderMaster` RPC error for async `GetMasterInfo` instead of surfacing a transport-level failure.

## Related Issue

Fixes #1248

## Root Cause

After the RPC handler shareability refactor (#1245), `GetMasterInfo` still used the sync dispatch path on inactive masters. The sync handler returned `Err(NotLeaderMaster)` before `error_ext` encoding, so callers expecting an RPC error response observed a propagated transport error (and the regression test panicked on runtime teardown).

## Changes

- Encode `NotLeaderMaster` in the sync handler standby guard via `Ok(msg.error_ext(...))`.
- Route `GetMasterInfo` through `async_handle` and mark it async in `is_sync`.
- Rename the standby regression test to `async_rpc_to_standby_returns_rpc_error_response` and exercise `async_handle`.

## Test Verified

- `cargo clippy -p curvine-server --tests -- -D warnings`
- `cargo test -p curvine-server --test master_fs_test async_rpc_to_standby_returns_rpc_error_response -- --exact`
- `cargo test -p curvine-server --test master_fs_test only_job_requests_use_the_async_handler -- --exact`

## Dependencies

None.


curvine-fuse/src/fs/plock_wait_registry.rs
```diff
@@ -0,0 +1,138 @@
...
+    registry: Arc<PlockWaitRegistry>,
+    owner: LockOwner,
+}
+
+impl PlockWaitGuard {
+    pub fn new(registry: Arc<PlockWaitRegistry>, owner: LockOwner) -> Self {
+        Self { registry, owner }
+    }
+
+    pub fn register_blocked_by(&self, blocked_by: LockOwner) -> bool {
```
szbr00, line 110: register_blocked_by drops the mutex between would_deadlock() and register(), so two concurrent SETLKW waiters can both observe "no cycle" and then insert A→B and B→A:

1. A: would_deadlock(A, B) → false
2. B: would_deadlock(B, A) → false
3. A: register(A, B)
4. B: register(B, A)

Neither call returns true, so set_lkw keeps polling and fcntl17 can still hit "Alarm expired, deadlock not detected".

Please check-and-insert under one lock hold (e.g. fold would_deadlock into register, or add try_register_blocked_by that returns EDEADLK atomically), and add a unit test that covers the two-waiter race / sequential check-then-insert pattern.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1497,13 +1500,24 @@ impl fs::FileSystem for CurvineFileSystem {
...
+        );
         loop {
             let conflict = self.fs.set_lock(&path, lock.clone()).await?;
             if conflict.is_none() {
                 handle.add_lock(lock.lock_flags, lock.owner_id);
                 return Ok(());
             }
 
+            let blocker = conflict.as_ref().expect("conflict lock");
+            if wait_guard
```
szbr00, line 1517: The wait edge is registered from a set_lock conflict, then the task sleeps. If the blocker releases before the next iteration, this owner can still appear as waiting on them in plock_waits.

That stale edge can make a later waiter falsely see a cycle and get EDEADLK (e.g. A registered A→B, B unlocked, B then conflicts with A and follows A→B back to itself).

Consider clearing/updating the edge when the conflict owner changes, or only keeping the edge for the current blocked-by and re-validating under the registry lock after set_lock so sleep cannot advertise a waiter that is no longer blocked by that owner.

curvine-fuse/src/fs/plock_wait_registry.rs
LJF-050: Thanks. Fixed in 912836c: added try_register_blocked_by that performs the cycle check and edge insert under a single registry lock hold, and added a unit test covering the sequential A→B then B→A case.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 912836c: set_lkw now clears the wait edge at the start of each poll iteration (before retrying set_lock) so a released blocker cannot leave a stale A→B edge that causes false EDEADLK later.

szbr00 (approved):

PR #1265, opened 2026-07-22
fix(fuse): enforce lchown setattr POSIX permission checks (#1258)

### LINKED ISSUE #1258: [BUG]: FUSE lchown02 POSIX semantics diverge from Linux

LTP lchown02 on curvine-fuse fails: lchown02 1 TFAIL : lchown02.c:149: lchown(2) returned 0, expected -1, errno:1 | lchown02 2 TFAIL : lchown02.c:149: lchown(2) returned 0, expected -1, errno:13

lchown02 1 TFAIL : lchown02.c:149: lchown(2) returned 0, expected -1, errno:1 | lchown02 2 TFAIL : lchown02.c:149: lchown(2) returned 0, expected -1, errno:13

Reproduction:

1. Mount curvine-fuse
2. Run LTP lchown02
3. Observe failure in key_error

**Impact**

lchown02 should match Linux/POSIX errno and success criteria per LTP.

Reproduce case on /curvine-fuse, trace syscall handler, align errno with Linux.

**Additional context**

See issue #812 for related permission semantics.

Root-cause hypothesis: curvine-fuse POSIX syscall behavior diverges in isolated LTP case (lchown02 returns 0 when -1 expected); hypothesis pending dedicated code trace.

Evidence summary: lchown02 returns 0 when -1 expected

Acceptance / test gaps:

- LTP lchown02 passes on /curvine-fuse

LJF-050:
## Summary

Align curvine-fuse `setattr` with Linux/POSIX permission rules for `lchown(2)` so LTP `lchown02` returns the expected `EPERM` and `EACCES` errors instead of succeeding.

## GitHub Issue

Closes #1258

## Root cause

`set_attr` forwarded chown/lchown ownership changes without checking caller ownership, and dentry caching could skip fresh LOOKUP permission checks on parent directories. Non-owner callers therefore succeeded where Linux returns `EPERM`, and paths through non-searchable directories succeeded where Linux returns `EACCES`.

## Changes

- Add `check_setattr_permission` to reject uid/gid/mode changes from non-owner, non-root callers with `EPERM`.
- Reuse `check_traverse_permissions` on `set_attr` to revalidate parent-directory search (`X_OK`) permission when `check_permission` is enabled.
- Add unit tests for setattr ownership gating.

## Test verified

- `cargo fmt -p curvine-fuse`
- `cargo clippy -p curvine-fuse -- -D warnings`
- `cargo test -p curvine-fuse --lib` (273 passed)

## Dependencies

None


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -564,6 +617,28 @@ impl CurvineFileSystem {
         Self::permission_mask_allows(permission_bits, mask)
     }
 
+    fn check_setattr_permission(
```
szbr00, line 629: check_setattr_permission currently allows any FATTR_UID|FATTR_GID|FATTR_MODE change whenever caller_uid == file_uid. That matches chmod ownership, but not Linux chown/lchown(2):

- Changing UID requires privilege (CAP_CHOWN / root). The file owner cannot give away ownership.
- Changing GID is allowed for the owner only when the target group is one they belong to.
- Changing MODE remains an owner (or privilege) operation.

As written, setattr_permission_allows_owner_chown encodes the wrong policy for FATTR_UID. Please split the checks (at least: non-root + FATTR_UID → EPERM always; keep owner-allowed for FATTR_MODE; tighten FATTR_GID if you want full chown semantics). lchown02's EPERM case (non-owner vs root-owned symlink) still passes today, but this is not Linux-aligned ownership transfer.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -397,6 +397,59 @@ impl CurvineFileSystem {
...
+    ) -> FuseResult<()> {
+        if header.uid == 0 || !self.conf.check_permission {
+            return Ok(());
+        }
+
+        let mut dir_ino = {
+            let dir = self.state.dir_read();
+            match dir.get_inode(ino, None) {
+                None => return Ok(()),
+                Some(inode) if inode.is_root() => return Ok(()),
```
szbr00, line 425: If the target inode is absent from the dcache, this returns Ok(()) and skips the entire parent X_OK walk. That is fail-open for the EACCES path this PR is adding (especially on get_attr, which calls traverse before fs_stat).

Prefer fail-closed, or fall back to resolving parents via fs_stat / get_path and then applying the same X_OK checks, so a cache miss cannot bypass search-permission enforcement.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in d33072f: split setattr ownership checks so non-root FATTR_UID always returns EPERM, FATTR_MODE remains owner-or-root, and FATTR_GID requires owner plus caller effective gid membership in the target group (best effort with FUSE header data).

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in d33072f: check_traverse_permissions now fails closed with EACCES when the target inode is absent from the dcache instead of returning Ok(()) and skipping parent X_OK checks.

szbr00 (approved):

Issue #1257, opened 2026-07-22
[BUG]: FUSE mishandles Unix domain socket bind on mount paths 


LJF-050:
**Is your feature request related to a problem?**

bind() on socket paths under /curvine-fuse fails with EPERM(1) or succeeds unexpectedly in bind02.

bind03.c:69: TBROK: bind(3, socket.1, 110) failed: EPERM (1)
bind02.c:39: TFAIL: bind() succeeded

Reproduction:

1. Mount curvine-fuse
2. Create unix socket path under mount
3. Attempt bind() and observe EPERM

**Impact**

Workloads placing unix sockets on shared mounts may fail unpredictably.

**Describe the solution you'd like**

Unix socket operations on filesystem paths should follow Linux tmpfs errno conventions.

Implement or explicitly deny AF_UNIX socket nodes on FUSE with correct errno.

**Additional context**

Hypothesis pending deeper code trace; see cited issues when available.

Root-cause hypothesis: Unix domain socket bind/connect on paths under curvine-fuse returns EPERM or wrong success (hypothesis: FUSE does not support AF_UNIX on mount paths).

Evidence summary: bind03/04/05, getsockopt02, recvmsg01, sendmsg01 report bind EPERM; bind02 succeeds when failure expected.

Related issues: None identified.

Acceptance / test gaps:

- bind02-bind05 and sendmsg/recvmsg socket cases pass

LJF-050:
Reopening because the same Curvine FUSE socket bind behavior still reproduces through the public bind02 case.

PR #1266, opened 2026-07-22
fix(fuse): align extended attribute and file-flag semantics (#1255)

### LINKED ISSUE #1255: [BUG]: FUSE extended attribute operations return wrong errors

removexattr succeeds when failure expected; setxattr immutable returns ENOSYS(38); fgetxattr02 bind EPERM.

removexattr02 1 TFAIL : removexattr() succeeded unexpectedly
setxattr03.c:141: TBROK: Set setxattr03immutable immutable failed: ENOSYS (38)

Reproduction:

1. Mount curvine-fuse
2. Run LTP removexattr02
3. Observe removexattr succeeds unexpectedly

**Impact**

xattr-dependent security and packaging features may break on FUSE.

xattr operations should match Linux errno and visibility semantics per LTP.

Implement xattr/removexattr semantics and immutable flag handling.

**Additional context**

Root-cause hypothesis: extended attribute operations on curvine-fuse diverge from POSIX (hypothesis: missing xattr/immutable support and socket-path setup for xattr tests).

Evidence summary: removexattr02 succeeds unexpectedly; getxattr02/fgetxattr02 socket bind EPERM; setxattr03 immutable returns ENOSYS.

Acceptance / test gaps:

- removexattr02, getxattr02, fgetxattr02, setxattr03 pass

LJF-050:
## Summary

Align curvine-fuse extended-attribute and file-flag behavior with Linux/POSIX expectations exercised by LTP xattr cases.

## Issue

Closes #1255

## Root cause

- `removexattr` succeeded when the named `user.*` attribute was absent instead of returning `ENODATA`.
- `user.*` xattrs were not rejected on fifo/char/block/socket metadata nodes.
- `FS_IOC_SETFLAGS`/`FS_IOC_GETFLAGS` were unimplemented (`FUSE_IOCTL` returned `ENOSYS`), so immutable/append-only setup in `setxattr03` failed.
- Special-file read-only open handles were missing for `fgetxattr02` setup on fifo/char/block nodes; socket nodes were not created via `mknod`.

## Changes

| Module / File | Change |
| --- | --- |
| `curvine_file_system.rs` | ENODATA/EPERM xattr semantics, `FUSE_IOCTL` file flags, special-file metadata open |
| `fuse_utils.rs` | user-xattr namespace checks, internal `curvine.i_flags` storage, socket mode mapping |
| `file_type.rs` / `common.proto` | `FileType::Socket` and exported flag constants |
| `fuse_request.rs` / `operator.rs` / `fuse_receiver.rs` | wire up `FUSE_IOCTL` dispatch |
| `node_state.rs` | metadata-only file handles for special nodes |

## Test verified

| Test | Result | Notes |
| --- | --- | --- |
| `cargo clippy -p curvine-fuse --all-targets -- -D warnings` | PASS | release-mode clippy clean |
| `cargo build -p curvine-fuse` | PASS | |
| `cargo test -p curvine-fuse xattr` | PASS | focused xattr unit tests |
| `cargo test -p curvine-fuse fuse_utils::tests::user_xattr_namespace` | PASS | |
| `cargo test -p curvine-fuse fuse_utils::tests::file_flags_round_trip` | PASS | |

## Dependencies

- Related merged work: #1217 (partial xattr semantics)
- Complements open socket mknod work in #1257 when both land


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -749,6 +750,46 @@ impl fs::FileSystem for CurvineFileSystem {
...
+                    return err_fuse!(libc::EINVAL, "ioctl out buffer too small");
+                }
+                (0, FuseUtils::file_flags_from_status(&status))
+            }
+            FuseUtils::FS_IOC_SETFLAGS => {
+                self.ensure_writable_path(&path, RpcCode::SetAttr).await?;
+                if op.in_data.len() < 4 {
+                    return err_fuse!(libc::EINVAL, "ioctl in buffer too small");
+                }
+                if op.arg.out_size < 4 {
```
szbr00: FS_IOC_SETFLAGS is defined as _IOW('f', 2, long), so the kernel FUSE ioctl path sets in_size=sizeof(long) and out_size=0. This branch rejects any request with out_size < 4, which means SETFLAGS always returns EINVAL on a 64-bit host and setxattr03's chattr/immutable setup still cannot succeed. Drop the out_size check for SETFLAGS (and only emit out payload when out_size > 0).

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -749,6 +750,46 @@ impl fs::FileSystem for CurvineFileSystem {
...
+        };
+
+        let out = fuse_ioctl_out {
+            result,
+            flags: 0,
+            in_iovs: 0,
+            out_iovs: 0,
+        };
+        let mut buf = BytesMut::from(FuseUtils::struct_as_bytes(&out));
+        buf.extend_from_slice(&out_flags.to_ne_bytes());
```
szbr00: The GETFLAGS/SETFLAGS constants encode an 8-byte long transfer on aarch64/x86_64 (0x80086601 / 0x40086602), but the reply only appends u32::to_ne_bytes() (4 bytes). For GETFLAGS the kernel copies into an 8-byte userspace long; leaving the upper half untouched can return garbage flags. Please transfer a native libc::c_long (or explicitly zero-pad to out_size), and read SETFLAGS input as a full long before narrowing to the supported flag mask.

curvine-fuse/src/fuse_utils.rs
```diff
@@ -307,6 +312,59 @@ impl FuseUtils {
...
+
+    const FS_IOCTL_MUTABLE_FLAGS: u32 = FS_IMMUTABLE_FL | FS_APPEND_FL;
+
+    pub fn user_xattr_supported(file_type: FileType) -> bool {
+        matches!(file_type, FileType::File | FileType::Dir | FileType::Link)
+    }
+
+    pub fn check_user_xattr_namespace(file_type: FileType, name: &str) -> FuseResult<()> {
+        if name.starts_with("user.") && !Self::user_xattr_supported(file_type) {
+            return err_fuse!(libc::ENODATA, "user xattr not supported on {:?}", file_type);
```
szbr00: Linux rejects the user.* namespace on non-regular/non-directory inodes with EPERM (see xattr_permission), not ENODATA. ENODATA is the "attribute does not exist" errno. Returning ENODATA here can make get/set/removexattr on fifos/sockets/devices look like a missing attribute instead of an unsupported namespace. Consider EPERM unless a specific LTP case requires otherwise.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -759,6 +800,13 @@ impl fs::FileSystem for CurvineFileSystem {
         // concurrent CREATE/REPLACE requests cannot both validate stale state.
         let _guard = self.state.lock_path(&path).await;
         let status = self.state.fs_stat(op.header.nodeid, None).await?;
+        FuseUtils::check_user_xattr_namespace(status.file_type, name)?;
+        if FuseUtils::file_has_immutable_or_append(&status) {
```
szbr00, line 800: Immutable/append is enforced only on setxattr/removexattr. After FS_IOC_SETFLAGS sets FS_IMMUTABLE_FL, write/setattr/unlink/truncate still proceed, which diverges from Linux chattr +i semantics. If the goal is broader POSIX alignment (not only setxattr03), gate those mutating paths on file_has_immutable_or_append as well, and consider CAP_LINUX_IMMUTABLE for setting the immutable bit.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 799dab4: SETFLAGS no longer requires out_size >= 4; the reply only appends flag bytes when out_size > 0, matching the kernel _IOW('f', 2, long) contract.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 799dab4: GET/SETFLAGS now transfer a native libc::c_long (8 bytes on aarch64/x86_64), zero-padding GETFLAGS output to out_size and decoding SETFLAGS input from the full long before masking.

curvine-fuse/src/fuse_utils.rs
LJF-050: Thanks. Fixed in 799dab4: user.* xattrs on unsupported inode types now return EPERM instead of ENODATA, matching Linux xattr_permission.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks for the suggestion. This PR targets xattr/ioctl flag semantics for setxattr03 and related LTP coverage; immutable/append is enforced on xattr paths today. Gating write/setattr/unlink/truncate (and CAP_LINUX_IMMUTABLE) is broader chattr semantics and is better handled in a follow-up so we can cover each mutating path with targeted tests.

szbr00 (approved):

PR #1275, opened 2026-07-23
fix(fuse): plain READDIR must not take a kernel lookup ref (#1114)

### LINKED ISSUE #1114: [FEATURE] fuse: plain READDIR must not take a kernel lookup reference (inode leak) (by LiAuto)

`read_dir_common_inner` materializes every child into the dcache via `dir.lookup(...)` for BOTH plain `READDIR` and `READDIRPLUS`, taking a kernel lookup reference in both cases. For plain `READDIR` the kernel never sends a matching `FORGET`, so those inode references never drop back to zero — a steady inode leak on repeated directory listing.

- `curvine-fuse/src/fs/curvine_file_system.rs:174` (`dir.lookup(header.nodeid, &status.name, status.clone())` — called regardless of `plus`)
- `curvine-fuse/src/fs/dcache/dir_tree.rs` (`DirTree::lookup` calls `add_lookup(1)`; new inode via `Inode::with_status` starts at `n_lookup=1, ref_ctr=1`)
- `curvine-fuse/src/raw/fuse_dirent_list.rs:71` (`add_dirent`: `plus` only selects `add` = `fuse_dirent` vs `add_plus` = `fuse_direntplus`+`fuse_entry_out` — the wire format; the dcache lookup is identical for both)

**Verified FUSE protocol semantics**

- `READDIRPLUS` returns entry attributes, so the kernel treats each entry like a `LOOKUP`: it increments the inode lookup count and WILL send a `FORGET` on eviction.
- Plain `READDIR` returns only name (+optional ino), no attributes: the kernel does NOT populate the inode cache, does NOT increment the lookup count, and does NOT send `FORGET`.

**Impact**

Because plain `READDIR` also runs `dir.lookup` → `add_lookup(1)` (or creates an inode with `n_lookup=1`), those inodes have `n_lookup >= 1` with no `FORGET` ever arriving. `Inode::should_unref` requires `n_lookup == 0 && ref_ctr == 0`, so it stays false forever, and the dcache cleaner's `can_evict` cannot reclaim them. Repeatedly `ls`-ing large directories accumulates dcache inodes without bound.

- Distinguish `READDIR` from `READDIRPLUS` in `read_dir_common_inner`:
  - `READDIRPLUS` may take the lookup-ref path (the kernel balances it with `FORGET`);
  - plain `READDIR` must only cache status/path WITHOUT incrementing the kernel lookup reference — add a "no lookup ref" insert/update API for this.
- Re-examine the meaning of `ref_ctr` (dcache-local dentry ref) vs `n_lookup` (kernel lookup ref) so the two are not conflated.
- Split the ambiguous `get_cached_status(add_lookup: bool)` accessor (`node_state.rs`) into `get_cached_status_for_lookup()` / `get_cached_status_no_lookup()`, and have each call site pick based on whether the FUSE op actually establishes a kernel lookup reference — so no non-LOOKUP path silently leaks like READDIR does.

**Additional context**

From a curvine-fuse NodeState/inode-cache/handle review. Real resource leak; protocol semantics confirmed against the FUSE readdir/readdirplus lookup-count contract.

Acceptance / test gaps:
- `readdir_does_not_increment_lookup_refs`: after a plain READDIR, the child inode is reclaimable by the cleaner without a FORGET.
- `readdirplus_increments_lookup_refs`: after READDIRPLUS, the inode is retained until the matching FORGET.

yangyang233333:
# Summary

Plain READDIR was taking a kernel lookup ref (n_lookup += 1) for every
child, but the kernel never sends a FORGET for READDIR entries — so the
daemon-side lookup count grew with no way to balance it, defeating
unlink's immediate reclaim. This splits the dcache lookup path so plain
READDIR materializes children without a kernel ref, while READDIRPLUS
keeps the real-LOOKUP semantics.

# Issue Describe / Design

Closes #1114

Root cause:
- `read_dir_common_inner` used one path (`dir.lookup`) for both READDIR
  and READDIRPLUS, which unconditionally calls `add_lookup(1)`.
- FUSE contract: plain READDIR returns names only — the kernel does not
  cache the child, does not increment its lookup count, and never sends
  a FORGET. READDIRPLUS returns attrs, caches the dentry, bumps the
  lookup count, and later sends a FORGET.
- `sub_lookup` is only reachable from `forget`. So the n_lookup that
  READDIR raised was never balanced.

Impact (corrected severity — not an unbounded leak):
- The background cleaner reclaims via `can_evict`, which is purely
  TTL-based and does not consult n_lookup, so inodes are still reclaimed
  on TTL expiry — no unbounded growth.
- The real defect: a file the kernel no longer references, if seen by a
  READDIR, has n_lookup >= 1, so `should_unref()` stays false and
  unlink's *immediate* reclaim degrades to TTL fallback; the counter
  also diverges from the kernel's real value.

Fix approach:
- Extract `lookup_impl(..., bump_kref)`; keep public `lookup`
  (bump_kref=true) for real LOOKUP and READDIRPLUS, add `lookup_no_kref`
  (bump_kref=false) for plain READDIR.
- `lookup_no_kref` still takes the local dcache dentry ref (ref_ctr) so
  READDIR keeps warming the dcache, but leaves n_lookup untouched
  (Path C drops the with_status seed back to 0 via saturating sub_lookup).
- `read_dir_common_inner` dispatches on `plus`.

# Changes

| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Split `lookup` into `lookup`/`lookup_no_kref` over shared `lookup_impl(bump_kref)` | Public `lookup` semantics unchanged (bump_kref=true); new no-kref variant for READDIR |
| `curvine-fuse/src/fs/curvine_file_system.rs` | `read_dir_common_inner` dispatches READDIRPLUS to `lookup`, READDIR to `lookup_no_kref` | READDIR no longer inflates n_lookup |

# Test verified

| `cargo test -p curvine-fuse dir_tree::` | PASS | 26 passed, incl. 3 new |
| `readdir_does_not_increment_lookup_refs` | PASS | READDIR: ref_ctr=1, n_lookup=0 |
| `readdirplus_increments_lookup_refs` | PASS | READDIRPLUS still +1 each |
| `unlink_reclaims_inode_seen_only_by_readdir` | PASS | fix-proof: FAILs on old behavior |
| `cargo clippy -p curvine-fuse --all-targets -- --deny=warnings` | PASS | clean |

# Dependencies

- Related PRs: None
- Related issues: #1114
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/dcache/dir_tree.rs
+++ b/curvine-fuse/src/fs/dcache/dir_tree.rs
@@ -176,11 +176,42 @@ impl DirTree {
     }
 
     // LOOKUP: create inode and parent directory entry as needed.
+    // Establishes a kernel lookup ref (n_lookup += 1); the kernel will later
+    // balance it with a FORGET. Used by the real LOOKUP op and READDIRPLUS.
     pub fn lookup(
         &mut self,
         parent: u64,
         name: &str,
         status: FileStatus,
+    ) -> FuseResult<&mut Inode> {
+        self.lookup_impl(parent, name, status, true)
+    }
+
+    // READDIR: materialize the child into the dcache (local dentry ref) WITHOUT
+    // taking a kernel lookup ref. Plain READDIR returns names only; the kernel
+    // does not cache the child, does not increment its lookup count, and never
+    // sends a FORGET for it (unlike READDIRPLUS). Bumping n_lookup here would
+    // inflate the daemon-side count past the kernel's real value with no FORGET
+    // to balance it, which defeats unlink's immediate `should_unref` reclaim
+    // (issue #1114).
+    pub fn lookup_no_kref(
+        &mut self,
+        parent: u64,
+        name: &str,
+        status: FileStatus,
+    ) -> FuseResult<&mut Inode> {
+        self.lookup_impl(parent, name, status, false)
+    }
+
+    // `bump_kref` = whether to take a kernel lookup ref (n_lookup += 1). The
+    // local dcache dentry ref (ref_ctr) is taken regardless, since either op
+    // materializes the child into the dcache.
+    fn lookup_impl(
+        &mut self,
+        parent: u64,
+        name: &str,
+        status: FileStatus,
+        bump_kref: bool,
     ) -> FuseResult<&mut Inode> {
         let ino = match self.get_inode_mut(parent, Some(name)) {
             Some(inode) => {
@@ -193,7 +224,9 @@ impl DirTree {
                     );
                 }
 
-                inode.add_lookup(1);
+                if bump_kref {
+                    inode.add_lookup(1);
+                }
                 inode.update_status(status);
                 inode.ino
             }
@@ -220,7 +253,9 @@ impl DirTree {
                                 inode.ino
                             );
                         }
-                        inode.add_lookup(1);
+                        if bump_kref {
+                            inode.add_lookup(1);
+                        }
                         inode.add_ref(1);
                         inode.update_status(status);
                         inode.parent = parent;
@@ -231,8 +266,14 @@ impl DirTree {
                     // Path C: brand-new inode.
                     None => {
                         let ino = self.next_id(status.id);
-                        self.inodes
-                            .insert(ino, Inode::with_status(ino, parent, name, status));
+                        let mut inode = Inode::with_status(ino, parent, name, status);
+                        // with_status seeds n_lookup=1 (the LOOKUP case). READDIR
+                        // takes no kernel lookup ref, so drop it back to 0 while
+                        // keeping ref_ctr=1 for the local dcache dentry.
+                        if !bump_kref {
+                            inode.sub_lookup(1);
+                        }
+                        self.inodes.insert(ino, inode);
                         ino
                     }
                 }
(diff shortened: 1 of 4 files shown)


curvine-fuse/src/fs/dcache/dir_tree.rs
```diff
@@ -231,8 +266,14 @@ impl DirTree {
                     // Path C: brand-new inode.
                     None => {
                         let ino = self.next_id(status.id);
-                        self.inodes
-                            .insert(ino, Inode::with_status(ino, parent, name, status));
+                        let mut inode = Inode::with_status(ino, parent, name, status);
+                        // with_status seeds n_lookup=1 (the LOOKUP case). READDIR
+                        // takes no kernel lookup ref, so drop it back to 0 while
+                        // keeping ref_ctr=1 for the local dcache dentry.
+                        if !bump_kref {
```
szbr00: Path C depends on Inode::with_status always seeding n_lookup=1, then undoing it with sub_lookup(1) when bump_kref is false. That is correct today, but it is easy to break if with_status's initial lookup count ever changes. Prefer setting n_lookup=0 explicitly here (or adding a constructor/helper that takes the initial kernel lookup count) so the READDIR case does not rely on a saturating undo.

yangyang233333: Good catch. Fixed in 4412e4b: added an explicit `n_lookup` param to `with_status` (1 for LOOKUP/READDIRPLUS, 0 for READDIR) and dropped the `sub_lookup(1)` undo, so the READDIR case no longer relies on a saturating undo of a hard-coded seed.

lzjqsdd: Looks like it's a compatibility workaround. Can we just modify the origin lookup function to accept more parameters

yangyang233333: Done in 12e5828: folded lookup/lookup_no_kref/lookup_impl into one `lookup(parent, name, status, bump_kref)`. READDIRPLUS/real LOOKUP pass true, plain READDIR passes false. No behavior change; 26 dir_tree tests + clippy pass.

szbr00 (approved):

Issue #1221, opened 2026-07-21
[SUB] fuse: retry and observe failed deferred-delete cleanup (by LiAuto)


ligeng8848:
**Is your feature request related to a problem?**

Parent: #1112  
Follow-up to: #1218

#1218 preserves file handles, writer references, lock ownership, and
`pending_delete` state when FUSE release cleanup fails. This prevents the
failure from being silently discarded.

However, FUSE normally sends `RELEASE` only once for each file handle. A
retained handle therefore may never receive another release request. A failed
writer completion, unlock, or deferred backend deletion can remain pending
indefinitely, keep `has_open_handles` true, and prevent the deferred delete
from completing.

The current warning logs make individual failures visible, but there is no
background retry mechanism or aggregate metric for stuck cleanup work.

**Describe the solution you'd like**

Add a bounded background cleanup mechanism for release work retained after a
failure.

The mechanism should:

- Retry writer completion or flush before releasing the associated shared
  writer reference.
- Retry failed lock cleanup while preserving locally tracked lock ownership
  until the backend unlock succeeds.
- Retry deferred backend deletion while keeping `pending_delete` set.
- Use the existing per-path serialization to avoid racing with open, unlink,
  or another cleanup attempt.
- Treat successful deletion and `FileNotFound` as terminal success.
- Use bounded exponential backoff and avoid a busy retry loop.
- Expose metrics and structured logs for pending work, retries, successes, and
  repeated failures.
- Define shutdown and restart behavior for retained cleanup state.

**Describe alternatives you've considered**

- Relying on a second FUSE `RELEASE`: not reliable because the kernel normally
  sends it only once.
- Retrying indefinitely inside the original `RELEASE`: can block the FUSE
  request and filesystem teardown.
- Dropping retained state after reporting the error: recreates the silent
  orphan and leaked-cleanup problem fixed by #1218.
- Logging only: improves visibility but does not eventually reclaim resources.

**Additional context**

Raised during review of #1218:

- https://github.com/CurvineIO/curvine/pull/1218#discussion_r3620242990
- https://github.com/CurvineIO/curvine/pull/1218#discussion_r3620243647

# Goal

| Sub-requirement | Problem solved |
| --------------- | -------------- |
| Queue failed release cleanup for retry | FUSE may not send another `RELEASE` |
| Preserve handles, writers, locks, and pending-delete state until cleanup succeeds | Prevent silent state loss and backend orphans |
[...]

PR #1277, opened 2026-07-23
fix(fuse): release handles after cleanup errors

ligeng8848:
# Summary

Release FUSE file handles and shared-writer references even when writer cleanup fails. Cleanup errors are still returned to the caller, while in-memory ownership now continues to match the handle lifecycle reported by the kernel.

# Issue Describe / Design

This is a standalone bug fix discovered after #1218; there is no dedicated issue to close.

## Root cause

#1218 retained the file handle and shared-writer reference when `flush` or `complete` failed. However, the kernel sends `RELEASE` once and does not retry or wait for its result. Keeping those active references therefore made `has_open_handles` and the writer reference count permanently larger than the kernel-owned handle count.

## Reproduction steps

1. Open a writable FUSE handle and make `flush` or `complete` fail during `RELEASE`.
2. Observe that the kernel has released the handle, but the old path retains both the handle and writer reference in memory.
3. Observe stale open-handle state and incorrect writer accounting because no second `RELEASE` is expected.

## Fix approach

- Attempt writer cleanup before releasing the reference, preserving the required flush-versus-complete decision.
- Consume the shared-writer reference regardless of the cleanup result.
- Remove the FUSE handle regardless of the cleanup result.
- Return the cleanup error separately so the release path still reports and logs the failure.
- Keep retriable backend cleanup separate from active kernel-handle ownership; the broader retry mechanism remains tracked by #1221.

# Changes

| `orpc/src/sync/async_map.rs` | Return release status and cleanup result separately, and consume the reference after every cleanup attempt | Failed cleanup no longer leaves a stale active reference |
| `curvine-fuse/src/fs/state/node_state.rs` | Always remove a released handle while continuing to propagate `flush`/`complete` errors | Handle and writer counts remain aligned with kernel ownership |
| `curvine-fuse/src/fs/state/node_state.rs` tests | Cover single-handle and shared-writer cleanup failures | Prevents stale-handle and incorrect reference-count regressions |

# Test verified

| `make format` | PASS | Run in the remote Linux development environment |
| Workspace Clippy | PASS | Run in the remote Linux development environment |
| `cargo test -p orpc sync::async_map::tests -- --nocapture` | PASS | 12 passed |
| `cargo test -p curvine-fuse --lib fs::state::node_state::test:: -- --nocapture --test-threads=1` | PASS | 9 passed, including `/dev/full` cleanup-failure injection |
| Real FUSE mount E2E | NOT RUN | A normal mount does not deterministically trigger the cleanup-failure branch; focused fault-injection tests cover the changed path |

# Dependencies

- Related PRs: #1218
- Related issues: #1221
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -588,7 +588,8 @@ impl NodeState {
         let close_result: FuseResult<()> = match handle.as_ref() {
             FileHandle::Backend(_) if handle.has_writer() => {
                 let cleanup_handle = handle.clone();
-                self.writers
+                let (_, cleanup_result) = self
+                    .writers
                     .release_with_cleanup(handle.ino(), move |last| async move {
                         if last {
                             cleanup_handle.complete(None).await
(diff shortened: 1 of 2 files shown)


ligeng8848:
@bigbigxu Could you please help review this fix? It ensures the writer reference and FUSE handle are released even when flush/complete cleanup fails, while the cleanup error is still propagated. Thanks!

bigbigxu (approved):

PR #1278, opened 2026-07-23
fix(fuse): enforce POSIX file open permissions

ljy1-lixing:
**Summary**

Fix FUSE file-open permission handling for execute-only files and non-root write opens, and correct traversal checks at the mounted backend root.

Linux opens an executable through an internal read-only FUSE handle even when the caller only requires execute permission. Curvine previously interpreted that `FUSE_OPEN(O_RDONLY)` as an ordinary read and required `R_OK`, which incorrectly rejected valid execute-only files. Separately, path traversal could check the synthetic FUSE root inode before its real backend metadata had been loaded, causing valid non-root operations to fail against a synthetic mode of zero.

**Minimal Reproducer**

Create a small executable and make it executable but not readable by the test user:

```bash
cat >/tmp/execute-only.c <<'EOF'
#include <stdio.h>

int main(void)
{
    puts("execute-only file ran successfully");
    return 0;
}
EOF

gcc -Wall -O2 /tmp/execute-only.c -o /mnt/curvine/execute-only
chown 99:99 /mnt/curvine/execute-only
chmod 0100 /mnt/curvine/execute-only
```

Execute it as the owner:

```bash
setpriv --reuid=99 --regid=500 --clear-groups \
    /mnt/curvine/execute-only
```

Before the fix, the kernel sent `FUSE_OPEN(O_RDONLY)` while loading the executable. Curvine mapped that request to `R_OK`, so the command failed with `EACCES` even though owner execute permission was present.

The related write failure can be reproduced with a writable file owned by a non-root user:

```bash
touch /mnt/curvine/write-only
chown 99:99 /mnt/curvine/write-only
chmod 0200 /mnt/curvine/write-only

setpriv --reuid=99 --regid=500 --clear-groups \
    sh -c 'printf data > /mnt/curvine/write-only'
```

The writable open succeeded, but the following truncate-related `SETATTR` could fail while traversing the synthetic FUSE root inode with mode zero.

**Root Cause**

Curvine performed file-open permission checks entirely in userspace based on the access mode in `fuse_open_in.flags`:

```text
O_RDONLY -> R_OK
O_WRONLY -> W_OK
O_RDWR   -> R_OK | W_OK
```

This mapping is valid for ordinary `open(2)` calls, but it cannot represent an executable load. The Linux VFS checks execute permission before loading the file, then sends an internal read-only open to the FUSE daemon. The FUSE open request does not expose a reliable flag that distinguishes this internal open from a normal read. Requiring `R_OK` in the daemon therefore rejects execute-only files, while simply removing read checks without kernel permission enforcement would allow ordinary reads of those files.

The correct distinction is available inside the kernel permission path. The FUSE `default_permissions` mount option lets the VFS enforce the original operation's permission mask before dispatching the request to userspace.

Write-only and read-write opens remain unambiguous in the FUSE request, so Curvine keeps their userspace `W_OK` and `R_OK | W_OK` checks as defense in depth and to preserve fresh backend permission validation.

The traversal failure had a separate cause. `Inode::new_root()` creates a synthetic dcache root before backend metadata is available. Its initial `FileStatus` has mode zero. The traversal check used that synthetic status as if it were authoritative, so non-root operations below the mount root could be denied even when the real exported directory was searchable.

**Changes**

- Add `default_permissions` automatically when Curvine permission checking is enabled, without duplicating an explicitly configured mount option.
- Delegate read-only open permission decisions to the kernel so executable loading is distinguished from ordinary reading.
- Retain userspace permission checks for `WriteOnly` and `ReadWrite` opens.
- Resolve the real backend root status before using it for traversal checks.
- Cache the resolved root status in the dcache even when the general metadata cache is disabled, avoiding an extra root metadata RPC on every traversal.
- Add unit coverage for mount-option generation and per-open userspace masks.

**Deployment Note**

`default_permissions` is fixed when the FUSE connection is mounted. Upgrading from a version that did not enable this option requires stopping and remounting the FUSE client. A hot reload that inherits the existing `/dev/fuse` file descriptor does not change the original mount options.

Master and worker processes do not need to restart for this FUSE-only change.

**Verification**

- `cargo test -p curvine-fuse --lib -- --test-threads=1` passes: `295 passed`.
- FUSE configuration tests pass: `36 passed`.
- `cargo build --release -p curvine-fuse` passes.
- `git diff --check` passes.

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::fs::dcache::CleanerTask;
+use crate::fs::dcache::{CleanerTask, Inode};
 use crate::fs::operator::*;
 use crate::fs::plock_wait_registry::{LockOwner, PlockWaitGuard, PlockWaitRegistry};
 use crate::fs::state::{FileHandle, NodeState};
@@ -406,6 +406,10 @@ impl CurvineFileSystem {
         }
     }
 
+    fn traversal_cached_status(inode: &Inode, meta_ttl: u64) -> Option<FileStatus> {
+        inode.cache_valid(meta_ttl).then(|| inode.clone_status())
+    }
+
     /// Linux root access(2) bypasses R_OK/W_OK but still validates X_OK against any
     /// execute bit in the file mode, not the caller's owner/group/other class.
     fn check_root_access_permissions(status: &FileStatus) -> FuseResult<()> {
@@ -459,9 +463,7 @@ impl CurvineFileSystem {
                 match dir.get_inode(dir_ino, None) {
                     Some(inode) => {
                         let is_root = inode.is_root();
-                        let status = inode
-                            .cache_valid(meta_ttl)
-                            .then(|| inode.clone_status());
+                        let status = Self::traversal_cached_status(inode, meta_ttl);
                         (is_root, status)
                     }
                     None => (false, None),
@@ -1943,8 +1945,9 @@ impl fs::FileSystem for CurvineFileSystem {
 
 #[cfg(test)]
 mod tests {
+    use crate::fs::dcache::Inode;
     use crate::{FATTR_ATIME_NOW, FATTR_GID, FATTR_MODE, FATTR_MTIME, FATTR_MTIME_NOW, FATTR_UID};
-    use curvine_common::state::{FileAllocMode, FileType, INTERNAL_CTIME_XATTR};
+    use curvine_common::state::{FileAllocMode, FileStatus, FileType, INTERNAL_CTIME_XATTR};
 
     #[test]
     fn userspace_open_checks_only_unambiguous_write_modes() {
@@ -1963,6 +1966,27 @@ mod tests {
             CurvineFileSystem::open_userspace_acl_mask(OpenAction::ReadWrite),
             Some((libc::R_OK | libc::W_OK) as u32)
         );
+    }
+
+    #[test]
+    fn traversal_ignores_synthetic_root_until_backend_status_is_loaded() {
+        use super::CurvineFileSystem;
+
+        let mut root = Inode::new_root();
+        let meta_ttl = 60_000;
+        assert!(CurvineFileSystem::traversal_cached_status(&root, meta_ttl).is_none());
+
+        root.update_status(FileStatus {
+            is_dir: true,
+            mode: 0o755,
+            ..Default::default()
+        });
+
+        let status = CurvineFileSystem::traversal_cached_status(&root, meta_ttl).unwrap();
+        assert_eq!(status.mode, 0o755);
+
+        root.invalid_cache();
+        assert!(CurvineFileSystem::traversal_cached_status(&root, meta_ttl).is_none());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


szbr00 (commented):
Nice fix — the exec-load vs. ordinary-read distinction can only be resolved kernel-side, and the traversal fix for the synthetic mode-0 root is well targeted.

One deployment consideration worth surfacing in the release notes alongside the remount requirement: enabling `default_permissions` broadens kernel permission enforcement beyond opens. Namespace-mutating ops such as `unlink`, `rmdir`, `rename`, `link`, `mkdir`, and `create` previously had no userspace POSIX permission check (only the cluster-level `ensure_writable_path`), so on multi-user (`allow_other`) mounts the kernel will now enforce parent-directory `MAY_WRITE|MAY_EXEC` for them. This is the correct POSIX behavior and a security improvement, but it is a user-visible behavior change: operations that previously succeeded for non-owners (e.g., deleting another user's file in a `0755` directory) will start returning `EACCES` after remount.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -434,22 +445,29 @@ impl CurvineFileSystem {
...
             };
-            let cached_status = {
+            let (is_root, cached_status) = {
                 let dir = self.state.dir_read();
-                dir.get_inode(dir_ino, None)
-                    .map(|inode| inode.clone_status())
+                match dir.get_inode(dir_ino, None) {
+                    Some(inode) => {
+                        let is_root = inode.is_root();
+                        let status = if is_root && inode.lifecycle == Lifecycle::Invalid {
```
szbr00: The traversal fix (skipping the synthetic root status when `lifecycle == Lifecycle::Invalid`) and the companion root-status caching in `NodeState::fs_stat` are the core of the SETATTR/traverse bug fix, but they have no regression test — the new coverage is for mount-option generation and the open mask only.

I realize there is no mock-backend infrastructure for constructing a full `CurvineFileSystem`/`NodeState` today (the integration tests are stubs), so this may not be cheap. If feasible, even a focused test asserting that `fs_stat(FUSE_ROOT_ID, None)` materializes the root status into the dcache when the metadata cache is disabled would guard the caching behavior; otherwise it may be worth tracking test infrastructure for this path as a follow-up.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -386,6 +386,17 @@ impl CurvineFileSystem {
         self.check_access_permissions(&status, header, mask)
     }
 
+    /// Read-only FUSE opens cannot distinguish an executable load from a normal
+    /// read, so the kernel handles those through `default_permissions`. Write
+    /// modes remain unambiguous and keep the userspace check as defense in depth.
+    fn open_userspace_acl_mask(action: OpenAction) -> Option<u32> {
```
szbr00, line 401: Minor observation on the split enforcement model, not a blocker: the kernel's `default_permissions` path (`generic_permission`) considers supplementary groups via `in_group_p`, while the retained userspace write-mode check only matches the caller's primary gid (`get_effective_permission_bits` compares `header.gid` against the file gid). For a file whose group matches a caller's supplementary group with mode `0660`, a read-only open is now allowed (kernel, group bits) while a write open is still denied (userspace falls through to the other bits).

The read side is more POSIX-correct than before, so this is fine for this PR — but resolving supplementary groups (e.g., via `getgroups`) in the userspace check would make the two paths consistent, if full POSIX compliance is a goal for a follow-up.

szbr00 (approved):

PR #1281, opened 2026-07-23
fix(server): prevent partial parallel loads from becoming valid cache

gongxun0928:
# Summary

Prevent a cancelled or incomplete parallel UFS-to-Curvine load from stamping `ufs_mtime` on a pre-resized target and becoming a false cache hit.

# Issue Describe / Design

## Root cause

`run_parallel` pre-creates the target and calls `resize(src_len)` before copying. This is required for disjoint multi-writer ranges, but it also means the target metadata reports the full source length before every range contains data.

Each join-loop iteration checks `task.is_cancel()` before awaiting a stream handle. A race remains for the final handle:

1. The parent checks cancellation before awaiting the final stream and sees `false`.
2. The task is cancelled while the parent is awaiting that stream.
3. The stream observes cancellation, cancels its writer, and returns `Ok(0)` as a soft exit.
4. The join loop accepts `Ok(0)` and ends; there is no next iteration to recheck cancellation.
5. The runner stamps the source `ufs_mtime` on the target even though the final range is incomplete.
6. `FileStatus::cv_valid` can then accept the target because its pre-resized length and stamped mtime match the UFS source.

The result is a false cache hit: later load or read paths may skip reloading an incomplete file.

## Trigger conditions

The race requires all of the following:

- a UFS-to-Curvine load using parallel fan-out (`effective_streams > 1`);
- cancellation while the parent is awaiting the final stream;
- the final stream observing cancellation before completing its assigned range.

Cancellation can come from a user cancel request, a duplicate task superseding the current runner, or master cleanup after a partial dispatch failure.

## Fix approach

- Recheck `task.is_cancel()` immediately after all stream handles join and before stamping cache validity. Cancellation remains a soft `Ok(())` exit, preserving existing task-state semantics.
- Require `written == src_len` for the non-cancelled path. Any mismatch is reported as a data-integrity error and refuses to stamp `ufs_mtime`.
- Include the task ID in the incomplete-load error for consistency with sibling errors.
- Use fault-injection points to deterministically reproduce cancellation after the final pre-await check.
- Stamp `ufs_mtime` only after both integrity gates pass.

# Changes

| `curvine-server/src/worker/task/load_task_runner.rs` | Add final cancellation and copied-length gates, task context in the error, and test-only fault points | Cancelled or incomplete parallel loads can no longer become valid cache entries |
| `curvine-server/tests/load_task_runner_fault_test.rs` | Add a deterministic MiniCluster regression test for final-stream cancellation | Regression coverage only |
| `curvine-tests/regression/daily_regression_test.sh` | Enable `curvine-server/fault-injection` for Nextest discovery and specific server tests | Daily regression now executes feature-gated fault tests |

# Test verified

| `cargo fmt --all --check` | PASS | |
| `cargo clippy -p curvine-server --all-targets --features fault-injection -- -D warnings --allow clippy::uninlined-format-args` | PASS | |
| `cargo test -p curvine-server --lib worker::task::load_task_runner` | PASS | 8 passed |
| `cargo nextest run --workspace --profile ci-no-ufs --features curvine-server/fault-injection -E 'test(canceled_final_parallel_stream_does_not_mark_cache_valid)'` | PASS | 1 passed; 15.217s |
| `daily_regression_test.sh ... curvine-server load_task_runner_fault_test canceled_final_parallel_stream_does_not_mark_cache_valid` | PASS | Structured result reports 1/1 passed |
| Negative control with both final integrity gates temporarily removed | EXPECTED FAIL | Test observed the invalid cache-validity stamp |
| `bash -n curvine-tests/regression/daily_regression_test.sh` | PASS | |
| `make format` | BASELINE FAILURE | Existing unused imports in `curvine-fuse/src/fs/state/node_state.rs` |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/task/load_task_runner.rs
+++ b/curvine-server/src/worker/task/load_task_runner.rs
@@ -450,6 +450,21 @@ impl LoadTaskRunner {
             }
         }
 
+        // Cancellation may happen after the loop's pre-await check while the
+        // final stream is still running. That stream returns Ok(0), so recheck
+        // here before stamping the pre-resized target as a valid cache entry.
+        if self.task.is_cancel() {
+            info!("task {} was cancelled", self.task.info.task_id);
+            return Ok(());
+        }
+        if written != src_len {
+            return err_box!(
+                "parallel load incomplete: wrote {} of {} bytes; refusing to mark cache valid",
+                written,
+                src_len
+            );
+        }
+
         // ufs -> cv: stamp the source mtime onto the cached file (cache validity).
         let ufs_mtime = self.get_ufs()?.get_status(source_path).await?.mtime;
         let attr_opts = SetAttrOptsBuilder::new().ufs_mtime(ufs_mtime).build();
--- a/curvine-server/src/worker/task/load_task_runner.rs
+++ b/curvine-server/src/worker/task/load_task_runner.rs
@@ -80,6 +80,15 @@ impl LoadTaskRunner {
                 warn!("report task {}", e)
             }
         }
+
+        crate::fault_point! {
+            async,
+            name: "worker.load_task.after_run",
+            description: "After a worker load task runner has fully exited",
+            context: {
+                "task_id" => self.task.info.task_id.clone(),
+            },
+        }
     }
 
     async fn run0(&self) -> FsResult<()> {
@@ -359,6 +368,18 @@ impl LoadTaskRunner {
                 let mut writer = fs.open_for_write(&dst, false).await?;
                 writer.seek(off).await?;
 
+                crate::fault_point! {
+                    async,
+                    name: "worker.load_task.parallel.before_segment_copy",
+                    description: "After a parallel load segment opens its streams and before it copies data",
+                    context: {
+                        "task_id" => task.info.task_id.clone(),
+                        "stream_index" => i as i64,
+                        "segment_offset" => off,
+                        "segment_len" => len,
+                    },
+                }
+
                 let mut remaining = len;
                 while remaining > 0 {
                     // Honor cancellation and the shared deadline INSIDE the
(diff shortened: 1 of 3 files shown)


curvine-server/src/worker/task/load_task_runner.rs
```diff
@@ -450,6 +450,21 @@ impl LoadTaskRunner {
...
         }
 
+        // Cancellation may happen after the loop's pre-await check while the
+        // final stream is still running. That stream returns Ok(0), so recheck
+        // here before stamping the pre-resized target as a valid cache entry.
+        if self.task.is_cancel() {
+            info!("task {} was cancelled", self.task.info.task_id);
+            return Ok(());
+        }
+        if written != src_len {
```
szbr00, line 490: The two new gates close a real race, but neither is covered by a regression test — the existing tests exercise only the pure planning math. A deterministic unit test would need injectable reader/writer streams (the runner currently takes concrete `CurvineFileSystem`/`UfsFactory`), so I understand if this is deferred; flagging it in case a fault-injection point (e.g. delay-before-`set_attr` + concurrent cancel) is feasible as a follow-up.

curvine-server/src/worker/task/load_task_runner.rs
```diff
@@ -450,6 +450,21 @@ impl LoadTaskRunner {
...
+            return err_box!(
+                "parallel load incomplete: wrote {} of {} bytes; refusing to mark cache valid",
```
szbr00: Minor: the sibling timeout error includes `task.info.task_id` in its message while this one doesn't. In practice `run()` logs the task ID around any failure, so this is purely a consistency nit — feel free to ignore.

curvine-server/src/worker/task/load_task_runner.rs
gongxun0928: Thanks. Addressed in 96cc408f: added a deterministic fault-injection MiniCluster regression that pauses the final segment and the parent after its final pre-await cancellation check, delivers a real job cancellation, waits for the runner to exit, and verifies that ufs_mtime remains zero and the cache entry remains invalid. The test is included in the daily Nextest suite via curvine-server/fault-injection and passed with the ci-no-ufs profile. A negative-control run without the final integrity gates failed as expected.

curvine-server/src/worker/task/load_task_runner.rs
gongxun0928: Thanks. Fixed in 96cc408f: the incomplete parallel-load error now includes task.info.task_id, consistent with the sibling timeout error.

szbr00 (approved):

szbr00 (approved):

Issue #1243, opened 2026-07-22
[BUG]: curvine-cli should not require RDMA runtime libraries on non-RDMA clients


jlon:
**Describe the bug**

The Rocky 9 Curvine runtime package currently makes `curvine-cli` depend on RDMA native runtime libraries such as `libibverbs.so.1` and `librdmacm.so.1`.

This is an architecture/dependency-boundary problem: a client-side CLI invocation should not require RDMA userspace libraries unless the command actually uses an SPDK/RDMA data path. In production handoff scenarios where `/opt/curvine` is exported as a self-contained client package, business hosts fail before the CLI can even start if those RDMA libraries are not installed on the host.

Observed error from a business host:

```text
/home/service/app/gx-inference/curvine/lib/curvine-cli: error while loading shared libraries: libibverbs.so.1: cannot open shared object file: No such file or directory
```

A previous similar failure was also observed for OpenSSL 3:

```text
/home/service/app/gx-inference/curvine/lib/curvine-fuse: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
```

The OpenSSL case can be handled by packaging runtime libraries, but the RDMA case exposes a deeper issue: optional transport/storage features are leaking into binaries that should remain usable in non-RDMA client environments.

**To Reproduce**

1. Build/export a Rocky 9 Curvine package with the current Docker build args that include SPDK/RDMA support.
2. Copy the exported Curvine directory to a business host that does not have RDMA userspace libraries installed.
3. Run the CLI through the normal wrapper:

```bash
cd /home/service/app/gx-inference/curvine
bin/cv --help
```

4. The process fails during dynamic loading before CLI logic runs:

```text
libibverbs.so.1: cannot open shared object file: No such file or directory
```

Local evidence from the exported package:

```text
$ readelf -d /opt/curvine/lib/curvine-cli | egrep 'NEEDED|RPATH|RUNPATH'
NEEDED Shared library: [libstdc++.so.6]
NEEDED Shared library: [libjindosdk_c.so.6]
NEEDED Shared library: [libnuma.so.1]
NEEDED Shared library: [libuuid.so.1]
NEEDED Shared library: [libfuse3.so.3]
NEEDED Shared library: [libisal.so.2]
NEEDED Shared library: [libssl.so.3]
NEEDED Shared library: [libcrypto.so.3]
NEEDED Shared library: [libibverbs.so.1]
NEEDED Shared library: [librdmacm.so.1]
```

[...]

PR #1290, opened 2026-07-24
refactor(common): split lightweight API crates

lzjqsdd:
# Summary

[GT-1243] Split lightweight API and domain surfaces out of `curvine-common` while keeping compatibility re-exports for existing downstream users. This is intended as a structural dependency cleanup with no runtime behavior change.

# Issue Describe / Design

Related to #1243

This implements the Stage 2 crate split from the common dependency cleanup plan. The design keeps `curvine-common` as the compatibility facade while moving low-level API/domain types into smaller workspace crates with empty default features.

- Keep `curvine-common` as the legacy import surface by re-exporting the new crates from the old module paths.
- Keep raft, RocksDB, native storage, server-side config, and local filesystem implementation in `curvine-common`.
- Move `FsError`, non-raft protobufs, domain model types, filesystem API traits/types, storage API types, and client/fuse/job/ufs config into dedicated crates.
- Keep existing Java/Python SDK protobuf source paths under `curvine-common/proto` unchanged for compatibility, while adding the non-raft proto inputs to `curvine-proto`.
- Gate the server-only HTTP response integration for `FsError` behind an optional `axum-response` feature.

# Changes

| `crates/common/curvine-error` | Adds the lightweight filesystem error crate with empty default features and optional Axum response support. | No intended behavior change; server enables the optional response feature explicitly. |
| `crates/common/curvine-proto` | Adds the non-raft protobuf crate and display implementations for mount responses. | No intended behavior change; raft protobuf generation remains in `curvine-common`. |
| `crates/common/curvine-model` | Moves domain state, options, mount logic, job display, and proto conversion helpers behind lightweight model traits. | No intended behavior change; avoids depending on filesystem/config concrete types. |
| `crates/common/curvine-fs-api` | Moves `Path`, `FileSystem`, reader/writer/list stream, `FsKind`, and `RpcCode`. | No intended behavior change; old `curvine_common::fs::*` imports still work. |
| `crates/common/curvine-storage-api` | Re-exports storage-related model types for lightweight storage consumers. | No intended behavior change. |
| `crates/common/curvine-config` | Moves client/fuse/job/ufs/cli config and implements model defaults. | No intended behavior change; old `curvine_common::conf::*` imports still work. |
| `curvine-common` | Converts moved modules into compatibility re-exports and retains heavyweight/common-only modules. | Preserves existing public import paths. |
| `scripts/check-api-crate-deps.sh` | Adds a dependency boundary guard for the new API/domain crates. | Build/test helper only. |

# Test verified

| `make format` | PASS | Runs `cargo clippy --release --fix`, `cargo fmt`, and `cargo clippy --release --all-targets --deny warnings`. |
| `cargo test -p curvine-error -p curvine-model -p curvine-fs-api -p curvine-config -p curvine-common --lib --tests` | PASS | Covers moved error/model/fs/config/common compatibility tests. |
| `scripts/check-api-crate-deps.sh` | PASS | Confirms the new API/domain crates do not pull forbidden heavyweight dependencies with no default features. |
| `cargo check -p curvine-lancedb` | PASS | Verifies the optional lancedb package against the split API surface. |

# Dependencies

- Related PRs: None
- Related issues: #1243
- Blocks / blocked by: None


scripts/check-api-crate-deps.sh
```diff
@@ -0,0 +1,81 @@
...
+
+ROOT="$(git rev-parse --show-toplevel)"
+cd "$ROOT"
+
+for bin in cargo awk grep sort; do
+  if ! command -v "$bin" >/dev/null 2>&1; then
+    echo "Missing required command: $bin" >&2
+    exit 2
+  fi
+done
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 71cf4519fae66ec39437bd2ddbe1ba147b122af4: required-command checks now include `git` and run before `git rev-parse --show-toplevel`.

crates/common/curvine-proto/build.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    let base = env::var("OUT_DIR").unwrap_or_else(|_| ".".to_string());
+    let output = format!("{}/protos", base);
+    fs::create_dir_all(&output).unwrap();
+
+    let mut build = prost_build::Config::new();
+    build.type_attribute(".", "#[derive(serde::Serialize, serde::Deserialize)]");
+    build
+        .out_dir(&output)
+        .compile_protos(&src, &["proto/"])
```
szbr00: The six non-raft protos now live in two places: here and under `curvine-common/proto/`, which the Java SDK still consumes via `protoSourceRoot` in `curvine-libsdk/java/pom.xml`. Two follow-ups worth considering:

1. `curvine-common/build.rs` still compiles these same six protos into its own `OUT_DIR`, but that output is now dead — `curvine-common/src/lib.rs` only includes `raft.rs` and re-exports `curvine_proto` for the rest. The non-raft compilation there can be dropped.
2. Nothing keeps the two copies in sync. The Rust runtime uses this crate's generated types while the Java/Python SDKs generate from `curvine-common/proto`, so a one-sided edit would silently diverge the wire format. Consider making this directory the single source of truth (or adding a CI diff check between the two copies).

scripts/check-api-crate-deps.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
+#
+# Stage-2 dependency boundary checks for lightweight API/domain crates.
```
szbr00, line 3: This guard only runs manually — it isn't referenced by any workflow, Makefile target, or doc. A future PR could reintroduce a forbidden heavyweight dependency without anything catching it. Consider wiring it into CI (or a `make` target that CI invokes) so the boundary is actually enforced.

crates/common/curvine-model/src/mount.rs
```diff
@@ -513,8 +517,110 @@ impl TryFrom<&str> for WriteType {
 
 #[cfg(test)]
 mod tests {
-    use crate::fs::Path;
     use crate::state::{AccessMode, MountInfo, MountOptions, WriteType};
+    use crate::CurvinePath;
+
+    #[derive(Clone)]
+    struct TestPath {
```
szbr00: `TestPath` re-implements ~90 lines of `Path` parsing/normalization to avoid depending on `curvine-fs-api`. If `Path`'s normalization ever changes, this mock won't track it and these tests would silently validate stale behavior. Since Cargo permits dev-dependency cycles, `curvine-model` could add `curvine-fs-api` as a `dev-dependency` and test against the real `Path` (the `-e normal` boundary check ignores dev-deps), removing the duplicated logic.

crates/common/curvine-proto/build.rs
lzjqsdd: Thanks. Fixed in f938a66214f90387e69c8214f93c4c860ce5f9b2: `curvine-proto` now compiles the six non-raft protos from `curvine-common/proto` (single source of truth shared with Java/Python SDK generation), the duplicate copies under `crates/common/curvine-proto/proto/` were removed, and `curvine-common/build.rs` only generates raft bindings plus VERSION metadata.

scripts/check-api-crate-deps.sh
lzjqsdd: Thanks. Fixed in f938a66214f90387e69c8214f93c4c860ce5f9b2: added `make check-api-crate-deps` and wired it into `.github/workflows/build.yml` next to the existing dependency guardrails step so CI enforces the lightweight API/domain crate boundary.

crates/common/curvine-model/src/mount.rs
lzjqsdd: Thanks. Fixed in f938a66214f90387e69c8214f93c4c860ce5f9b2: `curvine-model` now takes `curvine-fs-api` as a `dev-dependency`, and the mount tests use a thin `TestPath` wrapper over the real `Path` parser/normalizer instead of re-implementing path logic. Directly using `Path` as the `CurvinePath` implementor in `curvine-model` tests hits a crate-cycle visibility issue, so the wrapper keeps production parsing while remaining a local `CurvinePath` impl.

szbr00 (approved):

PR #1292, opened 2026-07-24
fix(master): prevent oversized fallocate stalls

ljy1-lixing:
**Summary**

Reject file preallocation requests that exceed the cluster's currently available physical capacity, and make master-side inode extension linear in the number of blocks being added.

Previously, an oversized `fallocate(2)` request could be accepted as long as the requested file length remained below Curvine's maximum file size. The master then attempted to materialize metadata for every block while holding the global filesystem metadata write lock. Large requests could therefore consume a CPU core for minutes and block unrelated metadata RPCs.

**Minimal Reproducer**

Create an empty file on a Curvine FUSE mount and request twice the capacity reported as available by `df`:

```bash
file=/mnt/curvine/fallocate-too-large
available_kib=$(df -P /mnt/curvine | awk 'END {print $4}')
too_large_kib=$((available_kib * 2))

: >"$file"
time fallocate -l "${too_large_kib}K" "$file"
```

Before the fix, the request did not fail promptly with `ENOSPC`. For a cluster using 128 MiB blocks, a request around 110 TiB caused the master to create metadata for roughly 900,000 blocks. While this work was in progress, other filesystem operations could time out because they also needed the global metadata lock.

After the fix, the same request returns `ENOSPC` immediately and the master continues serving unrelated metadata operations.

**Root Cause**

`FileAllocOpts::validate()` checked argument validity and the maximum supported file size, but it did not compare preallocation growth with available worker capacity. A request could therefore be valid as a file length while still being impossible to allocate on the cluster.

Once accepted, `InodeFile::extend()` processed the requested range one block at a time. For every position it called `search_block_mut_by_pos()`, which scanned the block vector from the beginning. Appending `n` blocks consequently required O(n^2) cumulative block comparisons.

The resize operation runs while the master holds the global `fs_dir` write lock. The quadratic loop therefore affected the entire metadata control plane, not only the client that issued the oversized request.

**Changes**

- Add a `DiskOutOfSpace` constructor for errors that map to POSIX `ENOSPC`.
- Aggregate currently available bytes from registered workers using the same capacity source exposed through filesystem statistics.
- Validate non-truncate allocation growth before expanding inode block metadata, including the file's replication factor in physical capacity requirements.
- Preserve sparse truncate behavior; growing a logical file with truncate does not require immediate physical capacity.
- Update only the existing tail block once and append subsequent block metadata directly, reducing inode extension from O(n^2) to O(n).
- Add unit coverage for insufficient capacity, exact-capacity allocation, sparse truncate growth, and ordered tail-block extension.

**Deployment Note**

The capacity check and inode extension logic run in the master process. Deploy the updated server binary and restart or fail over the active master for the change to take effect.

Worker and FUSE processes do not require code changes for this fix. Existing clients already map the master's `DiskOutOfSpace` error to `ENOSPC`.

**Verification**

- `cargo test -p curvine-server --lib` passes: `114 passed`, `1 ignored`.
- `cargo test -p curvine-common fs_error --lib` passes: `8 passed`.
- `cargo check --workspace` passes.
- An oversized preallocation integration reproducer returns `ENOSPC` and completes in approximately one second instead of timing out.
- `cargo fmt --all -- --check` passes.
- `git diff --check` passes.

### DIFF
--- a/curvine-server/src/master/fs/master_filesystem.rs
+++ b/curvine-server/src/master/fs/master_filesystem.rs
@@ -82,6 +82,28 @@ impl MasterFilesystem {
     // Max block-report location updates applied under a single fs_dir write lock.
     const BLOCK_REPORT_WRITE_CHUNK: usize = 4096;
 
+    fn validate_alloc_capacity(
+        current_len: i64,
+        replicas: u8,
+        opts: &FileAllocOpts,
+        available: i64,
+    ) -> FsResult<()> {
+        if opts.truncate || opts.len <= current_len {
+            return Ok(());
+        }
+
+        let logical_growth = opts.len - current_len;
+        let required = logical_growth.saturating_mul(i64::from(replicas));
+        if required > available {
+            return err_ext!(FsError::disk_out_of_space(format!(
+                "fallocate requires {} bytes for {} replicas, but only {} bytes are available",
+                logical_growth, replicas, available
+            )));
+        }
+
+        Ok(())
+    }
+
     pub fn new(
         conf: &ClusterConf,
         fs_dir: SyncFsDir,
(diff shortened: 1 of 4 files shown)


curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -1186,10 +1208,18 @@ impl MasterFilesystem {
         opts.validate()?;
 
         let path = path.as_ref();
+        let available = if opts.truncate {
+            i64::MAX
+        } else {
+            self.worker_manager.read().available_bytes()
```
szbr00, line 1217: The snapshot is taken outside the `fs_dir` write lock, so concurrent fallocates can each pass against the same `available` value and over-commit in aggregate. That seems intentional here (worker-side block allocation remains the hard enforcement point), but a one-line note that this is a best-effort guard rather than a reservation would help future readers avoid mistaking it for an exact check.

curvine-server/src/master/fs/worker_manager.rs
```diff
@@ -130,6 +130,14 @@ impl WorkerManager {
         res
     }
 
+    pub fn available_bytes(&self) -> i64 {
```
szbr00, line 133: Minor: the negative-value clamping (`max(0)`) and `saturating_add` are the only non-trivial behaviors of this helper and are not covered by a unit test. A small test with one negative-`available` worker contributing 0 would lock that in.

szbr00 (approved):

PR #1296, opened 2026-07-24
fix(fuse): update ctime for futimens

ljy1-lixing:
# Summary

Fix FUSE `futimens()` failing to update `ctime` when metadata changes without changing `mtime`.

Curvine now tracks an independent metadata change timestamp across Master persistence, journal replay, RPC transport, and FUSE attribute replies.

## Root cause

Curvine's metadata model only exposed `atime` and `mtime`. The FUSE attribute conversion derived `ctime` directly from `mtime`.

When `futimens()` updated `atime` while specifying `UTIME_OMIT` for `mtime`, the metadata operation succeeded, but the returned `ctime` remained unchanged because `mtime` had not changed.

The failure is reproduced by xfstests `generic/221`:

```c
struct timespec times[2] = {
    { .tv_sec = 1000000000, .tv_nsec = 0 },
    { .tv_sec = 0, .tv_nsec = UTIME_OMIT },
};

fstat(fd, &before);
sleep(1);
futimens(fd, times);
fstat(fd, &after);
```

Before this fix:

```text
ctime before: 1784884658.899000000
ctime after:  1784884658.899000000
FAIL: futimens did not update ctime
```

## Fix approach

- Record an independent `ctime` whenever the Master applies `setattr`.
- Persist it through the existing inode metadata extension map instead of changing the bincode inode layout.
- Store the operation timestamp in the journal entry so replay preserves the original `ctime`.
- Add an optional `ctime` field to `FileStatusProto`.
- Fall back to `mtime` when communicating with legacy peers or backends that do not provide `ctime`.
- Return the independent timestamp through FUSE attributes.
- Prevent active-writer metadata merging from overwriting a newer `ctime`.

The internal persisted key starts with a NUL byte, which cannot occur in a FUSE xattr name, preventing collisions with user-visible extended attributes.

# Changes

| `curvine-common/proto/common.proto` | Add optional `ctime` to `FileStatusProto` | Backward-compatible RPC extension |
| `curvine-common/src/state/file_status.rs` | Add internal ctime encoding and legacy fallback | Existing statuses without ctime continue using mtime |
| `curvine-common/src/state/opts.rs` | Preserve internal ctime during recursive setattr | Journal replay retains the original operation time |
| `curvine-common/src/utils/proto_utils.rs` | Encode and decode ctime separately from user xattrs | Internal metadata is not exposed in protobuf xattr maps |
| `curvine-server/src/master/meta/fs_dir.rs` | Assign ctime when applying setattr and include it in the journal | Metadata-changing setattr operations now update ctime |
| `curvine-server/src/master/meta/inode/*` | Read and persist independent inode ctime | No change to the existing bincode inode structure |
| `curvine-fuse/src/fuse_utils.rs` | Populate FUSE ctime from independent metadata | `stat(2)` now reports POSIX-compatible ctime |
| `curvine-fuse/src/fs/state/node_state.rs` | Preserve newer ctime while merging active-writer metadata | Active writers no longer hide setattr ctime updates |

# Test verified

| Standalone `futimens` C reproducer | PASS | ctime advances while mtime is omitted |
| `generic/221` | PASS | `Ran: generic/221`, `Passed all 1 tests` |
| `cargo test -p curvine-common --lib` | PASS | 100 passed |
| `cargo test -p curvine-server --lib` | PASS | 112 passed, 1 ignored |
| `cargo test -p curvine-fuse --lib -- --test-threads=1` | PASS | 297 passed |
| `cargo check --workspace` | PASS | All workspace crates checked |
| `make format` | PASS | Full pre-commit checks passed |

After the fix:

```text
ctime before: 1784885529.607000000
ctime after:  1784885530.608000000
PASS: futimens updated ctime
```

### DIFF
--- a/curvine-common/src/state/file_status.rs
+++ b/curvine-common/src/state/file_status.rs
@@ -19,6 +19,10 @@ use orpc::ternary;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+// NUL cannot appear in a FUSE xattr name, so this persisted metadata key cannot
+// collide with a user-visible extended attribute.
+pub const INTERNAL_CTIME_XATTR: &str = "\0curvine.ctime";
+
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct FileStatus {
     pub id: i64,
@@ -48,6 +52,14 @@ pub struct FileStatus {
 }
 
 impl FileStatus {
+    pub fn ctime(&self) -> i64 {
+        self.x_attr
+            .get(INTERNAL_CTIME_XATTR)
+            .and_then(|bytes| bytes.as_slice().try_into().ok())
+            .map(i64::from_le_bytes)
+            .unwrap_or(self.mtime)
+    }
+
     pub fn with_name(id: i64, name: String, is_dir: bool) -> Self {
         FileStatus {
             id,
--- a/curvine-common/src/utils/proto_utils.rs
+++ b/curvine-common/src/utils/proto_utils.rs
@@ -216,7 +216,9 @@ impl ProtoUtils {
         }
     }
 
-    pub fn file_status_to_pb(status: FileStatus) -> FileStatusProto {
+    pub fn file_status_to_pb(mut status: FileStatus) -> FileStatusProto {
+        let ctime = status.ctime();
+        status.x_attr.remove(INTERNAL_CTIME_XATTR);
         FileStatusProto {
             id: status.id,
             path: status.path,
@@ -238,10 +240,18 @@ impl ProtoUtils {
             mode: status.mode,
             target: status.target,
             nlink: status.nlink,
+            ctime: Some(ctime),
         }
     }
 
     pub fn file_status_from_pb(status: FileStatusProto) -> FileStatus {
+        let mut x_attr = status.x_attr;
+        if let Some(ctime) = status.ctime {
+            x_attr.insert(
+                INTERNAL_CTIME_XATTR.to_string(),
+                ctime.to_le_bytes().to_vec(),
+            );
+        }
         FileStatus {
             id: status.id,
             path: status.path,
@@ -255,7 +265,7 @@ impl ProtoUtils {
             replicas: status.replicas,
             block_size: status.block_size,
             file_type: FileType::from(status.file_type),
-            x_attr: status.x_attr,
+            x_attr,
             storage_policy: Self::storage_policy_from_pb(status.storage_policy),
             owner: status.owner,
             group: status.group,
@@ -724,8 +734,9 @@ impl ProtoUtils {
 
 #[cfg(test)]
 mod tests {
-    use crate::state::{AccessMode, MountInfo, MountOptions};
+    use crate::state::{AccessMode, FileStatus, MountInfo, MountOptions, INTERNAL_CTIME_XATTR};
     use crate::utils::ProtoUtils;
+    use std::collections::HashMap;
 
     #[test]
     fn test_mount_info_proto_round_trip_auto_cache_and_access_mode() {
@@ -757,5 +768,25 @@ mod tests {
 
         assert_eq!(round_trip.auto_cache, Some(false));
         assert_eq!(round_trip.access_mode, Some(AccessMode::ReadWrite));
+    }
+
+    #[test]
+    fn file_status_proto_preserves_ctime_and_falls_back_for_legacy_peers() {
+        let status = FileStatus {
+            mtime: 1_000,
+            x_attr: HashMap::from([(
+                INTERNAL_CTIME_XATTR.to_string(),
+                2_000_i64.to_le_bytes().to_vec(),
+            )]),
+            ..Default::default()
+        };
+
+        let mut pb = ProtoUtils::file_status_to_pb(status);
+        assert_eq!(pb.ctime, Some(2_000));
+        assert!(!pb.x_attr.contains_key(INTERNAL_CTIME_XATTR));
+        assert_eq!(ProtoUtils::file_status_from_pb(pb.clone()).ctime(), 2_000);
+
+        pb.ctime = None;
+        assert_eq!(ProtoUtils::file_status_from_pb(pb).ctime(), 1_000);
     }
 }
--- a/curvine-common/src/utils/proto_utils.rs
+++ b/curvine-common/src/utils/proto_utils.rs
@@ -217,6 +217,7 @@ impl ProtoUtils {
     }
 
     pub fn file_status_to_pb(mut status: FileStatus) -> FileStatusProto {
+        let has_ctime = status.x_attr.contains_key(INTERNAL_CTIME_XATTR);
         let ctime = status.ctime();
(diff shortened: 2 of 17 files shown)


curvine-common/src/state/file_status.rs
```diff
@@ -19,6 +19,10 @@ use orpc::ternary;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+// NUL cannot appear in a FUSE xattr name, so this persisted metadata key cannot
+// collide with a user-visible extended attribute.
+pub const INTERNAL_CTIME_XATTR: &str = "\0curvine.ctime";
```
szbr00, line 24: The NUL prefix prevents *collision* with user xattrs in get/set/removexattr (kernel-supplied names cannot contain NUL), but it does not prevent *exposure*: FUSE `list_xattr` (curvine_file_system.rs:1017) iterates `status.x_attr.keys()` and only skips names rejected by `FuseUtils::check_xattr`, which does not recognize this key. Since `file_status_from_pb` re-inserts the key into every status and the master now always sends `ctime`, every file that has had a setattr will advertise two phantom entries in `listxattr(2)`: an empty name `""` (from the leading NUL) and `"curvine.ctime"` — the latter then returns ENODATA from `getxattr`, which breaks tools that enumerate xattrs (getfattr, backup software, xfstests xattr tests).

Suggested fix: reject names containing `'\0'` in `check_xattr` (such names can never be legitimate kernel requests, so get/set/remove behavior is unchanged); the existing `is_err() => continue` filter in `list_xattr` then hides the internal key automatically. A regression test asserting `list_xattr` output does not contain the internal key would be valuable, since generic/221 does not cover this.

curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -848,12 +849,19 @@ impl FsDir {
...
-    pub fn set_attr(&mut self, inp: InodePath, opts: SetAttrOpts) -> FsResult<FileStatus> {
+    pub fn set_attr(&mut self, inp: InodePath, mut opts: SetAttrOpts) -> FsResult<FileStatus> {
         let inode = match inp.get_last_inode() {
             Some(v) => v,
             None => return err_ext!(FsError::file_not_found(inp.path())),
         };
 
+        // Persist the operation timestamp inside the existing xattr map so journal
+        // replay restores the original ctime without changing the bincode layout.
+        let ctime = LocalTime::mills() as i64;
```
szbr00, line 868: ctime is now only refreshed on the setattr path, but several other metadata-mutating operations advance `mtime` without touching the internal ctime: `InodeFile::complete` (data write/flush sets `mtime = now`) and the `update_mtime` calls on parent dirs for create/delete/rename/link. Once an inode has a pinned ctime from any setattr, subsequent writes or child operations produce `ctime < mtime`, which is impossible on a POSIX filesystem and is a behavior change from the previous `ctime == mtime` derivation (e.g. `touch -a f; echo x >> f` now leaves ctime behind). If full POSIX parity is out of scope for this PR, it would be good to at least document the limitation; otherwise consider bumping the internal ctime in `update_mtime` and `complete` as well.

Minor related note: there is no xattr name validation at the master RPC boundary, so a non-FUSE client can send `remove_x_attr = ["\0curvine.ctime"]`; `InodeView::set_attr` processes removals after inserting the fresh ctime, so the key is stripped and ctime silently degrades to the mtime fallback. Filtering internal keys out of client-supplied `add_x_attr`/`remove_x_attr` would close this off.

curvine-server/src/master/meta/inode/inode_view.rs
```diff
@@ -416,6 +426,25 @@ impl InodeView {
     }
 
     pub fn set_attr(&mut self, opts: SetAttrOpts) -> CommonResult<()> {
+        let ctime = opts
+            .add_x_attr
+            .get(INTERNAL_CTIME_XATTR)
+            .cloned()
+            .unwrap_or_else(|| (LocalTime::mills() as i64).to_le_bytes().to_vec());
```
szbr00: For journal entries written before this PR (no `INTERNAL_CTIME_XATTR` in `add_x_attr`), this fallback stamps ctime with replay wall-clock time. Replay happens on every restart/failover and on every Raft follower, so during the transition window (until the journal truncates past legacy entries) the same legacy entry yields different persisted ctimes per node and per restart — i.e., nondeterministic metadata divergence between master replicas, and ctimes that jump to "restart time" without any user operation. Consider leaving the key absent when the opts do not carry it: `ctime()` then falls back to `mtime`, which is deterministic and matches the pre-upgrade visible behavior.

curvine-common/src/utils/proto_utils.rs
```diff
@@ -216,7 +216,9 @@ impl ProtoUtils {
         }
     }
 
-    pub fn file_status_to_pb(status: FileStatus) -> FileStatusProto {
+    pub fn file_status_to_pb(mut status: FileStatus) -> FileStatusProto {
+        let ctime = status.ctime();
+        status.x_attr.remove(INTERNAL_CTIME_XATTR);
         FileStatusProto {
```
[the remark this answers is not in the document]
ljy1-lixing: Thanks. Fixed in 1ebbfa4: FileStatusProto.ctime is now only populated when the source status contains an independent ctime marker.

curvine-common/src/utils/proto_utils.rs
```diff
@@ -238,10 +240,18 @@ impl ProtoUtils {
             mode: status.mode,
             target: status.target,
             nlink: status.nlink,
+            ctime: Some(ctime),
```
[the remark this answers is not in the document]
ljy1-lixing: Thanks. Fixed in 1ebbfa4: legacy statuses now keep ctime unset and continue falling back to mtime after the protobuf round trip.

curvine-common/src/state/file_status.rs
```diff
@@ -19,6 +19,10 @@ use orpc::ternary;
...
+
```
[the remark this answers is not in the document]
ljy1-lixing: Thanks. Fixed in 1ebbfa4: FUSE now rejects xattr names containing NUL, preventing internal ctime metadata from being exposed through listxattr.

curvine-common/src/state/file_status.rs
ljy1-lixing: Thanks. Fixed in 1ebbfa4: listxattr now filters the internal ctime key, with regression coverage for the encoded xattr name list.

curvine-server/src/master/meta/inode/inode_view.rs
ljy1-lixing: Thanks. Fixed in 1ebbfa4: legacy journal entries without an explicit ctime leave the internal marker absent, preserving the deterministic mtime fallback.

curvine-server/src/master/meta/fs_dir.rs
ljy1-lixing: Thanks. Fixed in 1ebbfa4: ctime now advances for file completion, parent directory changes, rename, link and unlink operations. Master-owned ctime metadata is also protected from client add/remove requests.

szbr00 (approved):

PR #1301, opened 2026-07-25
perf(client): prefetch next block on sequential write boundaries (#1291)

### LINKED ISSUE #1291: [BUG] curvine-fuse single-node writes stall 2–70s (BW 52–630 MiB/s) — synchronous block alloc/commit serialized through one write_future per file
## Description

Single-node writes through `curvine-fuse` stall severely: **2–70 s per operation**, bandwidth
**52–630 MiB/s** (highly variable), `clat` avg 2–22 s, 94–98 % of latencies ≥ 2 s, while the
node sits at **~97 % idle, 0 % iowait**, with **no errors and no retries** in any log.

Profiling (strace + `perf record -g`) and code tracing localize the cause to the write-commit
path: at every `block_size` (128 MB) boundary, `FsWriterBase::get_writer` runs a **synchronous
block-allocation + commit RPC sequence** — `old.complete()` (`complete_file_by_id`, commit RPC
to master) → `add_block_by_id` (allocate-block RPC to master) → `BlockWriter::new` — all
serialized through the **single per-file background `write_future` task**. While `write_future`
awaits these RPCs, the bounded `chunk_sender` channel fills and every concurrent writer parks on
`futex`; the worker starves on `epoll_wait`. Disk writes occur in bursts (256 MiB/s for ~1 s,
then idle for seconds), matching the block-boundary cadence.

The stall is **workload-independent** (reproduced with both `O_DIRECT` and buffered fio,
`write_back_cache` on/off — see below) and **not addressable by config**.

## Steps to reproduce

On a multi-node curvine cluster (masters + workers + fuse), with `v0.3.6-alpha` deployed
and a worker+fuse co-located on a node with a local NVMe data dir:

```bash
mkdir -p /cv_fuse/fio
fio --name=sn --filename=/cv_fuse/fio/sn --rw=write --bs=4M --ioengine=libaio \
    --iodepth=64 --numjobs=8 --direct=1 --size=32G --runtime=60 --time_based \
    --end_fsync=1 --group_reporting
```

## Expected behavior

Single-node writes through `curvine-fuse` to a local NVMe-backed worker should sustain
near line-rate of the device (hundreds of MiB/s to GiB/s) with sub-second latency, not
2–70 s `clat` at <630 MiB/s with the node 97 % idle.

## Actual behavior

| Workload | write_back_cache | BW | clat avg | lat ≥2 s |
| --- | --- | --- | --- | --- |
| direct fio | false | 52–630 MiB/s (variable) | 2–22 s | 94–98 % |
| direct fio | true | 129 MiB/s | 15.3 s | 98 % |
| buffered fio | true | 602 MiB/s | 3.24 s | 97 % |
| buffered fio | false | 600 MiB/s | 3.24 s | 97 % |

- `clat` max up to 70 s; `err=0`; `sender_dropped = 0` (the unrelated "sender dropped" brick
  was fixed by #993 / included in v0.3.6-alpha — this is a different, distinct problem).
- `strace` on `curvine-fuse`: threads park on `futex` 1.4–6.4 s (tokio `.await`); on the worker:
  `epoll_wait` (tokio parked idle, starved).
- `perf record -g` on the fuse (on-CPU, 43 k samples): time is spread across per-operation
  kernel overhead — `__alloc_pages` / `alloc_iova_fast` / `rmqueue` (IOMMU DMA mapping + page
  alloc), `tcp_sendmsg_locked` (RPC send to master at each block boundary), `pipe_read`,
  `ext4_block_write_begin` (short-circuit local write), tokio runtime — not a single hot
  curvine function.
- On the worker (1.4 % CPU): metrics instrumentation overhead (`std::time::Instant::now`,
  `clock_gettime`, `bvar_sampler`) + spinlock contention.
- Config levers `write_chunk_num` (16 → 128, 512 MiB buffer) and `write_back_cache`
  (false → true) had **no effect** on either direct or buffered fio. The stall is not
  `O_DIRECT`-specific (buffered stalls identically) and not chunk-channel backpressure
  (larger buffer didn't help).

## Environment

- Curvine: official release `v0.3.6-alpha` (2026-07-17 build, includes #993/#987/#988/#1002).
- Multi-node cluster: raft masters + workers + fuse; worker data dir on local NVMe (ext4),
  `short_circuit = true`, `block_size = 128 MB`, `write_chunk_size = 4 MB`, `write_chunk_num = 16`.
- OS: Ubuntu 22.04, kernel 5.15.0-113-generic.
- fio 3.x, libaio, bs 4 M, iodepth 64, numjobs 8.

## Logs

No errors, no retries, no timeouts in `fuse.out`, `worker.log`, or `master.log` during writes.
`rpc_timeout_ms = 120000`; the 70 s `clat` is under the RPC timeout, so it is genuine slow
completion, not a retry storm. `master.log` is silent at INFO during writes (the alloc/commit
RPCs are not logged at INFO).

## Root cause (code)

`FsWriterBase::write` calls `get_writer().await` on every write
(`curvine-client/src/file/fs_writer_base.rs`, `write` ~line 93). At each 128 MB boundary
(`v.has_remaining()` false), `get_writer` (line ~193) runs synchronously through the single
per-file `write_future` (spawned in `FsWriterBuffer::new`,
`curvine-client/src/file/fs_writer_buffer.rs` ~line 124):

1. `update_writer(None, false)` → `old.complete().await` → `complete_file_by_id` RPC (commit).
2. `add_block_by_id(...).await` — allocate-block RPC to master.
3. `BlockWriter::new(...).await`.

The bounded `chunk_sender` (`AsyncChannel::new(chunk_num)`, `fs_writer_buffer.rs` ~line 119)
fills while `write_future` is parked on 1–3, so all writers block on `chunk_sender.send().await`.

## Proposed fix

Decouple block allocation/commit from the single `write_future` drain path so the writer does
not block synchronously at each 128 MB boundary. Lowest-risk candidate: **prefetch/pipeline the
next block** — when the current block is nearing full, kick off `add_block_by_id` for the next
block on a separate task so it is ready (or in-flight) when the boundary is hit. (Async
`complete_file_by_id` or per-file parallel writers are larger alternatives.)

Note: which RPC (`add_block_by_id` vs `complete_file_by_id` vs `BlockWriter::new`) dominates the
1.4–70 s await, and whether the master's raft-journal append is the slow sub-step, was not
distinguishable with plain `perf` on the cluster (off-CPU user call-graphs need bcc/BPF
`offcputime` or a frame-pointer debug build). The structural cause is confirmed regardless.

## Related

- #993 — the "sender dropped" `FsWriterBuffer` brick (already fixed in v0.3.6-alpha); this
  issue is a separate, distinct write-path performance problem.

lzjqsdd:
# Summary

Prefetch the next block (`add_block` + `BlockWriter::new`) while writing the tail of the current block so sequential appends through `curvine-fuse` / buffered writers no longer stall the single per-file `write_future` for a full master+worker round-trip at every `block_size` boundary.

# Issue Describe / Design

Closes #1291

## Root cause

At each block boundary, `FsWriterBase::get_writer` serialized `old.complete()` → `add_block_by_id` (commit piggybacked) → `BlockWriter::new` on the only drain task. The bounded chunk channel filled and concurrent writers parked, matching the observed multi-second idle stalls.

## Fix approach

1. Prefetch when the current write overflows the active block, **or** when `FsWriterBuffer` has more queued chunks and the current block is within the prefetch threshold (keeps overlap for aligned `write_chunk_size` / `block_size`).
2. Exact EOF with no queued follow-up data does not prefetch (avoids empty next-block orphans).
3. At the boundary, install the prefetched writer after completing the old block; flush commits via `CompleteFile(only_flush=true)` because piggybacked `AddBlock` commits are ignored once next is allocated.
4. `discard_prefetch` awaits and `cancel()`s unused writers; `Drop` aborts in-flight tasks or cancels finished writers.
5. Regression tests use non-divisible chunk sizes and/or `async_write` so the prefetch path is actually exercised.

# Changes

| `curvine-client/src/file/fs_writer_base.rs` | Prefetch gates, discard/Drop cleanup, `write_with_more_pending` | Sequential append overlaps next-block alloc/open; drop-without-cleanup no longer leaks finished prefetch writers |
| `curvine-client/src/file/fs_writer_buffer.rs` | Batch queued chunks and pass `more_data_pending` | Aligned chunk/block writes can prefetch from queued follow-up data |
| `curvine-tests/tests/block_test.rs` | Prefetch / EOF / discard regressions | None |

# Test verified

| `make format` | PASS | fmt + clippy (`-D warnings`) |
| `test_sequential_write_across_many_blocks_with_prefetch` | PASS | `write_chunk_size=1500`, `block_size=4096` |
| `test_exact_block_multiple_complete_has_no_extra_empty_block` | PASS | Exact 3×block_size → exactly 3 full blocks |
| `test_prefetch_discard_on_seek_and_flush` | PASS | `async_write` overflow + seek/flush |

# Dependencies

- Related PRs: None
- Related issues: #1291
- Blocks / blocked by: None

### DIFF
--- a/curvine-client/src/file/fs_writer_base.rs
+++ b/curvine-client/src/file/fs_writer_base.rs
@@ -138,7 +138,23 @@ impl FsWriterBase {
         )
     }
 
-    pub async fn write(&mut self, mut chunk: DataSlice) -> FsResult<()> {
+    pub async fn write(&mut self, chunk: DataSlice) -> FsResult<()> {
+        self.write_with_more_pending(chunk, false).await
+    }
+
+    /// Write `chunk`, optionally hinting that more application data is already
+    /// queued behind this chunk (e.g. `FsWriterBuffer` drained multiple items).
+    ///
+    /// Prefetch starts when either:
+    /// - this chunk still overflows the current block, or
+    /// - `more_data_pending` is true and the current block is within the
+    ///   prefetch threshold (so alloc/open overlaps the tail write).
+    /// Exact EOF with no queued follow-up data does not prefetch.
+    pub async fn write_with_more_pending(
+        &mut self,
+        mut chunk: DataSlice,
+        more_data_pending: bool,
+    ) -> FsResult<()> {
         if chunk.is_empty() {
             return Ok(());
         }
@@ -149,11 +165,11 @@ impl FsWriterBase {
 
         let mut remaining = chunk.len();
         while remaining > 0 {
-            // Only prefetch when this write still needs bytes beyond the current
-            // block. Speculative prefetch near EOF would allocate an unused
-            // 0-length next block that complete()/flush()/seek() must discard.
             if let Some(cur) = self.cur_writer.as_ref() {
-                if (remaining as i64) > cur.remaining() {
+                let overflows = (remaining as i64) > cur.remaining();
+                let queued_near_end =
+                    more_data_pending && cur.remaining() <= self.prefetch_threshold;
+                if overflows || queued_near_end {
                     self.maybe_start_prefetch();
                 }
             }
@@ -817,5 +833,33 @@ impl FsWriterBase {
         );
 
         Ok(())
+    }
+}
+
+impl Drop for FsWriterBase {
+    fn drop(&mut self) {
+        let Some(handle) = self.next_block_prefetch.take() else {
+            return;
+        };
+
+        // Prefer cancelling an already-finished prefetched writer. Abort alone
+        // would drop a completed BlockWriter without cancel() and leak the
+        // worker write session. In-flight tasks are aborted so they do not
+        // continue allocating after the writer is gone.
+        if handle.is_finished() {
+            let rt = self.fs_context.clone_runtime();
+            rt.spawn(async move {
+                if let Ok(Ok((_lb, mut writer))) = handle.await {
+                    if let Err(e) = writer.cancel().await {
+                        warn!(
+                            "failed to cancel prefetched writer on FsWriterBase drop: {}",
+                            e
+                        );
+                    }
+                }
+            });
+        } else {
+            handle.abort();
+        }
     }
 }
--- a/curvine-client/src/file/fs_writer_base.rs
+++ b/curvine-client/src/file/fs_writer_base.rs
@@ -16,18 +16,25 @@ use crate::block::BlockWriter;
 use crate::file::{FsClient, FsContext};
 use curvine_common::error::FsError;
 use curvine_common::fs::Path;
-use curvine_common::state::{CommitBlock, FileAllocOpts, FileBlocks, FileStatus, WriteFileBlocks};
+use curvine_common::state::{
+    CommitBlock, FileAllocOpts, FileBlocks, FileStatus, LocatedBlock, WriteFileBlocks,
+};
 use curvine_common::FsResult;
 use fxhash::FxHasher;
 use linked_hash_map::LinkedHashMap;
-use log::warn;
+use log::{debug, warn};
 use orpc::common::FastHashSet;
-use orpc::runtime::{RpcRuntime, Runtime};
+use orpc::runtime::{JoinHandle, RpcRuntime, Runtime};
 use orpc::sys::DataSlice;
 use orpc::{err_box, try_option_mut};
 use std::hash::BuildHasherDefault;
 use std::mem;
 use std::sync::Arc;
+use std::time::Instant;
+
+/// Background prefetch of the next sequential block so boundary switching does
+/// not serialize `add_block` + `BlockWriter::new` on the single write drain task.
+type NextBlockPrefetchHandle = JoinHandle<FsResult<(LocatedBlock, BlockWriter)>>;
 
 pub struct FsWriterBase {
     fs_context: Arc<FsContext>,
@@ -46,6 +53,15 @@ pub struct FsWriterBase {
     /// opened, blocks published by a successful flush, and blocks finalized on
     /// workers but still waiting for their commit metadata to reach the master.
     durable_blocks: FastHashSet<i64>,
+
+    /// Prefetch next block when current remaining bytes fall at or below this.
+    prefetch_threshold: i64,
+    next_block_prefetch: Option<NextBlockPrefetchHandle>,
+    /// Set when a prefetch task may have already allocated the next block on
+    /// the master. Sync allocation must flush commits via CompleteFile first,
+    /// because a repeated AddBlock returns the existing next block and ignores
+    /// piggybacked commit_blocks.
+    prefetch_alloc_attempted: bool,
 }
 
 impl FsWriterBase {
@@ -60,6 +76,10 @@ impl FsWriterBase {
                 .filter(|block| block.block.len > 0 && !block.locs.is_empty())
                 .map(|block| block.block.id)
                 .collect(),
+        );
+        let prefetch_threshold = Self::prefetch_threshold(
+            fs_context.conf.client.write_chunk_size as i64,
+            status.block_size,
         );
         let file_blocks = WriteFileBlocks::new(status);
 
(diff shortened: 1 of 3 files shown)


lzjqsdd:
Thanks for the review. Addressed in 64d9da20:

1. `discard_prefetch` now awaits the prefetch task and `cancel()`s any unused `BlockWriter` instead of bare `abort()`.
2. Prefetch starts only when the current write still overflows the active block (plus the existing remaining threshold), so exact `N * block_size` `complete()` no longer allocates a speculative empty next block.
3. Added regressions: `test_exact_block_multiple_complete_has_no_extra_empty_block` and `test_prefetch_discard_on_seek_and_flush`.

curvine-tests/tests/block_test.rs
```diff
@@ -335,6 +335,164 @@ fn test_remote_random_write_with_replication() -> CommonResult<()> {
...
+    let fs = testing.get_fs(Some(rt.clone()), Some(conf))?;
+    let path = Path::from_str("/sequential_prefetch_many_blocks.data")?;
+
+    rt.block_on(async move {
+        let block_size = 4 * 1024usize;
+        let total = block_size * 8 + 512; // cross 8 full block boundaries
+        let data = BytesMut::from(Utils::rand_str(total).as_bytes());
+
+        let mut writer = fs.create(&path, true).await?;
+        // Write in chunk-sized pieces so remaining() crosses the prefetch threshold.
```
szbr00: This comment is misleading — the prefetch is never actually triggered in any
of the three new tests.

`Writer::write` buffers data up to `write_chunk_size` (1024) and then calls
`write_chunk`, so `FsWriterBase::write` always receives 1024-byte chunks
(or 512 for the final partial chunk). Because 4096 is divisible by 1024,
`cur.remaining()` is always 4096, 3072, 2048, 1024, or 0. The prefetch gate
`(remaining as i64) > cur.remaining()` evaluates to `1024 > 1024` = false at
every block boundary, so `maybe_start_prefetch` is never called.

The same applies to `test_prefetch_discard_on_seek_and_flush`: writing 6144
bytes via `writer.write(&data)` is split into 1024-byte chunks by
`Writer::write`, so the prefetch gate is never true and `discard_prefetch`
is a no-op.

To actually exercise the prefetch path, use a `write_chunk_size` that does
not evenly divide `block_size` (e.g. 1500), or bypass the buffer with
`async_write` using a chunk larger than `block_size`.

curvine-client/src/file/fs_writer_base.rs
```diff
@@ -342,90 +391,320 @@ impl FsWriterBase {
...
+            flush_start.elapsed(),
+            self.path
+        );
+        if result.is_err() {
+            self.restore_commit_blocks(&commit_blocks);
+        }
+        result.map(|_| ())
+    }
+
+    async fn discard_prefetch(&mut self) {
```
szbr00, line 672: If the writer is dropped without calling complete()/flush()/cancel()
(e.g. a panic or early return in user code), the prefetch JoinHandle is
dropped without being awaited or aborted. The spawned task continues
running on the runtime, completing add_block + BlockWriter::new, and the
resulting block allocation and worker write session are leaked because no
one receives the (LocatedBlock, BlockWriter).

The proper cleanup paths (complete/flush/seek/resize/cancel) all call
discard_prefetch correctly, so this only affects the drop-without-cleanup
case. Consider aborting the handle in a Drop guard or documenting that
cancel/complete must be called before drop, since tokio JoinHandle::drop
does not abort the task by default.

curvine-tests/tests/block_test.rs
lzjqsdd: Thanks. Fixed in 86bfa62e / d68e1aba:

1. Regression tests now use `write_chunk_size=1500` (does not divide `block_size=4096`) and a single large `write`, plus `async_write` for the discard cases, so `FsWriterBase` actually enters the prefetch path.
2. `FsWriterBuffer` also batches already-queued chunks and passes `more_data_pending`, so aligned chunk/block configs can still prefetch from follow-up data in the buffer.

curvine-client/src/file/fs_writer_base.rs
lzjqsdd: Thanks. Fixed in 86bfa62e: `FsWriterBase::Drop` now aborts an in-flight prefetch task, or if the task already finished, spawns a cleanup that `cancel()`s the prefetched `BlockWriter` so the worker write session is not leaked when the writer is dropped without `complete`/`flush`/`cancel`.

curvine-client/src/file/fs_writer_base.rs
```diff
@@ -342,90 +378,284 @@ impl FsWriterBase {
...
+        if let Some(handle) = self.next_block_prefetch.take() {
+            handle.abort();
+        }
+    }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. This was addressed earlier in 64d9da20: `discard_prefetch` awaits the join handle and explicitly `cancel()`s a produced `BlockWriter` instead of bare-aborting. Follow-up 86bfa62e also adds `Drop` cleanup for the drop-without-explicit-complete case.

PR #1311, opened 2026-07-26
feat(spdk): add per controller qpair state tracking to qpair pool

thuongle2210:
# Summary

This PR introduces per-controller qpair state tracking in `QpairPool` as the first step toward fixing SPDK qpair exhaustion in issue #1076.

Currently, `QpairPool` caches idle qpairs per controller, but it does not track controller-specific active usage or capacity. This PR adds the basic state needed to model per-controller limits safely before wiring in blocking backpressure in follow-up PRs.

# Key Changes

- Added a new `CtrlQpairState` structure to define per-controller qpair state:
  - `active`: current in-use qpair count
  - `max_active`: per-controller qpair limit
- Extended `QpairPool` with:
  - `ctrl_state: Mutex<HashMap<usize, CtrlQpairState>>`
- Added helper methods:
  - `register_limit(ctrlr_ptr, actual_io_queues)`
  - `controller_stats(ctrlr_ptr) -> (active, max_active)`
- Updated `QpairPool::new()` to initialize controller state tracking
- Added unit coverage for the new controller state behavior:
  - `register_limit_overwrites`
  - `per_controller_active_tracking`
  - `register_limit_rejects_duplicate`

# Impact

- No blocking or retry behavior is introduced in this PR yet
- No qpair reservation or backpressure lifecycle is added yet
- Improves debuggability by making per-controller state visible through controller_stats()
- Lays the foundation for the next steps in the fix by making controller-specific qpair state explicit and testable

# Future Works

- Wire registration of negotiated I/O queue count from SPDK init path into register_limit
- Add reservation-based capacity control for `active < max_active`
- Rework acquire() and release() to support reservation-aware and blocking backpressure behavior under contention
- Add observability metrics for qpair contention, wait duration, and allocation failures

Issue #1076, opened 2026-07-13
[BUG]: 4k random reads fail with EIO under SPDK (QpairPool exhaustion via read-ahead fan-out)


gangump82:
### Describe the bug
Under an SPDK-backed worker, `fio --rw=randread --bs=4k` returns EIO **under any `numjobs`** (including `numjobs=1`); only large sequential reads (`bs=4M`) succeed. The fuse log floods `errno channel closed: : 5`, which is the downstream symptom of an SPDK qpair allocation failure:

```
QpairPool: failed to allocate I/O qpair for ctrlr ... (spdk_env.rs:61)
```

A small 4k read is amplified by read-ahead + the parallel sub-read fan-out in the client read path; each sub-read acquires an SPDK qpair, and once the per-controller qpair ceiling (`max_io_qpairs_per_ctrlr`, default 128 on the target) is hit, `QpairPool::acquire` **hard-fails** instead of blocking/retrying, the parallel sub-read task errors out (`fs_reader_buffer.rs:79 "buffer read(parallel id {}) error"`), the `CallChannel` drops, and the read loop receives `None` → `channel closed` → EIO.

### To Reproduce
1. SPDK-backed worker, target `max_io_qpairs_per_ctrlr = 128`, `io_queues` not tuned.
2. `fio --name=r --filename=/curvine-fuse/fio-spdk/r --rw=randread --bs=4k --ioengine=libaio --iodepth=64 --numjobs=1 --direct=1 --runtime=30 --time_based --group_reporting` → `err=5` EIO.
3. Same command with `--bs=4m --rw=read` succeeds.

Relevant code path:
- `orpc/src/io/spdk_env.rs:39` — `QpairPool::acquire` returns an error (line 61 `"failed to allocate I/O qpair"`) when the controller's qpair ceiling is reached, instead of blocking or retrying.
- `curvine-client/src/file/fs_reader_buffer.rs:72` — `args.rt.spawn(...)` fans a read into parallel sub-reads; `:79` logs the per-sub-read error.
- `curvine-server/src/worker/handler/read_handler.rs:44` — `MAX_READ_AHEAD = 16 * 1024 * 1024` caps `read_ahead_len`; `:81-86` rejects read-ahead above the cap. `read_ahead_len` is client-configurable (`curvine-common/src/conf/client_conf.rs:163`); `read_parallel` defaults to `1` (`client_conf.rs:291`) but is commonly raised for throughput.

### Expected behavior
[...]

PR #1312, opened 2026-07-26
feat(spdk): add cas based qpair reservation to qpair pool

thuongle2210:
# Summary

- This PR adds CAS-based per-controller reservation primitives in QpairPool to prepare capacity control for qpair usage.
- Resolves part of issue #1076.
- Based on pr #1311 

# Key Changes
- Added `try_reserve(ctrlr_ptr) -> bool`
  - uses a CAS loop on `active`
  - reserves a slot only when `active < max_active`
  - returns `false` when the controller is already at capacity
  - logs an error if called for an unregistered controller
- Added `release_reservation(ctrlr_ptr)`
  - decrements the active reservation count
  - warns when release is attempted for an unregistered controller
- Updated `release()` to drop the pool lock before reservation decrement
  - preserves correct lock scope
  - avoids holding the pool mutex longer than necessary
  - ensures active accounting is reduced when a qpair is returned or freed
- Added unit coverage for reservation behavior:
  - `try_reserve_respects_limit`
  - `release_returns_qpair_to_pool`
  - `release_active_count_decremented`

# Impact
- Introduces explicit per-controller capacity accounting in `QpairPool`
- Adds a reservation primitive that enforces active < max_active when used by callers
- Ensures release can decrement reservation count after slot return, assuming the slot was previously reserved
- Does not yet block, wait, or retry when capacity is exhausted
- Lays the groundwork for the next PR to wire reservation state into full backpressure behavior in `acquire()`

# Future Works
- Register the actual negotiated I/O queue count during `SpdkEnv::init()`
- Rework `acquire()` to use reservation failure as the trigger for blocking backpressure instead of immediate allocation failure
- Add wakeup and shutdown-aware coordination around reservation exhaustion
- Add observability metrics for reservation exhaustion, wait duration, and allocation failures,...

### DIFF
--- a/crates/adapters/curvine-storage-spdk/src/spdk_env.rs
+++ b/crates/adapters/curvine-storage-spdk/src/spdk_env.rs
@@ -10,26 +10,83 @@ use orpc::{err_box, err_msg, CommonResult};
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
-use std::sync::atomic::{AtomicU8, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
 use std::sync::{Mutex, OnceLock};
 
+// ---------------------------------------------------------------------------
 // Qpair pool - reuse NVMe qpairs across handles
-/// Lazy allocate, cache on release
+// Lazy allocate, cache on release.
+// Per-controller limits derived from negotiated io_queues of spdk's target
+// ---------------------------------------------------------------------------
+
+// Per-controller qpair state
+struct CtrlQpairState {
+    active: AtomicUsize,
+    max_active: usize,
+}
+
+impl CtrlQpairState {
+    fn new(max_active: usize) -> Self {
+        Self {
+            active: AtomicUsize::new(0),
+            max_active,
+        }
+    }
+}
+
 pub struct QpairPool {
     inner: Mutex<HashMap<usize, Vec<*mut spdk_ffi::spdk_nvme_qpair>>>,
+    /// Per-controller active count and max limit, keyed by controller pointer.
+    ctrl_state: Mutex<HashMap<usize, CtrlQpairState>>,
+    /// Idle cache limit per controller (soft - excess freed on release).
     max_per_ctrlr: usize,
 }
 // SAFETY: exclusive ownership
 unsafe impl Send for QpairPool {}
 unsafe impl Sync for QpairPool {}
 impl QpairPool {
-    /// Default max idle qpairs per controller
+    /// Default max idle qpairs per controller (soft cache limit).
     const DEFAULT_MAX_PER_CTRLR: usize = 16;
 
     fn new() -> Self {
         Self {
             inner: Mutex::new(HashMap::new()),
+            ctrl_state: Mutex::new(HashMap::new()),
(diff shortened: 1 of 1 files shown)


orpc/src/io/spdk_env.rs
```diff
@@ -74,26 +168,29 @@ impl QpairPool {
...
+                stack.push(qpair);
+                log::trace!(
+                    "QpairPool: returned qpair to pool for ctrlr {:p} ({}/{})",
+                    ctrlr,
+                    stack.len(),
+                    self.max_per_ctrlr
+                );
+            }
+        } // inner lock dropped here
+        self.release_reservation(key);
```
szbr00: `release()` always calls `release_reservation()`, but this PR never registers controllers in `SpdkEnv::init()` and `acquire()` never calls `try_reserve()`. In production every qpair return hits the unregistered branch and emits a warning, which will spam logs on the normal open/close path.

Please keep `release_reservation()` out of production `release()` until reservation is fully wired (register + acquire), or gate the call on an actual prior reserve. As written, this couples half of the accounting into the hot path too early.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -12,29 +12,123 @@ use serde::{Deserialize, Serialize};
...
+            error!(
+                "QpairPool: try_reserve called for unregistered ctrlr {:p}",
+                ctrlr_ptr as *const ()
+            );
+            false
+        }
+    }
+
+    /// Release a previously reserved slot.
+    fn release_reservation(&self, ctrlr_ptr: usize) {
```
szbr00, line 122: `release_reservation()` uses unconditional `fetch_sub(1)`. A double-release or release without a matching `try_reserve()` wraps `active` to `usize::MAX` and permanently breaks capacity checks. Consider refusing to decrement when `active == 0` (debug assert / warn + return), or using a saturating decrement, especially since `release()` will call this on every return once controllers are registered.

orpc/src/io/spdk_env.rs
```diff
@@ -12,29 +12,123 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
-use std::sync::atomic::{AtomicU8, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
```
szbr00: `AtomicBool` and `Duration` appear unused here (`poller_is_sleeping` already returns `std::sync::Arc<std::sync::atomic::AtomicBool>`, and other timing paths use fully qualified `std::time::Duration`). Please drop these imports to avoid unused-import warnings.

orpc/src/io/spdk_env.rs
thuongle2210: Ok, I will add release_reservation at later phase.

orpc/src/io/spdk_env.rs
thuongle2210: I refactored it.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
thuongle2210: Thanks, I converted it to safe path: load -> check -> refuse/fetch_sub.

szbr00 (approved):

PR #1313, opened 2026-07-26
feat(spdk): register per controller qpair limits

thuongle2210:
## Summary

- Bridges FFI query capability with reservation infrastructure by integrating per-controller qpair limit registration during SPDK environment initialization. Queries the actual negotiated I/O queue count from each controller after connection and registers it with the qpair pool's reservation system.
- Resolves part of issue #1076.
- Based on pr #1312 

### 1. **C FFI Helper** (`orpc/csrc/spdk_opts_helper.c`)
- New function: `curvine_spdk_ctrlr_get_num_io_queues(ctrlr)`
- Queries the negotiated I/O queue count from the controller's connected options
- Returns the actual negotiated value

### 2. **Rust FFI Binding** (`orpc/src/io/spdk_ffi.rs`)
- Declares `curvine_spdk_ctrlr_get_num_io_queues()` in the extern block

### 3. **Registration During Init** (`orpc/src/io/spdk_env.rs`)
- In `SpdkEnv::init()`, after successfully attaching each controller:
  - Calls the FFI helper to get actual negotiated queue count
  - Logs the comparison between configured and actual values
  - Calls `qpair_pool.register_limit(ctrlr, actual_io_queues)` to populate per-controller state
- This populates per-controller reservation limits at init time so later reservation/backpressure logic can enforce negotiated capacity

## Impact

- Per-controller negotiated capacity is now discovered and registered at runtime.
- Without this registration, reservations had no limits; now they are bounded by SPDK's actual negotiated queue count
- Foundation for blocking backpressure in the next PR

## Future Works
- Implement blocking lifecycle in `acquire()` — when capacity is exhausted (reservation fails), queue the request and unblock after a successful `release()` drops a slot
- Observability metrics


orpc/src/io/spdk_env.rs
```diff
@@ -74,26 +168,29 @@ impl QpairPool {
...
+                stack.push(qpair);
+                log::trace!(
+                    "QpairPool: returned qpair to pool for ctrlr {:p} ({}/{})",
+                    ctrlr,
+                    stack.len(),
+                    self.max_per_ctrlr
+                );
+            }
+        } // inner lock dropped here
+        self.release_reservation(key);
```
szbr00: `release()` always calls `release_reservation()`, but `acquire()` never calls `try_reserve()`. After `register_limit()` runs during init, every production release does `fetch_sub(1)` from 0 and underflows `active` to `usize::MAX`, corrupting per-controller accounting.

Either wire `try_reserve()` into `acquire()` (and roll back on alloc failure) in this PR, or keep `release_reservation()` out of the production `release()` path until the acquire side is landed (e.g. in #1314). As written, this PR is not safe to merge on its own.

orpc/src/io/spdk_env.rs
```diff
@@ -12,29 +12,123 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
-use std::sync::atomic::{AtomicU8, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicUsize, Ordering};
```
szbr00: `AtomicBool` and `Duration` appear unused in this change (`poller_is_sleeping` / shutdown paths already use fully qualified paths). Please drop these imports to avoid unused-import warnings.

orpc/src/io/spdk_env.rs
```diff
@@ -12,29 +12,123 @@ use serde::{Deserialize, Serialize};
...
         Self {
             inner: Mutex::new(HashMap::new()),
+            ctrl_state: Mutex::new(HashMap::new()),
             max_per_ctrlr: Self::DEFAULT_MAX_PER_CTRLR,
         }
     }
 
+    /// Register the per-controller qpair limit from the actual negotiated IO queue count.
+    /// `actual_io_queues` is queried from SPDK after controller initialization
+    fn register_limit(&self, ctrlr_ptr: usize, actual_io_queues: u32) {
```
szbr00, line 61: `register_limit()` replaces the entire `CtrlQpairState`, which resets `active` to 0. That is fine at init today, but the overwrite behavior (also covered by `register_limit_overwrites`) is unsafe if this is ever called again while qpairs are in flight. Consider updating only `max_active`, or documenting/asserting that registration is one-shot before any acquire.

thuongle2210: I updated it from upstream branch and resolved conflicts.

orpc/src/io/spdk_env.rs
thuongle2210: I updated it from upstream branch and resolved conflicts.

orpc/src/io/spdk_env.rs
thuongle2210: I kept the `release_reservation()` out of the production `release()` path from upstream branch and resolved conflicts. Thanks.

thuongle2210:
Hi @szbr9486 ,thanks for your feedback, I resolved conflicts for this PR.

szbr00 (approved):

PR #1314, opened 2026-07-26
feat(spdk): wire backpressure into qpair operations

thuongle2210:
# Summary
- Implements blocking backpressure in `QpairPool::acquire()` to handle qpair pool exhaustion gracefully. When a controller reaches its per-qpair capacity limit, acquisition is blocked (with 30-second timeout) until a previously acquired qpair is released back to the pool. This prevents immediate allocation failures under high concurrency.
- Resolves for issue #1076
- Based on pr #1313

# Change
### 1. Condvar-based Blocking (`QpairPool`)
- New field: `notify: Condvar` - wakes blocked acquirers when qpair is released
- New field: `shutdown: AtomicBool` - rejects new acquisitions during shutdown
- Timeout constant: `ACQUIRE_TIMEOUT = 30s` (configurable future work)

### 2. Two-Path Acquire Logic (`acquire()`)
- Fast path: tries to reserve + check pool (non-blocking, typical case)
- Slow path: waits on `Condvar` with timeout when at capacity
  - Checks shutdown flag to fail fast during graceful shutdown
  - Deadline ensures maximum wait of 30 seconds
  - Attempts reservation after wakeup; if reserved, rechecks cached pool first, then falls back to allocation
  - Falls back to FFI allocation if no cached qpair available

### 3. Release Notification (`release()`)
- After dropping inner pool lock and releasing reservation:
  - Calls `notify.notify_one()` to wake exactly one blocked thread
  - No lock held during notify to avoid contention with woken threads

### 4. Graceful Shutdown (`drain_all()`)
- Sets `shutdown = true` to reject new acquisitions
- Calls `notify.notify_all()` to unblock all waiters
- Logs warning if active/in-flight qpairs still exist

### 5. Test Coverage
- `acquire_fast_path_returns_cached_qpair`: verifies no blocking on pool hit
- `acquire_slow_path_increments_active`: verifies blocking, wakeup, and correct active count
- `acquire_rejects_after_shutdown`: confirms shutdown flag prevents 30s timeout delay
-  remove `cap()`, `drain()`: because their behavior is covered by the new tests.
# Impact
- Reduces instant failures under capacity pressure by blocking up to a bounded timeout before failing
- Best-effort scheduling: blocked threads are woken via notify_one, but strict FIFO ordering is not guaranteed
- Graceful shutdown: drain_all() marks shutdown and wakes waiters so they can fail fast instead of waiting for full timeout

### Scoped internals to pub(crate) to reduce public API exposure.
- make pub(crate) for fields: QpairPool.ctrl_state, QpairPool.max_per_ctrlr, CtrlQpairState.active, CtrlQpairState.max_active.

# Future works
- ACQUIRE_TIMEOUT hardcoded: make configurable via `SpdkConf`
- Mutex on ctrl_state per try_reserve: blocked CAS loop holds `ctrl_state` lock; future optimization with `Arc<CtrlQpairState>` + cache-line alignment (documented TODO)
- Re-check shutdown immediately after wakeup to minimize worst-case delay in missed-notify races
- Observability metrics
- Backoff on qpair contention
- Progressive backoff tuning
- Cap read_parallel by queue budget
- Propagate backpressure to clients
- Qpair warm-up at init

# Reproduction Steps Analysis
## Configuration Setup:
re-use the configuration: `curvine-docker/deploy/spdk/curvine-cluster-spdk.toml` and add/change some configs:
```
[[worker.spdk_disk.targets]]
io_queues = 2 <-- added this config

# enable trace log:

# Log configuration  

[worker]  
log = { level = "trace", log_dir = "stdout", file_name = "worker.log" }   <-- changed to trace

[log]  
level = "trace"  <-- changed to trace
log_dir = "stdout"  
file_name = "curvine.log" 
```
1. Successful Wait & Unblock
- client operation:
image

- worker log:
image

2. Timeout
- client operation:
image
- worker log:
image


crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -611,6 +804,40 @@ impl SpdkEnv {
...
+                            "Target[{}] {}: requested io_queues={}, actual negotiated={}",
+                            i,
+                            target.endpoint(),
+                            target.io_queues,
+                            actual_io_queues
+                        );
+                        self.qpair_pool
+                            .register_limit(first.ctrlr, actual_io_queues);
+                    }
+                    // Register per-controller qpair limit from SPDK's negotiated IO queue count.
```
szbr00: This `register_limit` block is duplicated verbatim from the block immediately above. Besides redundant FFI/log work, `register_limit()` inserts a fresh `CtrlQpairState::new(limit)`, which resets `active` to 0. Please keep a single registration site so a future mid-life re-register cannot silently clear in-flight accounting.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -12,36 +12,142 @@ use serde::{Deserialize, Serialize};
...
+    pub(crate) fn acquire(
         &self,
         ctrlr: *mut spdk_ffi::spdk_nvme_ctrlr,
     ) -> CommonResult<*mut spdk_ffi::spdk_nvme_qpair> {
         let key = ctrlr as usize;
-        {
+        let mut reserved = false;
+
+        // Fast path: check capacity, then try pool without blocking
+        if self.try_reserve(key) {
```
szbr00, line 163: `shutdown` is only checked on the Condvar wait path (`!reserved`). After `drain_all()`, if `active < max_active`, `acquire()` can still `try_reserve()` on the fast path and allocate a new qpair via FFI, which defeats graceful shutdown. Please reject acquires whenever `shutdown` is set (including the fast path and the reserved→allocate path), and add a unit test where capacity is available after `drain_all()`.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -53,10 +159,72 @@ impl QpairPool {
...
+                .unwrap_or_else(|p| p.into_inner())
+                .0;
+            // TODO: check shutdown here to avoid 30s delay if notify_all() was missed
+        }
+
+        // Slot reserved (by fast path or slow path) - allocate the qpair
+        drop(pool);
         let qpair = unsafe { spdk_ffi::curvine_spdk_alloc_io_qpair(ctrlr) };
         if qpair.is_null() {
+            self.release_reservation(key);
```
szbr00, line 255: On FFI alloc failure this path calls `release_reservation()` but never `notify_one()`. Waiters blocked in the slow path will not wake until timeout even though a slot just became free. Mirror `release()` here: release the reservation, then notify.

crates/adapters/curvine-storage-spdk/src/spdk_bdev_test.rs
```diff
@@ -189,6 +191,95 @@ fn spdk_full_lifecycle() {
...
+            pool.entry(ctrlr as usize).or_default().push(q1);
+            pool.entry(ctrlr as usize).or_default().push(q2);
+        }
+
+        // Allocate one more qpair to release when pool is full
+        let q3 = unsafe { spdk_ffi::curvine_spdk_alloc_io_qpair(ctrlr) };
+        assert!(!q3.is_null());
+
+        // release() with pool full -> should free via FFI, not push
+        p.release(ctrlr, q3);
```
szbr00: This calls `release()` without a matching `try_reserve()`/`acquire()`, so `release_reservation()` underflows `active` to `usize::MAX`. Also, `q1`/`q2` remain in the local pool and `QpairPool` has no `Drop` impl, so those real controller qpairs leak and reduce the controller's available IO queues for later phases. Prefer asserting the pool-full free path via a balanced reserve/release (or a test-only helper), and free/drain the cached qpairs before leaving the scope.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
thuongle2210: I added the notify the reduce the waiting time.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
thuongle2210: I made a mistake during conflict resolution and removed a registration

crates/adapters/curvine-storage-spdk/src/spdk_bdev_test.rs
thuongle2210: I updated the tests and applied `try_reserve` and `drain` the cached qpairs.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
thuongle2210: I checked if shutdown is in progress before rejecting the controller acquire request

crates/adapters/curvine-storage-spdk/src/spdk_bdev_test.rs
thuongle2210: I switched to use pure acquire() to release() through the real API

thuongle2210:
Hi @szbr9486 , I resolved all your threads. Thanks!

szbr00 (approved):

Issue #1121, opened 2026-07-14
[FEATURE] fuse: create must assign one ino for both writer map key and handle (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

On create of a new file, the inode is computed by `next_ino` TWICE — once for the writer-map key and once for the handle ino — and the two can diverge, so the writer registered at create time cannot be found later by the handle's ino.

- `curvine-fuse/src/fs/state/node_state.rs:339`-`360` (`get_or_create_writer(None, ...)`: else branch does `let ino = self.next_ino(writer.status()); self.writers.insert(ino, ...)`)
- `curvine-fuse/src/fs/state/node_state.rs:389`-`448` (`new_handle(None, ...)`: after opening the writer, `let ino = ino.unwrap_or(self.next_ino(&status))` with `status = writer.status().clone()`)
- `curvine-fuse/src/fs/dcache/dir_tree.rs` (`next_id(cv_id)`: returns `cv_id` if `cv_id > FUSE_ROOT_ID && cv_id != FUSE_UNKNOWN_INO && !contains_key(cv_id)`, else `id_creator.next()`)

**Verified — condition-triggered**

- `fs_create` → `new_handle(None, ...)` → for WRONLY/RDWR calls `get_or_create_writer(None, ...)`, which computes `next_ino(writer.status())` as the writers-map key.
- `new_handle` then computes a SECOND `next_ino(&status)` for the handle ino.
- When `writer.status().id` is a valid stable id (`> FUSE_ROOT_ID`, not UNKNOWN), both `next_ino` calls return that same `status.id` → keys match (the common case, once the backend has assigned an id).
- But when `status.id <= FUSE_ROOT_ID` or `== FUSE_UNKNOWN_INO` (a freshly created file whose backend id is not yet assigned), both calls fall through to `id_creator.next()`, returning **two different inos** → the writer-map key ≠ `handle.ino()`.

**Impact (when they diverge)**

- `find_writer(handle.ino())` / `release(handle.ino())` cannot find the writer registered at create → flush/complete/len updates are lost, and the writer map entry leaks.

**Describe the solution you'd like**

- Determine the unique ino once in the create flow, then create the writer AND the handle with that same ino.
- `get_or_create_writer(None, ...)` should not allocate an inode itself; the caller should pass the resolved ino.

**Additional context**

[...]

PR #1315, opened 2026-07-26
chore(fuse): trim redundant comments, keep load-bearing ones

yangcx000:
## What

Trim redundant doc-comments and inline comments across the `curvine-fuse` crate while keeping the load-bearing ones (SAFETY contracts, `#[repr(C)]`/ABI notes, issue-traceability references, and counterintuitive-decision rationale).

This is a **comments-only** change: no code, `#[cfg]` gate, attribute, or test logic is modified.

## Why

During review of the FUSE daemon it became clear that a large share of comments merely restated the code or narrated obvious control flow, adding noise without information. This cleanup removes that noise so the remaining comments (safety invariants, protocol/kernel behavior, "deliberately NOT doing X" rationale, bug-fix issue tags) stand out.

## Scope / guarantees

- 26 files under `curvine-fuse/src/`, comments only.
- Verified that comment-stripping both the base and this branch yields byte-identical source — i.e. **zero non-comment lines changed** in either direction.
- Load-bearing comments explicitly retained: `struct_as_bytes` SAFETY/`#[repr(C)]` contract, the `#1122` `FUSE_ATOMIC_O_TRUNC` / `FUSE_EXPORT_SUPPORT` allowlist rationale, and issue traceability (`#1116`/`#1118`/`#1121`/`#1122`/`#1215`/`#1221`).

## Testing

- `cargo fmt --check` — clean
- `cargo clippy -p curvine-fuse --all-targets` — 0 warnings
- `cargo build -p curvine-fuse` — clean
- `cargo test -p curvine-fuse --lib` — 302 passed

### DIFF
--- a/curvine-fuse/src/session/fuse_response.rs
+++ b/curvine-fuse/src/session/fuse_response.rs
@@ -158,7 +158,11 @@ impl FuseResponse {
             Some(slot) => slot,
         };
 
-        // Reserve first so bounded-channel cancellation cannot silently finish a request.
+        // Reserve first, without touching the slot: a bounded send().await can
+        // suspend on a full channel, and a cancel there would drop the ActiveGuard
+        // (decrementing active_requests) while emitting NO terminal metric. The
+        // reserve is the only suspend point; if cancelled the slot stays unfinished
+        // and the guard is dropped cleanly, no half-finished state.
         if self.sender.is_bounded() {
             let permit = match self.sender.reserve().await {
                 Ok(p) => p,
--- a/curvine-fuse/src/session/fuse_response.rs
+++ b/curvine-fuse/src/session/fuse_response.rs
@@ -55,15 +55,11 @@ impl ResponseData {
     pub fn as_iovec(&self) -> IOResult<(usize, Vec<IoSlice<'_>>)> {
         let mut iovec: Vec<IoSlice<'_>> = Vec::with_capacity(self.data.len() + 1);
 
-        // write header
         let header_bytes = FuseUtils::struct_as_bytes(&self.header);
         iovec.push(IoSlice::new(header_bytes));
 
-        // write data
         for data in &self.data {
-            // FUSE iovec responses can only contain memory-backed data. An
-            // IOSlice refers to an fd-backed region and cannot be represented
-            // as an std::io::IoSlice without a separate transfer path.
+            // FUSE iovec replies require memory-backed data, not fd-backed IOSlice regions.
             if matches!(data, DataSlice::IOSlice(_)) {
                 return orpc::err_box!(
                     "DataSlice::IOSlice is not supported in FUSE iovec responses"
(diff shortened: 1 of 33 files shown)


curvine-fuse/src/web_server.rs
```diff
@@ -31,17 +45,10 @@ impl WebServer {
...
-// event-driven, so the scrape is a pure `text_output()` with no map traversal
-// under read locks.
 async fn metrics_handler(State(_): State<Arc<NodeState>>) -> String {
     let fuse_metrics = FuseMetrics::get();
 
-    // Scrape hygiene: time the render + record output size. Last-scrape semantics
-    // — the values in *this* response reflect the PREVIOUS scrape (order
-    // text_output THEN record_scrape; do not reorder). Covers the `text_output()`
-    // render only, not Axum's `State` extraction (tiny, out of scope).
+    // Record scrape duration/size after rendering, so this response shows previous scrape values.
```
szbr00: The original comment here called out "do not reorder" explicitly — swapping text_output() and record_scrape() would silently make this response reflect the current scrape instead of the previous one. The compressed "after rendering" phrasing implies the order, but consider keeping an explicit "do not reorder" note since the semantic consequence of reordering is non-obvious.

curvine-fuse/src/session/fuse_response.rs
```diff
@@ -191,26 +158,17 @@ impl FuseResponse {
...
         };
 
-        // Cancellation safety on bounded channels: a bounded `send().await` can
-        // suspend when full, and cancellation there would drop the guard (dec'ing
-        // `active_requests`) while emitting NO terminal metric. So on bounded
-        // channels we `reserve()` a permit FIRST without touching the slot (the only
-        // suspendable point); if cancelled there the slot is untouched and the
-        // request is cleanly dropped. Once the permit is in hand we commit the slot
-        // and `permit.send(...)` synchronously. Unbounded `send` is already sync.
+        // Reserve first so bounded-channel cancellation cannot silently finish a request.
```
szbr00: The original comment described the specific failure mode this guards against: a bounded send().await suspended on a full channel, if cancelled, would drop the ActiveGuard (decrementing active_requests) without emitting a terminal metric. The compressed version captures the "what" but not the "why" — a future maintainer refactoring the reserve-first discipline may not realize the metric inconsistency it prevents. Consider keeping a brief note on the failure mode.

curvine-fuse/src/web_server.rs
yangcx000: Good catch — restored the explicit "do not reorder" note in f45b6fc. The compressed version now states that the text_output THEN record_scrape order is what yields last-scrape semantics, and that reordering would silently make the response reflect the current scrape instead of the previous one.

curvine-fuse/src/session/fuse_response.rs
yangcx000: Agreed — restored the failure mode in f45b6fc. The comment now spells out the "why": a bounded send().await can suspend on a full channel, and a cancel there would drop the ActiveGuard (decrementing active_requests) while emitting no terminal metric — which is exactly the silent finish the reserve-first discipline prevents.

szbr00 (approved):

PR #1316, opened 2026-07-26
fix(fuse): align mode and protected-link permission checks with POSIX (#1269)

### LINKED ISSUE #1269: [BUG]: FUSE permission and mode-bit semantics diverge from POSIX LTP expectations

Multiple LTP fs_perms and syscall cases on curvine-fuse report incorrect mode bits, unexpected success on restricted operations, or wrong errno.

chmod05 reports Incorrect modes; lchown02/lstat02 return success when EACCES expected; truncate03 succeeds when denied.

Reproduction:

1. Mount curvine-fuse.
2. Run LTP chmod05 and lchown02.
3. Observe mode/permission TFAIL.

**Impact**

Broad POSIX compatibility gap on FUSE mounts.

Permission and mode operations match Linux tmpfs behavior per LTP.

Align chmod/fchmod/stat/lchown/truncate permission enforcement on FUSE with Linux.

**Additional context**

Root-cause hypothesis: curvine-fuse chmod/fchmod/stat/lchown/truncate permission checks diverge from POSIX (hypothesis: incorrect mode masking and missing EACCES enforcement).

Evidence summary: 22 LTP mode-permission family cases fail on /curvine-fuse with wrong mode bits or unexpected success.

Acceptance / test gaps:

- mode-permission LTP suite passes on /curvine-fuse

LJF-050:
# Summary

Align Curvine FUSE path/mode permission checks with POSIX and Linux sticky-directory link policy so chmod/fchmod/stat/truncate/utime and protected symlink/hardlink cases behave correctly under `check_permission`.

# Issue Describe / Design

Closes #1269

## Root cause

1. Mount-root dcache used an invalid ACL placeholder (`mode=0`), so non-root path traverse always failed with `EACCES`.
2. Non-zero kernel entry/attr timeouts let LOOKUP/GETATTR succeed after a parent lost execute permission.
3. Truncate lacked a write check; utime(2) did not follow owner/`*_NOW`+write rules; readlink lacked path traverse checks.
4. Sticky world-writable `protected_symlinks` / `protected_hardlinks` were not enforced in userspace, so LTP `prot_hsymlinks` allowed forbidden follows and hardlinks.
5. Hardlink dcache rewrote the primary parent/name and marked shared inodes deleted too early, causing cleanup `ENOENT` after the functional checks passed.

## Fix approach

- Hydrate root ACL for traverse; trust only valid cached ACL bits.
- Zero kernel entry/attr timeouts when `check_permission` is enabled.
- Enforce truncate write permission, utime(2) rules, and readlink traverse checks.
- Apply Linux sticky-dir symlink follow and safe-hardlink policy in `readlink`/`link`.
- Keep hardlink primary path stable; mark_delete only on the last link; make dcache-miss unlink idempotent on the master.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | POSIX setattr/utime/truncate/traverse checks; protected symlink/hardlink helpers | Stricter permission denials matching Linux when `check_permission` is on |
| `curvine-fuse/src/fuse_utils.rs` | Zero kernel cache timeouts under `check_permission` | Parents revalidated after chmod -x |
| `curvine-fuse/src/lib.rs` | Expose `FATTR_FH` | FH-based setattr can skip path traverse |
| `curvine-fuse/src/fs/dcache/inode.rs` | Root placeholder mode `0o777` | Non-root traverse no longer fail-closed at mount root |
| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Hardlink primary path stable; last-link mark_delete | Correct hardlink cleanup / LOOKUP |
| `curvine-fuse/src/fs/state/node_state.rs` | Idempotent unlink on dcache miss; clear unlink state by name | Stale dentry cleanup no longer returns spurious `ENOENT` |

# Test verified

| `make format` | PASS | fmt + clippy gate |
| `cargo test -p curvine-fuse --lib` (permission/protected/hardlink helpers) | PASS | Unit tests for new helpers |
| LTP mode/permission focused suite (22 assigned cases) | PASS | Including `prot_hsymlinks` |
| Full LTP profile on fixed HEAD (3 independent runs) | PASS for assigned set | Assigned cases absent; no new failures vs baseline; historical unrelated failures unchanged |

# Dependencies

- Related PRs: None
- Related issues: #1269
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -456,17 +456,27 @@ impl CurvineFileSystem {
             }
         };
 
+        let meta_ttl = self.conf.meta_cache_ttl.as_millis() as u64;
         while dir_ino != 0 {
             let check_header = fuse_in_header {
                 uid: header.uid,
                 gid: header.gid,
+                pid: header.pid,
                 nodeid: dir_ino,
                 ..Default::default()
             };
+            // Only trust ACL bits from a valid dcache entry. The mount-root placeholder
+            // starts as Lifecycle::Invalid; using it fail-closes every non-root traverse
+            // with EACCES (breaks LTP chmod/chown under nobody).
             let cached_status = {
                 let dir = self.state.dir_read();
-                dir.get_inode(dir_ino, None)
-                    .map(|inode| inode.clone_status())
+                dir.get_inode(dir_ino, None).and_then(|inode| {
+                    if inode.cache_valid(meta_ttl) {
+                        Some(inode.clone_status())
+                    } else {
+                        None
+                    }
+                })
             };
             if let Some(status) = cached_status {
                 self.check_access_permissions(&status, &check_header, libc::X_OK as u32)?;
@@ -501,6 +511,7 @@ impl CurvineFileSystem {
             status.mode,
             header.uid,
             header.gid,
+            header.pid,
             file_uid,
             file_gid,
         );
@@ -572,19 +583,23 @@ impl CurvineFileSystem {
         }
     }
 
-    /// Determine which permission bits to check based on user relationship to file
+    /// Determine which permission bits to check based on user relationship to file.
+    ///
+    /// Group membership includes the caller's effective gid and supplementary groups
+    /// from `/proc/<pid>/status` (same source as setattr setgid checks).
     fn get_effective_permission_bits(
         &self,
         mode: u32,
         current_uid: u32,
         current_gid: u32,
+        caller_pid: u32,
         file_uid: u32,
         file_gid: u32,
     ) -> u32 {
         if current_uid == file_uid {
             // Owner permissions (bits 8-10)
             (mode >> 6) & 0o7
-        } else if current_gid == file_gid {
+        } else if FuseUtils::caller_in_file_group(current_gid, file_gid, caller_pid) {
             // Group permissions (bits 5-7)
             (mode >> 3) & 0o7
         } else {
@@ -1089,7 +1104,10 @@ impl fs::FileSystem for CurvineFileSystem {
 
         let mode_only = (op.arg.valid & (FATTR_UID | FATTR_GID | FATTR_SIZE)) == 0
             && (op.arg.valid & FATTR_MODE) != 0;
-        let skip_traverse = op.arg.fh != 0 || (mode_only && op.header.uid == file_uid);
+        // Path-based chmod must still traverse; fh-based fchmod/fchown must not
+        // (POSIX open-file descriptors do not re-check path search permission).
+        let skip_traverse =
+            (op.arg.valid & FATTR_FH) != 0 || (mode_only && op.header.uid == file_uid);
 
         if !skip_traverse {
             self.check_traverse_permissions(op.header.nodeid, op.header)
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1224,12 +1224,11 @@ impl fs::FileSystem for CurvineFileSystem {
             None
         };
 
-        let mode_only = (op.arg.valid & (FATTR_UID | FATTR_GID | FATTR_SIZE)) == 0
-            && (op.arg.valid & FATTR_MODE) != 0;
-        // Path-based chmod must still traverse; fh-based fchmod/fchown must not
-        // (POSIX open-file descriptors do not re-check path search permission).
-        let skip_traverse =
-            (op.arg.valid & FATTR_FH) != 0 || (mode_only && op.header.uid == file_uid);
+        // Path-based chmod(2)/chown(2)/truncate must still traverse the path prefix
+        // regardless of file ownership (POSIX search permission). Only fh-based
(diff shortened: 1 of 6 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1594,6 +1787,25 @@ impl fs::FileSystem for CurvineFileSystem {
             return err_fuse!(libc::EINVAL, "Not a symbolic link: {}", status.path);
         }
 
+        // fs.protected_symlinks (LTP prot_hsymlinks): VFS may_follow_link is not
```
szbr00, line 1713: FUSE READLINK is issued both when the kernel follows a symlink during path
resolution and when userspace calls readlink(2). Applying
check_protected_symlink_follow here means readlink(2) on a symlink owned by
another user in a sticky world-writable directory (e.g. /tmp) returns EACCES,
even though POSIX readlink(2) only requires search permission on the path
prefix, which is already checked above via check_traverse_permissions.

Since FUSE cannot distinguish follow from explicit readlink, this tradeoff is
understandable for prot_hsymlinks compatibility. Consider documenting the
readlink(2) side effect, or gating the check behind a condition that can
separate the two cases if one becomes available.

curvine-fuse/src/fuse_utils.rs
```diff
@@ -557,12 +578,39 @@ impl FuseUtils {
...
-        effective_gid == file_gid
+    /// `fuse_in_header` exposes only the effective gid; chmod(2) setgid handling and
+    /// chown gid changes must also consider supplementary groups of the caller pid.
+    pub fn caller_supplementary_groups(pid: u32) -> Vec<u32> {
+        if pid == 0 {
+            return Vec::new();
+        }
+
+        let status_path = format!("/proc/{pid}/status");
+        let Ok(content) = std::fs::read_to_string(status_path) else {
```
szbr00, line 563: caller_supplementary_groups reads /proc/<pid>/status on every call, and
caller_in_file_group is invoked from get_effective_permission_bits which runs
on every permission check. In check_traverse_permissions this is called once
per parent directory in the path prefix, so a deep path can trigger O(depth)
procfs reads per operation.

Consider caching the resolved groups per pid (e.g. a small LRU or thread-local
with a short TTL) to avoid repeated reads within the same request and across
requests for the same pid.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1088,21 +1223,54 @@ impl fs::FileSystem for CurvineFileSystem {
...
-            op.arg.valid,
-            target_gid,
-        )?;
+
+        let mode_only = (op.arg.valid & (FATTR_UID | FATTR_GID | FATTR_SIZE)) == 0
+            && (op.arg.valid & FATTR_MODE) != 0;
+        // Path-based chmod must still traverse; fh-based fchmod/fchown must not
+        // (POSIX open-file descriptors do not re-check path search permission).
+        let skip_traverse =
+            (op.arg.valid & FATTR_FH) != 0 || (mode_only && op.header.uid == file_uid);
```
szbr00: The skip_traverse condition includes `mode_only && op.header.uid == file_uid`,
which skips path-prefix search permission checks for owner-initiated chmod(2)
by path. POSIX chmod(2) requires search permission on all path components
regardless of file ownership — only fchmod(2) (by fd, covered by the FATTR_FH
case) is exempt.

The comment explains the FATTR_FH skip but not the mode_only-owner skip. If
this is needed for a specific LTP case, consider documenting the rationale;
otherwise the traverse check should apply to path-based chmod regardless of
ownership.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks for the careful analysis. Agreed: FUSE READLINK is used for both symlink follow and explicit readlink(2), so we cannot apply protected_symlinks only on follow.

We intentionally keep `check_protected_symlink_follow` here for `fs.protected_symlinks` / LTP `prot_hsymlinks` compatibility, accepting the stricter readlink(2) EACCES side effect on sticky world-writable directories. The inline comment already notes this FUSE limitation; if the kernel later exposes a follow-vs-readlink distinction we can gate the check. No code change in this round.

curvine-fuse/src/fuse_utils.rs
LJF-050: Thanks. Agreed that `caller_supplementary_groups` can cause O(depth) `/proc/<pid>/status` reads under path-prefix traverse checks.

For this correctness-focused change we are keeping the simple per-call procfs read to avoid stale-group races and extra caching complexity. A small per-pid LRU/TTL cache is a reasonable follow-up performance improvement; tracking it separately rather than expanding this PR.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks — good catch. POSIX chmod(2) requires search permission on every path component regardless of file ownership; only fchmod (FATTR_FH) should skip traverse.

Fixed in d6f966d1: removed the `mode_only && owner` skip so path-based chmod always runs `check_traverse_permissions`. Comment updated to match.

szbr00 (approved):

PR #1317, opened 2026-07-27
fix(fuse): enforce parent write/search checks before unlink (#1285)

### LINKED ISSUE #1285: [BUG]: FUSE unlink08 allows unlink on protected directories

Case unlink08 fails on curvine-fuse with POSIX semantic mismatch.

unlink08.c:41: TFAIL: unlink(<unwritable directory>) succeeded unexpectedly | unlink08.c:41: TFAIL: unlink(<unsearchable directory>) succeeded unexpectedly

Reproduction:

1. Start Curvine cluster and mount curvine-fuse.
2. Run LTP/FUSE case unlink08.
3. Observe failure versus Linux reference.

**Impact**

Operations match Linux reference behavior per LTP.

Align FUSE behavior with POSIX/Linux reference semantics for affected syscalls.

**Additional context**

Root-cause hypothesis: unlink08 indicates fuse unlink defect (hypothesis).

Evidence summary: unlink08.c:41: TFAIL: unlink(<unwritable directory>) succeeded unexpectedly | unlink08.c:41: TFAIL: unlink(<unsearchable directory>) succeeded unexpectedly

Acceptance / test gaps:

- FUSE-UNLINK affected cases pass

LJF-050:
# Summary

Enforce POSIX `may_delete` parent-directory write and search permission checks on Curvine FUSE `unlink` so protected parents reject unlink with `EACCES` under `check_permission`.

# Issue Describe / Design

Closes #1285

## Root cause

With `check_permission` enabled, FUSE `unlink` did not verify that the caller has write and search permission on the parent directory. As a result, LTP `unlink08` could remove files through unwritable (`0555`) or unsearchable (`0666`) parents.

## Fix approach

Before deleting the name, require `W_OK|X_OK` on the parent inode (`op.header.nodeid`), matching Linux VFS `may_delete` for the parent-permission portion of unlink. Sticky-bit ownership rules remain out of scope for this change.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | Call `check_permissions(..., W_OK\|X_OK)` at the start of `unlink`; add unit coverage for the parent mask | Unlink through 0555/0666 parents now returns `EACCES` when `check_permission` is on |

# Test verified

| `make format` | PASS | fmt + clippy gate |
| `cargo test -p curvine-fuse --lib unlink_parent_may_delete` | PASS | Parent W/X mask unit test |
| LTP `unlink08` on fixed HEAD | PASS | Previously failed parent-permission assertions now pass |
| Full LTP profile on fixed HEAD | PASS for assigned set | `unlink08` absent; no new failure identities vs baseline; cleanup passed |

# Dependencies

- Related PRs: None
- Related issues: #1285
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1489,6 +1489,10 @@ impl fs::FileSystem for CurvineFileSystem {
 
     async fn unlink(&self, op: Unlink<'_>) -> FuseResult<()> {
         let name = try_option!(op.name.to_str());
+        // Linux VFS may_delete requires MAY_WRITE|MAY_EXEC on the parent directory
+        // (LTP unlink08: unwritable 0555 and unsearchable 0666 parents must EACCES).
+        self.check_permissions(op.header, (libc::W_OK | libc::X_OK) as u32)
+            .await?;
         let path = self.state.get_path_common(op.header.nodeid, Some(name))?;
         self.ensure_writable_path(&path, RpcCode::Delete).await?;
         self.state.fs_unlink(op.header.nodeid, name).await?;
@@ -1829,6 +1833,39 @@ mod tests {
         assert!(CurvineFileSystem::posix_access_requires_mode_check(
             1000,
             libc::R_OK as u32
+        ));
+    }
+
+    #[test]
+    fn unlink_parent_may_delete_requires_write_and_search() {
+        use super::CurvineFileSystem;
+
+        // other class of 0555 (r-x): searchable but not writable
+        let unwritable_other = 0o5;
+        assert!(!CurvineFileSystem::permission_mask_allows(
+            unwritable_other,
+            (libc::W_OK | libc::X_OK) as u32
+        ));
+        assert!(CurvineFileSystem::permission_mask_allows(
+            unwritable_other,
+            libc::X_OK as u32
+        ));
+
+        // other class of 0666 (rw-): writable but not searchable
+        let unsearchable_other = 0o6;
+        assert!(!CurvineFileSystem::permission_mask_allows(
+            unsearchable_other,
+            (libc::W_OK | libc::X_OK) as u32
+        ));
+        assert!(CurvineFileSystem::permission_mask_allows(
+            unsearchable_other,
+            libc::W_OK as u32
+        ));
+
+        // other class of 0777 (rwx): may_delete allowed
+        assert!(CurvineFileSystem::permission_mask_allows(
+            0o7,
+            (libc::W_OK | libc::X_OK) as u32
         ));
     }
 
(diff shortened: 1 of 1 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1489,6 +1489,10 @@ impl fs::FileSystem for CurvineFileSystem {
 
     async fn unlink(&self, op: Unlink<'_>) -> FuseResult<()> {
         let name = try_option!(op.name.to_str());
+        // Linux VFS may_delete requires MAY_WRITE|MAY_EXEC on the parent directory
+        // (LTP unlink08: unwritable 0555 and unsearchable 0666 parents must EACCES).
+        self.check_permissions(op.header, (libc::W_OK | libc::X_OK) as u32)
```
szbr00, line 1494: The parent W/X check here is correct and matches VFS may_delete. However,
rm_dir (below) performs the same name-deletion operation and Linux vfs_rmdir
also calls may_delete, requiring MAY_WRITE|MAY_EXEC on the parent. rm_dir
currently has no check_permissions call, so rmdir through an 0555/0666 parent
still succeeds — the same class of gap this PR closes for unlink.

Consider applying the same check at the start of rm_dir for consistency.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1832,6 +1836,39 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn unlink_parent_may_delete_requires_write_and_search() {
```
szbr00: This test only calls CurvineFileSystem::permission_mask_allows, a pure static
utility. It never drives unlink or check_permissions, so it would still pass
if the permission check were removed from unlink. The existing
access01_mode_masks_deny_root_x_ok_on_non_executable_files test already
exercises permission_mask_allows with W_OK|X_OK.

Consider either renaming this test to reflect that it covers the bit-mask
helper, or adding a test that exercises the unlink path and asserts EACCES
for 0555/0666 parents (the LTP unlink08 scenario the PR description cites).

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Addressed in ca7e7df: `rm_dir` now calls `check_permissions` with `W_OK|X_OK` on the parent before delete, matching Linux `vfs_rmdir`/`may_delete` and the existing `unlink` path.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Partial address in ca7e7df: renamed the unit test to `may_delete_permission_mask_requires_write_and_search` so it clearly documents that it covers the `permission_mask_allows` bit-mask helper used by both `unlink` and `rm_dir`, not the full FUSE unlink path. A harness-level test that drives unlink/rmdir and asserts `EACCES` for 0555/0666 parents under `check_permission` is still better left to LTP/integration coverage (unlink08 / rmdir equivalents), since constructing a full `CurvineFileSystem` permission path in this unit-test module is out of scope for this PR.

szbr00 (approved):

PR #1320, opened 2026-07-27
fix(worker): preserve finalized blocks when writes abort (#1318)

### LINKED ISSUE #1318: [BUG]: Aborting an in-place rewrite can delete a previously finalized block
## Describe the bug

When an existing finalized block is reopened for append or random write, the Worker modifies the committed physical block file in place. If the write later fails or is cancelled, the abort path removes the physical block.

Failures can occur after `open_writer`, resize, short-circuit setup, partial writes, flush, or final-length validation. Because there is no rollback generation, aborting a rewrite can either delete the old committed block or leave it partially modified.

## To Reproduce

1. Create and complete a file containing at least one block.
2. Reopen the same block for append or random write.
3. Write part of the requested data.
4. Inject a Worker/client error or cancel the write before successful completion.
5. Read the previously committed block or inspect its local path.

The old block may be missing or contain partial new data even though the rewrite did not complete.

## Expected behavior

- Aborting a new block may remove its unpublished temporary generation.
- Aborting a rewrite must preserve the previously committed generation byte-for-byte.
- Failed or cancelled bytes must never become visible to readers.
- The Worker must explicitly distinguish a new write from a rewrite.

## Environment

- OS: Platform-independent for file-backed Worker storage
- Revision: `main` at `b5db9cb`
- Storage types: `DISK`, `SSD`, and `MEM` file-backed modes

## Additional context

Relevant paths include Worker writer opening/resizing, short-circuit setup, flush/final-length handling, and block abort/removal.

The recommended fix is generation-specific copy-on-write staging: copy or reflink the committed block into a new staging generation, modify only that generation, and publish it atomically after validation.

no-teet-on:
# Summary

Prevent aborted file-backed block rewrites from deleting or corrupting previously finalized data. New writes and rewrites now use staging files, while the committed block remains readable until the new write is successfully published.

# Issue Describe / Design

Closes #1318

Root cause:

- `Writing` and `Finalized` blocks shared the same active physical path.
- Rewrites modified committed data in place.
- `abort_block` unconditionally removed the current physical block.

Fix approach:

- Store file-backed `Writing` blocks under `staging/`.
- Copy committed data to staging before a rewrite.
- Keep the committed metadata readable and reportable while rewriting.
- On abort, delete only staging and restore committed metadata.
- On finalize, validate length and atomically rename staging to active.
- On restart, prefer an existing finalized active generation and discard an incomplete staging rewrite with the same block ID.
- Account for committed and staging capacity independently.
- Remove both active and staging files when deleting the logical block.
- Keep SPDK behavior unchanged.

Known trade-off:

- The initial committed-to-staging copy is serialized by the dataset write lock to fence concurrent metadata transitions. Existing open readers continue reading the active file, while new metadata operations can wait for a large copy to finish. Moving the copy outside the lock requires a separate fenced prepare/copy/commit lifecycle.

# Changes

| Worker file layout | Route writing blocks through staging and publish with rename | File-backed writes gain an explicit publication boundary |
| Worker startup scan | Reconcile active/staging rewrite conflicts in favor of committed active data | Interrupted rewrites no longer shadow committed data or reserve capacity twice |
| Worker dataset | Track committed rewrite state and restore it on abort | Failed rewrites preserve old data |
| Block store | Return committed data during an active rewrite | Readers do not observe partial rewrite data |
| Layout abstraction | Describe whether a layout preserves committed data | SPDK behavior remains unchanged |
| Worker tests | Add abort, finalize, capacity, deletion, and restart-recovery coverage | Prevents regression |

# Test verified

| `cargo test -p curvine-server restart_during_rewrite_discards_staging_and_keeps_committed --lib --no-default-features` | PASS | Linux/Rust 1.92, 1 passed |
| `cargo test -p curvine-server worker:: --lib --no-default-features` | PASS | Linux/Rust 1.92, 58 passed |
| `make format` | PASS | Linux; includes release Clippy and formatting |

# Dependencies

- Related PRs: None
- Related issues: #1318
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/storage/layout/file_layout.rs
+++ b/curvine-server/src/worker/storage/layout/file_layout.rs
@@ -21,7 +21,7 @@ use orpc::common::ByteUnit;
 use orpc::common::FileUtils;
 use orpc::io::{BlockDevice, IOError, IOResult, LocalFile};
 use orpc::{err_box, try_err, CommonResult};
-use std::fs::{self, File};
+use std::fs::{self, OpenOptions};
 use std::path::PathBuf;
 
 #[derive(Clone, Copy)]
@@ -47,7 +47,7 @@ impl FileLayout {
 
     fn block_dir(dir: &VfsDir, meta: &BlockMeta) -> CommonResult<PathBuf> {
         let path = match meta.state() {
-            BlockState::Finalized | BlockState::Writing => {
+            BlockState::Finalized => {
                 let uid = meta.id() as u64;
                 let d1 = (uid >> 48) & 0x1F;
                 let d2 = (uid >> 32) & 0x1F;
@@ -55,7 +55,7 @@ impl FileLayout {
                     .join(format!("b{}", d1))
                     .join(format!("b{}", d2))
             }
-            BlockState::Recovering => Self::staging_dir(dir),
+            BlockState::Writing | BlockState::Recovering => Self::staging_dir(dir),
         };
 
         if path.exists() {
@@ -89,10 +89,14 @@ impl FileLayout {
 }
 
 impl BlockLayout for FileLayout {
+    fn preserves_committed_on_write(&self) -> bool {
+        true
+    }
+
     fn allocate(&self, dir: &VfsDir, block: &ExtendedBlock) -> CommonResult<BlockMeta> {
         let meta = BlockMeta::with_tmp(block, dir);
         let file = Self::block_path(dir, &meta)?;
-        File::create(file)?;
+        OpenOptions::new().write(true).create_new(true).open(file)?;
         Ok(meta)
     }
 
@@ -106,10 +110,25 @@ impl BlockLayout for FileLayout {
             return err_box!("Invalid file block size: {}", block.len);
         }
         let mut prepared = BlockMeta::new(meta.id(), block.len, dir);
-        // The old file still occupies space until resize/finalize. Keep the
-        // larger physical charge so preparing a smaller write cannot temporarily
-        // overstate available capacity or under-release on abort.
         prepared.actual_len = meta.actual_len.max(block.len);
+
+        if meta.is_final() {
+            let committed_path = Self::block_path(dir, meta)?;
+            let staging_path = Self::block_path(dir, &prepared)?;
+            if staging_path.exists() {
+                return err_box!(
+                    "Cannot rewrite block {} because staging file {} already exists",
+                    meta.id(),
+                    staging_path.display()
+                );
+            }
+
+            if let Err(e) = fs::copy(&committed_path, &staging_path) {
+                let _ = fs::remove_file(&staging_path);
+                return Err(e.into());
+            }
+        }
+
         Ok(prepared)
     }
 
@@ -117,10 +136,22 @@ impl BlockLayout for FileLayout {
         &self,
         dir: &VfsDir,
         meta: &BlockMeta,
-        _committed_len: i64,
+        committed_len: i64,
     ) -> CommonResult<BlockMeta> {
-        let path = Self::block_path(dir, meta)?;
-        BlockMeta::with_final(meta, &path)
+        let staging_path = Self::block_path(dir, meta)?;
+        let final_meta = BlockMeta::with_final(meta, &staging_path)?;
+        if final_meta.len() != committed_len {
+            return err_box!(
+                "Block {} length mismatch, expected: {}, actual: {}",
+                meta.id(),
+                committed_len,
+                final_meta.len()
+            );
+        }
+
+        let active_path = Self::block_path(dir, &final_meta)?;
+        FileUtils::rename(staging_path, active_path)?;
+        Ok(final_meta)
     }
 
     fn scan(&self, dir: &VfsDir) -> CommonResult<Vec<BlockMeta>> {
--- a/curvine-server/src/worker/storage/layout/file_layout.rs
+++ b/curvine-server/src/worker/storage/layout/file_layout.rs
@@ -21,6 +21,7 @@ use orpc::common::ByteUnit;
 use orpc::common::FileUtils;
 use orpc::io::{BlockDevice, IOError, IOResult, LocalFile};
 use orpc::{err_box, try_err, CommonResult};
+use std::collections::HashSet;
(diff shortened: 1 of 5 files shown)


curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -47,15 +47,15 @@ impl FileLayout {
...
+            BlockState::Finalized => {
                 let uid = meta.id() as u64;
                 let d1 = (uid >> 48) & 0x1F;
                 let d2 = (uid >> 32) & 0x1F;
                 Self::active_dir(dir)
                     .join(format!("b{}", d1))
                     .join(format!("b{}", d2))
             }
-            BlockState::Recovering => Self::staging_dir(dir),
+            BlockState::Writing | BlockState::Recovering => Self::staging_dir(dir),
```
szbr00, line 59: Routing `Writing` blocks to `staging/` means a rewrite now has two on-disk files for the same block id: the committed file in `active/` and the staging copy in `staging/`. `scan()` and `initialize()` don't reconcile this. After a crash mid-rewrite, `scan()` returns the active file as `Finalized` and the staging file as `Recovering`; `initialize()` reserves capacity for **both**, and `meta.put` keeps only the last one (`Recovering`, since staging is scanned after active).

The result is a permanent capacity leak (the committed file's bytes stay reserved but unreachable), an orphaned active file that is never deallocated, and the intact committed block being shadowed by the partial staging block (reported to the master as `Writing`). Since this PR's goal is to preserve finalized data, the crash path deserves handling too — e.g. in `scan`/`initialize`, when a block id exists in both `active/` and `staging/`, prefer the committed active file and discard the incomplete staging copy. A crash-recovery test (simulate a crash between `prepare_write` and `finalize`, then re-init) would lock this down.

curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -106,21 +110,48 @@ impl BlockLayout for FileLayout {
...
+            let staging_path = Self::block_path(dir, &prepared)?;
+            if staging_path.exists() {
+                return err_box!(
+                    "Cannot rewrite block {} because staging file {} already exists",
+                    meta.id(),
+                    staging_path.display()
+                );
+            }
+
+            if let Err(e) = fs::copy(&committed_path, &staging_path) {
```
szbr00, line 132: `fs::copy` here runs while the caller holds the dataset write lock (`BlockStore::open_block` → `write()`), which excludes all concurrent readers (`get_block`, `open_reader`, `all_blocks`). For a large committed block, the rewrite copy can stall every read on this worker for the full copy duration. Before this change `prepare_write` did no bulk I/O under the lock. Worth considering whether the staging copy can be moved off the critical path, or at least documenting the expected latency impact for large blocks.

curvine-server/src/worker/storage/layout/file_layout.rs
no-teet-on: Good catch. Fixed in 41e1f61. FileLayout::scan now records valid active block IDs and, when a matching recovering staging file exists, removes the incomplete staging file and omits it before VfsDataset::initialize reserves capacity or inserts metadata. The new restart_during_rewrite_discards_staging_and_keeps_committed test simulates a crash after rewrite preparation and verifies that the committed bytes, finalized metadata, and capacity accounting are preserved after reinitialization. This intentionally handles the active-plus-staging interrupted-rewrite case; staging-only recovery remains a separate recovery-lifecycle concern.

curvine-server/src/worker/storage/layout/file_layout.rs
no-teet-on: Agreed on the lock scope and latency impact for new metadata operations. I documented the boundary both next to fs::copy and in the PR description. Existing readers that already hold an open file handle continue reading the committed active file, but get_block, all_blocks, and new reader setup can wait for the copy. For this correctness fix, the copy remains under the dataset write lock so the committed-to-staging transition and single-writer check stay serialized. Moving it out safely requires a fenced prepare/copy/commit lifecycle with rollback, which is intentionally left out of this PR rather than introducing a larger concurrency refactor.

curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -106,21 +110,48 @@ impl BlockLayout for FileLayout {
             return err_box!("Invalid file block size: {}", block.len);
         }
         let mut prepared = BlockMeta::new(meta.id(), block.len, dir);
-        // The old file still occupies space until resize/finalize. Keep the
-        // larger physical charge so preparing a smaller write cannot temporarily
-        // overstate available capacity or under-release on abort.
```
szbr00, line 111: The comment explaining why actual_len keeps the larger of the committed and new sizes was removed. Under the staging design the rationale is slightly different now (the staging file is a copy of the committed file, so it physically occupies at least meta.actual_len and may grow to block.len), and the line is no longer self-explanatory. A one-line comment would prevent a future reader from "simplifying" this to block.len and under-charging capacity.

szbr00:
BTW，could you change the status of PR from draft to ready for merging? Thank you for your contribution.

curvine-server/src/worker/storage/layout/file_layout.rs
no-teet-on: Thanks. Fixed in 376a5e9: restored a concise comment explaining that a staging rewrite starts from the committed allocation and may grow, so capacity must be charged using the larger size.

no-teet-on:
Sure. I addressed the latest inline comment in 376a5e9, reran make format and the Worker test suite (58 passed) in the Linux environment, and marked the PR ready for review. Thank you for the review.

szbr00:
> Sure. I addressed the latest inline comment in [376a5e9](https://github.com/CurvineIO/curvine/commit/376a5e9021c5fb0cc28bec02acd00e14dae97e4b), reran make format and the Worker test suite (58 passed) in the Linux environment, and marked the PR ready for review. Thank you for the review.
Where are from？If you are Chinese, could you contact me with my WeChat: 398713015. Thank you.

szbr00 (approved):

szbr00 (approved):

Issue #1238, opened 2026-07-22
[FEATURE] Improve load-job semantics for partial task success


xingnailu:
**Is your feature request related to a problem? Please describe.**

Directory-level `load` jobs are split into file-level tasks. Today, **any single task failure marks the entire job as `Failed`**, while **successfully completed tasks leave their cached files in Curvine** (no automatic rollback).

This creates confusing operator semantics:

1. `cv load-status` / job logs report `Failed`, which looks like a full failure.
2. A large portion of the directory may already be cached and readable from Curvine (file-level cache hit).
3. Failure logs can report `len = 0` even after hours of progress, which understates how much data was actually written.
4. There is no first-class way to distinguish “nothing cached” vs “partially cached”, or to **retry only failed tasks**.

Observed production example (cache_mode mount):

- Job planned: `8192` tasks, `~3.4TB`
- Workers started copying individual files into CV
- Job ended as `Failed` because one task hit `No available worker found`
- Cached files from successful tasks remained under the CV path

Client reads are already per-file (valid CV cache → CV; otherwise → UFS), so retaining successful files is useful for cache acceleration — but the **job-level status does not describe partial success**.

**Describe the solution you'd like**

1. **Richer terminal states** for load jobs, e.g.:
   - `Completed` — all tasks succeeded
   - `PartialSuccess` / `CompletedWithErrors` — some tasks succeeded, some failed
   - `Failed` — no usable progress / hard failure per policy

2. **Accurate progress on failure**: aggregate `loaded_size` / completed-task bytes in failure summaries (avoid misleading `len = 0`).

3. **Expose partial-result metrics** in `load-status` / `load-list` / Web UI:
   - `completed_files`, `failed_files`, `running_files`
   - cached bytes attributable to successful tasks
   - failed task paths (verbose)

4. **Configurable fail policy** (job or mount / load CLI):
   - `fail_fast` — current behavior (optional keep as default)
   - `continue` — run remaining tasks; finish as `PartialSuccess` if any failures
   - `all_or_nothing` (optional) — on failure, clean up files written by this job run

5. **Retry only failed tasks** (e.g. `cv load --retry-failed <job_id>` or skip-already-cached), so operators do not re-load the whole directory.

**Describe alternatives you've considered**

- Keep current fail-fast + no cleanup (status only): cheapest, but operators keep misreading `Failed` as “no cache”.
[...]

PR #1326, opened 2026-07-27
feat(transfer): add protocol and configuration

jlon:
# Summary
Define the Transfer service contract without starting the service itself.
This PR adds the public RPC schema, configuration, state model, and error surface consumed by later Transfer implementation PRs.

# Issue Describe / Design
Related to #1238.
The configuration retains compatibility inputs while normalizing runtime storage selection through `store_url`.

# Changes
| --- | --- | --- |
| `curvine-common/proto/transfer.proto` | Add Transfer request, status, and task-report messages. | New protocol only. |
| `curvine-common/src/conf/transfer_conf.rs` | Add Transfer configuration and validation. | Disabled by default. |
| `curvine-common/src/state/transfer.rs` | Add transfer state, task, and progress types. | New API surface only. |
| common protocol/state helpers | Add RPC codes, errors, and Worker capability fields. | Backward-compatible optional fields. |

# Test verified
| --- | --- | --- |
| `cargo check -p curvine-common` | PASS | |

# Dependencies
- Related PRs: Blocks the Transfer persistence, metadata, scheduler, client, and deployment PRs.
- Related issues: #1238

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,8 @@ prost = "0.11.9"
 prost-build = "0.11.9"
 bincode = "1.3.3"
 serde_json = "1.0.125"
+rusqlite = { version = "0.32.1", features = ["bundled"] }
+mysql = "25.0.1"
 
 
 once_cell = "1.18.0"
--- a/curvine-common/src/conf/cluster_conf.rs
+++ b/curvine-common/src/conf/cluster_conf.rs
@@ -15,7 +15,8 @@
 use crate::alloc::allocator_type_name;
 use crate::conf::CliConf;
 use crate::conf::{
-    ClientConf, FaultHttpConfig, FuseConf, JobConf, JournalConf, MasterConf, WorkerConf,
+    ClientConf, FaultHttpConfig, FuseConf, JobConf, JournalConf, MasterConf, TransferConf,
+    WorkerConf,
 };
 use crate::rocksdb::DBConf;
 use crate::version;
@@ -73,6 +74,8 @@ pub struct ClusterConf {
 
     pub job: JobConf,
 
+    pub transfer: TransferConf,
+
     pub cli: CliConf,
 }
 
@@ -124,7 +127,8 @@ impl ClusterConf {
             conf.master.hostname = ip.clone();
             conf.journal.hostname = ip.clone();
             conf.worker.hostname = ip.clone();
-            conf.client.hostname = ip;
+            conf.client.hostname = ip.clone();
+            conf.transfer.hostname = ip;
         } else {
             if let Ok(v) = env::var(Self::ENV_MASTER_HOSTNAME) {
                 conf.master.hostname = v.to_owned();
@@ -146,6 +150,7 @@ impl ClusterConf {
         conf.client.init()?;
         conf.fuse.init()?;
         conf.job.init()?;
+        conf.transfer.init()?;
 
         if conf.client.master_addrs.is_empty() {
             for peer in &mut conf.journal.journal_addrs {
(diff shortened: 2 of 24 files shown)


Cargo.toml
```diff
@@ -88,6 +88,8 @@ prost = "0.11.9"
 prost-build = "0.11.9"
 bincode = "1.3.3"
 serde_json = "1.0.125"
+rusqlite = { version = "0.32.1", features = ["bundled"] }
+mysql = "25.0.1"
```
szbr00, line 92: The committed `Cargo.lock` is inconsistent with the manifests in this PR: it records `curvine-server` depending on `mysql`/`rusqlite`/`crossbeam` and `curvine-cli` depending on `curvine-server`/`toml`, but no crate declares these. Verified locally: `cargo metadata --locked` fails with "the lock file needs to be updated", so any `--locked` CI/build breaks and every local build silently rewrites `Cargo.lock`. It looks like the lockfile was generated from a later revision of the branch stack that still contained the store implementation. Since nothing in this PR consumes `rusqlite`/`mysql` yet, please drop these two workspace dependencies for now and regenerate `Cargo.lock` so it matches the manifests.

curvine-common/src/conf/transfer_conf.rs
```diff
@@ -0,0 +1,524 @@
...
+        default = "TransferConf::default_ufs_max_concurrency_per_endpoint"
+    )]
+    pub ufs_max_concurrency_per_endpoint: usize,
+    #[serde(skip, default = "TransferConf::default_task_max_retries")]
+    pub task_max_retries: usize,
+
+    #[serde(skip)]
+    pub task_report_interval: Duration,
+    #[serde(skip, default = "TransferConf::default_task_report_interval_str")]
+    pub task_report_interval_str: String,
```
szbr00, line 81: Several duration knobs (`task_report_interval`, `task_stale_timeout`, `cluster_snapshot_max_staleness`, `metadata_replica_refresh_interval`, `metadata_replica_max_staleness`) mark their `_str` fields `#[serde(skip, ...)]`, so they cannot actually be set from `curvine-cluster.toml` — a user-supplied value would be silently ignored and `init()` always parses the hardcoded default. This is inconsistent with `lease_timeout`/`terminal_retention` here (and with `JobConf`'s convention), which use `#[serde(alias = ...)]` and are configurable. Please either make these configurable via the alias pattern, or drop the unused `_str` fields/constants (`task_stale_timeout` and `task_report_interval` are derived from `lease_timeout` anyway).

curvine-common/src/conf/journal_conf.rs
```diff
@@ -57,6 +57,14 @@ pub struct JournalConf {
     // How many entries are created after creating snapshots.
     pub snapshot_entries: u64,
 
+    // Internal ring buffer for Transfer metadata replica delta sync.
+    #[serde(
+        skip_serializing,
+        skip_deserializing,
+        default = "JournalConf::default_metadata_delta_log_capacity"
+    )]
+    pub metadata_delta_log_capacity: usize,
```
szbr00, line 66: With `skip_serializing` + `skip_deserializing` this field can never be configured — it behaves as a constant that happens to live in `JournalConf`. Either make it deserializable so operators can tune the delta-log capacity, or keep it as a plain `const` until the Transfer delta-sync PR that consumes it lands.

curvine-common/src/conf/cluster_conf.rs
```diff
@@ -146,6 +150,7 @@ impl ClusterConf {
         conf.client.init()?;
         conf.fuse.init()?;
         conf.job.init()?;
+        conf.transfer.init()?;
```
szbr00, line 153: `transfer.init()` runs full validation unconditionally, so an invalid transfer config (e.g. a malformed `store_url`, or `max_running_transfers = 0`) prevents any process loading the cluster config from starting, even when `transfer.enabled = false`. Consider gating the validation on `enabled` (the mysql-store check already works this way) so a disabled feature cannot break cluster startup.

jlon:
Superseded by the rebased split stack: #1342 and #1343.

PR #1327, opened 2026-07-27
feat(transfer): add local transfer stores

jlon:
# Summary

Adds in-memory and SQLite implementations of the transfer persistence contract. This PR deliberately stops at local persistence; scheduling, MySQL storage, and client or worker routing remain in follow-up PRs.

# Issue Describe / Design

Related to #1238 and the load-job isolation design.

The store owns request-id idempotency and target-path conflict checks. SQLite persists submitted transfers across process restarts, while the memory store is suitable for local and test deployment. Keeping these implementations separate from scheduling avoids coupling persistence rollout to task ownership and retry policy.

# Changes

| `curvine-server/src/transfer` | Adds store contract plus memory and SQLite implementations. | None until a later scheduler selects a transfer store. |
| `curvine-server/Cargo.toml` | Adds the existing workspace `rusqlite` dependency. | Enables local SQLite persistence. |
| `Cargo.lock` | Records the resolved SQLite dependency graph and removes unreachable legacy packages. | Reproducible locked builds. |
| `curvine-server/tests/transfer_store_test.rs` | Covers idempotency, target conflicts, and SQLite reopen persistence. | Test coverage only. |

# Test verified

| `cargo test --locked -p curvine-server --test transfer_store_test` | PASS | 2 tests passed. |

# Dependencies

- Related PRs: #1326
- Related issues: #1238
- Blocks / blocked by: Blocked by #1326; blocks the scheduler and MySQL store PRs.

### DIFF
--- a/curvine-server/src/transfer/memory_store.rs
+++ b/curvine-server/src/transfer/memory_store.rs
@@ -1,0 +1,735 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashMap;
+
+use curvine_common::error::FsError;
+use curvine_common::state::{
+    summarize_transfer_tasks, StaleTaskAttempt, TaskAttemptStart, TransferJobRecord, TransferLease,
+    TransferListFilter, TransferState, TransferStateUpdate, TransferTaskRecord, TransferTaskReport,
+    TransferTaskState, TransferTenantSummary,
+};
+use curvine_common::FsResult;
+use parking_lot::Mutex;
+
+use crate::transfer::{TransferRequeueUpdate, TransferStore};
+
+#[derive(Default)]
+pub struct MemoryTransferStore {
+    inner: Mutex<MemoryTransferState>,
+}
+
+#[derive(Default)]
+struct MemoryTransferState {
+    jobs: HashMap<String, TransferJobRecord>,
+    request_index: HashMap<(String, String), String>,
+    tasks: HashMap<(String, u64, String), TransferTaskRecord>,
+}
+
+impl MemoryTransferStore {
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl TransferStore for MemoryTransferStore {
+    fn check_available(&self) -> FsResult<()> {
+        Ok(())
+    }
+
+    fn create_or_get_by_request_id(&self, job: TransferJobRecord) -> FsResult<TransferJobRecord> {
+        self.create_or_get_by_request_id_checked(job)
+    }
+
+    fn create_or_get_by_request_id_checked(
+        &self,
+        job: TransferJobRecord,
+    ) -> FsResult<TransferJobRecord> {
+        let mut inner = self.inner.lock();
+        let request_key = (job.submitter.clone(), job.client_request_id.clone());
+        if let Some(job_id) = inner.request_index.get(&request_key) {
+            return inner
+                .jobs
+                .get(job_id)
+                .cloned()
+                .ok_or_else(|| FsError::job_not_found(job_id));
+        }
+
+        if let Some(existing) = inner
+            .jobs
+            .values()
+            .find(|existing| existing.job_key == job.job_key && !existing.state.is_terminal())
+        {
+            if existing.command_json == job.command_json {
+                return Ok(existing.clone());
+            }
+            return Err(FsError::transfer_already_running(format!(
+                "job_key {} has running job {} with different command",
+                existing.job_key, existing.job_id
+            )));
+        }
+
+        if let Some(existing) = inner.jobs.values().find(|existing| {
+            !(existing.state.is_terminal()
+                || (existing.submitter == job.submitter
+                    && existing.client_request_id == job.client_request_id))
+                && target_paths_conflict(&existing.target_path, &job.target_path)
+        }) {
+            return Err(FsError::transfer_target_conflict(format!(
+                "target {} conflicts with active transfer {} target {}",
+                job.target_path, existing.job_id, existing.target_path
+            )));
+        }
+
+        inner.request_index.insert(request_key, job.job_id.clone());
+        inner.jobs.insert(job.job_id.clone(), job.clone());
+        Ok(job)
+    }
+
+    fn get_transfer(&self, job_id: &str) -> FsResult<Option<TransferJobRecord>> {
+        Ok(self.inner.lock().jobs.get(job_id).cloned())
+    }
+
+    fn get_transfer_by_request(
+        &self,
+        submitter: &str,
+        client_request_id: &str,
+    ) -> FsResult<Option<TransferJobRecord>> {
+        let inner = self.inner.lock();
+        let Some(job_id) = inner
+            .request_index
+            .get(&(submitter.to_string(), client_request_id.to_string()))
+        else {
+            return Ok(None);
+        };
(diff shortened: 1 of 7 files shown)


curvine-server/src/transfer/sqlite_store.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+    fn update_transfer_state(&self, update: TransferStateUpdate) -> FsResult<bool> {
+        let states = state_list(&update.from_states);
+        let sql = format!(
+            "update transfer_jobs
+             set state = ?1, summary_json = ?2, updated_at = ?3
+             where job_id = ?4 and run_id = ?5 and owner = ?6 and lease_epoch = ?7
+               and state in ({states})"
+        );
+        let summary = transfer_progress_json(&TransferProgress {
```
szbr00, line 468: This rebuilds `summary_json` from `Default::default()`, so progress accumulated by task reports (`loaded_size`/`total_size`) is silently reset to zero on every state transition. The memory store preserves the existing summary and only updates the message. `requeue_transfer` below has the same pattern. Please preserve the stored summary (read-modify-write, or keep counters in separate columns) so both stores behave identically and progress is not lost.

curvine-server/src/transfer/memory_store.rs
```diff
@@ -0,0 +1,735 @@
...
+        let progress = summary.progress;
+
+        if let Some(job) = inner.jobs.get_mut(&task_job_id) {
+            job.summary = progress;
+            job.updated_at = report.now_ms;
+            if has_failed {
+                job.state = TransferState::Failed;
+            } else if has_task && all_completed {
+                job.state = TransferState::Completed;
+            } else if task_state == TransferTaskState::Completed {
```
szbr00, line 658: This transition diverges from `SqliteTransferStore`: the SQLite implementation never moves a job to `Running` from `update_task_report` (`update_job_summary` only sets `Failed`/`Completed`), so the two stores report different job states for the same event stream. It can also clobber `Canceling` — if a task completes after `request_cancel`, the job flips from `Canceling` back to `Running`. Please align both implementations on one transition rule and guard this transition so it cannot overwrite `Canceling` (or a terminal state).

curvine-server/src/transfer/sqlite_store.rs
```diff
@@ -0,0 +1,1372 @@
...
+    .map_err(sqlite_err)
+}
+
+fn select_conflicting_active_transfer(
+    conn: &Connection,
+    target_path: &str,
+    submitter: &str,
+    client_request_id: &str,
+) -> FsResult<Option<TransferJobRecord>> {
+    let child_prefix = format!("{}/%", target_path.trim_end_matches('/'));
```
szbr00, line 1131: Path-conflict matching is weaker than the memory store in three cases: (1) stored targets keep their raw form, so a stored `/target/` (trailing slash) never matches a new `/target/child` — neither the LIKE nor the ancestor checks hit, while `MemoryTransferStore` normalizes trailing slashes and detects it; (2) a new job targeting `/` only conflicts with an existing `/` job, not with any existing subtree job, whereas the memory store treats `/` as conflicting with everything; (3) the LIKE pattern is not escaped, so a target containing `%` or `_` can produce false conflicts. Consider normalizing `target_path` on insert and escaping the LIKE pattern so both stores enforce the same conflict contract.

curvine-server/src/transfer/store.rs
```diff
@@ -0,0 +1,156 @@
...
+    TransferTenantSummary,
+};
+use curvine_common::FsResult;
+
+pub trait TransferStore: Send + Sync + 'static {
+    fn check_available(&self) -> FsResult<()>;
+
+    fn create_or_get_by_request_id(&self, job: TransferJobRecord) -> FsResult<TransferJobRecord>;
+
+    fn create_or_get_by_request_id_checked(
```
szbr00, line 27: Both store implementations implement `create_or_get_by_request_id` by delegating to `create_or_get_by_request_id_checked`, so the trait exposes two methods with identical semantics. Consider making one a default trait method (or dropping the unchecked variant) so implementations cannot drift apart.

curvine-server/src/transfer/sqlite_store.rs
```diff
@@ -0,0 +1,1372 @@
...
+        mount_snapshot_json: row.get(7)?,
+        secret_ref_json: row.get(8)?,
+        cluster_snapshot_version: row.get::<_, i64>(9)? as u64,
+        cv_metadata_epoch: row.get::<_, Option<i64>>(10)?.map(|v| v as u64),
+        state: TransferState::from(row.get::<_, i32>(11)?),
+        owner: row.get(12)?,
+        lease_epoch: row.get::<_, i64>(13)? as u64,
+        lease_expire_at: row.get(14)?,
+        cancel_requested: row.get::<_, i64>(15)? != 0,
+        summary: serde_json::from_str(&summary_json).unwrap_or_default(),
```
szbr00, line 1223: `unwrap_or_default()` here (and in `sqlite_task_row` for `progress_json`) silently converts corrupt summary/progress JSON into zeroed progress — a job loses all reported progress without any error surfacing. Consider propagating a decode error, or at least logging a warning, so data corruption is visible.

curvine-server/src/transfer/memory_store.rs
```diff
@@ -0,0 +1,735 @@
...
+        Ok(true)
+    }
+
+    fn claim_pending_tasks(
+        &self,
+        job_id: &str,
+        run_id: u64,
+        limit: usize,
+    ) -> FsResult<Vec<TransferTaskRecord>> {
+        let inner = self.inner.lock();
```
szbr00, line 505: Task claim order differs between stores: SQLite claims pending tasks oldest-first (`order by updated_at asc`), while the memory store iterates a `HashMap` in arbitrary order (same for `mark_stale_attempts` and `purge_terminal_transfers`). If claim order affects fairness or retry priority, the memory store should sort by the same keys so the implementations stay interchangeable.

curvine-server/src/transfer/sqlite_store.rs
```diff
@@ -0,0 +1,1372 @@
...
+        submitter: &str,
+        client_request_id: &str,
+    ) -> FsResult<Option<TransferJobRecord>> {
+        select_job_by_request(&self.conn.lock(), submitter, client_request_id)
+    }
+
+    fn list_active_transfers(&self) -> FsResult<Vec<TransferJobRecord>> {
+        let conn = self.conn.lock();
+        let mut stmt = conn
+            .prepare(job_select_sql("where state not in (6, 7, 8)").as_str())
```
szbr00, line 124: SQL literals repeat the `TransferState`/`TransferTaskState` discriminants (`6,7,8` / `2,3,4,5` / `1,6`) throughout this file; if the enum values ever change, these queries silently change meaning. Also note `TransferState::from(i32)` maps any unknown stored value to the `#[default]` `Pending`, which could resurrect a terminal job into scheduling after reading a future or corrupt state. Consider deriving the state sets from the enum (shared constants or a helper like `state_list`) and handling unknown values explicitly.

curvine-server/src/transfer/sqlite_store.rs
```diff
@@ -0,0 +1,1372 @@
...
+
+    fn set_transfer_state(
+        &self,
+        job_id: &str,
+        run_id: u64,
+        state: TransferState,
+        message: impl Into<String>,
+        now_ms: i64,
+    ) -> FsResult<bool> {
+        let mut job = match self.get_transfer(job_id)? {
```
szbr00, line 500: `set_transfer_state` reads the job, releases the connection lock, then re-acquires it to write — a concurrent `update_task_report` can commit in between and have its summary overwritten by the stale `summary_json` written here (progress regression). The memory store performs this under a single lock. Consider wrapping the read-modify-write in one transaction, or updating `summary_json` atomically in SQL.

szbr00 (commented):
Overall the store abstraction looks solid, but the two implementations currently diverge on a few contract points (progress preservation on state transitions, job-state transitions on task reports, path-conflict normalization, claim/purge ordering). Since the scheduler will treat these stores as interchangeable, could you extend the shared `assert_store_contract` harness to cover those flows (lease acquire/renew, task attempt + report, requeue, conflict edge cases like trailing slashes) and run it against both stores? That would pin both implementations to identical behavior before the scheduler PR lands.

jlon:
Superseded by the rebased split stack: #1344.

PR #1328, opened 2026-07-27
feat(transfer): add MySQL transfer store

jlon:
# Summary

Adds the MySQL implementation of the transfer persistence contract, including schema initialization and forward migrations. Store selection and transfer scheduling remain outside this PR.

# Issue Describe / Design

Related to #1238 and the load-job isolation design.

The MySQL store persists both job records and task records, performs request-id idempotency and active-target conflict checks transactionally, and maps backend failures to a user-actionable transfer-store error. Scheduler ownership, lease fencing, and task outcome policy are intentionally deferred to the scheduler PR.

# Changes

| `curvine-server/src/transfer/mysql_store.rs` | Adds MySQL-backed transfer job/task store and schema migration. | None until a later transfer service selects MySQL storage. |
| `curvine-server/src/transfer/mod.rs` | Exports `MysqlTransferStore`. | Makes implementation available to the future backend factory. |
| `curvine-server/Cargo.toml` and `Cargo.lock` | Adds the workspace MySQL dependency and its resolved closure. | Enables MySQL-backed persistence. |

# Test verified

| `cargo check --locked -p curvine-server` | PASS | Validates the store against the current transfer contract and lockfile. |

# Dependencies

- Related PRs: #1327
- Related issues: #1238
- Blocks / blocked by: Blocked by #1327; blocks the transfer scheduler PR.

### DIFF
--- a/curvine-server/src/transfer/mysql_store.rs
+++ b/curvine-server/src/transfer/mysql_store.rs
@@ -1,0 +1,1329 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use curvine_common::error::FsError;
+use curvine_common::state::{
+    summarize_transfer_tasks, StaleTaskAttempt, TaskAttemptStart, TransferJobRecord, TransferLease,
+    TransferListFilter, TransferState, TransferStateUpdate, TransferTaskRecord, TransferTaskReport,
+    TransferTaskState, TransferTenantSummary,
+};
+use curvine_common::FsResult;
+use mysql::prelude::*;
+use mysql::{params, Params, Pool, PooledConn, TxOpts, Value as MysqlValue};
+
+use crate::transfer::{TransferRequeueUpdate, TransferStore};
+
+const TRANSFER_SCHEMA_VERSION: u64 = 4;
+const TRANSFER_SCHEMA_V3: u64 = 3;
+
+type TenantSummaryRow = (
+    String,
+    Option<u64>,
+    Option<u64>,
+    Option<u64>,
+    Option<u64>,
+    Option<u64>,
+    u64,
+);
+
+pub struct MysqlTransferStore {
+    pool: Pool,
+}
+
+impl MysqlTransferStore {
+    pub fn open(url: &str) -> FsResult<Self> {
+        let pool = Pool::new(url).map_err(mysql_err)?;
+        let mut conn = pool.get_conn().map_err(mysql_err)?;
+        init_schema(&mut conn)?;
+        Ok(Self { pool })
+    }
+
+    fn conn(&self) -> FsResult<PooledConn> {
+        self.pool.get_conn().map_err(mysql_err)
+    }
+}
+
+impl TransferStore for MysqlTransferStore {
+    fn check_available(&self) -> FsResult<()> {
+        self.conn()?.query_drop("select 1").map_err(mysql_err)
+    }
+
+    fn create_or_get_by_request_id(&self, job: TransferJobRecord) -> FsResult<TransferJobRecord> {
+        self.create_or_get_by_request_id_checked(job)
+    }
+
+    fn create_or_get_by_request_id_checked(
+        &self,
+        job: TransferJobRecord,
+    ) -> FsResult<TransferJobRecord> {
+        let mut conn = self.conn()?;
+        let mut tx = conn
+            .start_transaction(TxOpts::default())
+            .map_err(mysql_err)?;
+        if let Some(existing) =
+            select_job_by_request(&mut tx, &job.submitter, &job.client_request_id)?
+        {
+            tx.commit().map_err(mysql_err)?;
+            return Ok(existing);
+        }
+        if let Some(existing) = select_non_terminal_job_by_key(&mut tx, &job.job_key)? {
+            if existing.command_json == job.command_json {
+                tx.commit().map_err(mysql_err)?;
+                return Ok(existing);
+            }
+            return Err(FsError::transfer_already_running(format!(
+                "job_key {} has running job {} with different command",
+                existing.job_key, existing.job_id
+            )));
+        }
+        if let Some(existing) = select_conflicting_active_transfer(
+            &mut tx,
+            &job.target_path,
+            &job.submitter,
+            &job.client_request_id,
+        )? {
+            return Err(FsError::transfer_target_conflict(format!(
+                "target {} conflicts with active transfer {} target {}",
+                job.target_path, existing.job_id, existing.target_path
+            )));
+        }
+        insert_job(&mut tx, &job)?;
+        tx.commit().map_err(mysql_err)?;
+        Ok(job)
+    }
+
+    fn get_transfer(&self, job_id: &str) -> FsResult<Option<TransferJobRecord>> {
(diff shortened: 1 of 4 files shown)


curvine-server/src/transfer/mysql_store.rs
```diff
@@ -0,0 +1,1329 @@
...
+                    && job.lease_epoch == lease_epoch
+                    && !job.state.is_terminal() =>
+            {
+                job
+            }
+            _ => return Ok(false),
+        };
+        job.lease_expire_at = now_ms.saturating_add(lease_ms);
+        job.updated_at = now_ms;
+        exec_update_job(&mut self.conn()?, &job)
```
szbr00, line 396: `renew_lease` (and the other read-modify-write methods `request_cancel`, `set_transfer_state`, `update_task_state`) reads the record on one pooled connection, validates `owner`/`lease_epoch`/state against that stale snapshot, then rewrites every column via `exec_update_job` on a different connection — no transaction, row lock, or CAS guard. Under concurrent schedulers (the scenario this backend exists for) this loses updates: an `acquire_runnable_transfer` that bumped `lease_epoch` N→N+1 and changed owner can be overwritten back to N by this write, silently reverting the lease handover; and `request_cancel` can move a job that already reached a terminal state back to `Canceling`, because the UPDATE carries no `state not in (6,7,8)` guard. The SQLite and memory implementations are atomic here (single guarded UPDATE, or lock held across check-and-set). Please make these single-statement guarded UPDATEs (match owner/lease_epoch/state in the WHERE clause) or wrap read-check-write in a transaction with `SELECT ... FOR UPDATE`, and only write the columns that actually change.

curvine-server/src/transfer/mysql_store.rs
```diff
@@ -0,0 +1,1329 @@
...
+    let mut job = select_job_by_id(conn, job_id)?.ok_or_else(|| FsError::job_not_found(job_id))?;
+    let summary = summarize_transfer_tasks(&tasks, now_ms);
+    job.summary = summary.progress;
+    job.updated_at = now_ms;
+    if summary.has_failed {
+        job.state = TransferState::Failed;
+    } else if summary.has_task && summary.all_completed {
+        job.state = TransferState::Completed;
+    }
+    let _ = exec_update_job(conn, &job)?;
```
szbr00, line 1269: `update_job_summary` rewrites the whole job row from a snapshot read without `FOR UPDATE` (`select_job_by_id`), so a task report committing here overwrites concurrent changes to `owner`/`lease_epoch`/`lease_expire_at`/`cancel_requested` made by `renew_lease`, `acquire_runnable_transfer`, or `request_cancel` — e.g. a fresh `lease_expire_at` can regress to an older value and the job can then be spuriously re-acquired by another scheduler. The SQLite implementation deliberately updates only `state`/`summary_json`/`updated_at` here. Please restrict this UPDATE to the columns it owns, or lock the job row before the summary rewrite.

curvine-server/src/transfer/mysql_store.rs
```diff
@@ -0,0 +1,1329 @@
...
+        .exec_first::<u64, _, _>(
+            "select version from transfer_schema_version where id = 1",
+            Params::Empty,
+        )
+        .map_err(mysql_err)?
+        .ok_or_else(|| FsError::common("Missing mysql transfer schema version"))?;
+    migrate_mysql_schema(conn, version, now_ms)?;
+
+    conn.query_drop(
+        "create table if not exists transfer_jobs (
```
szbr00, line 750: The schema declares no `CHARACTER SET`/`COLLATE`, so the varchar columns inherit the database default — on MySQL 8 typically `utf8mb4_0900_ai_ci`, i.e. case-insensitive. `target_path`/`job_key` equality, the unique `(submitter, client_request_id)` key, and the range scan in `select_conflicting_active_transfer` would then fold case: a job for `/Data` and a job for `/data` are treated as the same target, producing false `transfer_target_conflict` errors and wrong request-id dedup. That diverges from the SQLite store (BINARY collation) and POSIX path semantics. Please declare an explicit binary collation (e.g. `COLLATE utf8mb4_bin`) on these tables/columns.

curvine-server/src/transfer/mysql_store.rs
```diff
@@ -0,0 +1,1329 @@
...
+        let mut tx = conn
+            .start_transaction(TxOpts::default())
+            .map_err(mysql_err)?;
+        if let Some(existing) =
+            select_job_by_request(&mut tx, &job.submitter, &job.client_request_id)?
+        {
+            tx.commit().map_err(mysql_err)?;
+            return Ok(existing);
+        }
+        if let Some(existing) = select_non_terminal_job_by_key(&mut tx, &job.job_key)? {
```
szbr00, line 80: This check (like `select_job_by_request` above) is a plain SELECT under REPEATABLE READ, so two concurrent creators with the same `job_key` but different `client_request_id`s can both observe "no non-terminal job", both pass the target-conflict check, and both insert — leaving two active jobs for the same `job_key`, which the store contract forbids. The SQLite backend is protected by its process-wide connection mutex; this one is not. Consider `SELECT ... FOR UPDATE` on the job_key lookup (InnoDB gap locks then serialize concurrent inserts of the same key) or enforcing one active job per job_key with a constraint. Related: a racing duplicate `(submitter, client_request_id)` insert currently surfaces as a 1062 error mapped to `transfer_store_unavailable` instead of returning the existing record — catching 1062 and re-selecting would preserve the idempotent-get semantics.

curvine-server/src/transfer/mysql_store.rs
```diff
@@ -0,0 +1,1329 @@
...
+            .map(|row| serde_json::from_str(&row).map_err(json_err))
+            .collect()
+    }
+
+    fn start_task_attempt(&self, start: TaskAttemptStart) -> FsResult<bool> {
+        let mut conn = self.conn()?;
+        let mut tx = conn
+            .start_transaction(TxOpts::default())
+            .map_err(mysql_err)?;
+        let Some(job) = select_job_by_id_for_update(&mut tx, &start.job_id)? else {
```
szbr00, line 643: `start_task_attempt` locks the job row then the task row (`select_job_by_id_for_update` → `select_task_for_update`), while `update_task_report` locks the task row first and then writes the job row via `update_job_summary`. Opposite lock ordering between these two transactions can deadlock under InnoDB (error 1213) when a worker report races an attempt start on the same job, surfacing as `transfer_store_unavailable`. Consider a consistent lock order (always job → task) or restructuring so both methods acquire rows in the same sequence.

curvine-server/src/transfer/mysql_store.rs
```diff
@@ -0,0 +1,1329 @@
...
+            key transfer_tasks_job_state_idx(job_id, run_id, state),
+            key transfer_tasks_worker_idx(worker_id, worker_session_id, state),
+            key transfer_tasks_stale_idx(state, stale_deadline_at)
+        )",
+    )
+    .map_err(mysql_err)?;
+    Ok(())
+}
+
+fn migrate_mysql_schema(conn: &mut PooledConn, mut version: u64, now_ms: u64) -> FsResult<()> {
```
szbr00, line 799: This is the first MySQL store in-tree, so MySQL schema versions 2 and 3 never existed here (the SQLite store is at version 2 with a different, column-normalized schema, and these migrations read from `record_json`). Are the v2→v3 and v3→v4 paths carried over from an out-of-tree deployment? If not, consider starting the MySQL lineage at version 1 and dropping the dead branches; if intentional, a short comment explaining the provenance would help future maintainers.

szbr00 (commented):
Thanks — the implementation is a faithful port of the SQLite store's query set, and the index layout looks well considered. My main concerns are the concurrency semantics (non-atomic read-modify-write methods, full-row overwrites, missing gap-lock protection in the create path, lock-order inversion) and the missing explicit collation; details are in the line comments. One more ask: the base PR introduced a reusable contract harness (`assert_store_contract` in `transfer_store_test.rs`) that this backend never runs — please consider adding a MySQL-backed contract test gated behind an env var (e.g. `CURVINE_TEST_MYSQL_URL`) so the contract can be verified against a real server, at least manually or in a later CI job.

jlon:
Superseded by the rebased split stack: #1346.

PR #1329, opened 2026-07-27
feat(master): add metadata snapshot journal

jlon:
# Summary

Adds the Master-side metadata change journal and paginated snapshot/delta readers needed by a transfer metadata replica. This PR only builds the read model; it does not expose new RPC endpoints yet.

# Issue Describe / Design

Related to the load-job isolation design.

Filesystem mutations are translated into path-level metadata changes held in a bounded in-memory journal. A replica can request a full snapshot or a delta up to a fixed epoch; when history is no longer retained, the API explicitly requires a full snapshot rather than returning an incomplete delta.

# Changes

| `master/journal/entry.rs` | Derives metadata changes from filesystem journal entries. | No external behavior until RPC exposure. |
| `master/journal/journal_writer.rs` | Keeps a bounded metadata delta log with a low watermark. | Provides deterministic fallback to full snapshots. |
| `master/fs/master_filesystem.rs` | Adds paginated metadata snapshot and delta readers. | No existing API behavior changed. |

# Test verified

| `cargo check --locked -p curvine-server` | PASS | Validates Master and journal integration. |

# Dependencies

- Related PRs: #1328
- Related issues: None
- Blocks / blocked by: Blocked by #1328; blocks the metadata RPC and replica-reader PRs.

### DIFF
--- a/curvine-server/src/master/fs/master_filesystem.rs
+++ b/curvine-server/src/master/fs/master_filesystem.rs
@@ -32,8 +32,32 @@ use orpc::runtime::GroupExecutor;
 use orpc::sync::ArcRwLock;
 use orpc::{err_box, err_ext, try_option, CommonResult};
 use parking_lot::Mutex;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::Arc;
+
+pub struct CvMetadataSnapshotEntry {
+    pub status: FileStatus,
+    pub blocks: Option<FileBlocks>,
+}
+
+pub struct CvMetadataSnapshotPage {
+    pub entries: Vec<CvMetadataSnapshotEntry>,
+    pub next_page_token: Option<String>,
+    pub epoch: u64,
+}
+
+pub struct CvMetadataDeltaEntry {
+    pub path: String,
+    pub entry: Option<CvMetadataSnapshotEntry>,
+}
+
+pub struct CvMetadataDeltaPage {
+    pub entries: Vec<CvMetadataDeltaEntry>,
+    pub next_page_token: Option<String>,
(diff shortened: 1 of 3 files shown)


curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -688,6 +729,279 @@ impl MasterFilesystem {
...
+        path: &str,
+        start_after: Option<&str>,
+        limit: usize,
+        entries: &mut Vec<CvMetadataSnapshotEntry>,
+    ) -> FsResult<()> {
+        if entries.len() >= limit {
+            return Ok(());
+        }
+        if let Some(token) = start_after {
+            if path != "/" && token > path && !snapshot_token_in_subtree(token, path) {
```
szbr00, line 953: The page-token comparisons here (and the emit check below) use raw byte ordering of full path strings, but traversal is pre-order DFS over name-sorted children. These orders disagree when one sibling name is a prefix of another and the next byte sorts before `/` (0x2F) — e.g. siblings `ab` (a dir containing `c`) and `ab.x`: DFS visits `/ab`, `/ab/c`, `/ab.x`, but byte-wise `/ab/c` > `/ab.x`. If a page boundary falls at token `/ab/c`, the next page prunes `/ab.x` here (`token > path` and not in subtree), so that entry — and its entire subtree — is silently dropped from the snapshot. This is common in practice (`data` + `data.txt`, `a` + `a.b`). Please compare paths component-wise (lexicographic over `split(PATH_SEPARATOR)` segments) in both the prune and emit checks, or switch to a skip-until-token-found cursor instead of string arithmetic.

jlon:
Superseded by the rebased split stack: #1347.

PR #1330, opened 2026-07-27
feat(master): expose metadata snapshot RPCs

jlon:
# Summary

Exposes the Master metadata snapshot and delta page operations through the existing synchronous read-RPC dispatch path.

# Issue Describe / Design

Related to the load-job isolation design.

The two endpoints are read-only and reuse the same Master handler execution model as existing block-location and master-info reads. This intentionally excludes the unrelated RPC lane refactor; a transfer metadata replica can now consume the journal-backed page model from #1329.

# Changes

| `master/master_handler.rs` | Adds snapshot/delta page request parsing, response conversion, audit labels, and dispatch. | Adds two new read-only RPC codes. |

# Test verified

| `cargo fmt --all -- --check` | PASS | Formatted with project rustfmt. |
| `cargo check --locked -p curvine-server` | PASS | Validates protocol-to-Master integration. |

# Dependencies

- Related PRs: #1329
- Related issues: None
- Blocks / blocked by: Blocked by #1329; blocks the metadata replica-reader PR.

### DIFF
--- a/curvine-server/src/master/master_handler.rs
+++ b/curvine-server/src/master/master_handler.rs
@@ -444,6 +444,54 @@ impl MasterHandler {
         fs.master_info()
     }
 
+    pub fn get_cv_metadata_snapshot_page(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+        let req: GetCvMetadataSnapshotPageRequest = ctx.parse_header()?;
+        ctx.set_audit(Some("cv-metadata-snapshot".to_string()), None);
+        let page = self
+            .fs
+            .cv_metadata_snapshot_page(req.page_token, req.page_size.unwrap_or(10_000) as usize)?;
+        ctx.response(GetCvMetadataSnapshotPageResponse {
+            entries: page
+                .entries
+                .into_iter()
+                .map(|entry| CvMetadataSnapshotEntryProto {
+                    status: ProtoUtils::file_status_to_pb(entry.status),
+                    blocks: entry.blocks.map(ProtoUtils::file_blocks_to_pb),
+                })
+                .collect(),
+            next_page_token: page.next_page_token,
+            epoch: page.epoch,
+        })
+    }
+
+    pub fn get_cv_metadata_delta_page(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+        let req: GetCvMetadataDeltaPageRequest = ctx.parse_header()?;
+        ctx.set_audit(Some("cv-metadata-delta".to_string()), None);
+        let page = self.fs.cv_metadata_delta_page(
+            req.from_epoch,
+            req.target_epoch,
+            req.page_token,
+            req.page_size.unwrap_or(10_000) as usize,
+        )?;
+        ctx.response(GetCvMetadataDeltaPageResponse {
+            entries: page
+                .entries
+                .into_iter()
+                .map(|entry| CvMetadataDeltaEntryProto {
+                    path: entry.path,
+                    entry: entry.entry.map(|entry| CvMetadataSnapshotEntryProto {
+                        status: ProtoUtils::file_status_to_pb(entry.status),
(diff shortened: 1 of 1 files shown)


curvine-server/src/master/master_handler.rs
```diff
@@ -444,6 +444,54 @@ impl MasterHandler {
         fs.master_info()
     }
 
+    pub fn get_cv_metadata_snapshot_page(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+        let req: GetCvMetadataSnapshotPageRequest = ctx.parse_header()?;
+        ctx.set_audit(Some("cv-metadata-snapshot".to_string()), None);
+        let page = self
+            .fs
+            .cv_metadata_snapshot_page(req.page_token, req.page_size.unwrap_or(10_000) as usize)?;
```
szbr00, line 452: `page_size` comes straight from the wire and is forwarded unbounded into the fs layer, where it feeds `Vec::with_capacity(page_size + 1)` in `cv_metadata_snapshot_page` / `cv_metadata_delta_page`. With this PR the fs calls become reachable over RPC, so a client sending e.g. `page_size = u32::MAX` makes the active master attempt a multi-hundred-GB allocation and abort — a remote master-availability risk. Please clamp the value at this trust boundary (e.g. `min(page_size, MAX_PAGE_SIZE)`), and apply the same bound in `get_cv_metadata_delta_page`.

curvine-server/src/master/master_handler.rs
```diff
@@ -444,6 +444,54 @@ impl MasterHandler {
         fs.master_info()
     }
 
+    pub fn get_cv_metadata_snapshot_page(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+        let req: GetCvMetadataSnapshotPageRequest = ctx.parse_header()?;
+        ctx.set_audit(Some("cv-metadata-snapshot".to_string()), None);
```
szbr00, line 449: These endpoints are designed for a metadata replica that polls `GetCvMetadataDeltaPage` continuously and resyncs through paginated snapshot reads, so every poll/page will emit an audit record. `audit_log` already skips the other high-frequency polling RPCs (`WorkerHeartbeat`, `WorkerBlockReport`, `GetMountTable`). Consider adding the two new codes to that skip list (or auditing only snapshot resyncs) to keep the audit log usable.

curvine-server/src/master/master_handler.rs
```diff
@@ -444,6 +444,54 @@ impl MasterHandler {
...
+    }
+
+    pub fn get_cv_metadata_delta_page(&self, ctx: &mut RpcContext<'_>) -> FsResult<Message> {
+        let req: GetCvMetadataDeltaPageRequest = ctx.parse_header()?;
+        ctx.set_audit(Some("cv-metadata-delta".to_string()), None);
+        let page = self.fs.cv_metadata_delta_page(
+            req.from_epoch,
+            req.target_epoch,
+            req.page_token,
+            req.page_size.unwrap_or(10_000) as usize,
```
szbr00, line 474: The `10_000` fallback is duplicated in both handlers, and the proto's `[default = 10000]` does not actually apply: prost generates `page_size` as `Option<u32>` and optional fields decode to `None` when absent (the `default` attribute is inert), so this `unwrap_or` is the only effective default. Consider a shared constant used by both handlers, and dropping the misleading proto default in the proto-owning change.

szbr00 (commented):
Consider adding coverage for the new RPC wiring. `curvine-server/tests/master_fs_test.rs` already provides a `new_handler_for_test` harness, so a round-trip test dispatching `RpcCode::GetCvMetadataSnapshotPage` / `GetCvMetadataDeltaPage` and verifying the proto conversion — including delta tombstone entries (`entry = None`) and `full_snapshot_required` — would protect the field mapping added here.

jlon:
Superseded by the rebased split stack: #1347.

PR #1331, opened 2026-07-27
chore(fuse): cap channel comments at 3 lines

yangcx000:
## What

Follow-up to #1315. Further compresses the long comments in the two FUSE
channel files so no comment block exceeds 3 lines:

- `curvine-fuse/src/session/channel/fuse_receiver.rs`
- `curvine-fuse/src/session/channel/fuse_sender.rs`

## Why

#1315 trimmed redundant comments crate-wide but left several multi-line blocks
in these two hot-path files. This tightens them to a 3-line cap while keeping
the load-bearing "why", so the invariants stay documented without the bulk.

Load-bearing facts deliberately preserved:
- the parse-success → RAII-scope "no `.await`/early return" window invariant;
- the writev-vs-splice EAGAIN root cause (the fuse device has no send-buffer
  watermark, so writev fails with a real errno instead of EAGAIN — hence
  `enable_splice=false` is sound and `splice_retry` is not applied to writev);
- the pipe-poisoning `drain_pipe` invariant;
- the shared process-global metric-label rationale for lower-bound assertions.

## Scope

Comment-only. No code, string literals, log strings, or behavior changed.

## Verification

- `cargo fmt -p curvine-fuse -- --check` ✓
- `cargo clippy -p curvine-fuse --all-targets` ✓
- `cargo test -p curvine-fuse --lib -- --test-threads=1 session::channel::` → 30 passed ✓

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -283,11 +283,9 @@ impl<T: FileSystem> FuseReceiver<T> {
         Self::send_stream_dispatch(&self.fs, req, rep).await
     }
 
-    /// Stream dispatch + IO attribution core, factored out of `send_stream` so
-    /// tests can drive it with a hand-built `FuseResponse` — no `FuseReceiver`/
-    /// `kernel_fd`/`Pipe2`/reactor. The metrics gate is derived solely from
-    /// `rep.metrics.is_some()` (not a separate flag), so a "metrics but no ctx"
-    /// split-brain state is unrepresentable.
+    /// Stream dispatch + IO attribution core, factored out of `send_stream` so tests
+    /// can drive it with a hand-built `FuseResponse`. The metrics gate derives solely
+    /// from `rep.metrics.is_some()`, so a "metrics but no ctx" state is unrepresentable.
     async fn send_stream_dispatch(
         fs: &Arc<T>,
         req: FuseRequest,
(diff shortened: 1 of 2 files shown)


curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -308,13 +306,9 @@ impl<T: FileSystem> FuseReceiver<T> {
...
-        // IO attribution timers, armed before the match so they also cover the
-        // error-reply enqueue below. One is `Some` per stream opcode: read/write ->
-        // io_dispatch timer, flush/fsync/release -> lifecycle scope.
-        //
-        // ⚠ INVARIANT: no `.await` / early return between `parse_operator()` and
-        // these two scopes — `stream_lifecycle_scope` counts the attempt and arms
-        // its guard atomically, so a suspension here would unbalance the family.
+        // IO attribution timers, armed before the match so they cover the error-reply
+        // enqueue too (one Some per stream opcode). ⚠ INVARIANT: after parse succeeds, no
+        // `.await`/early return before these scopes — the lifecycle scope arms its guard atomically.
```
szbr00: This line is 103 chars — the only line over the repo's `max_width = 100` convention in these two files (rustfmt doesn't wrap comments, so `fmt --check` stays green). Nit, but worth fixing in a comment-tidying PR. A re-flow that keeps 3 lines and the full wording:

```
// IO attribution timers, armed before the match so they cover the error-reply enqueue
// too (one Some per stream opcode). ⚠ INVARIANT: after parse succeeds, no `.await`/
// early return before these scopes — the lifecycle scope arms its guard atomically.
```

szbr00 (commented):
Comment-only diff verified locally: no non-comment lines changed; every trimmed block still matches the code it describes (parse→RAII-scope invariant, writev/EAGAIN rationale, `drain_pipe` contract, SETLKW timer hoisting). All blocks ≤3 lines apart from the license headers. `cargo fmt --check` clean and the 30 channel tests pass. One minor nit inline (a 103-char line vs `max_width = 100`). LGTM otherwise.

curvine-fuse/src/session/channel/fuse_receiver.rs
yangcx000: Fixed in d69b8ce — reflowed exactly as suggested, so no line exceeds max_width = 100 while keeping the 3 lines and the full wording. Thanks!

szbr00 (approved):

Issue #1332, opened 2026-07-27
[FEATURE] Add mount lifecycle APIs to Java SDK


jlon:
# Is your feature request related to a problem?

The Java SDK can submit and monitor load jobs, but applications cannot create, inspect, or remove the required UFS mount through the same SDK. They must shell out to the CLI or use a separate control path.

# Describe the solution you'd like

Add an explicit Java mount lifecycle client with mount, unmount, getMountInfo, and listMounts. Provide a typed MountRequest builder for the existing mount options protobuf and bridge the existing master RPCs through libsdk JNI.

# Describe alternatives you've considered

Automatically mount from the load client was rejected because a load request must not silently create or update cluster-wide UFS metadata. Shelling out to cv keeps a second, less composable client path.

# Additional context

The API should compose as mount -> load -> inspect/cancel -> unmount, while each operation remains explicit.

# Goal

| Sub-requirement | Problem solved |
| --------------- | -------------- |
| Java mount lifecycle client | Applications can manage UFS mounts without shelling out |
| Mount request builder | Java callers can express CLI-equivalent mount options |
| JNI and libsdk bridge | Existing Curvine mount RPC semantics are available from Java |

# Not Goal

- Do not add implicit mounting to CurvineLoadClient.
- Do not change master RPC, CLI mount behavior, or mount persistence semantics.
- Do not duplicate MountInfoProto as a second Java model.

# Test Plan

| Test case | Test objective | Modules involved |
| --------- | -------------- | ---------------- |
| MountRequest builder tests | Verify defaults and all mount option fields | curvine-libsdk/java |
| Mount response parsing tests | Verify mount info and table protocol handling | curvine-libsdk/java |
| Rust libsdk tests and clippy | Verify JNI bridge compiles cleanly | curvine-libsdk |
| Cluster E2E | Verify explicit mount -> load -> unmount flow | Java SDK, master |

PR #1333, opened 2026-07-27
feat(libsdk): add Java mount lifecycle APIs (#1332)

jlon:
# Summary

Adds an explicit Java UFS mount lifecycle client so applications can mount, inspect, and unmount through the SDK before submitting load jobs.

Mounting remains explicit: CurvineLoadClient does not create or update cluster-wide mount metadata.

# Issue Describe / Design

Related to #1332.

The Java SDK previously exposed load job lifecycle APIs but not the UFS mount lifecycle required by load. This change adds CurvineMountClient and MountRequest, reuses the existing MountOptionsProto and MountInfoProto protocol types, and bridges the existing master RPCs through libsdk JNI.

An implicit load-to-mount path was rejected because a load request must not silently mutate cluster-wide UFS metadata or credentials.

# Changes

| `curvine-libsdk/java` | Add CurvineMountClient, MountRequest, and contract tests | New explicit Java API only |
| `curvine-libsdk/src/core/filesystem.rs` | Expose mount, unmount, and mount-table wrappers | Reuses UnifiedFileSystem validation and cache refresh |
| `curvine-libsdk/src/java` | Add JNI bindings and protobuf bridge | No protocol or master behavior change |

# Test verified

| `cargo test -p curvine-libsdk` | PASS | 3 tests passed |
| `cargo clippy -p curvine-libsdk --all-targets -- --deny=warnings --allow clippy::uninlined_format_args` | PASS | |
| `mvn -B -ntp test -Dtest=MountRequestTest,CurvineMountClientTest,LoadJobClientTest` | PASS | 12 tests passed |
| `cargo build --release -p curvine-libsdk` | PASS | JNI symbols verified |
| `mvn -B -ntp package -DskipTests` | PASS | Java artifacts packaged |
| Full `mvn test` | BLOCKED | Existing JDK 21 CurvineNative reflection restriction plus local native library and localhost master prerequisites; failure occurs before mount APIs |

# Dependencies

- Related PRs: #1289
- Related issues: #1332
- Blocks / blocked by: Cluster E2E is pending an environment with the packaged native library and a reachable Curvine master.

### DIFF
--- a/curvine-libsdk/java/src/main/java/io/curvine/MountRequest.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/MountRequest.java
@@ -85,6 +85,13 @@ public final class MountRequest {
             return this;
         }
 
+        /**
+         * Select replacement-style update semantics for an existing mount.
+         *
+         * <p>All options in this request, including defaults for options not explicitly set,
+         * are sent to the master and can replace existing mount settings. Include each setting's
+         * current value in the request when it must be preserved.
+         */
         public Builder update(boolean update) {
             this.update = update;
             return this;
--- a/curvine-libsdk/java/src/main/java/io/curvine/MountRequest.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/MountRequest.java
@@ -1,0 +1,195 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io.curvine;
+
+import io.curvine.proto.AccessModeProto;
+import io.curvine.proto.MountOptionsProto;
+import io.curvine.proto.ProviderProto;
+import io.curvine.proto.StorageTypeProto;
+import io.curvine.proto.TtlActionProto;
+import io.curvine.proto.WriteTypeProto;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/** Immutable request for creating or updating a Curvine UFS mount. */
+public final class MountRequest {
+    private static final long DEFAULT_TTL_MS = 7L * 24L * 60L * 60L * 1000L;
+
+    private final String ufsPath;
+    private final String cvPath;
+    private final MountOptionsProto options;
+
+    private MountRequest(String ufsPath, String cvPath, MountOptionsProto options) {
+        this.ufsPath = ufsPath;
+        this.cvPath = cvPath;
+        this.options = options;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public String getUfsPath() {
+        return ufsPath;
+    }
+
+    public String getCvPath() {
+        return cvPath;
+    }
+
+    public MountOptionsProto getOptions() {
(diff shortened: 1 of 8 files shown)


curvine-libsdk/java/src/main/java/io/curvine/MountRequest.java
```diff
@@ -0,0 +1,195 @@
...
+            this.ufsPath = ufsPath;
+            return this;
+        }
+
+        public Builder cvPath(String cvPath) {
+            this.cvPath = cvPath;
+            return this;
+        }
+
+        public Builder update(boolean update) {
```
szbr00, line 95: When `update(true)` is used, this builder still populates every option with its default (7-day TTL, `autoCache=true`, `READ_ONLY`, `CACHE_MODE`-derived TTL action) unless the caller explicitly overrides it. Since `MountOptionsProto` fields are proto2 `optional`, the master's `merge_with` treats each populated field as intentional and overwrites the existing mount's settings — so an update meant to add one property also resets TTL/autoCache/accessMode to defaults. This matches the CLI behavior, but for a new API the contract is surprising. Could you document it in the `update(boolean)` javadoc (e.g. "unset options still carry defaults and replace existing mount settings on update"), or leave fields unset when `update=true` so the server preserves current values?

curvine-libsdk/src/java/java_filesystem.rs
```diff
@@ -108,6 +113,44 @@ impl JavaFilesystem {
...
+        ufs_path: JString,
+        cv_path: JString,
+        options: JByteArray,
+    ) -> FsResult<()> {
+        let ufs_path = JavaUtils::jstring_to_string(env, &ufs_path)?;
+        let cv_path = JavaUtils::jstring_to_string(env, &cv_path)?;
+        let options = env
+            .convert_byte_array(&options)
+            .map_err(FsError::from_error)?;
+        let options = try_err!(MountOptionsProto::decode(options.as_slice()));
```
szbr00: `prost` does not enforce proto2 `required` fields, so a direct `CurvineNative.mount` caller passing an empty or malformed `options` array decodes successfully into all-default options (`update=false`, `CACHE_MODE`, no `ttl_ms` — which the server maps to `ttl_ms = 0`) instead of failing. `CurvineMountClient` always serializes a valid proto, but `CurvineNative` is public API. Consider rejecting an empty array (or a minimal sanity check) here so malformed input cannot silently create a mount with unintended options.

szbr00 (commented):
`CurvineMountClient` duplicates the full lifecycle boilerplate from `CurvineLoadClient` (`from`/`close`/`ensureOpen`/`checkError`/`checkBytes`/`requirePath`, ~60 lines). With a third client likely to follow, consider extracting a package-private base (holding `nativeHandle` + `closed` and the shared helpers) so each client only implements its own operations. Non-blocking — fine to defer to a follow-up.

curvine-libsdk/src/java/java_filesystem.rs
jlon: Fixed in ee2079f7. The JNI path now rejects an empty mount-options payload before protobuf decoding, with a curvine-libsdk regression test. Nonempty malformed payloads continue to fail through protobuf decoding.

curvine-libsdk/java/src/main/java/io/curvine/MountRequest.java
jlon: Fixed in ee2079f7. MountRequest.update(true) is now documented as replacement-style: defaults are serialized and can replace existing settings. I retained the existing CLI-compatible behavior because the current protocol always supplies read_verify_ufs, so a partial-patch API would not safely preserve every field.

jlon:
Regarding the suggested package-private lifecycle base: deferred for now. CurvineMountClient and CurvineLoadClient have only two callers and retain distinct request validation and result parsing; a shared base would remove little code while adding a generic lifecycle contract. Revisit if a third native lifecycle client appears.

szbr00 (approved):

PR #1335, opened 2026-07-27
refactor(storage): move SPDK adapter out of ORPC (#1243)

lzjqsdd:
# Summary

Move SPDK storage implementation and link logic out of the shared `orpc` crate into a server-owned storage adapter. This prevents server SPDK/RDMA feature selection from being inherited by unrelated binaries through Cargo feature unification.

# Issue Describe / Design

Related to #1243

- Keep lightweight SPDK configuration types in `orpc` as a temporary compatibility surface.
- Move SPDK build script, C helper, FFI, environment, poller, and bdev implementation into `curvine-storage-spdk`.
- Make `curvine-server` explicitly opt into the leaf adapter through its own `spdk` and `spdk-rdma` features.
- Replace the server block device enum dependency with the `BlockIO` trait boundary so local files and SPDK bdevs can be selected by the storage layer.

# Changes

| `crates/adapters/curvine-storage-spdk` | Adds the SPDK storage adapter crate and moves the SPDK implementation and native link setup into it. | SPDK remains available when server SPDK features are enabled. |
| `orpc` | Removes SPDK features, native build dependencies, and SPDK implementation modules from the shared crate. | Shared ORPC consumers no longer inherit SPDK/RDMA native link requirements. |
| `curvine-server` | Routes `spdk` / `spdk-rdma` features through `curvine-storage-spdk` and opens storage through the `BlockIO` trait boundary. | Server storage behavior is preserved while SPDK selection becomes explicit. |
| `build/build.sh` / `scripts/check-deps.sh` | Updates feature forwarding and mixed-feature dependency guard behavior. | Build helpers no longer route legacy `orpc/spdk` feature names. |

# Test verified

| `make format` | PASS | Passed on the exact PR head. |
| `cargo check -p orpc -p curvine-server -p curvine-storage-spdk --no-default-features` | PASS | Verifies non-SPDK compile paths for the affected crates. |
| `cargo test -p orpc block_device_local_write_read_roundtrip` | PASS | 1 passed. |
| `cargo test -p curvine-server block_io_context` | PASS | 2 passed. |
| `scripts/check-deps.sh --mode report` | PASS | `mixed-spdk-feature-risk` is OK; remaining warnings match the current dependency migration stage. |
| `cargo tree -p curvine-server --no-default-features --features curvine-server/spdk-rdma -e features` | PASS | Confirms server SPDK/RDMA resolves to `curvine-storage-spdk` while `orpc` has no SPDK features. |
| Independent code review | PASS | Approved exact head SHA `9d9dbd724b5cc0df799cb20279da6d4a30a0eea6` with no blocking findings. |

# Dependencies

- Related PRs: #1334
- Related issues: #1243
- Blocks / blocked by: Blocked by #1334

# Remaining Risk

- Full SPDK feature compilation and runtime validation were not executed because the local environment does not provide SPDK headers. Feature boundaries were verified with Cargo feature tree checks and code review.


orpc/src/io/block_io.rs
```diff
@@ -79,6 +90,30 @@ impl BlockIO for BlockDevice {
     fn resize(&mut self, truncate: bool, off: i64, len: i64, mode: i32) -> IOResult<()> {
         delegate!(self, resize, truncate, off, len, mode)
     }
+
+    fn supports_read_ahead(&self) -> bool {
+        self.supports_read_ahead()
```
szbr00, line 95: These trait overrides call same-named inherent methods on `BlockDevice`, which only resolves correctly because inherent methods take priority over trait methods during lookup. It works, but it reads like unconditional recursion, and it silently becomes real recursion if the inherent helpers are ever removed. Consider making the delegation explicit (`BlockDevice::supports_read_ahead(self)`) or inlining the `matches!(self, BlockDevice::Local(_))` logic here and dropping the duplicate inherent surface.

crates/adapters/curvine-storage-spdk/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+spdk-rdma = ["spdk"]
+
+[dependencies]
+orpc = { workspace = true }
+bytes = { workspace = true }
+crossbeam = { workspace = true }
+libc = { workspace = true }
+log = { workspace = true }
+nix = { workspace = true }
+serde = { workspace = true }
```
szbr00, line 20: `serde` looks unused in this crate now that the config types carrying the derive impls live in `orpc::io::spdk_conf` — no file under `src/` imports it. Consider dropping it to keep the adapter's dependency surface minimal.

orpc/src/io/spdk_conf.rs
```diff
@@ -0,0 +1,311 @@
...
+            self.hugepage_mb,
+            self.reactor_mask,
+            iova_mode_label,
+            self.targets.len()
+        )
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct BdevInfo {
```
szbr00, line 305: This `BdevInfo` mirrors the old non-SPDK stub, but the adapter's `spdk_env::BdevInfo` has a different, richer shape (`num_blocks`/`ctrlr`/`ns`), and nothing references `orpc::io::BdevInfo` anymore. Two same-named types with different fields across the two crates is a drift trap. Consider either dropping this one or adding a short comment that it exists only as a compatibility placeholder.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ members = [
     "crates/infra/orpc-rpc",
     "crates/infra/curvine-sys",
     "crates/infra/curvine-io",
+    "crates/adapters/curvine-storage-spdk",
```
szbr00, line 27: The workspace manifest asks members to be mirrored in `default-members` unless intentionally opt-in, but this adapter isn't mirrored. Since its `build.rs` is a no-op without the `spdk` feature (the `link_spdk` call is feature-gated), adding it there would let root-level `cargo check`/`clippy` continuously verify the adapter's default configuration in CI. Intentional exclusion, or worth mirroring?

PR #1339, opened 2026-07-27
fix(fuse): skip Master plock unlock on FUSE_FLUSH without a held lock

bigbigxu:
## Summary
- Fix Spark hang when Curvine FUSE is used as YARN/Spark local-dirs: stop unconditional POSIX lock UnLock on every `FUSE_FLUSH`/close introduced by #1227.
- Only call Master UnLock when the file handle actually tracked `lock_owner` via `take_plock_if_owner`.
- Skip RocksDB/journal updates when `LockMeta` is unchanged; avoid redundant expire purges on the hot lock path.

## Test plan
- [ ] Reproduce previous hang with Spark + FUSE local-dirs on a build that includes #1227; confirm this branch no longer stalls around shuffle/MapOutput.
- [ ] Smoke LTP/fcntl cases that #1227 targeted (`F_GETLK`, partial unlock) still pass.
- [ ] Confirm Master `SetLock` QPS no longer spikes on file close storms without fcntl.

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1520,10 +1520,17 @@ impl fs::FileSystem for CurvineFileSystem {
                 .invalid_cache(op.header.nodeid, None, INVAL_REASON_FLUSH);
         }
 
-        if op.arg.lock_owner != 0 {
-            self.fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
-                .await?;
-            handle.take_plock_if_owner(op.arg.lock_owner);
+        // Only unlock when this handle held a plock for lock_owner. Linux fills
+        // lock_owner on almost every FUSE_FLUSH/close; unconditional unlock
+        // caused a Master SetLock+journal storm for Spark local-dirs (#1227).
+        if op.arg.lock_owner != 0 && handle.take_plock_if_owner(op.arg.lock_owner).is_some() {
+            if let Err(e) = self
+                .fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
+                .await
+            {
+                handle.add_lock(LockFlags::Plock, op.arg.lock_owner);
+                return Err(e);
+            }
         }
         handle.flush(Some(reply)).await
     }
--- a/curvine-server/src/master/meta/lock_meta.rs
+++ b/curvine-server/src/master/meta/lock_meta.rs
@@ -59,11 +59,10 @@ impl LockMeta {
         format!("{}{}", lock.client_id, lock.owner_id)
     }
 
-    fn is_lock_expired(lock: &FileLock, expire_ms: u64) -> bool {
+    fn is_lock_expired_at(lock: &FileLock, expire_ms: u64, now: u64) -> bool {
         if lock.acquire_time == 0 {
             return true;
         }
-        let now = LocalTime::mills();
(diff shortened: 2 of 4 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1521,9 +1521,15 @@ impl fs::FileSystem for CurvineFileSystem {
         }
 
         if op.arg.lock_owner != 0 {
-            self.fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
-                .await?;
-            handle.take_plock_if_owner(op.arg.lock_owner);
+            if handle.take_plock_if_owner(op.arg.lock_owner).is_some() {
```
szbr00: The nested `if` here fails CI: the `build` job runs clippy with `-D warnings` and this triggers `clippy::collapsible_if` ("this `if` statement can be collapsed"), so the PR currently cannot go green. Please collapse into a single condition:

```rust
if op.arg.lock_owner != 0
    && handle.take_plock_if_owner(op.arg.lock_owner).is_some()
{
    if let Err(e) = self
        .fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
        .await
    {
        handle.add_lock(LockFlags::Plock, op.arg.lock_owner);
        return Err(e);
    }
}
```

curvine-server/src/master/meta/lock_meta.rs
```diff
@@ -216,56 +220,75 @@ impl LockMeta {
...
+
             if lock.start == 0 && lock.end == u64::MAX {
-                self.plock.remove(&key);
-                return None;
+                changed |= self.plock.remove(&key).is_some();
+                return (None, changed);
             }
 
             if let Some(owner_locks) = self.plock.get_mut(&key) {
+                let before = owner_locks.clone();
```
szbr00: This clones the whole `Vec<FileLock>` (including the heap-allocated `client_id` of every entry) on every partial unlock just to detect whether anything changed — and `subtract_owner_regions` then drains/rebuilds with another round of clones even when the unlock range overlaps nothing. On a close storm over files that do hold locks this is avoidable allocation. Consider an overlap pre-check instead:

```rust
if owner_locks
    .iter()
    .any(|l| Self::regions_overlap(l.start, l.end, lock.start, lock.end))
{
    Self::subtract_owner_regions(owner_locks, lock.start, lock.end);
    changed = true;
}
```

If nothing overlaps, the region list is unchanged by definition, so `changed` can stay as-is without the clone/compare.

szbr00 (commented):
Overall the fix looks correct: the FUSE-side gate on `take_plock_if_owner` removes the per-close `SetLock` RPC for lock-free workloads, and the master-side `changed` flag correctly skips RocksDB + journal writes for no-op unlocks (I audited the purge/full-range/partial/flock paths for missed-change false negatives and didn't find any). Two things before merge:

1. **CI is red** — see the inline clippy comment on `curvine_file_system.rs`.
2. Test coverage: the two new tests cover empty-map and missing-owner full-range unlocks, but the trickier `changed`-detection branches added in this PR are untested — (a) partial unlock that actually removes/splits a region → `changed == true`, (b) partial unlock with no overlapping region → `changed == false` (exercises the `as_slice()` comparison), and (c) unlock whose only effect is purging expired locks → `changed == true` (exercises the `len()` before/after comparison). Adding these would protect the skip-persist logic in `FsDir::set_lock` from regressions.

szbr00 (approved):

PR #1340, opened 2026-07-27
refactor(storage): move SPDK adapter out of ORPC (#1243)

lzjqsdd:
# Summary

Move SPDK storage implementation and link logic out of the shared `orpc` crate into a server-owned storage adapter. This prevents server SPDK/RDMA feature selection from being inherited by unrelated binaries through Cargo feature unification.

This PR retargets the SPDK adapter work from #1335 onto current `main` after #1334 landed. The diff contains only the Stage 3 SPDK adapter changes.

# Issue Describe / Design

Related to #1243

- Keep lightweight SPDK configuration types in `orpc` as a temporary compatibility surface.
- Move SPDK build script, C helper, FFI, environment, poller, and bdev implementation into `curvine-storage-spdk`.
- Make `curvine-server` explicitly opt into the leaf adapter through its own `spdk` and `spdk-rdma` features.
- Replace the server block device enum dependency with the `BlockIO` trait boundary so local files and SPDK bdevs can be selected by the storage layer.

# Changes

| `crates/adapters/curvine-storage-spdk` | Adds the SPDK storage adapter crate and moves the SPDK implementation and native link setup into it. | SPDK remains available when server SPDK features are enabled. |
| `orpc` | Removes SPDK features, native build dependencies, and SPDK implementation modules from the shared crate. | Shared ORPC consumers no longer inherit SPDK/RDMA native link requirements. |
| `curvine-server` | Routes `spdk` / `spdk-rdma` features through `curvine-storage-spdk` and opens storage through the `BlockIO` trait boundary. | Server storage behavior is preserved while SPDK selection becomes explicit. |
| `build/build.sh` / `scripts/check-deps.sh` | Updates feature forwarding and mixed-feature dependency guard behavior. | Build helpers no longer route legacy `orpc/spdk` feature names. |

# Test verified

| `make format` | PASS | Passed on the exact PR head. |
| `cargo check -p orpc -p curvine-server -p curvine-storage-spdk --no-default-features` | PASS | Verifies non-SPDK compile paths for the affected crates. |
| `cargo test -p orpc block_device_local_write_read_roundtrip` | PASS | 1 passed. |
| `cargo test -p curvine-server block_io_context` | PASS | 2 passed. |
| `scripts/check-deps.sh --mode report` | PASS | `mixed-spdk-feature-risk` is OK; remaining warnings match the current dependency migration stage. |
| `cargo tree -p curvine-server --no-default-features --features curvine-server/spdk-rdma -i curvine-storage-spdk` | PASS | Confirms server SPDK/RDMA resolves to `curvine-storage-spdk`. |

# Dependencies

- Related PRs: #1334 (merged), #1335 (historical, base was `feat/orpc-error-infra`)
- Related issues: #1243
- Blocks / blocked by: None

# Remaining Risk

- Full SPDK feature compilation and runtime validation were not executed because the local environment does not provide SPDK headers. Feature boundaries were verified with Cargo feature tree checks.


orpc/src/io/block_io.rs
```diff
@@ -79,6 +90,30 @@ impl BlockIO for BlockDevice {
...
+        self.supports_resize()
+    }
+
+    fn as_local(&self) -> Option<&LocalFile> {
+        self.as_local()
+    }
+
+    fn as_local_mut(&mut self) -> Option<&mut LocalFile> {
+        self.as_local_mut()
+    }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in ec5a7fb9d41d63043e0c03435d7abc6eeeea7e3c: the `BlockIO` impl now delegates capability methods to the inherent `BlockDevice` helpers via UFCS (`BlockDevice::supports_*` / `as_local*`), so trait dispatch no longer recurses into itself.

orpc/src/io/spdk_conf.rs
```diff
@@ -0,0 +1,311 @@
...
+                let msg = format!("SpdkConf: targets[{}]: {}", i, e);
+                err_msg!(msg)
+            })?;
+            let resolved_ka_ms = if target.keep_alive_timeout_ms > 0 {
+                target.keep_alive_timeout_ms
+            } else {
+                self.keep_alive_timeout_ms
+            };
+            let min_ka_ms = self.poll_interval_ms * 3;
+            if resolved_ka_ms < min_ka_ms {
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in ec5a7fb9d41d63043e0c03435d7abc6eeeea7e3c: `poll_interval_ms` is validated before the targets loop, and `min_ka_ms` now uses `checked_mul(3)` so oversized values fail validation instead of panicking/wrapping.

orpc/src/io/spdk_conf.rs
```diff
@@ -0,0 +1,311 @@
...
+            self.hugepage_mb,
+            self.reactor_mask,
+            iova_mode_label,
+            self.targets.len()
+        )
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct BdevInfo {
```
szbr00, line 308: This `BdevInfo` looks like a leftover from the deleted `spdk_stub.rs`: it has no consumers anywhere in the workspace, and it duplicates — with a smaller field set — the real `BdevInfo` in `curvine-storage-spdk/src/spdk_env.rs` (which additionally carries `num_blocks`, `ctrlr`, and `ns`). With the `spdk` feature enabled, both types are public under the same name (`orpc::io::BdevInfo` vs `curvine_storage_spdk::BdevInfo`), which makes it easy to import the wrong one. Consider dropping this struct (and its re-export in `orpc/src/io/mod.rs`) from the compatibility surface.

crates/adapters/curvine-storage-spdk/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+spdk-rdma = ["spdk"]
+
+[dependencies]
+orpc = { workspace = true }
+bytes = { workspace = true }
+crossbeam = { workspace = true }
+libc = { workspace = true }
+log = { workspace = true }
+nix = { workspace = true }
+serde = { workspace = true }
```
szbr00, line 20: `serde` and `libc` appear to be unused in this crate — nothing under `src/`, `build.rs`, or `csrc/` references them (the serde-derived config types stayed behind in `orpc::io::spdk_conf`). `pkg-config` in `[build-dependencies]` is also never invoked by `build.rs`. Dropping the unused deps keeps the new adapter's manifest minimal.

orpc/src/io/block_io.rs
```diff
@@ -79,6 +90,30 @@ impl BlockIO for BlockDevice {
     fn resize(&mut self, truncate: bool, off: i64, len: i64, mode: i32) -> IOResult<()> {
         delegate!(self, resize, truncate, off, len, mode)
     }
+
+    fn supports_read_ahead(&self) -> bool {
```
szbr00, line 94: Two related observations on `BlockDevice`: (1) these trait-impl forwards like `self.supports_read_ahead()` only resolve correctly because inherent methods win over trait methods — if the inherent helpers on `impl BlockDevice` below are ever removed, each of these becomes silent infinite recursion; qualifying them (`BlockDevice::supports_read_ahead(self)`) removes the footgun. (2) After this refactor nothing in the workspace constructs or matches on `BlockDevice` anymore (server code passes `LocalFile`/`SpdkBdev` directly through `dyn BlockIO`), so it is a single-variant enum whose only remaining user is its own unit test. Consider removing it entirely, or leaving a note explaining why it stays as public API.

orpc/src/io/spdk_conf.rs
```diff
@@ -0,0 +1,311 @@
...
+#[serde(default)]
+pub struct SpdkConf {
+    pub enabled: bool,
+    pub app_name: String,
+    #[serde(alias = "hugepage")]
+    pub hugepage_str: String,
+    #[serde(skip)]
+    pub hugepage_mb: u32,
+    pub reactor_mask: String,
+    pub iova_mode: String,
```
szbr00, line 127: The move dropped the field-level docs from the old `spdk_env.rs` — e.g. the `iova_mode` guidance (empty = auto-detect; set to `va` when auto-detect fails on VM/virtio/AMD IOMMU hosts, spdk/spdk#2683; `pa` for bare metal) and the "0 = inherit global" semantics on the timeout fields. Since this file is now the canonical home of the operator-facing SPDK config, restoring those doc comments would preserve that guidance for people writing `curvine-cluster.toml`.

szbr00 (approved):

Issue #1040, opened 2026-07-09
[FEATURE] Decouple LoadJob execution from master


jlon:
## Is your feature request related to a problem?

Yes. The current LoadJob implementation is still tightly coupled to the master process. Even after making `SubmitJob` return earlier, the master still owns LoadJob planning, UFS metadata access, task dispatch, job state transitions, and worker cancellation coordination.

This is dangerous for production clusters because LoadJob is driven by cache misses and can be bursty. When many files are loaded from remote UFS such as S3, OSS, or HDFS, the master can spend CPU, memory, async runtime slots, connection resources, and scheduling attention on load-job work that is not part of the metadata control-plane core. That load can directly affect master latency and stability.

The master should remain focused on namespace metadata, journal/raft consistency, worker membership, and metadata RPCs. LoadJob execution should be isolated so a slow or high-volume UFS load workload cannot degrade core master service quality.

Related context:

- #984 describes the auto-cache metadata latency problem caused by master-side LoadJob/UFS work.
- #998 is a short-term mitigation that moves UFS planning out of the `SubmitJob` RPC critical path, but the work is still executed inside the master process.
- #997 tracks LoadJob state-machine and HA recovery hardening, which is complementary but does not by itself isolate execution from the master.

## Describe the solution you'd like

Introduce an independent LoadJob execution component so master only performs control-plane planning/coordination at a minimal level, while expensive LoadJob execution is handled outside the master critical path.

A reasonable long-term direction:

1. Master keeps only lightweight job metadata and scheduling decisions.
2. A separate LoadJob service, scheduler, or worker-side executor owns UFS-heavy planning and execution.
3. UFS `get_status`, `list_status`, recursive scanning, large task generation, retry, and throttling are moved out of the master process.
4. Job state is persisted in a way that supports failover and recovery without relying on in-memory master background tasks.
5. Backpressure and concurrency limits are applied at the LoadJob execution layer, not by consuming master runtime capacity.
6. Worker reports and cancel semantics are made run-aware so stale reports or old cancel requests cannot affect newer job runs.
7. Master exposes stable metadata APIs for job submission/status/cancel, but does not directly perform expensive UFS I/O for the job.

## Describe alternatives you've considered

- Keep LoadJob in master and only add more semaphores/rate limits.
  - This reduces peak concurrency but does not solve isolation. The master still pays the cost of UFS latency, runtime scheduling, memory pressure, and large planning workloads.

[...]

PR #1345, opened 2026-07-27
feat(transfer): add SQLite metadata store

jlon:
# Summary

Adds the durable SQLite Transfer metadata store and migration support.

# Issue Describe / Design

# Changes

| `curvine-server/src/transfer/sqlite_store.rs` | Adds the durable SQLite Transfer metadata store and migration support. | New Transfer behavior; legacy behavior remains available unless Transfer is enabled. |

# Test verified

| `cargo test -p curvine-server --test transfer sqlite_` | PASS | Rebased onto upstream main `497d078d`. |

# Dependencies

- Related PRs: #1344
- Related issues: #1040
- Blocks / blocked by: #1344

### DIFF
--- a/curvine-server/src/transfer/sqlite_store.rs
+++ b/curvine-server/src/transfer/sqlite_store.rs
@@ -1128,7 +1128,7 @@ fn select_conflicting_active_transfer(
     submitter: &str,
     client_request_id: &str,
 ) -> FsResult<Option<TransferJobRecord>> {
-    let child_prefix = format!("{}/%", target_path.trim_end_matches('/'));
+    let (child_lower, child_upper) = child_path_bounds(target_path);
     let exact_or_child = conn
         .query_row(
             job_select_sql(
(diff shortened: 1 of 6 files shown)


curvine-server/src/transfer/sqlite_store.rs
```diff
@@ -0,0 +1,1372 @@
...
+) -> FsResult<Option<TransferJobRecord>> {
+    let child_prefix = format!("{}/%", target_path.trim_end_matches('/'));
+    let exact_or_child = conn
+        .query_row(
+            job_select_sql(
+                "where state not in (6, 7, 8)
+               and not (submitter = ?2 and client_request_id = ?3)
+               and (
+                   target_path = ?1
+                   or target_path like ?4
```
lzjqsdd: `LIKE` treats `_` and `%` in the target path as wildcards, so paths such as `/a_b` or `/a%b` can reject unrelated transfers as conflicts (for example `/a_b/%` also matches `/axb/...`). Please escape the LIKE pattern with an `ESCAPE` clause, or use the same lexicographic range approach as the MySQL store so conflict detection is path-prefix based rather than pattern based.

jlon: Fixed. SQLite now uses the same lexicographic child-path range as MySQL instead of LIKE, so `%` and `_` are treated as literal path characters. Added a regression test for both wildcard characters.

PR #1354, opened 2026-07-27
feat(worker): execute transfer tasks

jlon:
# Summary

Registers Transfer-capable workers and executes/report tasks with attempt-safe outputs.

# Issue Describe / Design

# Changes

| `curvine-server/src/worker,curvine-server/src/master` | Registers Transfer-capable workers and executes/report tasks with attempt-safe outputs. | New Transfer behavior; legacy behavior remains available unless Transfer is enabled. |

# Test verified

| `MinIO load e2e after rebase` | PASS | Rebased onto upstream main `497d078d`. |

# Dependencies

- Related PRs: #1353
- Related issues: #1040
- Blocks / blocked by: #1353

### DIFF
--- a/curvine-server/src/worker/task/load_task_runner.rs
+++ b/curvine-server/src/worker/task/load_task_runner.rs
@@ -13,30 +13,51 @@
 // limitations under the License.
 
 use crate::common::UfsFactory;
+use crate::transfer::transfer_failure_message;
 use crate::worker::task::TaskContext;
-use curvine_client::file::CurvineFileSystem;
+use curvine_client::file::{CurvineFileSystem, FsReader};
 use curvine_client::rpc::JobMasterClient;
+use curvine_client::rpc::TransferClient;
 use curvine_client::unified::{UfsFileSystem, UnifiedReader, UnifiedWriter};
 use curvine_common::error::FsError;
 use curvine_common::fs::{FileSystem, Path, Reader, Writer};
 use curvine_common::state::{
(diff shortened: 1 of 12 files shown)


curvine-server/src/worker/task/load_task_runner.rs
```diff
@@ -605,137 +536,127 @@ impl LoadTaskRunner {
...
+    overwrite: bool,
+) -> FsResult<()> {
+    match ufs.get_status(final_path).await {
+        Ok(status) if status.is_dir => {
+            return err_box!(
+                "Transfer target {} is a directory; refusing to overwrite directory with file",
+                final_path.full_path()
+            );
+        }
+        Ok(_) if !overwrite => return Err(FsError::file_exists(final_path.full_path())),
```
lzjqsdd: This makes export retries non-idempotent when `overwrite=false`. If a worker successfully renames the temp object to the UFS target and then dies before the final report is accepted, the scheduler will retry the task; the retry reaches this branch, sees the already-committed target, and reports `FileAlreadyExists`, turning a successful export into a failed or partial job. The CV-target path has an `AlreadyCommitted` check with transfer xattrs, but UFS targets do not. Please add an idempotent committed-output check for UFS exports before treating the existing target as an error.

jlon: Fixed. For overwrite=false, an existing UFS target is accepted only when its length matches and the CV source carries the persisted job/run/task/source/target commit marker for this task. The marker is written before the UFS rename, so a retry after rename but before the final report recognizes the committed output; unrelated existing targets still fail normally.

PR #1355, opened 2026-07-27
feat(client): route automatic cache through transfer

jlon:
# Summary

Routes automatic unified filesystem cache submissions through Transfer when enabled.

# Issue Describe / Design

Part of the standalone Transfer service. This review layer is stacked deliberately so that protocol, metadata storage, scheduling, worker execution, client routing, and deployment can be reviewed independently.

# Changes

| Module / File | Change | Impact on existing behavior |
| ------------- | ------ | --------------------------- |
| `curvine-client/src/unified,curvine-client/src/file` | Routes automatic unified filesystem cache submissions through Transfer when enabled. | New Transfer behavior; legacy behavior remains available unless Transfer is enabled. |

# Test verified

| Test case | Result | Notes |
| --------- | ------ | ----- |
| `MinIO load e2e after rebase` | PASS | Rebased onto upstream main `497d078d`. |

# Dependencies

- Related PRs: #1354
- Related issues: #1040
- Blocks / blocked by: #1354

PR #1356, opened 2026-07-27
feat(cli): route load and export through transfer

jlon:
# Summary

Routes normal cv load/export commands through Transfer, adds transfer operations, and preserves mount -c configuration syntax.

# Issue Describe / Design

# Changes

| `curvine-cli/src` | Routes normal cv load/export commands through Transfer, adds transfer operations, and preserves mount -c configuration syntax. | New Transfer behavior; legacy behavior remains available unless Transfer is enabled. |

# Test verified

| `cargo test -p curvine-cli --bin curvine-cli; MinIO load/export e2e` | PASS | Rebased onto upstream main `497d078d`. |

# Dependencies

- Related PRs: #1355
- Related issues: #1040
- Blocks / blocked by: #1355

### DIFF
--- a/curvine-cli/src/cmds/load.rs
+++ b/curvine-cli/src/cmds/load.rs
@@ -15,9 +15,11 @@
 use crate::cmds::LoadStatusCommand;
 use crate::util::*;
 use clap::Parser;
-use curvine_client::rpc::JobMasterClient;
-use curvine_common::state::LoadJobCommand;
-use orpc::CommonResult;
+use curvine_client::rpc::{JobMasterClient, TransferClient};
+use curvine_client::unified::UnifiedFileSystem;
+use curvine_common::fs::Path;
+use curvine_common::state::{LoadJobCommand, TransferCommand, TransferKind};
+use orpc::{err_box, CommonResult};
 
 #[derive(Parser, Debug)]
 pub struct LoadCommand {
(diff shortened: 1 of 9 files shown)


curvine-cli/src/cmds/load.rs
```diff
@@ -30,10 +32,14 @@ pub struct LoadCommand {
     /// Watch load job status after submission
     #[arg(long, short = 'w')]
     watch: bool,
+
+    /// Do not overwrite an existing target file
+    #[arg(long = "no-overwrite", default_value_t = true, action = clap::ArgAction::SetFalse)]
+    overwrite: bool,
```
lzjqsdd, line 38: `--no-overwrite` is parsed for `cv load`, but the legacy path never applies it to `LoadJobCommand`. When Transfer is disabled, `cv load --no-overwrite ...` still builds the legacy command without setting `overwrite=false`, while the Transfer path honors the flag at line 138. Please pass `.overwrite(self.overwrite)` in `execute_legacy` as well, or avoid exposing this option on the legacy path.

jlon: Fixed. The legacy load path now forwards overwrite, and the equivalent legacy export path was corrected at the same time so --no-overwrite has identical semantics with and without Transfer.

PR #1366, opened 2026-07-28
fix(client): avoid sparse seek+write block rewrite amplification (#1364)

### LINKED ISSUE #1364: [BUG]: Curvine FUSE sparse seek+write stalls building large files for sendfile

Curvine FUSE makes LTP sendfile09 / sendfile09_64 unusable because the test setup cannot finish creating its ~4GiB sparse input file. The setup repeatedly seeks forward by nearly 1MiB and writes a single byte. On a normal Linux filesystem this completes quickly using holes. On curvine-fuse each one-byte write is answered only after multi-second FUSE WRITE handling that rewrites large (~128MiB) blocks, so the case never reaches the actual sendfile() assertions before external SIGTERM.

While building in_file, the process sleeps in the kernel FUSE path (request_wait_answer -> fuse_send_write_pages -> fuse_file_write_iter -> write). File size advances only about one MiB per 1.5-2.0 seconds. After the case timeout the process is terminated with SIGTERM and LTP reports TBROK. Adjacent small-file sendfile08 / sendfile08_64 pass immediately on the same mount.

Reproduction:

1. Build and start Curvine, then mount curvine-fuse.
2. On the mount, create a temp directory and run LTP sendfile09 (or sendfile09_64) from LTP 20210524.
3. Observe that setup spends many minutes writing in_file one byte every ~1MiB, while the same binary finishes in seconds on a local Linux filesystem such as ext4/xfs/tmpfs.
4. Optionally sample /proc/<pid>/stack during the stall and confirm the blocked path is a FUSE write (fuse_send_write_pages), not sendfile() itself.

**Impact**

Any workload that creates large sparse files via seek+write on Curvine FUSE becomes pathologically slow or times out. Large-file / large-offset sendfile use cases cannot be validated, and applications using similar sparse construction patterns are effectively blocked.

Sparse seek-beyond-EOF + single-byte write patterns should complete in seconds on curvine-fuse, matching local Linux filesystem behavior, so sendfile09 can exercise large-offset sendfile() without being killed for setup latency.

Make sparse seek-beyond-EOF writes hole-preserving (or otherwise avoid full-block rewrite amplification) in the Curvine FUSE/client write path so one-byte writes after large seeks complete quickly. Verify sendfile09/sendfile09_64 reach TPASS on curvine-fuse.

**Additional context**

sendfile08/sendfile08_64 (small files) pass on the same mount. Issue #813 covers broader mmap/splice/sendfile themes but does not describe this sparse-setup amplification failure for sendfile09.

Root-cause hypothesis: Root-cause hypothesis (medium-high confidence): Curvine FUSE/client write path materializes or rewrites large storage blocks for each sparse seek+write(1) instead of preserving holes efficiently, causing write amplification that grows with file offset and prevents sendfile09 setup from completing.

Evidence summary: Focused reproduction on Curvine commit 4ef12b9d6ba5dac09d63dd26400b5b09682be4fd: sendfile09 and sendfile09_64 each ran ~1200s then SIGTERM/TBROK with near-zero CPU time. Kernel stack showed blocked write via fuse_send_write_pages on fd to in_file. FUSE WRITE reply latency averaged ~1.5s (up to ~2.2s) per one-byte write while the worker logged repeated ~128MiB block rewrites. Reference run of the same binaries on local Linux FS: both cases TPASS quickly.

Related issues: https://github.com/CurvineIO/curvine/issues/813

Acceptance / test gaps:

- sendfile09 completes with TPASS on curvine-fuse without external SIGTERM during setup
- sendfile09_64 completes with TPASS on curvine-fuse without external SIGTERM during setup
- Creating a ~4GiB sparse file via 4096 seek(+1MiB-1)+write(1) cycles on curvine-fuse finishes in a timeframe comparable to local Linux FS (orders of magnitude faster than multi-second per write)
- No unrelated public API behavior changes

LJF-050:
# Summary

Fix sparse seek-past-EOF + write amplification that hung LTP `sendfile09` / `sendfile09_64` during setup on Curvine FUSE. Avoid forcing a full worker-block rewrite on every 1-byte hole write inside an already-allocated block slot.

# Issue Describe / Design

Closes #1364

## Root cause

`sendfile09` setup performs thousands of `lseek(~1MiB) + write(1)` hole writes. Each `pos > len` path called master `resize`/`extend`, which attached `alloc_opts` to an already-located tail block. The client then treated the block as needing resize and rewrote a growing multi-MiB (up to ~128MiB) worker block per byte, so the case timed out before reaching `sendfile()` itself.

## Reproduction steps

1. Mount Curvine FUSE and run LTP `sendfile09` / `sendfile09_64`.
2. Observe hang/timeout in setup sparse seek+write (not in `sendfile()`).
3. On a normal Linux filesystem the same cases pass quickly.

## Fix approach

1. Client: call master resize only when the write target has no existing block slot (`get_block(pos).is_none()`); hole writes inside an allocated slot rely on seek+write.
2. Master: on extend, do not attach `alloc_opts` to a tail block that already has `locs`; only unplaced blocks keep alloc opts for first open sizing.

# Changes

| `curvine-client/src/file/fs_writer_base.rs` | Gate sparse resize on missing block slot | Sparse hole writes inside an allocated block no longer force master resize |
| `curvine-server/src/master/meta/inode/inode_file.rs` | Skip `alloc_opts` on located tail extend; unit test | Located tail extends no longer force client full-block rewrite; unplaced blocks unchanged |

# Test verified

| `make format` | PASS | format + clippy gate |
| `extend_does_not_force_alloc_opts_on_located_tail_block` | PASS | master unit test |
| LTP `sendfile09` / `sendfile09_64` on fixed HEAD | PASS | offsets 0 and 3GB TPASS on Curvine FUSE |
| Full originating `ltp` profile before/after | PASS | original cases absent; semantic-attributed historical fingerprint drift only |
| Independent code review @ exact HEAD | PASS | approved, no blocking findings |

# Dependencies

- Related PRs: None
- Related issues: #1364
- Blocks / blocked by: None


curvine-server/src/master/meta/inode/inode_file.rs
```diff
@@ -440,9 +440,17 @@ impl InodeFile {
...
-            last_block
-                .alloc_opts
-                .replace(opts.clone_with_len(resize_len));
+            // Blocks already placed on workers must not get a fresh alloc_opts:
+            // that flag forces the client to open+rewrite the whole block on every
+            // sparse seek+write extend (e.g. LTP sendfile09 setup), which amplifies
+            // a 1-byte hole write into a multi-MiB/128MiB rewrite.
+            // Unplaced blocks still need alloc_opts so the first worker open can size
+            // the backing store.
+            if last_block.locs.is_none() {
```
szbr00: The guard `if last_block.locs.is_none()` does not mean “not yet placed on workers”.

`BlockMeta::commit()` clears `locs` (and `alloc_opts`) after locations are persisted in the location store. Client `resize()` also `complete()`s pending writers before calling master, so the common sparse-extend path sees a committed tail with `locs == None`.

With this check, extend still attaches fresh `alloc_opts`. Client `LocatedBlock::should_resize()` is `!locs.is_empty() && alloc_opts.is_some()`, and `get_block_locations` loads locs from the store — so each block-boundary sparse resize can still force a full open+rewrite of the previous block (up to 128MiB). The client `get_block().is_none()` gate avoids that inside a slot, but sendfile09 crosses many 128MiB boundaries (~32 times for a 4GiB file).

Prefer refreshing opts only while the block is still unplaced, e.g.:

```rust
if last_block.alloc_opts.is_some() {
    last_block.alloc_opts.replace(opts.clone_with_len(resize_len));
}
```

That updates sizing for unplaced/assigned blocks, and does not reintroduce `alloc_opts` after commit.

curvine-server/src/master/meta/inode/inode_file.rs
```diff
@@ -651,6 +659,34 @@ mod tests {
         assert_eq!(&file.block_ids()[..original_ids.len()], original_ids);
     }
 
+    #[test]
+    fn extend_does_not_force_alloc_opts_on_located_tail_block() {
```
szbr00, line 663: This test only simulates an assigned-but-uncommitted tail (`locs = Some(...)`). Please also cover the committed shape that `BlockMeta::commit()` produces: `locs = None` and `alloc_opts = None` with a non-zero `len`. That is the state after client `resize()` flushes pending writers, and it is where the current `locs.is_none()` guard regresses.

curvine-server/src/master/meta/inode/inode_file.rs
LJF-050: Thanks. Fixed in d7ae64845535799bf42a1d9fb4e39462c3ca321e: changed the guard from `last_block.locs.is_none()` to `last_block.alloc_opts.is_some()`, so sparse extend refreshes sizing only for unplaced/assigned-but-uncommitted tails and does not reintroduce `alloc_opts` after `BlockMeta::commit()`.

curvine-server/src/master/meta/inode/inode_file.rs
LJF-050: Thanks. Added in d7ae64845535799bf42a1d9fb4e39462c3ca321e: `extend_does_not_force_alloc_opts_on_committed_tail_block` covers the post-commit shape (`locs = None`, `alloc_opts = None`, non-zero `len`) and asserts sparse extend does not reattach `alloc_opts`.

szbr00 (approved):

PR #1367, opened 2026-07-28
fix(client): clear stale alloc_opts after sparse flush (#1336)

### LINKED ISSUE #1336: [BUG]: FUSE ftest read/readv returns EIO after writes

Curvine FUSE ftest workloads fail when reading back previously written data with errno EIO(5).

ftest cases report read/readv failure with errno=5 at valid offsets.

Reproduction:

1. Mount curvine-fuse.
2. Run LTP ftest01 against the mount.
3. Observe EIO on readback.

**Impact**

POSIX filesystem integrity tests and applications doing write-then-read on Curvine FUSE can fail.

ftest read/readv of written data should succeed without EIO.

Fix FUSE/worker read-after-write integrity for ftest-style workloads.

**Additional context**

Root-cause statement is a hypothesis based on repeated EIO evidence.

Root-cause hypothesis: Root-cause hypothesis: Curvine FUSE returns EIO(5) during ftest read/readv of previously written regions.

Evidence summary: ftest01/03/04/05/07/08 all fail with errno=5 / EIO on read or readv in ltp-isolated.log.

Related issues: https://github.com/CurvineIO/curvine/issues/808

Acceptance / test gaps:

- ftest01/03/04/05/07/08 pass on curvine-fuse
- No unrelated API changes

LJF-050:
# Summary

Fix FUSE write-then-read `EIO` failures in LTP `ftest01/03/04/05/07/08` after sparse seek/write and flush. The client kept stale resize `alloc_opts` across block commit, so later rewrites truncated staging blocks and failed length checks.

# Issue Describe / Design

Closes #1336

## Root cause

1. Sparse seek/write attaches resize `alloc_opts` on the block. Master commit clears those opts and updates length, but the client `WriteFileBlocks::add_commit` only updated `block.len` and kept stale `alloc_opts`.
2. The next rewrite reopen re-applied the old truncate, shrinking the staging copy and causing block length mismatch (`EIO`).

## Fix approach

# Changes

| `curvine-client/src/file/fs_writer_base.rs` | Finalize worker blocks on flush; refresh `file_blocks` after `only_flush` | Published sparse writes become durable/readable while lease stays open |
| `curvine-tests/tests/*` | Add flush/sparse/past-EOF/ftest stress coverage | Guards the regression |

# Test verified

| `make format` | PASS | fmt + clippy on exact HEAD |
| `cargo test -p curvine-tests --test ftest_flush_stress_test` | PASS | Reproduced length-mismatch before fix |
| `cargo test -p curvine-tests --test pwrite_flush_read_test` | PASS | |
| `cargo test -p curvine-tests --test sparse_hole_flush_read_test` | PASS | |
| `cargo test -p curvine-tests --test fuse_read_past_eof_test` | PASS | |
| LTP focused `ftest01/03/04/05/07/08` on fixed HEAD | PASS | All TPASS |
| Full LTP profile before (source SHA) vs after (fixed HEAD) | PASS | Assigned ftest cases gone; no new failures; cleanup passed |
| Independent code review on exact HEAD | PASS | Approved, no blocking findings |

# Dependencies

- Related PRs: None
- Related issues: #1336
- Blocks / blocked by: None


curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -121,7 +128,7 @@ impl BackendHandle {
                 self.refresh_status(new_reader.status().clone());
                 reader.replace(new_reader);
 
-                self.read_ver.set(writer_ver);
+                self.read_ver.set(ver_before);
```
szbr00: The loop still cannot close the `write_ver.incr()`-before-enqueue window in `FuseWriter::write`.

Sequence:
1. writer thread: `write_ver.incr()` then gets preempted before `send_queued_task`
2. dirty-read samples `ver_before`, `flush()` runs (queue has no Write yet), then `read_ver.set(ver_before)`
3. next loop iteration sees `read_ver == write_ver` and breaks
4. writer finally enqueues the Write — reader proceeds without that data

So we can still publish `read_ver` past data that was never queued/flushed. The "sample before flush / loop until stable" pattern only covers arrivals during flush, not this gap.

Please tie visibility to something that cannot advance before the Write is queued (or before the writer task observes it), e.g.:
- bump `write_ver` only after successful enqueue, plus a pending/queued flag for dirty-read; or
- have the writer task expose a flushed/processed generation and set `read_ver` from that after Flush completes.

curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -111,8 +111,15 @@ impl BackendHandle {
         };
 
         if let Some(writer) = state.find_writer(self.ino).await {
-            let writer_ver = writer.write_ver();
-            if self.read_ver.get() != writer_ver {
+            // `write_ver` advances when writes are enqueued, not when they land.
+            // Concurrent fork writers (LTP ftest) can enqueue after we schedule a
+            // flush; sample the version we flushed and loop until stable so we
+            // never publish `read_ver` past data that is still queued.
+            for _ in 0..16 {
```
szbr00: If `write_ver` keeps moving, this loop exits after 16 flushes with no error/warning and continues the read. Under sustained concurrent writers that can still return a reader that is behind the latest enqueued writes.

Consider logging when the cap is hit, and/or failing the read with EIO/EAGAIN instead of silently serving a possibly stale view. Also worth documenting why 16 is enough for the LTP fork-writer pattern.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Fixed in 824141a28a724db98c79a305db1e4b7fb6830164: `FuseWriter::write`/`resize` now bump `write_ver` only after a successful enqueue, and track `enqueue_inflight` around `send_queued_task`. Dirty-read waits for `enqueue_inflight == 0` before trusting a stable `write_ver`, which closes the incr-before-enqueue preemption window.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Addressed in 824141a28a724db98c79a305db1e4b7fb6830164: kept/merged main's dirty-read flush budget of 16 with `warn!` + `EAGAIN` when the version stays unstable (no silent stale serve). Documented that 16 is sized for LTP fork-writer bursts; sustained churn fails the read instead of returning behind still-queued writes.

szbr00 (approved):

PR #1369, opened 2026-07-28
feat(fuse): harden protocol I/O bounds (#1096)

### LINKED ISSUE #1096: [FEATURE] fuse: protocol I/O robustness — receive buffer reuse and response length/iovec bounds (by LiAuto)

Low-priority, defensive hardening items in the protocol I/O layer. None is an active bug today; grouped because they are small and live in the receive/response path.

**(a) Receive buffer reuse is fragile**
- `curvine-fuse/src/session/channel/fuse_receiver.rs:103` (and the `splice` path)
- `self.buf` is reused across reads via `reserve(fuse_len)` + unsafe `set_len(fuse_len)`, then `split_to(len)` leaves `self.buf` holding the tail of the previous allocation.
- It works today (the next read overwrites the beginning), but the pattern is hard to reason about, can grow capacity unnecessarily, and a future change could accidentally observe stale tail bytes.
- Fix: `clear()` + resize before each read/splice copy, or keep a fixed `Vec<u8>` raw read buffer and construct `BytesMut` only from the read length.

**(b) Response length and iovec count are not explicitly bounded**
- `curvine-fuse/src/session/fuse_response.rs:54` (`as_iovec`) and `:69` (`ResponseData::create`).
- `fuse_out_header.len` is computed as `(FUSE_OUT_HEADER_LEN + data_len) as u32` with no overflow assertion; `as_iovec` pushes one `IoSlice` per data slice with no `IOV_MAX` check.
- Current call sites return one buffer bounded by request size, so the bounds are not reached today. Future multi-slice responses could overflow the u32 length field or, from the sender's `writev`/`vmsplice` side, exceed platform `IOV_MAX` and fail with `EINVAL`.
- Fix: assert / return error if response length exceeds `u32::MAX` (or the negotiated max); coalesce or reject iovec lists exceeding `IOV_MAX` before the sender's `writev`.

**(c) `as_iovec` does not verify header.len equals the actual data bytes (from sender review)**
- `ResponseData::create` (`fuse_response.rs:69`) computes `header.len` from data length, but `ResponseData` fields are public and `as_iovec` (`fuse_response.rs:55`) returns `(header.len, iovec)` trusting `header.len` without checking that the sum of iovec fragment lengths equals `header.len`.
- The splice path uses `write_iov(len, &iovec)` with `len` taken from the header, not the actual fragment sum, so an inconsistent `ResponseData` makes the sender write/splice an incorrect byte count.
- Fix: make `ResponseData` fields private where possible; in `as_iovec`, compute the actual fragment-byte sum and validate it equals `header.len`, returning an error on mismatch.

**Additional context**

From curvine-fuse protocol-IO and sender reviews (P2 / defensive items). Can be split if a maintainer prefers. Related: the IOSlice-in-response panic is tracked separately (a real correctness boundary, not a bound check).

Acceptance / test gaps:
- Response builder rejects an oversized response length or excessive iovec count.
- Receive buffer does not expose stale tail bytes across reads.
- `as_iovec` rejects a `ResponseData` whose `header.len` does not match the summed fragment bytes.

ligeng8848:
# Summary

Harden FUSE protocol response framing against oversized response lengths, excessive iovec counts, and inconsistent header lengths. Ensure rejected success frames still complete the kernel request with an `EIO` reply, and add regression coverage for safe receive-buffer reuse.

# Issue Describe / Design

Closes #1096

- Make response construction fallible and reject frame lengths exceeding `u32::MAX`.
- Query and cache the platform `IOV_MAX`; if that query fails, use Linux `UIO_MAXIOV` or the conservative POSIX minimum on other platforms.
- Fall back to an empty `EIO` response when a success frame cannot be constructed, so the kernel request and metrics always finish.
- Keep `ResponseData` fields private and expose read-only accessors.
- Recompute and validate the actual frame length before constructing iovecs.
- Preserve the existing zero-copy response path; invalid frames are rejected rather than coalesced.
- The receive-buffer clearing logic already exists on current `main`; add regression coverage proving stale tail bytes are not returned.

# Changes

| `session/fuse_response.rs` | Add checked response construction, iovec bounds, frame-length validation, field encapsulation, Linux fallback handling, and an `EIO` reply fallback | Invalid success frames now produce an error reply instead of leaving the kernel request unfinished |
| `session/channel/fuse_sender.rs` | Use read-only response accessors | No behavior change |
| `session/channel/fuse_receiver.rs` | Add and strengthen receive-buffer reuse regression coverage | No behavior change |
| `fs/stream_result.rs` | Update tests to use response accessors | No behavior change |

# Test verified

| `cargo test -p curvine-fuse --lib fuse_response::tests -- --nocapture` | PASS | 35 passed on the remote machine |
| Receive-buffer regression test | PASS | 1 passed on the remote machine |
| `cargo test -p curvine-fuse --lib -- --test-threads=1` | PASS | 321 passed on the remote machine |
| `make format` | PASS | Full release Clippy and formatting checks on the remote machine |

A prior parallel full-suite run encountered one metrics global-state interference failure; that test passed both in isolation and as part of the serialized suite.

# Dependencies

- Related PRs: None
- Related issues: #1096
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -797,14 +797,25 @@ mod tests {
         assert!(!buf.is_empty(), "the first split leaves a stale tail");
 
         FuseReceiver::<TestFileSystem>::prepare_receive_buf(&mut buf, 8);
-        buf[..3].copy_from_slice(b"new");
+        assert_eq!(
+            buf.len(),
+            8,
+            "the next receive gets exactly its read window"
+        );
 
-        let returned = buf.split_to(3);
+        let read_len = 3;
+        buf[..read_len].copy_from_slice(b"new");
+        let returned = buf.split_to(read_len);
         assert_eq!(&returned[..], b"new");
         assert_eq!(
             returned.len(),
-            3,
+            read_len,
             "only bytes reported by read are returned"
+        );
+        assert_eq!(
+            buf.len(),
+            8 - read_len,
+            "the unused read window remains internal to the reusable buffer"
         );
     }
 
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -729,8 +729,10 @@ fn receive_error_labels(os_errno: Option<i32>) -> (&'static str, &'static str) {
 
 #[cfg(test)]
 mod tests {
-    use super::{receive_error_labels, PendingRequestGuard};
+    use super::{receive_error_labels, FuseReceiver, PendingRequestGuard};
+    use crate::fs::TestFileSystem;
     use crate::fuse_metrics::{RECEIVE_ACTION_CONTINUE, RECEIVE_ACTION_EXIT};
+    use bytes::BytesMut;
     use libc::{EAGAIN, ECONNABORTED, EINTR, EIO, ENODEV, ENOENT};
     use orpc::sync::FastDashMap;
     use std::sync::Arc;
@@ -782,6 +784,27 @@ mod tests {
         assert!(
             Arc::ptr_eq(current.value(), &replacement),
             "an older guard must not remove a same-unique replacement"
+        );
+    }
+
+    #[test]
+    fn prepare_receive_buf_does_not_return_stale_tail_bytes() {
+        let first = b"first-request-with-a-tail";
+        let mut buf = BytesMut::from(&first[..]);
+
+        let returned = buf.split_to(5);
+        assert_eq!(&returned[..], b"first");
+        assert!(!buf.is_empty(), "the first split leaves a stale tail");
+
+        FuseReceiver::<TestFileSystem>::prepare_receive_buf(&mut buf, 8);
+        buf[..3].copy_from_slice(b"new");
+
+        let returned = buf.split_to(3);
+        assert_eq!(&returned[..], b"new");
+        assert_eq!(
+            returned.len(),
+            3,
+            "only bytes reported by read are returned"
         );
     }
 
@@ -1943,7 +1966,8 @@ mod tests {
             // enabled path's `errno == EBADF` check, just read off the raw frame.
             match &tasks[0] {
                 FuseTask::Reply(d) => assert_eq!(
-                    d.header.error, -ERRNO,
+                    d.header().error,
+                    -ERRNO,
                     "disabled path propagates the pre-dispatch errno on the wire"
                 ),
                 FuseTask::RequestReply { .. } => {
--- a/curvine-fuse/src/session/fuse_response.rs
+++ b/curvine-fuse/src/session/fuse_response.rs
@@ -31,17 +31,24 @@ use orpc::ternary;
 use parking_lot::Mutex;
 use std::fmt::Debug;
 use std::io::IoSlice;
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use std::vec;
 
+const FALLBACK_IOV_MAX: usize = 16;
+
+#[derive(Debug)]
 pub struct ResponseData {
-    pub header: fuse_out_header,
-    pub data: Vec<DataSlice>,
+    header: fuse_out_header,
+    data: Vec<DataSlice>,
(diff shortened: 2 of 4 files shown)


curvine-fuse/src/session/fuse_response.rs
```diff
@@ -475,7 +534,7 @@ impl FuseResponse {
                     info!("send_data unique {}, data len: {}", self.unique, len);
                 }
                 (
-                    ResponseData::create(self.unique, FUSE_SUCCESS, v),
+                    ResponseData::create(self.unique, FUSE_SUCCESS, v)?,
```
szbr00: When `ResponseData::create` fails on the success path (`?` here, and the same pattern in `send_buf` / `send_rep_tagged`), this returns an `IOResult` error without enqueueing any reply. Callers such as `fuse_reader` only log the failure, so the kernel request can hang until interrupt/timeout.

`fuse_read` already returns one `DataSlice` per chunk, and this PR intentionally rejects oversized iovec lists rather than coalescing them — so this failure mode is reachable under a low IOV limit or a large multi-chunk read.

Please fall back to an empty error reply (e.g. `EIO`) via `finish_request` when success-frame construction fails, so the kernel always gets a response and metrics still finish cleanly.

curvine-fuse/src/session/fuse_response.rs
```diff
@@ -31,17 +31,24 @@ use orpc::ternary;
 use parking_lot::Mutex;
 use std::fmt::Debug;
 use std::io::IoSlice;
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use std::vec;
 
+const FALLBACK_IOV_MAX: usize = 16;
```
szbr00, line 40: `FALLBACK_IOV_MAX = 16` matches the POSIX minimum, but `fuse_read` emits one iovec per chunk (default 128KB). With a larger `max_readahead` / smaller `read_chunk_size`, a single READ can easily exceed 15 data slices. If `sysconf(IOV_MAX)` fails, those replies would be rejected even though Linux normally allows 1024.

Consider a Linux-oriented fallback (e.g. 1024 / `UIO_MAXIOV`) or coalescing when the count exceeds the limit, so a sysconf failure does not silently break large reads.

curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -785,6 +787,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn prepare_receive_buf_does_not_return_stale_tail_bytes() {
```
szbr00, line 791: This test overwrites 3 bytes after `prepare_receive_buf` and then `split_to(3)`, so it would also pass without `clear()` — it does not actually demonstrate that a stale tail cannot be returned.

A stronger regression would assert the prepared length (`buf.len() == 8`) and that a simulated short-read `split_to(n)` returns only `n` bytes from a buffer that previously retained a longer tail, matching the receive path contract (`split_to(read_len)`).

curvine-fuse/src/session/fuse_response.rs
ligeng8848: Thanks. Fixed in 2f9e00b: sysconf remains the primary source, while Linux now falls back to libc::UIO_MAXIOV (1024); non-Linux platforms retain the conservative POSIX minimum. The response path still rejects excessive lists instead of copying/coalescing data.

curvine-fuse/src/session/fuse_response.rs
ligeng8848: Thanks. Fixed in 2f9e00b: failed success-frame construction now falls back to an empty EIO response, enqueues it through finish_request, and finalizes request metrics. A regression test covers the excessive-iovec path.

curvine-fuse/src/session/channel/fuse_receiver.rs
ligeng8848: Thanks. Fixed in 2f9e00b: the regression now asserts the prepared receive-window length, simulates a short read after a retained stale tail, and verifies that only the reported bytes are returned while the unused window remains internal.

szbr00 (approved):

PR #1374, opened 2026-07-28
refactor(orpc): decouple infrastructure layer boundaries (#1243)

lzjqsdd:
# Summary

Decouple low-level `sys` primitives from upper-layer RPC/IO types so subsequent orpc infrastructure extractions can proceed without circular crate dependencies.

# Issue Describe / Design

Related to #1243

This is the first stacked PR in the orpc infrastructure extraction series. Low-level system APIs previously depended on `IOError`, `RpcFrame`, and `LocalFile`, which would create circular dependencies among `curvine-sys`, `curvine-io`, and `orpc-rpc` after extraction.

Design decisions:
- Introduce `SysResult` (`std::io::Result`) for low-level sys APIs so `sys` no longer depends on `io::IOError`.
- Move `send_file_full` specialization from `sys` into `RpcFrame`.
- Move file-slice construction into `LocalFile` instead of `DataSlice::from_file`.
- Remove unused `DataSlice::from_frame` / `DataSlice::from_file` helpers that pulled RPC/IO types into `sys`.
- Remove `TimeBondedRetryBuilder::from_conf(ClientConf)` so retry no longer depends on the client layer (`ClientConf` already builds retry policies via `TimeBondedRetry::new` / `TimeBondedRetryBuilder::new`).

# Changes

| `orpc/src/sys/**` | Switch sys APIs to `SysResult`; remove RPC/IO reverse deps | Call sites convert `std::io::Error` via `Into`/`?` where needed |
| `orpc/src/handler/rpc_frame.rs` | Host `send_file_full` locally | Behavior preserved |
| `orpc/src/io/local_file.rs` | Inline `DataSlice` construction for sendfile path | Behavior preserved |
| `orpc/src/io/retry/time_bounded_retry.rs` | Drop `from_conf(ClientConf)` | No callers; `ClientConf` already uses duration-based constructors |
| `curvine-fuse/**`, `curvine-server/**` | Adapt to `SysResult` error type | Equivalent error conversion |

# Test verified

| `make format` | PASS | format + workspace release clippy |
| `cargo test -p orpc` | PASS | unit, integration, and doc tests |

# Dependencies

- Related PRs: #1334 (orpc-error), #1340 (SPDK adapter)
- Related issues: #1243
- Blocks / blocked by: Base for subsequent stacked extraction PRs (runtime/sys/io/net/rpc)

### DIFF
--- a/orpc/src/io/local_file.rs
+++ b/orpc/src/io/local_file.rs
@@ -286,7 +286,7 @@ impl LocalFile {
 
     pub fn actual_size(&self) -> IOResult<u64> {
         let meta = self.inner.metadata()?;
-        Ok(sys::file_actual_size(meta)?)
+        sys::file_actual_size(meta).map_err(Into::into)
     }
 
     pub fn metadata(&self) -> IOResult<Metadata> {
--- a/orpc/src/io/local_file.rs
+++ b/orpc/src/io/local_file.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::io::IOResult;
-use crate::sys::{self, CacheManager, DataSlice, ReadAheadTask};
+use crate::sys::{self, CacheManager, DataSlice, RawIOSlice, ReadAheadTask};
 use crate::{err_box, try_err};
 use bytes::BytesMut;
 use serde::de::DeserializeOwned;
@@ -110,7 +110,19 @@ impl LocalFile {
             );
         }
 
-        let region = DataSlice::from_file(self, enable_send_file, Some(self.pos), chunk as i32)?;
+        #[cfg(not(target_os = "linux"))]
+        let region = DataSlice::Buffer(self.read_full(Some(self.pos), chunk as usize)?);
+
+        #[cfg(target_os = "linux")]
+        let region = if enable_send_file {
+            DataSlice::IOSlice(RawIOSlice::new(
+                sys::get_raw_io(self)?,
+                Some(self.pos),
+                chunk as usize,
+            ))
+        } else {
+            DataSlice::Buffer(self.read_full(Some(self.pos), chunk as usize)?)
+        };
 
         self.pos += chunk;
         Ok(region)
@@ -274,7 +286,7 @@ impl LocalFile {
 
     pub fn actual_size(&self) -> IOResult<u64> {
         let meta = self.inner.metadata()?;
-        sys::file_actual_size(meta)
+        Ok(sys::file_actual_size(meta)?)
     }
 
     pub fn metadata(&self) -> IOResult<Metadata> {
--- a/orpc/src/sys/fs_stats.rs
+++ b/orpc/src/sys/fs_stats.rs
@@ -81,7 +81,11 @@ impl FsStats {
         let m = self.path.metadata()?;
 
         if !m.is_dir() || m.permissions().readonly() {
-            return sys_error!("Directory is not writable: {:?}", self.path());
+            return sys_error!(
+                std::io::ErrorKind::PermissionDenied,
+                "Directory is not writable: {:?}",
+                self.path()
+            );
         }
 
         let new_device_id = sys::get_device_id(self.path());
--- a/orpc/src/sys/fs_stats.rs
+++ b/orpc/src/sys/fs_stats.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{err_box, sys, try_err, try_log, CommonResult};
+use crate::sys::{self, SysResult};
 use log::warn;
 use std::path::{Path, PathBuf};
 
@@ -61,15 +61,15 @@ impl FsStats {
     }
 
     pub fn total_space(&self) -> u64 {
-        try_log!(fs2::total_space(self.path()), 0)
+        Self::space_or_zero(fs2::total_space(self.path()))
     }
 
     pub fn free_space(&self) -> u64 {
-        try_log!(fs2::free_space(self.path()), 0)
+        Self::space_or_zero(fs2::free_space(self.path()))
     }
 
     pub fn available_space(&self) -> u64 {
-        try_log!(fs2::available_space(self.path()), 0)
+        Self::space_or_zero(fs2::available_space(self.path()))
     }
 
     pub fn used_space(&self) -> u64 {
@@ -77,11 +77,11 @@ impl FsStats {
         used.unwrap_or(0)
     }
 
-    pub fn check_dir(&self) -> CommonResult<()> {
-        let m = try_err!(self.path.metadata());
+    pub fn check_dir(&self) -> SysResult<()> {
+        let m = self.path.metadata()?;
 
         if !m.is_dir() || m.permissions().readonly() {
-            return err_box!("Directory is not writable: {:?}", self.path());
+            return sys_error!("Directory is not writable: {:?}", self.path());
         }
 
         let new_device_id = sys::get_device_id(self.path());
@@ -92,11 +92,20 @@ impl FsStats {
 
         Ok(())
     }
+
+    fn space_or_zero(result: SysResult<u64>) -> u64 {
+        match result {
+            Ok(value) => value,
+            Err(error) => {
+                warn!("{}", error);
+                0
+            }
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::common::ByteUnit;
     use crate::sys::FsStats;
 
     #[test]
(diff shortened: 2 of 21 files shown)


orpc/src/sys/pipe/async_fd.rs
```diff
@@ -147,12 +138,17 @@ impl AsyncFd {
...
             let res = self.async_write(|fd| sys::writev(fd.fd(), iov)).await;
 
-            io_loop_check!(res, len)?;
+            match res {
+                Ok(written) if written == len => break,
+                Ok(written) => return sys_error!("writev returned {}, expected {}", written, len),
+                Err(error) if error.kind() == std::io::ErrorKind::WouldBlock => continue,
+                Err(error) => return Err(error),
+            }
         }
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in d0989873.

One correction on the premise: the pre-refactor code did not loop to completion either. It went through `io_loop_check!(res, len)`, whose `Ok(v)` arm returned `err_box!("io return value error, res: {}, expect: {}", ...)` whenever `v != expect`, so a short `writev` was already a hard error before this PR. This refactor preserved that behavior rather than introducing it.

That said, the underlying concern is valid, so `AsyncFd::write_iov` now resumes instead of failing: it tracks how many bytes have been written, advances the iovec view for each subsequent attempt, and only returns an error when `writev` makes no progress at all (`WriteZero`, "writev returned 0"). This mirrors `Pipe2::write_iov`, which already had the correct advancing loop, and the advance helper is now shared as `sys::skip_iov_bytes` instead of being duplicated.

orpc/src/sys/mod.rs
```diff
@@ -12,6 +12,28 @@
...
+pub type SysResult<T> = std::io::Result<T>;
+
+macro_rules! sys_error {
+    ($message:expr) => {
+        Err(std::io::Error::other($message))
+    };
+    ($format:expr, $($arg:expr),+ $(,)?) => {
+        Err(std::io::Error::other(format!($format, $($arg),+)))
+    };
+}
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in d0989873. `sys_error!` now takes an optional leading `ErrorKind`, and the sys layer passes an explicit kind wherever the failure is categorizable:

- `ErrorKind::Unsupported` for the platform-gated `#[cfg(not(target_os = "linux"))]` paths and for the "no async fd available" cases.
- `ErrorKind::InvalidInput` for argument/precondition violations (blocking fd used for async IO, size exceeding the pipe buffer, `CString::new` on a string containing NUL, empty signal set).
- `ErrorKind::UnexpectedEof` / `ErrorKind::WriteZero` for zero-progress transfers (`splice` / `vmsplice` / `writev` returning 0).
- `ErrorKind::PermissionDenied` for a non-writable directory, `ErrorKind::BrokenPipe` for a closed signal stream.

The bare `sys_error!("...")` arm is kept for the remaining opaque cases only, as suggested.

orpc/src/sys/sys_libc.rs
```diff
@@ -175,46 +144,19 @@ pub fn splice_out_full(
...
+) -> SysResult<()> {
     let mut remaining = len;
     while remaining > 0 {
         let transferred = splice(fd_in, off_in.as_mut(), fd_out, off_out.as_mut(), remaining)?;
         if transferred == 0 {
-            return err_box!("unsupported operation");
+            return sys_error!("unsupported operation");
         }
         remaining -= transferred as usize;
     }
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in d0989873. `splice_out_full` now returns `ErrorKind::UnexpectedEof` with the message "splice returned 0", which matches the equivalent no-progress check already present in `Pipe2::read_io`.

orpc/src/io/local_file.rs
```diff
@@ -274,7 +286,7 @@ impl LocalFile {
 
     pub fn actual_size(&self) -> IOResult<u64> {
         let meta = self.inner.metadata()?;
-        sys::file_actual_size(meta)
+        Ok(sys::file_actual_size(meta)?)
     }
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in d0989873. `LocalFile::actual_size` now returns `sys::file_actual_size(meta).map_err(Into::into)` instead of the `Ok(...?)` wrapper.

lzjqsdd:
Regarding the low-confidence note on `posix_fadvise` in the review body: it is correct, and it is fixed in d0989873.

`posix_fadvise` reports failure by returning a positive errno directly, rather than returning `-1` and setting `errno`. `sys_call!` only treats values `<= -1` as failures, so every `posix_fadvise` failure was being reported as success. `read_ahead` now checks the return value explicitly:

```rust
if res == 0 {
    Ok(0)
} else {
    Err(std::io::Error::from_raw_os_error(res))
}
```

For the record, this was a pre-existing defect rather than a regression from this PR: the previous `err_io!` macro applied the same `<= -1` rule to the same call. It is fixed here because this PR is already touching that call site.

Verification for the whole review round: `make format` passes (rustfmt + clippy clean across the workspace) and `cargo test -p orpc` passes with 78 tests, 0 failures.

orpc/src/sys/fs_stats.rs
```diff
@@ -61,27 +61,31 @@ impl FsStats {
...
 
-    pub fn check_dir(&self) -> CommonResult<()> {
-        let m = try_err!(self.path.metadata());
+    pub fn check_dir(&self) -> SysResult<()> {
+        let m = self.path.metadata()?;
 
         if !m.is_dir() || m.permissions().readonly() {
-            return err_box!("Directory is not writable: {:?}", self.path());
+            return sys_error!(
+                std::io::ErrorKind::PermissionDenied,
```
szbr00, line 93: This branch covers both "not a directory" and "not writable", but always reports `ErrorKind::PermissionDenied`. For the `!m.is_dir()` case that kind is misleading and makes caller-side matching harder — especially now that this PR is standardizing explicit kinds.

Consider splitting the checks, e.g. `ErrorKind::NotADirectory` (or `InvalidInput`) when the path is not a directory, and keep `PermissionDenied` only for the readonly case.

orpc/src/sys/mod.rs
```diff
@@ -12,6 +12,56 @@
...
+
+macro_rules! sys_call {
+    ($expression:expr) => {{
+        let result = $expression;
+        if result as CInt <= ERRNO_SENTINEL {
```
szbr00: Unlike the existing `err_io!` macro, `sys_call!` refers to bare `CInt` / `ERRNO_SENTINEL` instead of `$crate::sys::…`. That works today because every call site is in `sys_libc` with `use crate::sys::*`, but it will become brittle once these macros move with the sys extraction.

Prefer `$crate::sys::CInt` and `$crate::sys::ERRNO_SENTINEL` so expansion does not depend on call-site imports.

orpc/src/sys/sys_libc.rs
```diff
@@ -268,16 +214,16 @@ pub fn pipe2(size: usize) -> IOResult<[RawIO; 2]> {
...
-            err_io!(set_buf_res)?;
+            sys_call!(set_buf_res)?;
 
             let set_buf_res = libc::fcntl(fds[0], libc::F_SETPIPE_SZ, size);
-            err_io!(set_buf_res)?;
+            sys_call!(set_buf_res)?;
 
             if (set_buf_res as usize) < size {
-                return err_box!(
+                return sys_error!(
```
szbr00, line 226: This failure is a clear argument/config violation (requested pipe size was not honored), but the no-kind `sys_error!` form maps it to `ErrorKind::Other`. For consistency with the rest of this PR, consider `ErrorKind::InvalidInput` here.

orpc/src/sys/fs_stats.rs
lzjqsdd: Thanks, agreed. Fixed in d62534e5.

`check_dir` now tests the two conditions separately: a non-directory path returns `ErrorKind::NotADirectory` with a "Not a directory" message, and `PermissionDenied` is kept only for the read-only case. `NotADirectory` is stable since Rust 1.83 and this repo pins 1.92.0, so it is available.

No caller matched on the previous kind (`vfs_dir.rs` and `dir_list.rs` only propagate or log the error), so this is a message/kind refinement with no behavioral impact on existing call sites.

orpc/src/sys/mod.rs
lzjqsdd: Good catch, fixed in d62534e5. `sys_call!` now expands to `$crate::sys::CInt` and `$crate::sys::ERRNO_SENTINEL`, matching what `err_io!` did.

I also changed both macros to use `::std::io::Error` instead of `std::io::Error`, since that is the same class of hygiene issue: the expansion should not depend on what the call site has in scope. Note that the `$kind` argument of `sys_error!` is deliberately left as a caller-supplied path, so call sites keep writing `ErrorKind::Unsupported` with their own import.

orpc/src/sys/sys_libc.rs
lzjqsdd: Agreed, fixed in d62534e5. The unhonored pipe size now reports `ErrorKind::InvalidInput` instead of falling back to `ErrorKind::Other`, consistent with the other argument/precondition violations in this PR.

szbr00 (approved):

PR #1375, opened 2026-07-28
fix(master): defer deletion of writing blocks (#1370)

### LINKED ISSUE #1370: [BUG]: Load job fails with ENOENT during worker complete block (oss-hdfs → CV)
This is **not** an OSS missing-object error. The failure happens on the **CV write path**: Worker `WriteHandler.complete` → `finalize_block` → `BlockMeta::with_final` → `path.metadata()`, when the local block file (`blk_<id>`) is already gone.

Observed impact: one failed task fails the entire job (e.g. 800 tasks, ~369GB partially loaded). Topology: single Master + single Worker.

**Suspected root cause**

Race / incorrect delete of an in-flight Writing block:

1. Worker reports Writing/Finalized blocks via heartbeat / incr block report.
2. Master `block_report` may classify a block as `Missing` / `NotFile` and schedule `DeleteBlock`.
3. Worker `async_remove_block` unlinks the file while the load writer still holds the write stream.
4. Later `complete` calls `metadata()` on the path → ENOENT.

Alternative operational causes (disk unmount / data-dir wiped) produce the same symptom and should be ruled out.

Relevant code:

- `curvine-client/src/block/block_writer.rs` (`failed to complete block`)
- `curvine-client/src/block/block_client.rs` (`rpc failed to worker`)
- `curvine-server/src/worker/handler/write_handler.rs` (`complete` → `finalize_block`)
- `curvine-server/src/worker/block/block_meta.rs` (`with_final` / `get_len`)
- `curvine-server/src/master/fs/master_filesystem.rs` (Missing/NotFile → `delete_blocks`)
- `curvine-server/src/worker/block/heartbeat_task.rs` (`async_remove_block`)

**To Reproduce**

1. Deploy Curvine with 1 Master + 1 Worker; mount oss-hdfs in `cache_mode`.
2. Submit a large directory load job (many files / multi-hundred GB), e.g. hundreds of tasks.
3. Observe intermittent task failure near end of write/complete with ENOENT + `failed to complete block on worker`.
4. Master aggregates: `job ... execute failed, tasks N, ... error task ... failed: No such file or directory (os error 2)...`

(Exact repro may require concurrent heartbeat delete vs write complete; check Master logs for `block_report found ... missing-inode` / DeleteBlock around the same timestamp.)

**Expected behavior**

- In-flight Writing blocks must not be deleted by Master/Worker reconciliation.
- Load tasks should not fail with ENOENT on complete when OSS read succeeded and data was written.
- If a block is invalid, fail with an explicit error (and avoid deleting under an open writer), rather than a bare OS ENOENT.

- Cluster: single Master + single Worker
- Workload: oss-hdfs → CV load job
- Observed: 2026-07-28; job with ~800 tasks / ~369GB; worker `10.41.54.241:7997`
- Curvine version/branch: (please fill)

**Additional context**

Master log excerpt:
WARN job_context.rs: job ... execute failed, tasks 800, len = 369.0GB, cost 158165 ms, error task ..._task_624 failed: No such file or directory (os error 2): rpc failed to worker ... addr = 10.41.54.241/10.41.54.241:7997 failed to complete block on worker_id = ...

Worker (load runner) log excerpt:
ERROR fs_writer_buffer.rs: buffer writer error: Common([rpc failed to worker ... failed to complete block on worker ...]No such file or directory (os error 2)) ERROR load_task_runner.rs: task ... execute failed: No such file or directory (os error 2): ...

**Suggested investigation / fix directions**

1. Do not issue `DeleteBlock` for blocks still in Writing / leased by an active write session.
2. Correlate Master delete commands with Worker complete ENOENT in the same window.
3. Improve error message when finalize fails (include block id + path) to distinguish delete-race vs disk failure.
4. Consider partial-success job semantics separately (already tracked if related feature exists).

jlon:
# Summary

Protect in-flight Worker writes reported incrementally, while preserving full-report cleanup of stale Worker staging blocks.

# Issue Describe / Design

Closes #1370.

Root cause: the original fix deferred deletion for every `Writing` block whose inode was Missing or NotFile. Worker restart recovery reports local `Recovering` blocks as `Writing` in a full block report, so unowned staging blocks could remain indefinitely and consume capacity.

Reproduction: report a `Writing` block for a deleted file inode in a full block report. The previous implementation returned no DeleteBlock command.

Fix: defer deletion only for incremental `Writing` reports, where the Master must not remove an in-flight write. A full report continues through the existing Missing/NotFile deletion path, removing stale recovered staging blocks.

# Changes

| `curvine-server/src/master/fs/master_filesystem.rs` | Limit Writing-block deletion deferral to incremental reports | Full reports again delete Missing/NotFile orphan blocks; incremental write protection remains |
| `curvine-server/tests/master_fs_test.rs` | Cover Writing + Missing for incremental and full reports | Guards both sides of the report-mode contract |

# Test verified

| `cargo test -p curvine-server --test master_fs_test -- --test-threads=1` | PASS | 35 passed |
| `make format` | PASS | Workspace formatting and release clippy checks |

# Dependencies

- Related PRs: None
- Related issues: #1370, #1024
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/master/fs/master_filesystem.rs
+++ b/curvine-server/src/master/fs/master_filesystem.rs
@@ -1232,6 +1232,8 @@ impl MasterFilesystem {
         for item in list.blocks {
             match item.status {
                 BlockReportStatus::Finalized | BlockReportStatus::Writing => {
+                    let defer_writing_delete =
+                        item.status == BlockReportStatus::Writing && !list.full_report;
                     let state = match self.block_inode_state(item.id) {
                         Ok(v) => v,
                         Err(e) => {
@@ -1241,6 +1243,18 @@ impl MasterFilesystem {
                     };
                     match state {
                         BlockInodeState::File => checked.push((item, Some(BlockInodeState::File))),
+                        BlockInodeState::Missing if defer_writing_delete => {
+                            warn!(
+                                "block_report deferred deletion for writing block {} on worker {} because its inode is missing",
+                                item.id, list.worker_id
+                            );
+                        }
+                        BlockInodeState::NotFile if defer_writing_delete => {
+                            warn!(
+                                "block_report deferred deletion for writing block {} on worker {} because its inode is not a file",
+                                item.id, list.worker_id
+                            );
+                        }
                         BlockInodeState::Missing => {
                             missing_blocks += 1;
                             delete_blocks.push(item.id);
--- a/curvine-server/tests/master_fs_test.rs
+++ b/curvine-server/tests/master_fs_test.rs
@@ -346,6 +346,90 @@ fn block_report_for_non_file_inode_schedules_worker_delete() -> CommonResult<()>
 }
 
 #[test]
+fn block_report_for_writing_non_file_inode_defers_worker_delete() -> CommonResult<()> {
+    let _serial = master_fs_test_serial();
+    let fs = new_fs(true, "block-report-writing-non-file");
+    fs.mkdir("/dir-block", true)?;
+    let dir_status = fs.file_status("/dir-block")?;
+    let block_id = InodeId::create_block_id(dir_status.id, 0)?;
+
+    let result = fs.block_report(
+        BlockReportList {
+            cluster_id: "curvine".into(),
+            worker_id: 0,
+            full_report: false,
+            total_len: 1,
+            blocks: vec![BlockReportInfo::new(
+                block_id,
+                BlockReportStatus::Writing,
+                StorageType::Disk,
+                1,
+            )],
+        },
+        None,
+    )?;
+
+    assert!(result.delete_blocks.is_empty());
+    Ok(())
+}
+
+#[test]
+fn block_report_for_writing_missing_inode_defers_worker_delete() -> CommonResult<()> {
+    let _serial = master_fs_test_serial();
+    let fs = new_fs(true, "block-report-writing-missing");
+    let file = fs.create("/missing", false)?;
+    fs.delete("/missing", false)?;
+    let block_id = InodeId::create_block_id(file.id, 0)?;
+
+    let result = fs.block_report(
+        BlockReportList {
+            cluster_id: "curvine".into(),
+            worker_id: 0,
+            full_report: false,
+            total_len: 1,
+            blocks: vec![BlockReportInfo::new(
+                block_id,
+                BlockReportStatus::Writing,
+                StorageType::Disk,
+                1,
+            )],
+        },
+        None,
+    )?;
+
+    assert!(result.delete_blocks.is_empty());
+    Ok(())
+}
+
+#[test]
+fn full_block_report_for_writing_missing_inode_schedules_worker_delete() -> CommonResult<()> {
+    let _serial = master_fs_test_serial();
+    let fs = new_fs(true, "full-block-report-writing-missing");
+    let file = fs.create("/missing", false)?;
+    fs.delete("/missing", false)?;
+    let block_id = InodeId::create_block_id(file.id, 0)?;
+
+    let result = fs.block_report(
+        BlockReportList {
+            cluster_id: "curvine".into(),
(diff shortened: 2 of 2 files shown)


curvine-server/src/master/fs/master_filesystem.rs
```diff
@@ -924,6 +924,18 @@ impl MasterFilesystem {
                     };
                     match state {
                         BlockInodeState::File => checked.push((item, Some(BlockInodeState::File))),
+                        BlockInodeState::Missing if item.status == BlockReportStatus::Writing => {
```
szbr00: This unconditionally skips DeleteBlock for every Writing report with a Missing/NotFile inode, including full-report orphans.

On worker restart, staging files are scanned as `BlockState::Recovering` and reported as `BlockReportStatus::Writing` (see `master_client::full_block_report`). Those Recovering blocks never transition to Finalized, so the "later Finalized report cleans orphans" assumption does not hold — they can sit forever and keep consuming capacity.

Please narrow the deferral, for example:
- still schedule deletion when `list.full_report` is true (keep #1024 orphan cleanup), and only defer on incremental reports; or
- report Recovering as a distinct status and delete Missing/NotFile Recovering blocks; or
- discard unowned Recovering staging locally on the worker before reporting.

As written, this regresses the intentional Missing/NotFile DeleteBlock cleanup from #1024 for the Writing/Recovering path.

curvine-server/tests/master_fs_test.rs
```diff
@@ -341,6 +341,34 @@ fn block_report_for_non_file_inode_schedules_worker_delete() -> CommonResult<()>
     Ok(())
 }
 
+#[test]
+fn block_report_for_writing_non_file_inode_defers_worker_delete() -> CommonResult<()> {
```
szbr00, line 349: This covers Writing + NotFile (directory inode), but #1370's production path is more likely Writing + Missing.

Please add a sibling test that reports Writing for a block_id whose file inode does not exist and asserts `delete_blocks` is empty. If orphan cleanup on full_report is preserved (see the master_filesystem comment), also cover Writing + Missing with `full_report: true` still scheduling deletion.

jlon: Thanks. Confirmed: the current test only covers Writing + NotFile and misses the #1370 Writing + Missing path. I will add coverage for incremental deferral and for full-report cleanup together with the fix.

curvine-server/src/master/fs/master_filesystem.rs
jlon: Thanks. Confirmed: Recovering blocks are reported as Writing during a full report, so the current unconditional deferral would retain unowned staging blocks and regress the #1024 cleanup. I will limit deferral to incremental reports and keep Missing/NotFile deletion for full reports.

jlon: Fixed in b9e62bff. Deletion is now deferred only for incremental Writing reports; full reports retain the Missing/NotFile deletion path, so Recovering orphan staging blocks are cleaned up. Covered by master_fs_test (35 passed).

curvine-server/tests/master_fs_test.rs
jlon: Fixed in b9e62bff. Added Writing + Missing regression coverage for both incremental reports (no DeleteBlock) and full reports (DeleteBlock scheduled). master_fs_test passes with 35 tests.

jlon:
Review follow-up:

- The full-report orphan cleanup concern is fixed in b9e62bff: deletion is deferred only for incremental `Writing` reports; full reports still delete `Missing` and `NotFile` blocks.
- Added `Writing + Missing` coverage for both incremental deferral and full-report deletion.

Verified with `cargo test -p curvine-server --test master_fs_test -- --test-threads=1` (35 passed) and `make format`.

szbr00 (approved):

PR #1377, opened 2026-07-28
fix(client): restore pwrite01 readback without EIO (#1337)

### LINKED ISSUE #1337: [BUG]: FUSE pwrite01 readback returns EIO

After pwrite on Curvine FUSE, the verification read fails with EIO(5).

pwrite01 reports read() failed nread=-1 errno=EIO(5).

Reproduction:

2. Run LTP case pwrite01 against the mount.
3. Observe the failure versus Linux reference.

**Impact**

Write-then-read workloads via pwrite can fail on Curvine FUSE.

Readback after pwrite should return written bytes without EIO.

Fix pwrite/write commit and readback integrity on curvine-fuse.

**Additional context**

Root-cause statement is a hypothesis. LTP resume 20260727T101138Z-9c0d2985 completed the post-sendfile09 remainder.

Root-cause hypothesis: Root-cause hypothesis: Curvine FUSE pwrite path leaves data unreadable and returns EIO on verification read.

Evidence summary: pwrite01 1 TBROK : pwrite01.c:294: read() failed: nread=-1: errno=EIO(5): Input/output error | pwrite01 2 TBROK : pwrite01.c:294: Remaining cases broken

Acceptance / test gaps:

- Case passes on curvine-fuse
- Linux-compatible errno/behavior
- No unrelated API changes

LJF-050:
# Summary

Fix FUSE `pwrite01` / `pwrite01_64` write-then-read `EIO` by clearing stale resize `alloc_opts` after block commit and publishing rewrite blocks on flush, so verification reads see the committed data.

# Issue Describe / Design

Closes #1337

## Root cause

1. Sparse / `pwrite` paths attach resize `alloc_opts` on the block. Master commit clears those opts and updates length, but the client `WriteFileBlocks::add_commit` only updated `block.len` and kept stale `alloc_opts`.
2. The next rewrite reopen re-applied the old truncate, shrinking the staging copy and causing block length mismatch (`EIO`) on the LTP verification read.

## Fix approach

# Changes

| `curvine-client/src/file/fs_writer_base.rs` | Finalize worker blocks on flush; refresh `file_blocks` after `only_flush` | Published sparse/`pwrite` data becomes durable/readable while lease stays open |
| `curvine-tests/tests/pwrite_flush_read_test.rs` | New integration coverage | Locks in pwrite-after-flush readback |
| `curvine-tests/tests/*flush*eof*.rs` | Related sparse/flush/EOF coverage | Guards sibling write-then-read paths |

# Test verified

| `make format` | PASS | On exact HEAD `9145c76c` |
| `cargo test -p curvine-tests --test pwrite_flush_read_test --test sparse_hole_flush_read_test --test fuse_read_past_eof_test --test ftest_flush_stress_test` | PASS | |
| Full LTP profile before on `734b19de` | FAIL (expected) | Contains `pwrite01` / `pwrite01_64` EIO signatures |
| Full LTP profile after on `9145c76c` | PASS for assigned cases | `pwrite01` / `pwrite01_64` absent; no new failures vs baseline |
| Independent review | APPROVED | Bound to exact HEAD `9145c76c` |

# Dependencies

- Related PRs: #1367 (same root-cause fix for LTP ftest EIO)
- Related issues: #1336
- Blocks / blocked by: None


curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -111,8 +111,15 @@ impl BackendHandle {
         };
 
         if let Some(writer) = state.find_writer(self.ino).await {
-            let writer_ver = writer.write_ver();
-            if self.read_ver.get() != writer_ver {
+            // `write_ver` advances when writes are enqueued, not when they land.
+            // Concurrent fork writers (LTP ftest) can enqueue after we schedule a
+            // flush; sample the version we flushed and loop until stable so we
+            // never publish `read_ver` past data that is still queued.
+            for _ in 0..16 {
```
szbr00: If concurrent writers keep advancing `write_ver` faster than this loop can flush, we exit after 16 iterations and still call `reader.read(...)` with `read_ver` possibly behind `write_ver`. That reintroduces the stale/unpublished dirty-read the loop is meant to close, just bounded instead of open-ended.

Please either keep waiting with a time/deadline budget, or fail the read (e.g. EAGAIN/EIO) / at least log when the budget is exhausted so we never silently serve a caught-up-looking reader that is still behind.

curvine-client-core/src/file/fs_writer_base.rs
```diff
@@ -295,23 +295,26 @@ impl FsWriterBase {
...
+            // Always finalize on the worker. `only_flush` only keeps the master
+            // write lease open; it must still publish block data.
+            //
+            // A prior resize/flush may already have finalized an older
+            // generation. Later writes reopen as a staging rewrite while
+            // `get_readable_block` prefers the committed generation. Calling
+            // `flush()` here without `complete()` leaves that rewrite
+            // unpublished, so FUSE dirty-read / LTP ftest/pwrite see EIO or
+            // stale holes after sparse write-then-read.
+            let commit_block = writer.complete().await?;
```
szbr00, line 311: Always calling `writer.complete()` on `only_flush` is the right fix for rewrite generations (`get_readable_block` prefers the committed copy until finalize). For first-generation Writing blocks, the old `flush()` + commit-metadata path was already readable, so unconditional finalize + `cache_writers.clear()` adds an extra worker finalize/reopen on every FUSE flush.

If rewrite vs first-write is distinguishable here, consider finalizing only rewrite/staging blocks and keeping `flush()` for first-generation writers to limit write-after-flush reopen cost on the hot path.

curvine-fuse/src/fs/state/backend_handle.rs
szbr00: Each loop iteration now triggers a full publish flush (worker finalize + master `only_flush` + reader reopen). Under forked writers this can amplify to many publish cycles per FUSE read.

Worth capping with a clear failure (see sibling comment on the loop budget) and/or coalescing dirty-read flushes so concurrent readers share one in-flight publish instead of stacking up to 16× each.

szbr00 (commented):
Heads-up: this PR’s four fix commits (`89c2af60`, `66b7934c`, `a46380f3`, `9145c76c`) are the same commits as open PR #1367 (ftest/#1336). Please coordinate so only one lands, or rebase one onto the other to avoid duplicate merges / conflicting follow-ups.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Fixed in 0c309ecc8bc6f7e647085286cc9de0688f47c510: dirty-read now fails with `warn!` + `EAGAIN` when the flush budget is exhausted with an unstable `write_ver`, instead of silently serving a reader that may still be behind. Also aligned with #1367: `write_ver` bumps only after enqueue and `enqueue_inflight` closes the incr-before-enqueue window.

curvine-client-core/src/file/fs_writer_base.rs
LJF-050: Thanks for the suggestion. Kept unconditional `complete()` on the `only_flush` path for now: rewrite/staging generations are not reliably distinguishable here yet, and skipping finalize for first-generation writers would reintroduce the unpublished-rewrite EIO/hole class of bugs this PR targets (pwrite01 / sparse write-then-read). A rewrite-only finalize optimization needs a clear staging signal and can be a follow-up once that is available.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Addressed the failure mode in 0c309ecc8bc6f7e647085286cc9de0688f47c510: budget exhaustion now returns `EAGAIN` (with a warn) rather than stacking silent publish cycles and serving stale data. Shared in-flight publish coalescing across concurrent readers is a good follow-up optimization; not done in this minimal correctness fix (same approach as #1367).

LJF-050:
Thanks for the heads-up on overlap with #1367. Acknowledged: the shared fix commits and the dirty-read follow-ups in 0c309ecc8bc6f7e647085286cc9de0688f47c510 are intentionally aligned with #1367 (enqueue_inflight + flush-budget EAGAIN). Please land only one of these PRs first (or rebase the other afterward) to avoid duplicate merges; happy to close/rebase the remaining branch once the first one merges.

PR #1378, opened 2026-07-28
fix(fuse): stop pread01 EIO on prepared file reads (#1338)

### LINKED ISSUE #1338: [BUG]: FUSE pread01 returns EIO at valid offsets

LTP pread01 fails with error:5 when reading previously prepared file content on Curvine FUSE.

pread01 reports pread() nread=-1, error:5.

Reproduction:

2. Run LTP case pread01 against the mount.
3. Observe the failure versus Linux reference.

**Impact**

Applications using pread on Curvine FUSE can observe EIO on valid reads.

pread at valid offsets should return data without EIO.

Fix FUSE pread/read data path integrity.

**Additional context**

Root-cause statement is a hypothesis. LTP resume 20260727T101138Z-9c0d2985 completed the post-sendfile09 remainder.

Root-cause hypothesis: Root-cause hypothesis: Curvine FUSE pread path returns EIO for readable offsets.

Evidence summary: pread01 1 TFAIL : pread01.c:122: pread() at off. K2 failed: nread=-1, error:5

Acceptance / test gaps:

- Case passes on curvine-fuse
- Linux-compatible errno/behavior
- No unrelated API changes

LJF-050:
# Summary

Fix FUSE `pread01` / `pread01_64` returning `EIO` at valid offsets by clearing stale resize `alloc_opts` after block commit and publishing rewrite blocks on flush, so prepared-file verification reads see the committed data.

# Issue Describe / Design

Closes #1338

## Root cause

1. Sparse / write paths attach resize `alloc_opts` on the block. Master commit clears those opts and updates length, but the client `WriteFileBlocks::add_commit` only updated `block.len` and kept stale `alloc_opts`.
2. The next rewrite reopen re-applied the old truncate, shrinking the staging copy and causing block length mismatch (`EIO`) on the LTP `pread01` / `pread01_64` verification reads.

## Fix approach

# Changes

| `curvine-client/src/file/fs_writer_base.rs` | Finalize worker blocks on flush; refresh `file_blocks` after `only_flush` | Published sparse/write data becomes durable/readable while lease stays open |
| `curvine-tests/tests/*flush*eof*.rs` | Integration coverage for flush/read and past-EOF | Guards write-then-read and pread verification paths |

# Test verified

| `make format` | PASS | On exact HEAD `9145c76c` |
| `cargo test -p curvine-tests --test pwrite_flush_read_test --test sparse_hole_flush_read_test --test fuse_read_past_eof_test --test ftest_flush_stress_test` | PASS | Related write-then-read / EOF coverage |
| LTP before (`734b19de`) | FAIL as expected | Contains `pread01` / `pread01_64` EIO signatures |
| LTP after (`9145c76c`) | PASS for assigned cases | `pread01` / `pread01_64` absent; `new_failures=[]`; cleanup passed |
| Independent review | APPROVED | Bound to exact HEAD `9145c76c` |

# Dependencies

- Related PRs: #1367, #1377 (same root-cause fix for ftest/pwrite)
- Related issues: #1336, #1337
- Blocks / blocked by: None


curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -111,8 +111,15 @@ impl BackendHandle {
         };
 
         if let Some(writer) = state.find_writer(self.ino).await {
-            let writer_ver = writer.write_ver();
-            if self.read_ver.get() != writer_ver {
+            // `write_ver` advances when writes are enqueued, not when they land.
+            // Concurrent fork writers (LTP ftest) can enqueue after we schedule a
+            // flush; sample the version we flushed and loop until stable so we
+            // never publish `read_ver` past data that is still queued.
+            for _ in 0..16 {
```
szbr00: If write_ver keeps advancing under concurrent fork writers, this loop exits after 16 iterations and still serves the read. That can return data that is behind writes still queued after the last flush, with no error to the caller.

Please fail the read (e.g. EAGAIN/EIO) when the version is still unstable after the budget, or keep looping with a time-based deadline instead of silently falling through. At minimum, log when the cap is hit so this path is visible in the field.

curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -111,8 +111,15 @@ impl BackendHandle {
...
+                let ver_before = writer.write_ver();
+                if self.read_ver.get() == ver_before {
+                    break;
+                }
                 writer.flush(None).await?;
 
                 let path = reader.path().clone();
```
szbr00, line 146: Each loop iteration calls flush() and then new_reader() (a master open). Under contended writers this can mean up to 16 flushes and 16 reopens per FUSE read, and flush now finalizes worker sessions via complete(), so the reopen cost is higher than before.

Consider flushing in a tight loop until write_ver is stable (or the budget fails), then reopen/replace the reader once. That preserves the stability guarantee with far less metadata traffic on the dirty-read path.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Fixed in 4a0adc4: after the flush budget is exhausted with an unstable write_ver, the dirty-read path now returns EAGAIN and logs a warning instead of silently serving the read.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Fixed in 4a0adc4: flush now runs in a tight loop until write_ver is stable (or the budget fails), then the reader is reopened/replaced once, avoiding repeated master opens on each iteration.

szbr00 (approved):

PR #1380, opened 2026-07-28
fix(build): fix build script failing on macOS due to sed incompatibility (#1379)

### LINKED ISSUE #1379: [BUG]: build/build.sh fails on macOS due to BSD sed -i syntax
**Describe the bug**
`build/build.sh` fails on macOS when generating Python protobuf stubs. 
The `sed -i -E` command uses GNU sed syntax which is incompatible with BSD sed (the default sed on macOS).

**To Reproduce**
Run `./build/build.sh` on macOS

**Expected behavior**
Build succeeds on both Linux and macOS.

**Screenshots**
Image

 macOS (any version)

**Additional context**

codenohup:
# Summary

Fix build script failing on macOS due to `sed` incompatibility between BSD and GNU implementations.

# Issue Describe / Design

Closes #1379

**Root cause**: `build/build.sh` uses `sed -i -E`, which is GNU sed syntax. macOS ships BSD sed, where `-i` requires an explicit backup extension argument. The command is parsed as `-i` with backup suffix `-E`, breaking `\1` backreferences.

See attached `man sed` screenshots confirming BSD sed behavior on macOS and Linux.
macOS:
image
Linux:
image

**Fix**: Use the portable `sed -i.bak -E` form supported by BSD and GNU sed, then remove the generated backup file. This also works when the `sed` implementation found on `PATH` does not match the host OS.

# Changes

| `build/build.sh` | Use a portable sed in-place edit and clean up its backup | None |

# Test verified

| macOS: `./build/build.sh` | PASS | image |
| `make format` | PASS | Includes repository pre-commit checks |
| `bash -n build/build.sh` | PASS | Shell syntax validation |
| BSD sed import rewrite and backup cleanup | PASS | Verified generated import output and removal of `.bak` file |

# Dependencies

- Related PRs: None
- Related issues: #1379
- Blocks / blocked by: None

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -883,11 +883,8 @@ if [ $BUILD_PYTHON_SDK -eq 1 ]; then
   protoc -I"$PROTO_DIR" --python_out="$PROTO_PKG" "$PROTO_DIR"/*.proto
   for f in "$PROTO_PKG"/*_pb2.py; do
     if [ -f "$f" ]; then
-      if [[ "$OSTYPE" == "darwin"* ]]; then
-        sed -i '' -E 's/^import ([A-Za-z0-9_]+_pb2)( as .*)$/from . import \1\2/' "$f"
-      else
-        sed -i -E 's/^import ([A-Za-z0-9_]+_pb2)( as .*)$/from . import \1\2/' "$f"
-      fi
+      sed -i.bak -E 's/^import ([A-Za-z0-9_]+_pb2)( as .*)$/from . import \1\2/' "$f"
+      rm -f "$f.bak"
     fi
   done
 
--- a/build/build.sh
+++ b/build/build.sh
@@ -883,7 +883,11 @@ if [ $BUILD_PYTHON_SDK -eq 1 ]; then
   protoc -I"$PROTO_DIR" --python_out="$PROTO_PKG" "$PROTO_DIR"/*.proto
   for f in "$PROTO_PKG"/*_pb2.py; do
     if [ -f "$f" ]; then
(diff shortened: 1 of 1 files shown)


build/build.sh
```diff
@@ -883,7 +883,11 @@ if [ $BUILD_PYTHON_SDK -eq 1 ]; then
   protoc -I"$PROTO_DIR" --python_out="$PROTO_PKG" "$PROTO_DIR"/*.proto
   for f in "$PROTO_PKG"/*_pb2.py; do
     if [ -f "$f" ]; then
-      sed -i -E 's/^import ([A-Za-z0-9_]+_pb2)( as .*)$/from . import \1\2/' "$f"
+      if [[ "$OSTYPE" == "darwin"* ]]; then
```
szbr00: The `$OSTYPE` branch fixes the reported macOS/BSD sed issue and matches the existing pattern in `get_os_version()`, so this looks good for #1379.

One optional hardening: branching on OS still assumes the system `sed` matches the OS (e.g. Homebrew GNU `sed` earlier on `PATH` would break the Darwin arm). A single portable form avoids the duplicate regex and works on both BSD and GNU sed:

```bash
sed -i.bak -E 's/^import ([A-Za-z0-9_]+_pb2)( as .*)$/from . import \1\2/' "$f"
rm -f "$f.bak"
```

Alternatively, detect sed flavor once (`sed --version` → GNU) and set a `SED_INPLACE` array. Not a blocker for this PR.

szbr00: Thanks. Fixed in 9879566f: replaced the OS-specific branches with the portable `sed -i.bak` form and removed the backup after the rewrite.

szbr00 (approved):

PR #1381, opened 2026-07-28
fix(fuse): align link/symlink permission checks with POSIX (#1253)

### LINKED ISSUE #1253: [BUG]: FUSE link and symlink operations ignore POSIX permission boundaries

linkat01 fails with ENOENT where link should succeed; prot_hsymlinks and link01 allow symlink/hardlink creation that should be denied.

linkat01 2 TFAIL linkat failed unexpectedly TEST_ERRNO=ENOENT(2)
link01 1 TFAIL symlink(object, symbolic) was succesful

Reproduction:

1. Mount curvine-fuse
2. Run LTP linkat01
3. Observe ENOENT on valid linkat

**Impact**

Incorrect link semantics can bypass directory permission models.

linkat should succeed for valid paths; symlink/hardlink across owners should be denied per LTP.

Align link/linkat/symlink permission checks with POSIX.

**Additional context**

Root-cause hypothesis: link/linkat/symlink permission semantics on curvine-fuse are too permissive (hypothesis: missing owner/mode checks).

Evidence summary: linkat01 ENOENT on expected success; prot_hsymlinks and link01 allow forbidden hard/symlink operations.

Acceptance / test gaps:

- linkat01, prot_hsymlinks, link01 pass

LJF-050:
# Summary

Align Curvine FUSE `link`/`symlink` permission checks with Linux POSIX semantics so directory-entry creation is denied when the parent lacks write/search permission, sticky hard-link ownership rules are enforced, and hard links to dangling symlink inodes succeed.

# Issue Describe / Design

Closes #1253

- **Root cause:** FUSE `link`/`symlink` omitted parent `W_OK|X_OK` checks; sticky `may_link` was not applied; `protected_hardlinks`/`lookup_link` used `fs_stat` on dangling symlink sources (ENOENT); master rejected hard links to `FileType::Link`.
- **Reproduction:** Mount curvine-fuse with permission checks enabled; run LTP `link01` (and related link/symlink cases). Observe symlink succeeding where denial is expected, and/or hard-link-to-dangling-symlink failures.
- **Fix approach:** Enforce parent `W_OK|X_OK` on link/symlink; apply sticky hard-link ownership checks; prefer dcache status and synthesize post-link attrs for dangling symlink targets; allow master hard links to symlink inodes.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | Parent ACL + sticky checks on `link`; dcache-based protected hardlink/status synthesis; parent ACL on `symlink` | Permission denials now match Linux for denied parents; dangling-symlink hard links succeed |
| `curvine-fuse/src/fuse_utils.rs` | Add `check_sticky_hardlink` helper + unit test | None for callers that skip sticky dirs |
| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Unit test that hard links preserve canonical source path | Test-only |
| `curvine-server/src/master/meta/fs_dir.rs` | Allow hard links to `FileType::Link` | Symlink inodes can receive additional hard links |
| `curvine-server/tests/master_fs_test.rs` | Cover hard link to dangling symlink | Test-only |

# Test verified

| `make format` | PASS | Format + clippy gates |
| `cargo test -p curvine-fuse --lib -- check_sticky_hardlink_denies_cross_owner_link hard_link_preserves_source_path` | PASS | |
| `cargo test -p curvine-server --test master_fs_test test_hardlink_to_dangling_symlink_inode` | PASS | |
| LTP originating profile on fixed HEAD | PASS for assigned `link01` | `link01`/`link06`/`linkat01`/`symlink03` TPASS; original failure absent; no new failure identities after semantic attribution of volatile historical cases |

# Dependencies

- Related PRs: None
- Related issues: #1253
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1615,17 +1615,53 @@ impl fs::FileSystem for CurvineFileSystem {
     async fn link(&self, op: Link<'_>) -> FuseResult<fuse_entry_out> {
         let name = try_option!(op.name.to_str());
         let oldnodeid = op.arg.oldnodeid;
+        let parent_ino = op.header.nodeid;
+
+        // Linux vfs_link requires MAY_WRITE|MAY_EXEC on the destination parent.
+        self.check_permissions(op.header, (libc::W_OK | libc::X_OK) as u32)
+            .await?;
+
+        // Sticky may_link checks use dcache only. Missing dcache entries must not
+        // abort the link: the kernel may still hold oldnodeid after FORGET, and
+        // master link of dangling symlink inodes is valid (LTP link01 case 2).
+        {
+            let dir = self.state.dir_read();
+            if let (Some(src), Some(dest_dir)) = (
+                dir.get_inode(oldnodeid, None),
+                dir.get_inode(parent_ino, None),
+            ) {
+                let src_uid = self.resolve_file_uid(&src.status.owner);
+                let dest_dir_uid = self.resolve_file_uid(&dest_dir.status.owner);
+                FuseUtils::check_sticky_hardlink(
+                    self.conf.check_permission,
+                    op.header.uid,
+                    dest_dir_uid,
+                    src_uid,
+                    dest_dir.mode,
+                )?;
+            }
+        }
 
         self.state.fs_fsync(oldnodeid, None).await?;
 
-        let des_path = self.state.get_path_common(op.header.nodeid, Some(name))?;
+        let des_path = self.state.get_path_common(parent_ino, Some(name))?;
         let src_path = self.state.get_path(oldnodeid)?;
         self.ensure_writable_path(&src_path, RpcCode::Link).await?;
         self.ensure_writable_path(&des_path, RpcCode::Link).await?;
 
         // fs.protected_hardlinks (LTP prot_hsymlinks): deny unsafe non-owner hardlinks.
+        // Prefer dcache status: fs_stat of a dangling symlink source can ENOENT even
+        // though the symlink inode itself is a valid hardlink target (LTP link01 case 2).
         if self.conf.check_permission {
-            let src_status = self.state.fs_stat(oldnodeid, None).await?;
+            let cached_src = {
+                let dir = self.state.dir_read();
+                dir.get_inode(oldnodeid, None)
+                    .map(|inode| inode.clone_status())
+            };
+            let src_status = match cached_src {
+                Some(status) => status,
+                None => self.state.fs_stat(oldnodeid, None).await?,
+            };
             let file_uid = self.resolve_file_uid(&src_status.owner);
             let has_read_write = self
                 .check_access_permissions(&src_status, op.header, (libc::R_OK | libc::W_OK) as u32)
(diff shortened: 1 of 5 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1642,14 +1678,30 @@ impl fs::FileSystem for CurvineFileSystem {
...
 
         self.fs.link(&src_path, &des_path).await?;
-        let attr = self
-            .state
-            .lookup_link(op.header.nodeid, name, oldnodeid)
-            .await?;
+        // After a successful master link, prefer synthesizing the new dentry from the
+        // source inode already in dcache. Fresh fs_stat("nick") can race or miss when
+        // the source was a dangling symlink (LTP link01 case 2).
+        let attr = {
```
szbr00: This post-link synthesize path regresses nlink correctness versus the previous always-`lookup_link` flow.

`DirTree::link` does `add_link(1)` then `update_status(status)`, so the status you pass must already carry the final nlink (hence the local `saturating_add(1)`). That means the attr returned to the kernel is `dcache_nlink + 1`, not the master's authoritative count. Problems:
1. Concurrent hard links on the same source can both observe nlink=N and both reply N+1 while master is at N+2.
2. If dcache nlink was already stale, the entry_out nlink stays wrong until a later getattr.
3. There is a TOCTOU window between `dir_read` (clone) and `dir_write` (link); if the inode disappears, master link has already succeeded but this returns an error.

Unless `lookup_link`/`fs_stat(parent, name)` is proven broken for the dangling-symlink case after the master `FileType::Link` fix, prefer always calling `lookup_link` here (master is source of truth for nlink). If synthesize must stay, do the read-modify under one write lock and/or fall back to `lookup_link` on failure, and set nlink from master rather than `dcache + 1`.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1710,6 +1762,10 @@ impl fs::FileSystem for CurvineFileSystem {
         let link_path = self.state.get_path_common(id, Some(linkname))?;
         self.ensure_writable_path(&link_path, RpcCode::Symlink)
```
szbr00: `link`/`unlink` check parent `W_OK|X_OK` before `ensure_writable_path`; `symlink` does the reverse. Please move this `check_permissions` above `ensure_writable_path` so denial errno prefers EACCES (POSIX parent ACL) over mount read-only errors, consistent with the other directory-entry creation paths.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 82958e2: after a successful master link we always call lookup_link so nlink comes from master fs_stat(parent, name), rather than synthesizing dcache_nlink+1. The master FileType::Link hardlink support covers the dangling-symlink case.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 82958e2: symlink now checks parent W_OK|X_OK before ensure_writable_path, matching link/unlink so ACL denial prefers EACCES over mount read-only errors.

szbr00 (approved):

Issue #1322, opened 2026-07-27
[FEATURE] Fetch block locations by range and return flush deltas


jlon:
## Is your feature request related to a problem?

`open` and `only_flush` currently return a complete `FileBlocks` list. Master
walks every block in the file and serializes every location even when a client
reads a small range or flushes only newly changed metadata. Control-plane CPU,
RPC size, and latency therefore grow with historical file size rather than the
requested I/O range.

## Describe the solution you'd like

Introduce a backward-compatible block-location protocol that supports
range/page retrieval for reads and version/delta responses for flushes.
Clients should retain a block map and expand it on demand, instead of requiring
the full file location list at open or flush.

The protocol must define snapshot/version semantics so a client cannot combine
incompatible pages or deltas.

## Describe alternatives you've considered

- Raising client reader caches only changes local reuse and cannot reduce a
  full response generated by the master.
- Splitting open and flush into separate issues would duplicate one API and
  compatibility design; both are consequences of the same full-list contract.

## Additional context

The FUSE I/O audit identifies open as W9 and `only_flush` as W12. This is
especially relevant to the sequential-write prefetch direction in #1301:
frequent flushes must not replace block-allocation stalls with full-file
location transfers.

# Goal

| Sub-requirement | Problem solved |
| --- | --- |
| Range/page location API with stable version | Avoid full-file location responses for small reads |
| Flush response with version/delta | Avoid resending unchanged locations |
| Demand-driven client block-map expansion | Keep normal reads transparent to callers |
| Compatibility path for existing clients | Permit staged rollout |

# Not Goal

- Changing replica placement policy.
- Changing file consistency, flush, or fsync visibility semantics.
- Rewriting the entire metadata RPC protocol in one change.

# Test Plan

| Test case | Test objective | Modules involved |
| --- | --- | --- |
| Read a small range from a file with many blocks | Response size and master work scale with requested range | master, client |
| Repeated flush of a large file | Verify unchanged locations are not resent | master, client, fuse |
| Concurrent append and range read | Verify version/page consistency | master, client |
| Old client and new client interoperability | Verify staged compatibility | RPC, master, client |
| Large-file metrics | Report block count, bytes, page/delta count, latency | master, client |

PR #1384, opened 2026-07-29
perf(master): skip unconsumed flush block snapshots (#1322)

jlon:
## Summary

Avoid building full block-location snapshots for flush requests whose caller does not consume them. This shortens the Master write-lock critical section for large files while preserving normal complete and flush responses.

## Issue / Design

Related to #1322.

The optional return_file_blocks protocol flag lets an internal no-result flush opt out of the FileBlocks snapshot. Legacy clients omit the field and receive the historical response.

This branch is rebased on #1386. Field 7 remains set_attr_opts; return_file_blocks is field 8. Public complete APIs always request FileBlocks because only_flush callers use that response to refresh local block state. The no-snapshot path is limited to the internal cleanup flush and batch requests, where the response is not observable. Batch requests now explicitly send false rather than relying on the legacy default.

## Changes

| Area | Change |
| --- | --- |
| master.proto | Add optional return_file_blocks field 8 with legacy-compatible default true |
| FsClient and FsWriterBase | Add internal no-result flush request while retaining set_attr and public complete behavior |
| MasterFilesystem and MasterHandler | Skip FileBlocks construction only for eligible flushes |
| master_fs_test | Cover persistence, Handler wire contract, compatibility, and the 4096-block benchmark |

## Test verified

- make format, including workspace release clippy: PASS
- cargo test -p curvine-server --test master_fs_test -- --test-threads=1: PASS, 34 passed, 1 ignored
- cargo test -p curvine-client-core: PASS, 25 passed
- cargo test -p curvine-server --test lock_order_deadlock_stress_test -- --nocapture --test-threads=1: PASS, 4000 add-block and 4000 block-report operations completed
- 4096-block flush benchmark: PASS, 21.688 ms to 1.890 ms; response payload 184414 B to 0 B

## Dependencies

- Integrates with #1386 complete_with_attr changes.
- Ranged block-location reads remain outside this PR.

### DIFF
--- a/curvine-client-core/src/file/fs_client.rs
+++ b/curvine-client-core/src/file/fs_client.rs
@@ -35,6 +35,12 @@ pub struct FsClient {
     connector: Arc<ClusterConnector>,
 }
 
+struct CompleteFileOptions {
+    only_flush: bool,
+    set_attr_opts: Option<SetAttrOpts>,
+    return_file_blocks: bool,
+}
+
 impl FsClient {
     pub fn new(context: Arc<FsContext>) -> Self {
         let connector = context.connector.clone();
@@ -359,8 +365,18 @@ impl FsClient {
         only_flush: bool,
         set_attr_opts: Option<SetAttrOpts>,
     ) -> FsResult<Option<FileBlocks>> {
-        self.complete_file0(path, None, len, commit_blocks, only_flush, set_attr_opts)
-            .await
+        self.complete_file0(
+            path,
+            None,
+            len,
+            commit_blocks,
+            CompleteFileOptions {
+                only_flush,
+                set_attr_opts,
+                return_file_blocks: true,
+            },
+        )
+        .await
     }
 
     pub async fn complete_file_by_id(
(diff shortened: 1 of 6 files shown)


curvine-client-core/src/file/fs_client.rs
```diff
@@ -422,6 +436,7 @@ impl FsClient {
                     commit_blocks,
                     only_flush,
                     inode_id: None,
+                    return_file_blocks: None,
```
szbr00: Batch completion hardcodes snapshot skip on the server (`complete_file0(req, false)`), but this client still sends `return_file_blocks: None`, which means “legacy default = true” on the wire.

Please set `Some(false)` here so the request is self-describing. Otherwise a future change that starts honoring the per-request field (to match the single-file path) could silently reintroduce the full-snapshot hotspot for batch flushes.

curvine-server/tests/master_fs_test.rs
```diff
@@ -1805,6 +1851,95 @@ fn resize_rejects_extreme_file_size() {
     assert_eq!(status.len, 0);
 }
 
+#[test]
+fn only_flush_persists_block_without_returning_file_snapshot() -> CommonResult<()> {
```
szbr00, line 1884: This covers `MasterFilesystem::flush_file` and the legacy `complete_file(..., only_flush=true)` snapshot path well, but the production writer path goes through `MasterHandler` with `CompleteFileRequest.return_file_blocks = false`.

Consider adding a small handler/RPC test that sends the protobuf flag and asserts the response has no `file_blocks` while metadata still persists. That would lock the wire contract, not only the internal helper.

jlon: Thanks. Fixed in e0dae913: added handler_only_flush_without_snapshot_persists_block, which sends the protobuf flag through MasterHandler and verifies an empty FileBlocks response with persisted block metadata.

curvine-client-core/src/file/fs_client.rs
jlon: Thanks. Fixed in e0dae913: batch CompleteFileRequest now sets return_file_blocks to Some(false), keeping the wire contract explicit.

szbr00 (approved):

PR #1386, opened 2026-07-29
feat(client): add complete_with_attr to Writer trait for merged set_attr on file complete

bigbigxu:
# Summary

Add \complete_with_attr\ method to the Writer trait, allowing file attributes to be set atomically when completing a file write operation. This merges the \set_attr\ operation with \complete_file\, reducing two store writes and two journal entries to one each.

**Note:** This PR provides the API/RPC plumbing. FUSE attr-on-complete integration will be a follow-up PR.

# Issue Describe / Design

This optimization improves write performance by:
- Adding \complete_with_attr(opts: Option<SetAttrOpts>)\ to the Writer trait
- Passing \set_attr_opts\ through the client RPC to master
- Applying attributes directly to the inode during \complete_file\ on the server side
- Attribute order: first \ile.complete()\ sets length/mtime/ctime, then \set_attr(opts)\ overrides user-specified owner/group/mode/mtime/xattr
- TTL bucket index is refreshed in \pply_complete_file\ to handle TTL changes
- \set_attr_opts\ is ignored when \only_flush=true\ to keep flush semantics narrow

# Changes

| \crates/common/curvine-fs-api/src/writer.rs\ | Add \complete_with_attr\ to Writer trait with doc comments | New method with default impl calling \complete()\ |
| \curvine-client-core/src/file/fs_client.rs\ | Add \set_attr_opts\ param to \complete_file_by_id\ | None |
| \curvine-client-core/src/file/fs_writer*.rs\ | Implement \complete_with_attr\ in writer stack | None |
| \curvine-common/proto/master.proto\ | Add \set_attr_opts\ field to CompleteFileRequest | Protocol extension |
| \curvine-fuse/src/fs/fuse_writer.rs\ | Plumb \complete_with_attr\ through write task queue | API ready, FUSE integration follow-up |
| \curvine-server/src/master/meta/fs_dir.rs\ | Handle \set_attr_opts\ in complete_file, ignore on flush | Merged attr setting |
| \curvine-server/src/master/meta/store/inode_store.rs\ | Refresh TTL bucket in \pply_complete_file\ | TTL index consistency |
| \curvine-unified-fs/src/unified/macros.rs\ | Add \complete_with_attr\ to enum macro | None |

# Test verified

| \cargo test -p curvine-server master_fs_test\ | PASS | Includes new complete_with_attr test |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/common/curvine-fs-api/src/writer.rs
+++ b/crates/common/curvine-fs-api/src/writer.rs
@@ -104,12 +104,21 @@ pub trait Writer {
 
     fn complete(&mut self) -> impl Future<Output = FsResult<()>>;
 
+    /// Completes the write operation and optionally applies file attributes.
+    ///
+    /// # Default Implementation
+    ///
+    /// The default implementation **ignores** `opts` and simply calls [`complete()`](Writer::complete).
+    /// Backends that support atomic attribute setting on complete should override this method.
+    ///
+    /// Callers should not assume attributes were applied unless the backend explicitly
+    /// supports this method.
     fn complete_with_attr(
         &mut self,
         opts: Option<SetAttrOpts>,
     ) -> impl Future<Output = FsResult<()>> {
         async move {
-            let _ = &opts;
+            drop(opts);
             self.complete().await
         }
     }
--- a/crates/common/curvine-fs-api/src/writer.rs
+++ b/crates/common/curvine-fs-api/src/writer.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::fs::Path;
-use crate::state::{FileAllocOpts, FileStatus};
+use crate::state::{FileAllocOpts, FileStatus, SetAttrOpts};
 use crate::FsResult;
 use bytes::{BufMut, BytesMut};
 use orpc::err_box;
@@ -103,6 +103,16 @@ pub trait Writer {
     fn flush(&mut self) -> impl Future<Output = FsResult<()>>;
 
     fn complete(&mut self) -> impl Future<Output = FsResult<()>>;
+
+    fn complete_with_attr(
+        &mut self,
+        opts: Option<SetAttrOpts>,
+    ) -> impl Future<Output = FsResult<()>> {
+        async move {
+            let _ = &opts;
+            self.complete().await
+        }
+    }
 
     fn cancel(&mut self) -> impl Future<Output = FsResult<()>>;
 
--- a/curvine-fuse/src/fs/fuse_writer.rs
+++ b/curvine-fuse/src/fs/fuse_writer.rs
@@ -20,7 +20,7 @@ use curvine_client::unified::UnifiedWriter;
 use curvine_common::conf::FuseConf;
 use curvine_common::error::FsError;
 use curvine_common::fs::{Path, Writer};
-use curvine_common::state::{FileAllocOpts, FileStatus};
+use curvine_common::state::{FileAllocOpts, FileStatus, SetAttrOpts};
 use curvine_common::FsResult;
 use log::{error, warn};
 use orpc::common::LocalTime;
@@ -33,7 +33,11 @@ use std::sync::Arc;
 enum WriteTask {
     Write(i64, Bytes, Option<FuseResponse>),
     Flush(CallSender<FsResult<()>>, Option<FuseResponse>),
-    Complete(CallSender<FsResult<()>>, Option<FuseResponse>),
+    Complete(
+        CallSender<FsResult<()>>,
+        Option<FuseResponse>,
+        Option<SetAttrOpts>,
+    ),
     Resize(CallSender<FsResult<()>>, FileAllocOpts),
 }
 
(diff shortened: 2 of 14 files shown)


curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -593,11 +594,17 @@ impl FsDir {
...
         client_name: impl AsRef<str>,
         only_flush: bool,
+        set_attr_opts: Option<SetAttrOpts>,
     ) -> FsResult<bool> {
         let file = inode.as_file_mut()?;
+        let id = file.id();
         file.complete(len, &commit_block, client_name, only_flush)?;
 
-        self.evictor.on_access(file.id());
+        if let Some(opts) = set_attr_opts {
```
szbr00: Applying `set_attr_opts` here updates the in-memory inode and persists it via `apply_complete_file`, but unlike `FsDir::set_attr` → `apply_set_attr`, this path never refreshes `TtlBucketList`.

`SetAttrOpts` can introduce or change `ttl_ms` / `ttl_action`, and `file.complete()` also advances `mtime` (TTL expiration is `mtime + ttl_ms`). Without a TTL index update, the inode can stay in the wrong bucket (or never be indexed), so TTL cleanup runs incorrectly until a master restart rebuilds the index from RocksDB. Journal replay of `CompleteFileEntry` has the same gap because it also goes through `apply_complete_file`.

Please refresh the TTL index on this path — ideally inside `InodeStore::apply_complete_file` (so primary complete and journal replay stay consistent), matching what `apply_set_attr` already does. A test that sets TTL via `complete_file(..., Some(opts))` and asserts bucket membership would help lock this in.

szbr00: `set_attr_opts` is applied unconditionally, including when `only_flush` is true. The client currently always passes `None` on flush, but the RPC allows a flush request to mutate owner/mode/mtime/TTL/xattr as a side effect of a durability flush.

Consider rejecting `set_attr_opts.is_some() && only_flush` (or explicitly ignoring attrs on flush) so attribute changes are only accepted on real complete/close. That keeps flush semantics narrow and avoids surprising partial-write attribute updates.

crates/common/curvine-fs-api/src/writer.rs
```diff
@@ -104,6 +104,16 @@ pub trait Writer {
 
     fn complete(&mut self) -> impl Future<Output = FsResult<()>>;
 
+    fn complete_with_attr(
```
szbr00, line 116: The default `complete_with_attr` silently discards `opts` and calls `complete()`. That is convenient for backends that do not support attrs, but it is easy to miss when adding a new `Writer` impl or when routing through `UnifiedWriter` to a UFS backend — callers may believe attrs were applied atomically.

Please document this clearly on the trait method (opts may be ignored unless overridden), and consider logging a debug/warn when `opts.is_some()` in the default path so misuse is visible during development/testing.

curvine-fuse/src/fs/fuse_writer.rs
```diff
@@ -158,9 +162,17 @@ impl FuseWriter {
     }
 
     pub async fn complete(&self, reply: Option<FuseResponse>) -> FsResult<()> {
+        self.complete_with_attr(reply, None).await
+    }
+
+    pub async fn complete_with_attr(
```
szbr00, line 168: `FuseWriter::complete_with_attr` is plumbed through the write task queue, but every in-tree call site still passes `None` (`complete()` → `None`, cleanup → `None`, and `backend_handle` / `node_state` still call `complete`). So this PR does not yet deliver the FUSE “set attrs on close” behavior described in the PR summary.

If that merge is intentional for this change, please wire the FUSE release/setattr path to pass real `SetAttrOpts` here; otherwise it would help to clarify that this PR is API/RPC plumbing only and FUSE attr-on-complete is a follow-up.

curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -593,11 +594,17 @@ impl FsDir {
...
+            inode.set_attr(opts)?;
+        }
+
+        self.evictor.on_access(id);
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 51de7be7: TTL bucket index is now refreshed in pply_complete_file via self.ttl_bucket_list.add(file), matching what pply_set_attr does.

crates/common/curvine-fs-api/src/writer.rs
```diff
@@ -104,6 +104,16 @@ pub trait Writer {
...
+        &mut self,
+        opts: Option<SetAttrOpts>,
+    ) -> impl Future<Output = FsResult<()>> {
+        async move {
+            let _ = &opts;
+            self.complete().await
+        }
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 51de7be7: Changed to drop(opts) for clearer intent.

curvine-server/src/master/meta/fs_dir.rs
bigbigxu: Thanks for the detailed analysis. Fixed in 51de7be7: TTL bucket is now refreshed in InodeStore::apply_complete_file via self.ttl_bucket_list.add(file). This ensures both primary complete and journal replay stay consistent. The dd method is safe to call unconditionally - it returns early if TTL is not set.

bigbigxu: Thanks. Fixed in 51de7be7: set_attr_opts is now ignored when only_flush=true. Added comment explaining flush semantics should remain narrow (durability only).

crates/common/curvine-fs-api/src/writer.rs
bigbigxu: Thanks. Fixed in 51de7be7: Added doc comments explaining that the default implementation ignores opts and callers should not assume attributes were applied unless the backend explicitly supports this method. Also changed let _ = &opts to drop(opts) for clearer intent.

curvine-fuse/src/fs/fuse_writer.rs
bigbigxu: Thanks for pointing this out. Updated PR description to clarify: this PR provides the API/RPC plumbing only. FUSE attr-on-complete integration (wiring the release/setattr path to pass real SetAttrOpts) will be a follow-up PR.

szbr00 (approved):

PR #1387, opened 2026-07-29
perf(worker): move FileLayout I/O outside dataset lock (#1323)

### LINKED ISSUE #1323: [FEATURE] Remove FileLayout I/O from the worker BlockDataset global write lock
## Is your feature request related to a problem?

Worker `BlockStore::open_block` and `finalize_block` take the global
`BlockDataset` write lock. New FileLayout allocation performs directory/file
creation while holding the lock, and finalization reads file metadata under the
same lock. Local filesystem metadata latency therefore serializes allocation
and completion of all blocks on that worker.

## Describe the solution you'd like

Split BlockDataset metadata mutation from FileLayout filesystem work. Reserve
capacity and publish an explicit block state under synchronization, perform
slow filesystem allocation or metadata reads outside the global lock, then
finalize or roll back the reservation atomically.

The design must keep capacity accounting, duplicate allocation handling,
block visibility, and cleanup correct.

## Describe alternatives you've considered

- More worker RPC threads cannot remove the global dataset lock.
- Sharding the lock by block/directory may help, but only after reservation and
  state-transition invariants are defined.
- This is intentionally separate from master metadata locking because it
affects one worker, not all workers in the cluster.

## Additional context

The FUSE I/O audit identifies this as W8, a worker-level structural
bottleneck. It is most relevant to high-concurrency small-file workloads,
frequent block boundaries, and replica finalization.

# Goal

| --- | --- |
| Define reserved, allocated, finalized, and aborted block states | Preserve capacity and visibility invariants |
| Move FileLayout create/stat outside the dataset-wide lock | Remove filesystem latency from global serialization |
| Handle concurrent duplicate/open/finalize requests | Prevent leaks and inconsistent block state |
| Add lock and filesystem latency metrics | Prove contention and regressions |

# Not Goal

- Changing replica count or placement policy.
- Replacing FileLayout with SPDK.
- Relaxing completed-block durability or visibility semantics.

# Test Plan

| --- | --- | --- |
| Concurrent new-block allocations | Verify independent blocks no longer serialize on filesystem create | worker |
| Concurrent finalization with delayed metadata calls | Verify lock scope and correct final state | worker |
| Capacity exhaustion and allocation failure | Verify reservations are released | worker |
| Duplicate/retry requests | Verify idempotence and no leaked files | worker, RPC |
| Restart during reserved or allocated state | Verify cleanup/recovery | worker |
| Before/after metrics | Report lock wait/hold and FileLayout latency | worker |

jlon:
## Summary

Move FileLayout create, copy, and stat work outside the worker BlockDataset write lock. Keep only the atomic rename and metadata publication under the existing global lock.

## Issue / Design

Closes #1323.

A reservation publishes an internal Allocating or Finalizing state under the dataset lock. File preparation runs after that lock is released. Finalize first performs stat and path preparation outside the lock, then takes the short write lock for rename, capacity accounting, and metadata publication as one operation.

Per-block stripe locks serialize same-block mutations without replacing the existing global dataset lock. ReadHandler and replication now open the current generation while holding a dataset read lock, so rename cannot replace the selected path before its file descriptor opens. During Finalizing, short-circuit reads fall back to the server read path. Finalizing blocks remain visible to full reports as Writing; only Allocating blocks are hidden.

## Changes

| Area | Change |
| --- | --- |
| BlockStore | Two-phase file open/finalize flow, per-block stripe serialization, atomic current-generation reader open |
| VfsDataset | Reservation states, finalize publish with non-failing post-rename metadata accounting, stable report/read visibility |
| FileLayout | Split finalize preparation from atomic rename publish |
| Read and replication | Open the selected generation under the dataset read lock |
| Metrics | Record stripe wait, dataset lock wait/hold, and external FileLayout time |

## Test verified

- make format, including workspace release clippy: PASS
- cargo test -p curvine-server --lib -- --test-threads=1: PASS, 136 passed, 2 ignored
- BlockStore rollback, capacity, finalizing visibility, and reader-generation tests: PASS, 8 passed
- Manual real-file benchmark, four concurrent 32 MiB rewrites: old median 45.27 ms; final median 17.47 ms; about 2.6x faster

## Dependencies

None.

### DIFF
--- a/curvine-server/src/worker/block/block_store.rs
+++ b/curvine-server/src/worker/block/block_store.rs
@@ -12,18 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::worker::block::BlockMeta;
+use crate::worker::block::{BlockMeta, BlockState};
 use crate::worker::storage::{
     BlockDataset, BlockLayout, BlockReadContext, BlockWriteContext, Dataset,
 };
+use crate::worker::Worker;
 use curvine_common::conf::ClusterConf;
 use curvine_common::state::{ExtendedBlock, StorageInfo};
 use orpc::CommonResult;
+use parking_lot::{Mutex, MutexGuard};
 use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
+use std::time::{Duration, Instant};
+
+const BLOCK_LOCK_STRIPES: usize = 256;
 
 #[derive(Clone)]
 pub struct BlockStore {
     state: Arc<RwLock<BlockDataset>>,
+    block_locks: Arc<Vec<Mutex<()>>>,
 }
 
 impl BlockStore {
@@ -31,6 +37,7 @@ impl BlockStore {
         let dataset = BlockDataset::from_conf(cluster_id, conf)?;
         let block_store = BlockStore {
             state: Arc::new(RwLock::new(dataset)),
+            block_locks: Arc::new((0..BLOCK_LOCK_STRIPES).map(|_| Mutex::new(())).collect()),
         };
 
         Ok(block_store)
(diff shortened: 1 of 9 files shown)


curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -207,19 +234,223 @@ impl VfsDataset {
...
+                    .or_else(|| (!meta.is_transitioning()).then_some(meta))
             })
             .collect()
     }
 
     pub(crate) fn get_readable_block(&self, id: i64) -> Option<&BlockMeta> {
-        self.committed_rewrites
-            .get(&id)
-            .or_else(|| self.meta.get(id))
+        if let Some(committed) = self.committed_rewrites.get(&id) {
```
szbr00, line 248: During rewrite finalize, `committed_rewrites` keeps advertising the old Finalized meta while `FileLayout::finalize` renames staging over the same active path outside the dataset lock. `get_block` / `ReadHandler` do not take the per-block stripe lock, so a reader can interleave after rename and before `complete_file_finalize` and open the active path with the old `len` / accounting.

Previously the dataset write lock covered the rename, so readers could not observe that window. Please close it—for example by dropping or switching the readable generation under the dataset lock before the rename becomes visible, or by ensuring readers of this block cannot race the rename→commit gap (without blocking committed reads for the whole rewrite copy).

curvine-server/src/worker/storage/vfs_dataset.rs
```diff
@@ -207,19 +234,223 @@ impl VfsDataset {
         self.meta
             .all_blocks()
             .into_iter()
-            .map(|meta| {
+            .filter_map(|meta| {
                 self.committed_rewrites
                     .get(&meta.id())
                     .cloned()
-                    .unwrap_or(meta)
+                    .or_else(|| (!meta.is_transitioning()).then_some(meta))
```
szbr00: `all_blocks` treats `Finalizing` like `Allocating` and omits it when there is no `committed_rewrites` entry. For a first-time write that means a block that was `Writing` disappears from the snapshot used by `BlockActor::full_block_report`.

Master full-report reconcile treats the accumulated worker set as authoritative; a transient omission can look like the worker lost the block and trigger delete/under-replication. Hiding `Allocating` for new blocks matches the PR intent; `Finalizing` should still remain reportable/readable as the prior Writing generation (or an equivalent stable view) until commit publishes Finalized.

curvine-server/src/worker/block/block_store.rs
```diff
@@ -56,16 +63,134 @@ impl BlockStore {
...
+
+        let Some(reservation) = reservation else {
+            return self.with_dataset_write("finalize", |state| state.finalize_block(block));
+        };
+
+        let started = Instant::now();
+        let finalized = reservation.finalize(block.len);
+        self.observe_file_layout_operation("finalize", started.elapsed());
+        match finalized {
+            Ok(meta) => self.with_dataset_write("finalize", |state| {
```
szbr00: On the Ok path, `reservation.finalize` may already have renamed staging→active before `complete_file_finalize` runs. If complete fails (state check, `release_space`, or `update_final_space`), this returns the error without rollback, leaving `Finalizing` meta, a possible partial `committed_rewrites`/capacity update, and an active file that no longer matches Writing.

`rollback_file_finalize` only restores Writing metadata and cannot undo the rename. Please make commit failure safe: treat post-rename failure as a compensated path (force-publish Finalized / repair accounting), or restructure so rename and metadata commit cannot diverge.

jlon: Thanks. Fixed in ea6cef10: finalization is split into a pre-rename plan and a publish step. After rename, capacity and metadata publication use the already-validated reservation without fallible directory lookups, so rollback is only used when preparation or rename fails before publication.

curvine-server/src/worker/storage/vfs_dataset.rs
jlon: Thanks. Fixed in ea6cef10: finalize now performs stat and path preparation outside the lock, then keeps only rename and metadata publication in the short dataset write lock. ReadHandler and replication open the selected generation while holding the dataset read lock; finalizing short-circuit reads fall back to that path.

curvine-server/src/worker/storage/vfs_dataset.rs
jlon: Thanks. Fixed in ea6cef10: Finalizing is now retained in all_blocks and get_readable_block. It reports as Writing through the existing master mapping, while only Allocating remains hidden. Added finalizing_new_block_remains_reportable_and_readable coverage.

lzjqsdd (approved):

PR #1389, opened 2026-07-29
fix(fuse): readdir seekable offsets

ljy1-lixing:
**Summary**

Make FUSE directory offsets seekable after forward iteration, EOF, and backward `lseek()` while keeping per-directory-handle memory bounded.

Curvine previously backed each open directory with a forward-only `ListStream`. Sequential readdir calls worked, but entries that had already been consumed could no longer be recovered. Seeking to an earlier `d_off`, especially after reaching EOF, could therefore return the wrong entry or an empty result.

The fix adds a bounded sliding window to each directory handle. Recent entries are served directly from the window, while offsets older than the retained window rebuild and replay the backend stream. The window stores `Arc<FileStatus>` values so constructing a response batch only clones shared references instead of deeply cloning paths, names, and xattrs.

**Background**

Linux directory offsets are opaque resume cookies. Applications can save a position using `telldir()` or the `d_off` returned by `getdents64()`, then restore it with `seekdir()` or `lseek()` on the directory file descriptor.

This behavior is used by directory scanners, backup and synchronization tools, filesystem gateways, indexers, and applications that paginate or resume large directory walks. A directory implementation must therefore support more than strictly forward iteration: a previously returned cookie can be supplied again while the directory handle remains open.

**Root Cause**

The old `DirHandle` tracked:

```text
stream  forward-only backend ListStream
buf     entries fetched but not yet emitted
index   number of entries consumed from the stream
```

`get_batch()` moved the current buffer out of the handle, filled it by consuming the stream, and returned it to the caller. If the FUSE response buffer filled, the un-emitted suffix was written back with `set_buf()`.

This retained only the current response remainder. Entries successfully emitted in earlier calls were permanently discarded from the forward stream. For example:

```text
readdir(offset=0) -> a, b, c
readdir(offset=3) -> d, e, f
lseek(offset=1)
readdir(offset=1) -> expected b, c, d
```

At the final request, `b` and `c` were no longer available. Once the stream had reached EOF, the handle could not recover any historical position at all.

An initial implementation retained every consumed `FileStatus`, which restored seek behavior but made memory grow with the total number of entries visited by every open directory handle. It also deeply cloned up to `list_limit` statuses for each response batch.

**Changes**

- Replace the one-shot response buffer with a sliding history window containing `Arc<FileStatus>` entries.
- Track `base_off`, `next_off`, and EOF state so offsets can be translated into positions inside the retained window.
- Retain at most four readdir batches per directory handle:

  ```text
  cache_limit = list_limit * 4
  ```

- Serve forward reads and recent backward seeks directly from the retained window.
- Recreate and replay the backend `ListStream` when a requested cookie is older than the window.
- Remove `set_buf()`: entries not emitted because the kernel response buffer is full remain available in the sliding window and are selected by the next request's offset.
- Clone `Arc` references when building a batch instead of deeply cloning every `FileStatus`.
- Add focused regression tests for forward batching, EOF, window-local backward seeks, replay outside the window, rewind to zero, nonzero initial offsets, and the cache-size bound.

**Memory and Performance**

The retained history is bounded per open directory handle. With the default `list_limit=1000`, a handle keeps at most approximately 4000 directory statuses, regardless of the total directory size or how long the handle remains open. The cache is released when the kernel sends `RELEASEDIR` and Curvine drops the corresponding `DirHandle`.

Using four batches instead of one preserves locality for seek-heavy directory workloads. A one-batch prototype remained correct but caused the random-seek regression to increase from approximately 47 seconds to 141 seconds because most historical cookies required a complete replay. Four batches completed the same workload in approximately 42 seconds while retaining a strict entry-count limit.

Offsets older than the window still require replay from the start of the directory. Repeated long-distance random seeks in directories substantially larger than the window can therefore generate additional Master RPCs and CPU work. This is the intentional tradeoff for bounded handle memory without requiring a new persistent directory-cookie protocol.

**Concurrent Directory Changes**

Replaying an old offset creates a new view of the current directory. If entries are created, removed, or renamed between the original read and the replay, the ordering associated with a numerical cookie can change.

This change does not introduce directory snapshots or versioned cookies. It restores seek behavior for a stable directory and retains the usual Linux readdir limitations when a directory is modified concurrently.

**Verification**

- Focused `DirHandle` seek and bounded-cache tests pass: `2 passed`.
- FUSE readdir termination and split-response tests pass: `6 passed`.
- Full Curvine FUSE library test suite passes: `317 passed`.
- Directory offset, reverse-seek, and random-seek regression cases pass.
- The random-seek case completes in approximately 42 seconds with the bounded four-batch window.
- `make format` passes.
- Clippy passes with warnings denied.
- `git diff --check` passes.

**Deployment**

The change is entirely inside the FUSE client. Master and Worker protocols and persistent metadata formats are unchanged.

Deploy the updated `curvine-fuse` binary and restart or remount each FUSE mount that should use seekable directory offsets. Master and Worker processes do not need to be restarted solely for this change.

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -376,13 +376,17 @@ impl CurvineFileSystem {
         plus: bool,
     ) -> FuseResult<(FuseDirentList, u64)> {
         let handle = self.state.find_dir_handle(header.nodeid, arg.fh)?;
+        let path = Path::from_str(&handle.path)?;
 
         let mut res = FuseDirentList::new(arg);
         let mut index = arg.offset;
-        let mut batch = handle.get_batch(arg.offset as usize).await?;
+        let mut batch = handle
+            .get_batch(arg.offset as usize, || self.state.list_stream(&path))
+            .await?;
         {
             let mut dir = self.state.dir_write();
             while let Some(status) = batch.pop_front() {
+                let status = status.as_ref();
                 let attr = if status.name != FUSE_CURRENT_DIR && status.name != FUSE_PARENT_DIR {
                     // READDIRPLUS takes a kernel lookup ref (kernel caches the
                     // dentry and will send a FORGET); plain READDIR must not
@@ -394,22 +398,19 @@ impl CurvineFileSystem {
                     self.state.record_status_put();
                     attr
                 } else {
-                    FuseUtils::status_to_attr(&self.conf, &status)?
+                    FuseUtils::status_to_attr(&self.conf, status)?
                 };
 
                 let entry = FuseUtils::create_entry_out(&self.conf, attr);
                 // dirent `off` is the resume cookie = position of the NEXT entry.
                 // See `readdir_next_cookie` (the infinite-loop guard).
                 let next_off = Self::readdir_next_cookie(index);
-                if !res.add_dirent(plus, next_off, &status, entry) {
-                    batch.push_front(status);
+                if !res.add_dirent(plus, next_off, status, entry) {
                     break;
                 }
                 index += 1;
             }
         }
-        handle.set_buf(batch).await?;
-
         let entries = index.saturating_sub(arg.offset);
         Ok((res, entries))
     }
@@ -2832,7 +2833,12 @@ mod tests {
                         1000,
                         ListStream::from_vec(entries(names)),
                     );
-                    let batch = handle.get_batch(offset as usize).await.unwrap();
+                    let batch = handle
+                        .get_batch(offset as usize, || async {
+                            Ok(ListStream::from_vec(entries(names)))
+                        })
+                        .await
+                        .unwrap();
                     if batch.is_empty() {
                         return Ok(seen); // kernel sees 0 entries => readdir done
                     }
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -376,13 +376,17 @@ impl CurvineFileSystem {
         plus: bool,
     ) -> FuseResult<(FuseDirentList, u64)> {
         let handle = self.state.find_dir_handle(header.nodeid, arg.fh)?;
+        let path = Path::from_str(&handle.path)?;
```
szbr00: The replay path is permanently taken from `DirHandle.path`, but `NodeState::fs_rename` updates only the dcache and does not update existing directory handles. If this directory or one of its ancestors is renamed, seeking behind `base_off` rebuilds the stream from the old path and can return `ENOENT` or list a replacement directory. An open directory fd should remain usable across rename. Please resolve the current path from `header.nodeid` when rebuilding (or recreate the stream by inode), and add a rename-then-rewind regression test.

curvine-fuse/src/fs/state/dir_handle.rs
```diff
@@ -68,44 +91,201 @@ impl DirHandle {
...
-                match guard.stream.next().await {
-                    Some(Ok(_)) => guard.index += 1,
-                    Some(Err(e)) => return Err(e),
-                    None => return Ok(VecDeque::new()),
-                }
-            }
+        // Rename keeps an open directory fd valid, so a stream bound to an old
+        // path must be rebuilt even when the requested cookie is still cached.
+        if guard.stream_path != current_path || off < guard.base_off {
+            *guard = InnerStream::new(new_stream().await?, current_path.to_string());
```
szbr00, line 114: **Major:** `stream_path` is updated after a rename, but it is skipped by serialization while `DirHandle.path` remains the original open-time path. `NodeState::restore` reconstructs the stream from `handle.path` before any readdir can resolve the inode's current path, so a rename followed by persist/restore can fail with `ENOENT` (or reopen a replacement directory at the old path). Please persist/update the current path, or resolve it from the restored inode before reopening the stream, and add a rename → persist/restore regression test.

szbr00 (commented):
Re-reviewed the latest revision. The original runtime rename issue is fixed, but one persistence/restore gap remains.

szbr00 (approved):

Issue #1091, opened 2026-07-13
[FEATURE] fuse: tighten request length validation, clarify is_meta, and parse-failure replies (by LiAuto)


yangcx000:
**Is your feature request related to a problem?**

Four lower-priority correctness/clarity items in request parsing and classification, grouped:

**(a) Request length validation accepts trailing bytes for non-write ops**
- `parse_header` checks `self.buf.len() < header.len` but not equality (`fuse_request.rs:48` / `:54`).
- `parse_operator` does not verify the decoder consumed exactly `header.len` (`fuse_request.rs:111`).
- Only `get_write_bytes` is strict (`fuse_request.rs:93`).
- Non-write ops with trailing bytes are silently accepted. FUSE device reads normally return one request, so low-risk, but stricter validation catches malformed input earlier.
- Fix: require `buf.len() == header.len` in `from_bytes` (or slice to `header.len` and reject/record trailing bytes); add tests for trailing bytes on representative non-write ops.

**(b) `is_meta` classification is misleading**
- `is_stream` is true for READ/WRITE/FLUSH/RELEASE/FSYNC (`fuse_request.rs:81`); `is_meta` is false only for READ/WRITE (`fuse_request.rs:89`). So FLUSH/RELEASE/FSYNC are BOTH `is_stream()==true` and `is_meta()==true`.
- Dispatch checks stream first so routing is correct, but audit logging uses `is_meta` (`fuse_receiver.rs:383`), so flush/release/fsync are audited as metadata despite being stream lifecycle ops. Future code could wrongly assume the two are mutually exclusive.
- Fix: rename `is_meta` → `should_audit_as_meta`, or define `is_meta = !is_stream()` and adjust audit separately; add a test documenting the intended classification.

**(c) Metadata parse failures are not surfaced as kernel replies**
- If `parse_operator` fails after a `FuseResponse` exists, `finish_early` records metrics and sends no reply (`fuse_receiver.rs:546`, `fuse_response.rs:428`).
- Correct for structurally invalid input the daemon cannot safely reply to; but if header/unique are valid and only the body is malformed, no reply can leave the kernel request hanging.
- Fix: decide per parse-failure class whether an error reply is safe; for requests with valid header/unique and a known opcode, consider replying `EINVAL` instead of no-reply; keep no-reply only where replying would violate the protocol. (Needs careful protocol judgment, not a mechanical change.)

[...]

PR #1405, opened 2026-07-29
fix(fuse): tighten request parsing and audit classification (#1091)

### LINKED ISSUE #1392: [BUG]: Bound FUSE batch-forget count before allocation [LiAuto]
**Describe the bug**

`FuseDecoder::get_struct_vec` calls `Vec::with_capacity(count)` before proving that the bounded request body contains `count` elements. For `FUSE_BATCH_FORGET`, `count` comes from an untrusted `u32` request field. A malformed frame can therefore request a multi-gigabyte allocation and abort the process before normal decoder bounds checks run.

**To Reproduce**

1. Construct a valid FUSE header and `fuse_batch_forget_in` body.
2. Set `count` to `u32::MAX` while providing only a short request body.
3. Parse the request through `FuseRequest::parse_operator`.
4. Observe allocation is attempted before the remaining body is validated.

**Expected behavior**

The decoder must calculate the maximum element count from the bounded remaining frame and reject an excessive count before allocation.

- Linux
- Component: `curvine-fuse/src/session/fuse_decoder.rs`

**Additional context**

Scope:

- Calculate `max_count = remaining_bytes / size_of::<T>()` before allocation.
- Reject `count > max_count` with a decode error.
- Allocate only after validation.
- Add tests for zero, exact, one-too-many, and `u32::MAX` counts.

Related: #1091 owns the separate FUSE request framing-boundary validation.

yangcx000:
# Summary

Tighten FUSE request parsing by enforcing declared frame boundaries, rejecting trailing request data, and validating batch-forget counts before allocation. Parse the Linux `fuse_getattr_in` argument, use the decoder cursor as the single source of truth for WRITE payloads, and clarify audit and interruptible-wait classifications.

# Issue Describe / Design

Related to #1091  
Closes #1392

Root causes:

- Request parsing accepted declared lengths smaller than the FUSE header and buffers whose actual length differed from `header.len`.
- Supported operators did not verify that the complete request body was consumed.
- `FUSE_BATCH_FORGET` allocated from an untrusted count before validating the remaining body.
- GETATTR ignored the required Linux `fuse_getattr_in` request argument.
- WRITE payload extraction recomputed a fixed offset independently from the decoder cursor.
- `is_meta` and `is_interrupt` did not describe their audit-only and interruptible-wait meanings.

Fix approach:

- Require a valid exact request-frame length.
- Reject trailing body data for supported operators.
- Validate structure counts before allocation.
- Add and parse the 16-byte Linux `fuse_getattr_in` structure.
- Slice WRITE data from the decoder-consumed range.
- Rename the request classifications without changing dispatch or audit behavior.

# Changes

| `curvine-fuse/src/session/fuse_request.rs` | Validate frames, require full body consumption, parse GETATTR arguments, and unify WRITE cursor use | Malformed requests are rejected earlier; valid GETATTR requests are parsed with their kernel argument |
| `curvine-fuse/src/session/fuse_decoder.rs` | Bound structure-array counts and add `ensure_empty` | Prevents excessive allocation and trailing data |
| `curvine-fuse/src/raw/fuse_abi.rs` | Add Linux `fuse_getattr_in` ABI structure | Aligns GETATTR parsing with Linux UAPI |
| `curvine-fuse/src/fs/operator.rs` | Carry the GETATTR argument in the operator | Makes GETATTR flags and optional fh available to handlers |
| `curvine-fuse/src/session/channel/fuse_receiver.rs` | Use explicit audit and interruptible-wait naming; update GETATTR test request | No dispatch or audit behavior change |

# Test verified

| `make format` | PASS | Full repository pre-commit checks |
| `cargo test -p curvine-fuse session::fuse_request --no-fail-fast` | PASS | 9 request tests |
| `cargo test -p curvine-fuse session::fuse_decoder --no-fail-fast` | PASS | 2 decoder tests |
| GETATTR receiver dispatch integration test | PASS | Uses Linux `fuse_getattr_in` payload |
| `cargo test -p curvine-fuse --lib -- --test-threads=1` | PASS | 339 tests |

A parallel full test run before the follow-up commit had one existing global-metrics assertion observe cross-test counter increments; that test passed when rerun alone. The complete library suite passes with one test thread.

# Dependencies

- Related PRs: None
- Related issues: #1091, #1392
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/fuse_request.rs
+++ b/curvine-fuse/src/session/fuse_request.rs
@@ -53,8 +53,20 @@ impl FuseRequest {
         }
 
         let header: &fuse_in_header = FuseDecoder::parse(&self.buf[..FUSE_IN_HEADER_LEN])?;
-        if self.buf.len() < header.len as usize {
-            return err_box!("Not enough data for arguments (short read).");
+        let declared_len = header.len as usize;
+        if declared_len < FUSE_IN_HEADER_LEN {
+            return err_box!(
+                "Invalid FUSE request length {}, expected at least {}",
+                declared_len,
+                FUSE_IN_HEADER_LEN
+            );
+        }
+        if self.buf.len() != declared_len {
+            return err_box!(
+                "FUSE request length mismatch, declared {}, actual {}",
+                declared_len,
+                self.buf.len()
+            );
         }
 
         Ok(header)
@@ -87,7 +99,7 @@ impl FuseRequest {
         )
     }
 
-    pub fn is_meta(&self) -> bool {
+    pub fn should_audit(&self) -> bool {
         !matches!(self.opcode, FUSE_READ | FUSE_WRITE)
     }
 
@@ -219,12 +231,10 @@ impl FuseRequest {
             }),
 
             FUSE_WRITE => {
-                let arg = decoder.get_struct()?;
-                FuseOperator::Write(Write {
-                    header,
-                    arg,
-                    data: self.get_write_bytes(arg.size as usize)?,
-                })
+                let arg: &fuse_write_in = decoder.get_struct()?;
+                let data = self.get_write_bytes(arg.size as usize)?;
+                let _ = decoder.get_bytes(arg.size as usize)?;
+                FuseOperator::Write(Write { header, arg, data })
             }
 
             FUSE_MKNOD => FuseOperator::MkNod(MkNod {
@@ -333,6 +343,10 @@ impl FuseRequest {
             _ => FuseOperator::Notimplemented,
         };
 
+        if !matches!(&op, FuseOperator::Notimplemented) {
+            decoder.ensure_empty()?;
+        }
+
         Ok(op)
     }
 }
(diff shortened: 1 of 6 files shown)


curvine-fuse/src/session/fuse_request.rs
```diff
@@ -333,6 +335,10 @@ impl FuseRequest {
             _ => FuseOperator::Notimplemented,
         };
 
+        if !matches!(&op, FuseOperator::Notimplemented) {
+            decoder.ensure_empty()?;
```
szbr00, line 359: This blanket check rejects valid `FUSE_INIT` requests from Linux kernels using ABI 7.36+: the kernel sends the 64-byte extended `fuse_init_in` with `FUSE_INIT_EXT`, while Curvine consumes only its 16-byte local structure, leaving 48 bytes here. The INIT request therefore fails before Curvine can negotiate down to ABI 7.31 (and the current parse-error path sends no reply). Please handle the extended INIT input explicitly—or permit its defined compatibility tail—and add a test using a 7.36+ INIT frame.

yangcx000: Thanks. Fixed in a913a9b: the parser now recognizes FUSE_INIT_EXT, consumes the ABI 7.36+ 48-byte compatibility tail before the strict trailing-data check, and includes coverage for both legacy 16-byte and extended 64-byte INIT frames.

szbr00 (approved):

PR #1413, opened 2026-07-29
fix(worker): materialize sparse block length after resize (#1390)

### LINKED ISSUE #1390: [BUG]: Worker rejects post-resize rewrite with block length mismatch

Writing into a Curvine file just after a block-size boundary following ResizeFile causes the worker to reject block completion. The client sees an IO error when the writer channel drops. This is a data-path defect in block length accounting after resize, not a client API misuse.

After creating a file, writing near and across the 1024-byte block boundary, resizing, and writing again into an earlier block, the worker returns Block length mismatch (expected 1024, actual 200) and the client write fails with sender dropped.

Reproduction:

1. Check out Curvine commit aeb134905a1ca3917be533fc5f49f3f7ce64d9f4.
2. Run the curvine-tests block_test case resize_file_read_write (cargo nextest run -p curvine-tests resize_file_read_write).
3. Observe the worker Block length mismatch and client sender-dropped failure after the post-resize rewrite near the block boundary.

**Impact**

Applications that resize files and rewrite near block boundaries can lose writes or fail IO. Integration and daily unit coverage for block resize currently fail on this path.

Valid rewrite and append sequences after ResizeFile complete successfully with consistent block lengths on the worker and readable file contents.

In worker FileLayout / complete-block handling, reconcile expected block length with partial rewrite and post-resize append paths so a valid client rewrite after ResizeFile does not trip length mismatch; add a regression covering write-just-after-block-boundary after resize.

**Additional context**

Observed on Curvine commit aeb134905a1ca3917be533fc5f49f3f7ce64d9f4 in both integration and daily profiles with cleanup passed. Same short fingerprint has appeared in earlier daily runs, so treat as pre-existing. Root-cause statement is a hypothesis grounded in the worker error site.

Root-cause hypothesis: Root-cause hypothesis: after ResizeFile across a block boundary, a subsequent write into an existing remote block is completed with a shorter length than the worker FileLayout expects, so the worker rejects the commit with Block length mismatch and the client writer drops.

Evidence summary: curvine-tests::block_test::resize_file_read_write fails in both integration and daily on the same signature. Log shows writes at 0/924/1024/2048 succeed, then a write into blk_16793993217 fails with Block 16793993217 length mismatch, expected: 1024, actual: 200 from file_layout.rs:150, followed by sender dropped.

Related issues: https://github.com/CurvineIO/curvine/issues/591

Acceptance / test gaps:

- resize_file_read_write passes on the fixed commit
- Worker does not report Block length mismatch for the documented resize-then-rewrite sequence
- No unrelated public API behavior changes

LJF-050:
# Summary

Fix worker block length mismatch after sparse seek-past-EOF resize: finalize extends short staging files to the committed logical length, and reads synthesize sparse zeros when the master logical length exceeds physical worker bytes.

# Issue Describe / Design

Closes #1390

## Root cause

1. Sparse seek-past-EOF triggers `ResizeFile`. Master `InodeFile::extend` raises a committed last-block logical length (for example from 100 to 1024) without rewriting worker data and without re-attaching `alloc_opts` (intentional, to avoid rewrite amplification).
2. A later rewrite into that block opens with `ExtendedBlock.len` equal to the inflated logical length, copies the short committed staging file, writes a partial range, then `FileLayout::finalize` compared filesystem size to `committed_len` and failed with `Block length mismatch`.
3. Even after a successful write path, a later resize can inflate another block's master length without a rewrite. Reads then open with the larger logical length against a shorter physical file and fail past EOF.

## Fix approach

- In `FileLayout::finalize`, when staging size is below `committed_len`, `set_len(committed_len)` before the mismatch check so partial post-resize rewrites can commit.
- In worker read open/read, accept the client-visible logical length and return zeros past the physical file length; disable short-circuit when logical length exceeds physical so local readers cannot bypass that path.
- Do not rematerialize every grown committed block on the client during resize (that approach fixed the assigned case but introduced FUSE fio regressions).

# Changes

| `curvine-server/.../file_layout.rs` | Extend short staging to `committed_len` on finalize | Post-resize partial rewrites can finalize |
| `curvine-server/.../block_io_context.rs` | Sparse zero synthesis for logical > physical reads | Reads of inflated logical lengths succeed with holes |
| `curvine-server/.../read_handler.rs` | Open with logical length; force remote when sparse | Short-circuit skipped only for sparse logical tails |
| `curvine-server/.../vfs_dataset.rs` | Unit test for sparse finalize materialization | Guards the rewrite finalize path |

# Test verified

| `make format` | PASS | On exact HEAD `8e5f0d62` |
| `cargo test -p curvine-server --lib finalize_rewrite_materializes_sparse_committed_length` | PASS | Worker unit coverage |
| `cargo test -p curvine-server --lib read_context_sparse_logical_tail_returns_zeros` | PASS | Sparse read unit coverage |
| `cargo test -p curvine-tests --test block_test resize_file_read_write` | PASS | Assigned case on fixed HEAD |
| Profile before (`aeb13490`) / after (`8e5f0d62`) for fast, integration, daily, fuse, ltp, csi | PASS gate | Assigned `resize_file_read_write` absent after; `new_failures=[]`; cleanup passed; image digest matched |
| Independent review | APPROVED | Bound to exact HEAD `8e5f0d62` |

# Dependencies

- Related PRs: None
- Related issues: #1390
- Blocks / blocked by: None

### DIFF
--- a/curvine-server/src/worker/handler/read_handler.rs
+++ b/curvine-server/src/worker/handler/read_handler.rs
@@ -63,19 +63,22 @@ impl ReadHandler {
                 .ctx(format!("worker block store lookup failed: {}", e))
         })?;
 
+        // Master sparse ResizeFile may raise logical block len above the
+        // worker's physical file size. Serve the logical length and synthesize
+        // zeros for the sparse tail instead of rejecting the open.
+        let logical_len = context.len.max(meta.len);
         if context.off < 0 {
             return err_box!(
                 "Invalid read offset: {}, block length: {}",
                 context.off,
-                meta.len
+                logical_len
             );
         }
-        if context.off > meta.len {
+        if context.off > logical_len {
             return err_box!(
-                "The length of the requested data exceeds the maximum length of the block file, \
-            request off {}, file len {}",
+                "The length of the requested data exceeds the maximum length of the block file,             request off {}, file len {}",
                 context.off,
-                meta.len
+                logical_len
             );
         }
 
@@ -85,14 +88,15 @@ impl ReadHandler {
 
         if context.enable_read_ahead && context.read_ahead_len > Self::MAX_READ_AHEAD {
             return err_box!(
-                "The pre-read size exceeds the maximum value allowed by the system.\
-                 The current value is {}. The maximum allowed value is: {}",
+                "The pre-read size exceeds the maximum value allowed by the system.                 The current value is {}. The maximum allowed value is: {}",
                 context.read_ahead_len,
                 Self::MAX_READ_AHEAD
             );
         }
 
-        let sc_path = if context.short_circuit {
+        // Short-circuit local reads cannot synthesize sparse tails; force the
+        // remote path when logical length exceeds physical worker bytes.
+        let sc_path = if context.short_circuit && logical_len <= meta.len {
             self.store.short_circuit(&meta)?
         } else {
             None
@@ -101,7 +105,7 @@ impl ReadHandler {
         let (is_short_circuit, path, file) = if let Some(path) = sc_path {
             (true, path, None)
         } else {
-            let file = self.store.open_reader(&meta, context.off)?;
+            let file = self.store.open_reader(&meta, context.off, logical_len)?;
             let path = file.path().to_string();
             (false, path, Some(file))
         };
@@ -127,7 +131,7 @@ impl ReadHandler {
 
         let response = BlockReadResponse {
             id: context.block_id,
-            len: meta.len,
+            len: logical_len,
             path: ternary!(is_short_circuit, Some(path), None),
             storage_type: meta.storage_type().into(),
         };
--- a/curvine-server/src/worker/handler/read_handler.rs
+++ b/curvine-server/src/worker/handler/read_handler.rs
@@ -58,15 +58,29 @@ impl ReadHandler {
 
     pub fn open(&mut self, msg: &Message) -> FsResult<Message> {
         let context = ReadContext::from_req(msg)?;
+        let conf = Worker::get_conf()?;
+        let max_block_size = conf.master.max_block_size;
         let meta = self.store.get_block(context.block_id).map_err(|e| {
             FsError::block_not_found(context.block_id)
                 .ctx(format!("worker block store lookup failed: {}", e))
         })?;
 
         // Master sparse ResizeFile may raise logical block len above the
-        // worker's physical file size. Serve the logical length and synthesize
-        // zeros for the sparse tail instead of rejecting the open.
-        let logical_len = context.len.max(meta.len);
+        // worker's physical file size. Only synthesize a sparse tail when the
+        // client advertises a validated logical length above physical bytes.
+        let physical_len = meta.len;
+        let logical_len = if context.len > physical_len {
+            if context.len > max_block_size {
+                return err_box!(
+                    "Advertised block length {} exceeds max_block_size {}",
+                    context.len,
+                    max_block_size
+                );
+            }
+            context.len
+        } else {
+            physical_len
+        };
         if context.off < 0 {
             return err_box!(
                 "Invalid read offset: {}, block length: {}",
@@ -84,6 +98,14 @@ impl ReadHandler {
 
         if context.chunk_size <= 0 {
             return err_box!("chunk_size must be greater than 0");
+        }
+
+        if context.chunk_size as i64 > Self::MAX_READ_AHEAD {
+            return err_box!(
+                "chunk_size {} exceeds maximum allowed value {}",
+                context.chunk_size,
+                Self::MAX_READ_AHEAD
+            );
         }
 
         if context.enable_read_ahead && context.read_ahead_len > Self::MAX_READ_AHEAD {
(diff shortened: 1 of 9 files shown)


curvine-server/src/worker/handler/read_handler.rs
```diff
@@ -63,19 +63,22 @@ impl ReadHandler {
                 .ctx(format!("worker block store lookup failed: {}", e))
         })?;
 
+        // Master sparse ResizeFile may raise logical block len above the
+        // worker's physical file size. Serve the logical length and synthesize
+        // zeros for the sparse tail instead of rejecting the open.
+        let logical_len = context.len.max(meta.len);
```
szbr00: Both `context.len` and `chunk_size` are client-controlled. A request can advertise a huge logical length, open beyond `meta.len`, and cause `BytesMut::zeroed` to allocate up to approximately 2 GiB on the next read. Please enforce server-side logical-length and chunk-size limits against authoritative/configured bounds before enabling sparse synthesis.

curvine-server/src/worker/storage/layout/bdev_layout.rs
```diff
@@ -190,7 +190,14 @@ impl BlockLayout for BdevLayout {
         }
     }
 
-    fn open_reader(&self, dir: &VfsDir, meta: &BlockMeta, off: i64) -> IOResult<BlockReadContext> {
+    fn open_reader(
+        &self,
+        dir: &VfsDir,
+        meta: &BlockMeta,
+        off: i64,
+        _logical_len: i64,
```
szbr00: `_logical_len` is ignored here, while `ReadHandler` advertises that larger length for every storage type. After a sparse resize on SPDK, an open within the logical tail still fails validation against `meta.len`, and a crossing read stops at the old length. Please make bdev reads sparse-aware using `meta.len` as the initialized-data boundary, without exposing the remaining allocated extent, and add SPDK coverage.

curvine-server/src/worker/storage/layout/file_layout.rs
```diff
@@ -187,8 +144,34 @@ impl BlockLayout for FileLayout {
...
+        // rewrite then completes with committed_len larger than the staging
+        // file size. Materialize that logical length (including holes) before
+        // the mismatch check so valid post-resize rewrites can finalize.
+        if committed_len >= 0 {
+            let current_len = staging_path.metadata()?.len() as i64;
+            if committed_len > current_len {
+                OpenOptions::new()
+                    .write(true)
+                    .open(&staging_path)?
+                    .set_len(committed_len as u64)?;
```
szbr00: This extension applies to every file-backed finalize, not only rewrites of a committed sparse block. It removes the existing short-write integrity check: an ordinary write whose staging file is shorter than the claimed completion length is silently published with zero-filled missing bytes. Please restrict extension to a known committed rewrite and retain exact-length validation for normal writes.

curvine-server/src/worker/storage/block_io_context.rs
```diff
@@ -217,10 +268,30 @@ impl BlockReadContext {
...
+            .device
+            .read_region(enable_send_file, physical_chunk as i32)?;
+        let got = region.len() as i64;
+        self.block_pos += got;
+        if got == read_len {
+            return Ok(region);
+        }
+
+        let mut buf = Self::into_buffer(region)?;
+        buf.resize(read_len as usize, 0);
```
szbr00, line 301: Zero padding should start only after the complete `physical_chunk` has been read. If the device returns fewer bytes because of truncation or corruption, this currently masks the missing physical data as valid zeros. Please require `got == physical_chunk`, then append exactly `read_len - physical_chunk` sparse bytes.

curvine-server/src/worker/handler/read_handler.rs
LJF-050: Thanks. Fixed in b04d5914: sparse synthesis now requires a client-advertised logical length that is <= master.max_block_size, and chunk_size is capped at MAX_READ_AHEAD (16 MiB) before open proceeds.

curvine-server/src/worker/storage/layout/bdev_layout.rs
LJF-050: Thanks. Fixed in b04d5914: BdevLayout::open_reader now treats meta.len as the physical/initialized boundary, allows opens in the logical sparse tail, and uses BlockReadContext::with_physical so SPDK reads synthesize zeros without exposing unused allocated extent.

curvine-server/src/worker/storage/layout/file_layout.rs
LJF-050: Thanks. Fixed in b04d5914: prepare_finalize only extends the staging file when an active finalized block already exists (committed rewrite). Ordinary first writes keep the exact-length mismatch check.

curvine-server/src/worker/storage/block_io_context.rs
LJF-050: Thanks. Fixed in b04d5914: read_region now requires got == physical_chunk, then appends exactly read_len - physical_chunk sparse zeros. Short physical reads are rejected instead of being masked as valid zeros.

szbr00 (approved):

PR #1414, opened 2026-07-29
fix(fuse): support renameat2 flags and fix false OFD SETLKW EDEADLK (#1305)

### LINKED ISSUE #1305: [BUG]: FUSE rename and renameat2 semantics diverge from Linux

Curvine FUSE rename handling in curvine-fuse/src/fs/curvine_file_system.rs diverges from Linux for renameat2 and protected-directory cases.

renameat201 and renameat202 fail with EINVAL where success is expected, and rename12 succeeds unexpectedly.

Reproduction:

2. Run LTP renameat201 and rename12 against the mount.
3. Observe EINVAL on a valid renameat2 operation and unexpected success on a protected rename.

**Impact**

Applications using atomic rename, exchange, or permission-sensitive rename operations may fail or bypass expected checks on Curvine FUSE mounts.

Valid renameat2 operations should succeed, and renames that violate permissions should fail with the expected POSIX errno.

Implement renameat2 flag handling and enforce POSIX permission checks for protected rename scenarios.

**Additional context**

Root-cause statement is a hypothesis based on the observed Curvine FUSE behavior and source inspection.

Root-cause hypothesis: Root-cause hypothesis: Curvine FUSE rename and renameat2 semantics are incomplete for valid flags and protected-directory cases.

Evidence summary: renameat201 and renameat202 return EINVAL for valid renameat2 operations, while rename12 succeeds when failure is expected; rename09 cannot complete setup because chmod returns EACCES.

Related issues: https://github.com/CurvineIO/curvine/issues/1252, https://github.com/CurvineIO/curvine/issues/814

Acceptance / test gaps:

- renameat201 and renameat202 pass on curvine-fuse.
- rename09 and rename12 return Linux-compatible permission results.
- The rename profile cases introduce no new directory-entry inconsistencies.

LJF-050:
# Summary

This PR fixes Curvine FUSE renameat2 semantics for `RENAME_NOREPLACE` / `RENAME_EXCHANGE` and sticky-bit / parent ACL checks required by LTP `renameat201` / `renameat202`. It also stops concurrent OFD `F_OFD_SETLKW` waiters from getting a false `EDEADLK` under multi-thread lock/unlock races (`fcntl34_64`) while preserving true multi-resource deadlock detection (`fcntl17` / `fcntl17_64`).

# Issue Describe / Design

Closes #1305

- Root cause (rename): FUSE rename did not honor renameat2 `NOREPLACE`/`EXCHANGE` flags and did not enforce sticky-directory / parent ACL constraints expected by POSIX rename tests.
- Root cause (OFD SETLKW): `PlockWaitRegistry` treated transient waiter↔prior-holder cycles during unlock/re-lock races as deadlock and returned `EDEADLK` incorrectly. Clearing the wait edge before re-sampling Master also weakened true deadlock detection.
- Fix approach: implement renameat2 flag handling and rename ACL checks in FUSE/Master paths; resolve wait-graph edges to the root non-waiter holder; on a cycle, re-sample Master once while keeping the wait edge so free locks can be acquired and true deadlocks still return `EDEADLK`.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | renameat2 flag/ACL handling; SETLKW cycle confirm keeps wait edge | Behavior change: POSIX-aligned rename and OFD lock wait |
| `curvine-fuse/src/fs/plock_wait_registry.rs` | Resolve wait edges to root non-waiter | Behavior change: fewer false deadlocks |
| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Directory tree support for rename paths | Supports corrected rename |
| `curvine-fuse/src/fs/state/node_state.rs` | Node state updates for rename/lock owners | Supports corrected rename/lock |
| `curvine-server/src/master/meta/fs_dir.rs` | Master rename ACL / flag checks | Behavior change: reject illegal renames |
| `curvine-server/src/master/meta/store/inode_store.rs` | Inode store helpers for rename checks | Supports Master checks |
| `curvine-client-core/src/file/curvine_filesystem.rs` | Client rename API plumbing | Passes flags through |
| `curvine-client-core/src/file/fs_client.rs` | Client RPC rename flags | Passes flags through |
| `curvine-unified-fs/src/unified/unified_filesystem.rs` | Unified FS rename flag forwarding | Passes flags through |
| `curvine-server/tests/master_fs_test.rs` | Master rename coverage | Test-only |

# Test verified

| `make format` | PASS | Exact HEAD `84cda8a0037385e7bcefd8d83a72fae10ff236a5` |
| LTP `renameat201` | PASS | Assigned failure cleared on fixed HEAD |
| LTP `renameat202` | PASS | Assigned failure cleared on fixed HEAD |
| LTP `fcntl34_64` | PASS | False EDEADLK regression addressed |
| LTP `fcntl17` / `fcntl17_64` | PASS | True deadlock still returns EDEADLK |
| Profiles `fast`, `integration`, `daily`, `fuse`, `ltp`, `csi` | PASS gate | Compared against archived baseline; no new failure identities |

# Dependencies

- Related PRs: None
- Related issues: #1305
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -2132,7 +2132,34 @@ impl fs::FileSystem for CurvineFileSystem {
             if wait_guard
                 .register_blocked_by(LockOwner::new(blocker.client_id.clone(), blocker.owner_id))
             {
-                return err_fuse!(libc::EDEADLK);
+                // OFD multi-thread unlock/re-lock (LTP fcntl34) can transiently
+                // form a waiter↔prior-holder cycle in the local wait graph while
+                // the Master lock is already free or held by someone else.
+                // Drop the edge, re-sample once, and only then commit to EDEADLK
+                // so true multi-resource deadlocks (fcntl17) still fail fast.
+                wait_guard.clear_blocked_by();
+                let sleep_ms =
+                    check_interval_max_ms.min(check_interval_min_ms.saturating_mul(ticks + 1));
+                tokio::time::sleep(std::time::Duration::from_millis(sleep_ms)).await;
+                let conflict2 = self.fs.set_lock(&path, lock.clone()).await?;
+                if conflict2.is_none() {
+                    wait_guard.clear_blocked_by();
+                    if is_unlock {
+                        if full_range_unlock && lock.lock_flags == LockFlags::Plock {
+                            handle.take_plock_if_owner(lock.owner_id);
+                        }
+                    } else {
+                        handle.add_lock(lock.lock_flags, lock.owner_id);
+                    }
+                    return Ok(());
+                }
+                let blocker2 = conflict2.as_ref().expect("conflict lock");
+                if wait_guard.register_blocked_by(LockOwner::new(
+                    blocker2.client_id.clone(),
+                    blocker2.owner_id,
+                )) {
+                    return err_fuse!(libc::EDEADLK);
+                }
             }
 
             ticks += 1;
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -234,7 +234,7 @@ impl CurvineFileSystem {
             return err_fuse!(libc::ENAMETOOLONG);
         }
 
-        self.check_rename_permissions(header, old_id, old_name, new_dir)
+        self.check_rename_permissions(header, old_id, old_name, new_dir, new_name, flags)
             .await?;
 
         let (old_path, new_path) = self.state.get_path2(old_id, old_name, new_dir, new_name)?;
@@ -457,6 +457,8 @@ impl CurvineFileSystem {
         old_parent: u64,
         old_name: &str,
         new_parent: u64,
+        new_name: &str,
+        flags: RenameFlags,
     ) -> FuseResult<()> {
         if header.uid == 0 || !self.conf.check_permission {
             return Ok(());
@@ -465,20 +467,41 @@ impl CurvineFileSystem {
         let dir_mask = (libc::W_OK | libc::X_OK) as u32;
         self.check_permissions_on_node(header, old_parent, dir_mask)
             .await?;
-        if new_parent != old_parent {
+        let new_parent_status = if new_parent != old_parent {
             self.check_permissions_on_node(header, new_parent, dir_mask)
                 .await?;
-        }
+            self.state.fs_stat(new_parent, None).await?
+        } else {
+            self.state.fs_stat(old_parent, None).await?
+        };
+        let old_parent_status = if new_parent == old_parent {
+            new_parent_status.clone()
+        } else {
+            self.state.fs_stat(old_parent, None).await?
+        };
 
-        let parent_status = self.state.fs_stat(old_parent, None).await?;
-        if parent_status.mode & libc::S_ISVTX as u32 != 0 {
+        if old_parent_status.mode & libc::S_ISVTX as u32 != 0 {
             let file_status = self.state.fs_stat(old_parent, Some(old_name)).await?;
-            let parent_uid = self.resolve_file_uid(&parent_status.owner);
+            let parent_uid = self.resolve_file_uid(&old_parent_status.owner);
             let file_uid = self.resolve_file_uid(&file_status.owner);
             if header.uid != parent_uid && header.uid != file_uid {
                 return err_fuse!(
                     libc::EPERM,
                     "sticky directory: cannot rename file owned by another user"
+                );
+            }
+        }
+
+        let dest_exists =
+            flags.exchange_mode() || self.state.fs_stat(new_parent, Some(new_name)).await.is_ok();
+        if dest_exists && new_parent_status.mode & libc::S_ISVTX as u32 != 0 {
+            let dest_status = self.state.fs_stat(new_parent, Some(new_name)).await?;
+            let parent_uid = self.resolve_file_uid(&new_parent_status.owner);
+            let dest_uid = self.resolve_file_uid(&dest_status.owner);
+            if header.uid != parent_uid && header.uid != dest_uid {
+                return err_fuse!(
+                    libc::EPERM,
+                    "sticky directory: cannot rename over file owned by another user"
                 );
             }
         }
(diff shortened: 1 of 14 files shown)


curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -349,7 +349,8 @@ impl FsDir {
             Some(v) => v,
         };
         if flags.exchange_mode() {
-            return err_box!("Rename failed, because exchange mode is not supported");
+            self.unprotected_exchange(src_inp, dst_inp, mtime)?;
```
szbr00: `RENAME_EXCHANGE` also needs to reject cases where the source is below the destination. The common validation only rejects `dst` under `src`, so exchanging `/a/b` with `/a` reaches this code and can make `/a` its own descendant. Please validate ancestry in both directions for EXCHANGE.

curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -349,7 +349,8 @@ impl FsDir {
...
+            return Ok(None);
```
szbr00, line 369: Exchange is not idempotent under journal replay: replaying the same rename entry a second time finds both paths and swaps them back. An apply-batch retry after a later entry fails can therefore revert an already-applied exchange. Please make EXCHANGE replay idempotent and add a duplicate-replay regression test.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -420,6 +438,54 @@ impl CurvineFileSystem {
...
+        self.check_permissions_on_node(header, old_parent, dir_mask)
+            .await?;
+        if new_parent != old_parent {
+            self.check_permissions_on_node(header, new_parent, dir_mask)
+                .await?;
+        }
+
+        let parent_status = self.state.fs_stat(old_parent, None).await?;
+        if parent_status.mode & libc::S_ISVTX as u32 != 0 {
+            let file_status = self.state.fs_stat(old_parent, Some(old_name)).await?;
```
szbr00, line 484: Sticky-directory rename authorization must also be checked for an existing destination entry. This currently checks only the old parent/source owner, allowing overwrite or EXCHANGE to remove a destination owned by another user in a sticky directory. Please pass the destination name/flags and apply the same owner check to the destination side when it exists.

curvine-server/src/master/meta/fs_dir.rs
```diff
@@ -414,6 +415,65 @@ impl FsDir {
...
+
+        let mut at_src = dst_inode.as_ref().clone();
+        at_src.change_name(src_name.clone());
+        at_src.set_parent_id(src_parent.id());
+
+        let mut at_dst = src_inode.as_ref().clone();
+        at_dst.change_name(dst_name.clone());
+        at_dst.set_parent_id(dst_parent.id());
+
+        src_parent.update_mtime(mtime);
```
szbr00, line 497: For a cross-parent EXCHANGE between a directory and a non-directory, the parent gaining the directory must increment `nlink` and the parent losing it must decrement `nlink`. This path only updates mtimes, leaving both parent link counts incorrect. Please adjust and persist the counts based on the exchanged inode types.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -2060,6 +2132,23 @@ impl fs::FileSystem for CurvineFileSystem {
...
+                    if is_unlock {
+                        if full_range_unlock && lock.lock_flags == LockFlags::Plock {
+                            handle.take_plock_if_owner(lock.owner_id);
+                        }
+                    } else {
+                        handle.add_lock(lock.lock_flags, lock.owner_id);
+                    }
+                    return Ok(());
+                }
                 return err_fuse!(libc::EDEADLK);
```
szbr00: The second Master response may report a different blocker that does not participate in the detected cycle, but any non-`None` result returns `EDEADLK`. Please re-register/re-evaluate `conflict2` and continue waiting when the new owner does not close a cycle.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -420,6 +438,54 @@ impl CurvineFileSystem {
...
+
+        let dir_mask = (libc::W_OK | libc::X_OK) as u32;
+        self.check_permissions_on_node(header, old_parent, dir_mask)
+            .await?;
+        if new_parent != old_parent {
+            self.check_permissions_on_node(header, new_parent, dir_mask)
+                .await?;
+        }
+
+        let parent_status = self.state.fs_stat(old_parent, None).await?;
```
szbr00: The old parent is fetched once for the W|X check and immediately fetched again here for the sticky-bit check. With metadata caching disabled this adds a sequential metadata RPC to every non-root rename. Consider returning/reusing the first `FileStatus`.

curvine-server/src/master/meta/fs_dir.rs
LJF-050: Thanks. Fixed in ebabd5e1: EXCHANGE now rejects ancestry in both directions, so exchanging a path with one of its ancestors (e.g. /a/b with /a) returns EINVAL.

curvine-server/src/master/meta/fs_dir.rs
LJF-050: Thanks. Fixed in ebabd5e1: RenameEntry now records pre-swap src/dst inode ids (serde default for legacy entries). EXCHANGE replay is a no-op when the paths already reflect the post-swap state, with a duplicate-replay regression test.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in ebabd5e1: sticky-directory rename authorization now also checks an existing destination (including EXCHANGE) and denies overwrite/exchange of another user's entry unless the caller owns the sticky parent.

curvine-server/src/master/meta/fs_dir.rs
LJF-050: Thanks. Fixed in ebabd5e1: cross-parent EXCHANGE between a directory and a non-directory now increments nlink on the parent gaining the directory and decrements it on the parent losing it before apply_exchange persists the parents.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in ebabd5e1: after the confirmatory set_lock, if a different blocker is reported we re-register that edge and only return EDEADLK when it still closes a cycle; otherwise waiting continues.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Addressed in ebabd5e1 while updating sticky checks: same-parent renames reuse one parent FileStatus for sticky evaluation. Cross-parent still stats each parent once for the sticky bit (W|X checks remain via check_permissions_on_node).

szbr00 (approved):

szbr00 (approved):

PR #1434, opened 2026-07-30
fix(fuse): publish dirty-read snapshot without EAGAIN on blocking reads

### LINKED ISSUE #1429: [BUG]: FUSE blocking read returns EAGAIN under concurrent dirty-read (breaks LTP fcntl36)
## Describe the bug

On Curvine FUSE, LTP `fcntl36` / `fcntl36_64` fail during OFD vs POSIX lock races because blocking `read()` returns `EAGAIN` after a dirty-read flush budget is exhausted. Concurrent writers on the same inode keep advancing `write_ver`, so readers that already hold an OFD read lock still hit errno 11 (`EAGAIN`/`EWOULDBLOCK`). POSIX blocking reads on a regular file must not return `EAGAIN` unless the descriptor is non-blocking.

`fork09` already passes on the current runner (`OPEN_MAX=8192`) and does not need a product change, but daily defaults previously excluded it together with the fcntl36 cases via a trimmed suite.

## To Reproduce

1. Build Curvine at `upstream/main` before this fix and start a single-node cluster with FUSE mounted.
2. Run LTP 20210524 `fcntl36` (or `fcntl36_64`) against a file on the mount.
3. Observe TBROK: `read(...) failed, returned -1: EAGAIN/EWOULDBLOCK (11)` during `OFD read lock vs OFD write lock`.

## Expected behavior

Blocking `read`/`readv` wait for a coherent published snapshot and return data (or a real I/O error), never `EAGAIN`, so `fcntl36` / `fcntl36_64` PASS under concurrent OFD/POSIX lock traffic.

## OS Version

- OS: Alibaba Cloud Linux / Rocky Linux 9 container (aarch64), kernel 5.10.134
- Curvine: upstream/main prior to dirty-read snapshot fix
- LTP: 20210524

## Additional context

Root cause is in `BackendHandle::read`: requiring global `write_ver` quiescence and returning `EAGAIN` after a flush budget. Fix publishes a sampled snapshot when no enqueue is in flight (flush once, reopen, pin `read_ver`), matching the intended dirty-read semantics without inventing `EAGAIN` for blocking regular-file reads.

LJF-050:
## Summary
- Blocking FUSE reads no longer return `EAGAIN` when a shared per-inode writer keeps advancing `write_ver` under concurrent dirty-read load.
- Dirty-read now waits for no inflight enqueue, samples `write_ver`, flushes once, reopens the reader, and pins `read_ver` to that snapshot.
- This unblocks LTP `fcntl36` / `fcntl36_64` OFD vs POSIX lock races that previously TBROK on `read()` errno 11.

## Linked issue
Fixes #1429

## Root cause
`BackendHandle::read` required global `write_ver` quiescence across a flush budget and returned `EAGAIN` when concurrent writers (same inode, multiple OFD/POSIX lock holders) kept the version moving. POSIX blocking reads on regular files must not return `EAGAIN` unless `O_NONBLOCK` is set.

## Changes
- `curvine-fuse/src/fs/state/backend_handle.rs`: replace budgeted `EAGAIN` dirty-read path with single-snapshot publish.

## Test verified
- `make format` — passed on exact HEAD `7b07e9e9436d2bbe88c9a37e9f125eb79157fab4`
- LTP `fcntl36` alone — before FAIL (EAGAIN); after PASS (7/7 TPASS), multi-round PASS
- LTP `fcntl36_64` — PASS (7/7 TPASS) in combined runs
- LTP `fork09` — PASS (OPEN_MAX=8192)
- Full `syscalls-jfs` on exact HEAD — PASS (1068 PASS / 0 FAIL / 138 SKIP); `fcntl36`/`fcntl36_64`/`fork09` all executed with TPASS
- Independent review of exact HEAD — no blocking findings

## Dependencies
None

### DIFF
--- a/curvine-fuse/src/fs/state/backend_handle.rs
+++ b/curvine-fuse/src/fs/state/backend_handle.rs
@@ -20,7 +20,6 @@ use crate::session::FuseResponse;
 use crate::{err_fuse, FuseError, FuseResult, FuseUtils};
 use curvine_common::fs::{Path, StateReader, StateWriter};
 use curvine_common::state::{CreateFileOptsBuilder, FileStatus, LockFlags, OpenFlags};
-use log::warn;
 use orpc::err_box;
 use orpc::sync::AtomicCounter;
 use orpc::sys::RawPtr;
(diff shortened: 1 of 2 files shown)


curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -133,54 +132,29 @@ impl BackendHandle {
...
-                };
+            // One FuseWriter is shared per inode. LTP ftest04 forks several O_RDWR
+            // children against that writer, so write_ver keeps moving under load.
+            // Requiring global quiescence (stable write_ver across flush rounds) and
+            // returning EAGAIN when a budget is exhausted breaks blocking read/readv
+            // (errno=11). Instead publish a snapshot taken when no enqueue is in
+            // flight: flush once, reopen, and pin read_ver to that sampled target.
+            // Later concurrent writes leave write_ver ahead of read_ver so the next
+            // read publishes again; we never claim a newer ver than we flushed.
+            while writer.enqueue_inflight() != 0 {
```
szbr00: **Major:** This still requires a global zero-inflight instant, but `enqueue_inflight` is incremented before `send_queued_task()` and remains nonzero while a bounded channel waits for capacity. Under sustained concurrent writers, overlapping enqueue attempts can therefore keep the counter above zero indefinitely and make this blocking read spin/yield forever—the EAGAIN failure becomes a hang. Since `write_ver` advances only after a task is queued, please use a snapshot scheme that does not wait for global enqueue quiescence (or otherwise provide a starvation-free handoff), and add a stress test with continuous writers plus a bounded queue that asserts the read completes.

szbr00 (commented):
Found one major concurrency-progress issue in the dirty-read path.

curvine-fuse/src/fs/state/backend_handle.rs
LJF-050: Thanks. Fixed in 45817ba: replaced the global `enqueue_inflight == 0` wait with an `enqueue_gate` that serializes write/resize enqueue with dirty-read flush. `publish_dirty_read_snapshot` captures `write_ver` and queues Flush under the gate, so a bounded-channel Flush cannot overtake a mid-reserve Write, while continuous writers no longer starve the publish path. Added `dirty_read_snapshot_completes_under_continuous_bounded_writers` (capacity-1 queue + continuous writers) to assert the snapshot completes. `make format` and the targeted fuse_writer/backend_handle tests passed.

szbr00 (approved):

PR #1442, opened 2026-07-31
fix(fuse): handle receiver shutdown channel closure (#1405)

yangcx000:
# Summary

Handle closure of the receiver shutdown watch channel explicitly and improve receiver/sender task failure logs.

# Issue Describe / Design

Related to #1405.

- Root cause: `watch::Receiver::changed()` returns immediately with an error after all senders are dropped. Ignoring that result can keep the receiver loop selecting the already-closed shutdown branch repeatedly.
- Reproduction: drop the session shutdown sender while a receiver is still running.
- Fix approach: log the unexpected channel closure once at warning level and exit the receive loop. Recoverable FUSE receive errors remain metric-only to avoid noisy duplicate logs, while terminal receiver/sender errors continue to be logged once by the session task wrapper with clearer messages.

# Changes

| `curvine-fuse/src/session/channel/fuse_receiver.rs` | Handle shutdown watch closure explicitly and exit the receive loop | Prevents a potential busy loop after shutdown sender loss |
| `curvine-fuse/src/session/fuse_session.rs` | Replace generic accept/send failure messages with receiver/sender-specific errors | Improves operational diagnostics without duplicate logs |

# Test verified

| `make format` | PASS | Workspace formatting and pre-commit checks |
| `cargo test -p curvine-fuse session::channel::fuse_receiver::tests` | PASS | 29 tests passed |
| `cargo test -p curvine-fuse session::fuse_session::tests` | PASS | No matching unit tests; test binaries completed successfully |
| `cargo check -p curvine-fuse` | PASS | |

# Dependencies

- Related PRs: #1405
- Related issues: #1091
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -733,9 +733,47 @@ mod tests {
     use crate::fuse_metrics::{RECEIVE_ACTION_CONTINUE, RECEIVE_ACTION_EXIT};
     use bytes::BytesMut;
     use libc::{EAGAIN, ECONNABORTED, EINTR, EIO, ENODEV, ENOENT};
+    use orpc::runtime::{AsyncRuntime, RpcRuntime};
+    use orpc::sync::channel::AsyncChannel;
     use orpc::sync::FastDashMap;
+    use orpc::sys;
+    use orpc::sys::pipe::{AsyncFd, OwnedFd};
     use std::sync::Arc;
-    use tokio::sync::Notify;
+    use std::time::Duration;
+    use tokio::sync::{watch, Notify};
+
+    #[test]
+    fn receiver_exits_when_shutdown_channel_closes() {
+        let rt = Arc::new(AsyncRuntime::single());
+        rt.clone().block_on(async move {
+            let [read_fd, write_fd] = sys::pipe2(4096).unwrap();
+            let read_fd = OwnedFd::new(read_fd);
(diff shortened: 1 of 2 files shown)


curvine-fuse/src/session/channel/fuse_receiver.rs
```diff
@@ -498,10 +498,17 @@ impl<T: FileSystem> FuseReceiver<T> {
...
-                        break;
+                changed = shutdown_rx.changed() => {
+                    match changed {
+                        Ok(()) if *shutdown_rx.borrow() => {
+                            info!("receiver observed shutdown broadcast; exiting receive loop");
+                            break;
+                        }
+                        Ok(()) => {}
+                        Err(_) => {
+                            warn!("receiver shutdown channel closed; exiting receive loop");
```
szbr00, line 509: **Minor:** This branch fixes the closed-channel busy loop, but the PR does not add a regression test that drops the last shutdown sender and verifies the receiver exits promptly. Please add focused coverage for that path so a future change cannot reintroduce the immediate-error spin.

yangcx000: Thanks. Added focused regression coverage in 93dc9f6b. The test runs the real receiver loop against an idle nonblocking pipe, drops the last shutdown sender, and uses a timeout to verify start() exits promptly instead of spinning on the immediately-ready closed watch channel.

szbr00 (approved):

Issue #1321, opened 2026-07-27
[FEATURE] Shorten the master FsDir metadata write-lock critical path


jlon:
## Is your feature request related to a problem?

The master serializes unrelated file metadata mutations through the global
`SyncFsDir` write lock. `add_block` and `complete_file` hold this lock while
performing state changes, RocksDB batch commits, journal enqueue, and
conditionally checkpoint work. A slow operation for one file can therefore
raise tail latency for block lifecycle operations of unrelated files.

This is distinct from #951. The goal is to reduce the critical section in the
current stateful master, not to make the master stateless.

## Describe the solution you'd like

First instrument and characterize the exact lock dependency and ordering
requirements. Then design a versioned metadata commit path whose global lock
scope contains only the operations that must be serialized for correctness.
Slow persistence, journal, or checkpoint work may leave that scope only when
an explicit state transition, rollback path, and Raft recovery proof preserve
the current linearizable file/block lifecycle.

The design must not assume that every journal operation can be moved outside
the lock; it must retain whichever ordering is required by correctness.

## Describe alternatives you've considered

- Increasing master threads does not remove the shared write lock.
- Moving to an external metadata store is the broader discussion in #951 and
  is not required to solve this critical section.
- Keeping the current lock scope preserves correctness but limits concurrent
  multi-file writes.

## Additional context

The FUSE I/O audit identifies this as W3, an architecture-level bottleneck.
A full-block-location response and checkpoint/journal delays can further
extend the same lock hold time.

# Goal

| Sub-requirement | Problem solved |
| --- | --- |
| Measure lock wait and hold time by operation | Establish the actual contention source |
| Define lock ordering and state-transition invariants | Prevent an unsafe asynchronous rewrite |
| Reduce the critical section only where the invariants permit | Prevent unrelated files from sharing long critical sections |
| Preserve Raft failover and recovery behavior | Avoid turning a performance change into data loss |

# Not Goal

- Making the master stateless or replacing RocksDB.
- Weakening file/block visibility or durability semantics.
- Treating worker or FUSE local bottlenecks as master-lock fixes.

# Test Plan

| Test case | Test objective | Modules involved |
| --- | --- | --- |
| Concurrent multi-file block allocation and completion | Show reduced lock contention and no lost state updates | master, client |
| Failure between mutation and persistence | Verify rollback/recovery remains correct | master, journal, RocksDB |
| Raft leader change during block lifecycle | Verify consistency after failover | master, journal |
[...]

Issue #951, opened 2026-06-29
[FEATURE] Make the Master stateless to enable horizontal scaling, with metadata stored in an external distributed database


yangyang233333:
**Is your feature request related to a problem?**

Today the Master keeps its metadata locally (RocksDB) and relies on Raft for HA. This couples metadata capacity and request-serving capacity to a single leader, which limits how far a cluster can scale as the namespace and request volume grow.

**Describe the solution you'd like**

Explore making the Master **stateless** so it can be scaled out horizontally (linear scaling), by moving metadata into an external distributed database/KV service (e.g. FoundationDB or a cloud-managed KV service) as the source of truth.

I'd like to open this up for **community discussion first** — gathering thoughts on feasibility, trade-offs, and preferred direction — before any concrete design is proposed. Detailed design intentionally omitted at this stage.

**Describe alternatives you've considered**

(Open for discussion.)

**Additional context**

None yet — looking forward to community input.

PR #1451, opened 2026-07-31
docs(master): add metadata service design summary (#951)

yangcx000:
# Summary

Add a Chinese review document that summarizes Curvine metadata-service behavior and proposes an evolution path toward a stateless MDS with metadata storage separated by `WriteType`.

The document distinguishes current implementation facts from proposed design directions. It is intentionally submitted as a draft for community discussion; the per-operation protocol design in Chapter 4 remains to be completed after the high-level direction is reviewed.

# Issue Describe / Design

Related to #951 and #1321.

The current Master keeps CacheMode soft-state metadata and FsMode/native authoritative metadata in the same in-memory tree, lock domain, RocksDB store, and Raft journal. This couples request-serving capacity, metadata capacity, and persistence policy even though the two metadata classes have different consistency and durability requirements.

The document:

- summarizes Curvine use cases, clients, and metadata pressure patterns;
- derives metadata requirements from CacheMode and FsMode product semantics;
- proposes separating soft-state attribute caching from authoritative POSIX metadata;
- describes a stateless MDS computation layer with pluggable KV backends;
- outlines schema, transaction, routing, recovery, and inode-addressing requirements;
- records open design work for per-operation interaction and cross-shard rename.

This PR proposes a discussion document only. It does not change runtime behavior or commit the project to a specific storage backend.

# Changes

| `metadata-service-summary.md` | Add the MDS current-state summary and proposed metadata-service architecture | Documentation only; no runtime behavior change |

# Test verified

| `git diff --no-index --check /dev/null metadata-service-summary.md` | PASS | No whitespace errors |
| Manual document review | PASS | Checked headings, repository paths, terminology, and local-only references |

# Dependencies

- Related PRs: None
- Related issues: #951, #1321
- Blocks / blocked by: Community design review before implementation planning

### DIFF
--- a/metadata-service-summary.md
+++ b/metadata-service-summary.md
@@ -1,0 +1,484 @@
+# Curvine 元数据服务(MDS)总结文档
+
+> 状态:评审中(Review) · 版本:v0.1 · 最后更新:2026-07-31
+>
+> 本文自顶向下组织:先厘清 **使用场景与客户端**(第 1 章),再定义 **CacheMode / FsMode 的产品能力**(第 2 章),然后基于这两者反推 **新的元数据服务、存储与 Schema 设计**(第 3 章),最后 **逐操作定义元数据交互**(第 4 章)。
+>
+> 第 1、2 章是对现状的事实性描述(附代码位置),用于统一认知;第 3、4 章是元数据服务的 **设计与演进方向**,供评审与迭代。
+
+---
+
+## 0. 概览
+
+### 0.1 核心主线(Thesis)
+
+本文的核心主线是:**元数据的"权威性 / 持久性等级"应当由挂载的 `WriteType` 决定**。
+
+- **CacheMode**:UFS(S3/OSS/HDFS)是数据与元数据的 **权威**;Curvine 侧的元数据本质上是 UFS 元数据的 **可重建缓存(soft-state)**。丢失后可从 UFS 重新拉取,因此对 **强一致** 的要求较低,可以走"惰性填充 + 按 TTL / 内存压力淘汰单条"的路径。但"可重建"仅指 **正确性可回源**,后端 meta 存储 **仍需大容量 + 高可用**——一旦大面积丢失,请求会集中 fallback 至 UFS(S3),造成缓存击穿。
+- **FsMode**:Curvine 本身是 **权威**,UFS 只是异步刷写的 **持久副本**。此时元数据是 **硬状态(hard-state)**,必须强一致、强持久(如 Raft journal 等强一致持久化),不可丢。
+
+当前实现对两类子树 **一视同仁**:全部塞进同一棵内存树、同一把全局锁、同一份 Raft journal、同一个 RocksDB。这既让 cache-mode 路径背上了不必要的持久化 / 一致性开销,也让 fs-mode 路径与海量、易变的 cache 元数据争抢同一把锁。**按 WriteType 对元数据做分层治理**,是本设计的主线。
+
+### 0.2 名词表
+
+| 名词 | 含义 |
+|---|---|
+| **MDS / Master** | 元数据服务。当前是 `curvine-server` crate 内的 `master` 模块 |
+| **Worker** | 数据节点,负责多级缓存(mem/SSD/HDD)块的读写 |
+| **UFS** | Under File System,底层持久存储(S3 / OSS / GCS / HDFS) |
+| **Mount** | 将一个 Curvine 路径 `cv_path` 绑定到一个 UFS 路径 `ufs_path` 的映射 |
+| **WriteType** | 挂载的写入 / 权威模型,取值 `CacheMode` 或 `FsMode` |
+| **AccessMode** | 挂载的读写权限,取值 `ReadOnly` / `ReadWrite`(与 WriteType 正交) |
+| **inode** | 文件 / 目录的元数据节点 |
+| **dentry / edge** | 目录项,`(parent_id, name) -> child_id` 的命名空间边 |
+| **block** | 文件被切分成的定长数据块,由 Worker 缓存 |
+| **journal** | 编辑日志,当前经 Raft 复制 |
+
+### 0.3 代码位置索引(现状)
+
+| 组件 | 路径 |
+|---|---|
+| Master 模块根 | `curvine-server/src/master/` |
+| 内存树 + 变更逻辑 | `curvine-server/src/master/meta/fs_dir.rs` |
+| inode 类型 | `curvine-server/src/master/meta/inode/` |
+| RocksDB 持久化 | `curvine-server/src/master/meta/store/rocks_inode_store.rs` |
+| Raft journal | `curvine-server/src/master/journal/` |
+| 命名空间门面 | `curvine-server/src/master/fs/master_filesystem.rs` |
+| RPC 分发 | `curvine-server/src/master/master_handler.rs` |
+| WriteType / MountInfo | `crates/common/curvine-model/src/mount.rs` |
+| 元数据 proto | `curvine-common/proto/master.proto` |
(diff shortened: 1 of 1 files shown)


metadata-service-summary.md
```diff
@@ -0,0 +1,484 @@
...
+
+| 组件 | 路径 |
+|---|---|
+| Master 模块根 | `curvine-server/src/master/` |
+| 内存树 + 变更逻辑 | `curvine-server/src/master/meta/fs_dir.rs` |
+| inode 类型 | `curvine-server/src/master/meta/inode/` |
+| RocksDB 持久化 | `curvine-server/src/master/meta/store/rocks_inode_store.rs` |
+| Raft journal | `curvine-server/src/master/journal/` |
+| 命名空间门面 | `curvine-server/src/master/fs/master_filesystem.rs` |
+| RPC 分发 | `curvine-server/src/master/master_handler.rs` |
```
szbr00, line 47: Several paths in §0.3 are already wrong against current `main` (and one is wrong even on this PR branch):

- Master code lives under `curvine-master/src/master/...` now, not `curvine-server/src/master/...` (`curvine-server` is a thin binary shell after the master/worker split).
- Unified FS is at `crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs`, not the top-level `curvine-unified-fs/...` cited here.

Please refresh the index (and the inline citations that reuse these prefixes) to match `main`, otherwise the “附代码位置” sections will be misleading as soon as this merges.

metadata-service-summary.md
```diff
@@ -0,0 +1,484 @@
...
+| **rename** | 先 `rename` UFS 再删源缓存;**不支持 rename flags** | `unified_filesystem.rs:757-787` |
+| **TTL 默认** | `TtlAction::Delete`(淘汰即丢弃 Curvine 副本,UFS 不动) | `mount.rs:330-333` |
+| **free** | 丢弃 Curvine 的元数据 / 块(`cv.delete`),UFS 完好;根路径 recursive 时逐个删子项而不删根 | `unified_filesystem.rs:271-303` |
+| **set_attr(chmod/chown/utimes)** | cache-mode 下 **静默 no-op**,返回 `None`——不报错也不落 UFS | `unified_filesystem.rs:719-735` |
+| **不支持的 POSIX 操作** | `symlink` / `link` / `resize` / `get_lock` / `set_lock` / `mknod` 在 cache-mode 上直接返回 **unsupported 错误**(与 set_attr 的 no-op 不同) | `unified_filesystem.rs:305-368,737-755` |
+
+**元数据视角(小结,展开见 2.7)**:由于 `readdir` / `rename` / `hardlink` / `symlink` / `resize` / 锁 等操作要么不支持、要么直接 fallback 至 UFS(见上表),它们 **根本不经过 CacheMode 的元数据服务**。因此 CacheMode 侧需要维护的仅是 **目录 / 文件的属性元数据**(UFS 属性的投影,如 size/mtime/mode/owner),而非完整目录树、父子边、锁表等重型结构——这部分元数据是 **可从 UFS 重建的软状态(soft-state)**。
+
+### 2.3 FsMode 语义(Curvine 权威)
+
```
szbr00, line 215: This overstates CacheMode `set_attr` behavior. `fuse_set_attr` / `FileSystem::set_attr` do return a silent no-op once `get_mount_checked` succeeds for a cache-mode mount, but `get_mount` rejects `RpcCode::SetAttr` when `is_read_only_cache_mode()` is true (`unified_filesystem.rs` ~162–167, and `SetAttr` is in `is_mount_write_rpc`).

Since `AccessMode::ReadOnly` is the default, the common CacheMode path returns `unsupported`, not a silent no-op. Please split the table row by AccessMode (RO → error, RW → no-op), otherwise §2.2 and the soft-state “set_attr is harmless” implication in §2.7/§3 are inaccurate for the default product config.

metadata-service-summary.md
```diff
@@ -0,0 +1,484 @@
...
+Curvine 是真正的分布式文件系统权威,UFS 是异步刷写的持久副本。
+
+> **核心定位 —— FsMode 就是一个"完整、持久、全 POSIX 语义"的分布式文件系统。**
```
szbr00, line 218: The claim that CacheMode MDS only needs UFS attribute projection (size/mtime/mode/owner) is too strong for the current data path.

Even when `list`/`mkdir`/`rename` bypass or invalidate CV namespace state, a valid CacheMode read still goes through Curvine: `get_status` hits `cv.get_status` first, and the open/read path calls `cv.get_block_locations` and builds an `FsReader` from those locations (`unified_filesystem.rs` ~398–435, ~1046–1057). Load/async-cache jobs also materialize CV file inodes + block maps.

So soft-state storage cannot be “attribute KV only” if CacheMode is still expected to serve hot cached bytes from Workers. Please either (a) include block-layout / file-inode state in the soft-state schema (with TTL/`free` lifecycle), or (b) explicitly state that block maps stay on a separate cache-index service and attribute KV is only for `stat`. As written, §2.7 / §3.5.1 understate what must remain highly available to avoid UFS stampedes on open.

metadata-service-summary.md
```diff
@@ -0,0 +1,484 @@
+# Curvine 元数据服务(MDS)总结文档
```
szbr00, line 1: Please place this under `docs/` (e.g. `docs/metadata-service-summary.md`) rather than the repo root. Design / analysis docs already live there (`docs/CURVINE_METADATA_ANALYSIS.md`, `docs/fuse_metadata_cache_design.md`, `docs/P2P_DATA_READ_PROTOCOL_DESIGN.md`, etc.); the root should stay reserved for project entry docs. There is also related prior art in `docs/CURVINE_METADATA_ANALYSIS.md` — a short “Related docs” pointer would help reviewers.

metadata-service-summary.md
```diff
@@ -0,0 +1,484 @@
...
+
+#### 3.5.1 软状态 Schema(CacheMode:属性缓存)
+
+- **Key**:**`hash(parent_dir, name)`**——对路径身份取哈希,把条目均匀打散、**消除热点**(热目录 / 连续命名不再集中到同一分片)。可行前提:CacheMode 无 `list` / readdir,无需保留前缀顺序。
+- **Value**:UFS 属性的投影——`size / mtime / mode / owner / etag` 等;**不含** 父子边、nlink、符号链接目标、锁表(2.2 已论证这些操作在 CacheMode 下不经过 MDS)。
+- **操作面**:仅 `Get` / `Put` / `Delete` 点操作 + 原子点写;**无** `list`(readdir 回源 UFS)、**无** 目录树、**无** 跨键事务。
+- **生命周期**:懒填充、按 TTL / 内存压力单条淘汰。
+
+#### 3.5.2 硬状态 Schema(FsMode / native:完整 POSIX)
+
```
szbr00, line 438: `hash(parent_dir, name)` needs more precise identity rules before it can be the soft-state primary key:

1. Scope by mount (e.g. `mount_id` / namespace prefix); otherwise two mounts with overlapping relative names collide.
2. Define whether `parent_dir` is the CV path, UFS path, or inode id, and how it is canonicalized (trailing slash, mount-relative vs absolute).
3. Parent rename / mount remap: hashing a path string orphans all children with no reverse index; if CacheMode truly has no listing/scan, there is no cheap way to invalidate or rewrite the subtree.

If the goal is hotspot avoidance, consider `hash(mount_id, parent_id_or_ufs_key, name)` (or a content-addressed UFS object key) and document rename/invalidation as an explicit non-goal or as UFS-driven lazy miss, rather than implying path-string hashing alone is enough.

yangcx000:
暂时撤回此设计文档 PR，后续视需要再重新提交。感谢 @szbr9486 的详细 review。

PR #1470, opened 2026-08-03
fix(fuse): append at full block boundaries (#1291)

jlon:
# Summary

Fix Curvine FUSE writes that begin exactly at the logical EOF of a fully populated block. Previously, the writer looked up that offset as if it belonged to the previous block and returned `EIO`; it now follows the existing append path and allocates the next block when no block locator covers the EOF position.

# Issue Describe / Design

Related to #1291

## Root cause

`WriteFileBlocks::get_block` uses half-open block ranges and advances when `range.end <= pos`. At `pos == file_len == block_size`, a file with one full block has no locator for the next byte. `FsWriterBase::seek` only treated `pos > file_len` as an append, then called `get_block_check(pos)` for the equality case. That lookup failed and surfaced as FUSE `EIO`.

## Reproduction steps

1. Create a 128 MiB file, equal to the Curvine block size.
2. Write one 4 KiB page at offset 128 MiB.
3. Before this change, FUSE logs `Not found block for pos 134217728` and returns `EIO` on close.

## Fix approach

Use a local predicate for append allocation: preserve the existing `pos > file_len` behavior and additionally append only when `pos == file_len` has no block locator. A partial final block still accepts an EOF write through the existing in-block path; sparse writes keep their current behavior.

# Changes

| `crates/client/curvine-client-core/src/file/fs_writer_base.rs` | Allocate the next block for an EOF seek with no covering block locator; add a boundary regression test. | Fixes full-block append failure. Partial-block random writes and sparse writes are unchanged. |

# Test verified

| `cargo test -p curvine-client-core append_at_full_block_boundary_needs_a_new_block` | PASS | Covers 128 MiB full-block boundary and partial-block EOF behavior. |
| `cargo test -p curvine-client-core` | PASS | 21 unit tests and 5 integration tests passed. |
| Local FUSE 128 MiB boundary reproduction | Reproduced before fix | Fixed-image deployment and repeat run are pending. |

# Dependencies

- Related PRs: None
- Related issues: #1291
- Blocks / blocked by: Local image packaging and cluster regression verification

### DIFF
--- a/crates/client/curvine-client-core/src/file/fs_writer_base.rs
+++ b/crates/client/curvine-client-core/src/file/fs_writer_base.rs
@@ -50,6 +50,10 @@ pub struct FsWriterBase {
     durable_blocks: FastHashSet<i64>,
 }
 
+fn needs_append_block(file_blocks: &WriteFileBlocks, file_len: i64, pos: i64) -> bool {
+    pos > file_len || (pos == file_len && file_blocks.get_block(pos).is_none())
+}
+
 impl FsWriterBase {
     pub fn new(fs_context: Arc<FsContext>, path: Path, status: FileBlocks, pos: i64) -> Self {
         let fs_client = FsClient::new(fs_context.clone());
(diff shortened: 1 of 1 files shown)


szbr00 (commented):
Reviewed the full-block EOF append path and its regression coverage at `dc74cfa2a5eda3a8d8a38171522cbaa1da90df1c`. I found no actionable correctness, safety, or design issues. The change preserves the existing sparse-seek behavior while routing an uncovered EOF boundary to next-block allocation.

szbr00 (approved):

PR #1473, opened 2026-08-03
fix(fuse): stabilize POSIX lock deadlock victim (#1254)

### LINKED ISSUE #1254: [BUG]: FUSE fcntl17 deadlock detection fails and child processes die unexpectedly

LJF-050:
# Summary

This PR makes FUSE POSIX deadlock handling deterministic so the later request in a lock cycle receives EDEADLK regardless of task scheduling order. It also keeps wait edges consistent across retries and improves wake-up and cleanup behavior.

# Issue Describe / Design

Closes #1254

**Root cause:** The local POSIX wait registry returned EDEADLK to whichever SETLKW request completed cycle registration second. Tokio scheduling could therefore reject an earlier request instead of the later FUSE request expected by fcntl17.

**Reproduction:** Run fcntl17 or fcntl17_64 repeatedly through the FUSE mount. Under the affected scheduling order, the wrong process could become the deadlock victim.

**Fix approach:** Track direct waiter-to-blocker edges with FUSE request metadata, select the cycle member with the greatest request unique as the deterministic victim, re-sample the server before returning EDEADLK, and clear or notify wait state on success, error, cancellation, and ownership changes.

Compatibility risk is low: the change is limited to FUSE POSIX lock waiting and preserves existing bounded retry behavior. Rollback is the single commit 3835bbc6a498a35ce875d4bd3c1908ed72e648c0.

# Changes

| `curvine-fuse/src/fs/plock_wait_registry.rs` | Record direct wait edges, detect cycles, and choose the victim deterministically by request order. | Behavior change: stable EDEADLK victim selection and safer stale-edge cleanup. |
| `curvine-fuse/src/fs/curvine_file_system.rs` | Pass request and lock metadata into wait registration, re-sample before EDEADLK, and notify retries. | Behavior change limited to blocking POSIX lock acquisition. |

# Test verified

| `cargo fmt --all -- --check` | PASS | Exact HEAD 3835bbc6 |
| `cargo test -p curvine-fuse plock_wait_registry -- --nocapture` | PASS | 11 passed, 0 failed |
| `cargo clippy -p curvine-fuse --all-targets -- -D warnings` | PASS | No warnings |
| `fcntl17` | PASS | 500/500 iterations |
| `fcntl17_64` | PASS | 500/500 iterations |
| `fcntl34`, `fcntl34_64` | PASS | No false deadlock regression |
| Daily baseline/candidate comparison | PASS | Identical environment, profile, and failure signatures; no new failures |
| Independent review | PASS | Exact HEAD approved; no blocking findings |

# Dependencies

- Related PRs: None
- Related issues: #1254
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -353,6 +353,12 @@ impl CurvineFileSystem {
             return Err(e.into());
         }
 
+        // Notify the wait registry so any waiter blocked on this owner can
+        // re-sample Master and either acquire the lock or re-register.
+        if flags == LockFlags::Plock {
+            self.plock_waits.notify_waiters();
+        }
+
         Ok(())
     }
 
@@ -2106,10 +2112,10 @@ impl fs::FileSystem for CurvineFileSystem {
                 if full_range_unlock && flag == LockFlags::Plock {
                     handle.take_plock_if_owner(owner_id);
                 }
+                self.plock_waits.notify_waiters();
             } else {
                 handle.add_lock(flag, owner_id);
             }
-            self.plock_waits.notify_waiters();
             Ok(())
         } else {
             err_fuse!(libc::EAGAIN)
@@ -2160,7 +2166,6 @@ impl fs::FileSystem for CurvineFileSystem {
                 } else {
                     handle.add_lock(lock.lock_flags, lock.owner_id);
                 }
-                self.plock_waits.notify_waiters();
                 return Ok(());
             }
 
@@ -2198,7 +2203,6 @@ impl fs::FileSystem for CurvineFileSystem {
                     } else {
                         handle.add_lock(lock.lock_flags, lock.owner_id);
                     }
-                    self.plock_waits.notify_waiters();
                     return Ok(());
                 }
                 let blocker2 = conflict2.as_ref().expect("conflict lock");
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -14,7 +14,9 @@
 
 use crate::fs::dcache::{CleanerTask, Inode};
 use crate::fs::operator::*;
-use crate::fs::plock_wait_registry::{LockOwner, PlockWaitGuard, PlockWaitRegistry};
+use crate::fs::plock_wait_registry::{
+    LockOwner, LockWaitInfo, PlockWaitDecision, PlockWaitGuard, PlockWaitRegistry,
+};
 use crate::fs::state::{FileHandle, NodeState};
 use crate::fuse_metrics::{
     ReaddirTimer, INVAL_REASON_FLUSH, INVAL_REASON_FSYNC, INVAL_REASON_RELEASE, INVAL_REASON_RESIZE,
@@ -2107,6 +2109,7 @@ impl fs::FileSystem for CurvineFileSystem {
             } else {
                 handle.add_lock(flag, owner_id);
             }
+            self.plock_waits.notify_waiters();
             Ok(())
         } else {
             err_fuse!(libc::EAGAIN)
@@ -2137,6 +2140,14 @@ impl fs::FileSystem for CurvineFileSystem {
         let wait_guard = PlockWaitGuard::new(
             self.plock_waits.clone(),
             LockOwner::new(lock.client_id.clone(), lock.owner_id),
+            LockWaitInfo::new(
+                op.header.unique,
+                lock.pid,
+                op.header.nodeid,
+                path.to_string(),
+                lock.start,
+                lock.end,
+            ),
         );
         loop {
             let conflict = self.fs.set_lock(&path, lock.clone()).await?;
@@ -2149,13 +2160,29 @@ impl fs::FileSystem for CurvineFileSystem {
                 } else {
                     handle.add_lock(lock.lock_flags, lock.owner_id);
                 }
+                self.plock_waits.notify_waiters();
                 return Ok(());
             }
 
             let blocker = conflict.as_ref().expect("conflict lock");
-            if wait_guard
-                .register_blocked_by(LockOwner::new(blocker.client_id.clone(), blocker.owner_id))
-            {
+            let decision = wait_guard
+                .register_blocked_by(LockOwner::new(blocker.client_id.clone(), blocker.owner_id));
+            debug!(
+                "plock SETLKW wait decision unique={} pid={} owner_id={} nodeid={} path={} range=[{},{}] blocker_pid={} blocker_owner_id={} blocker_range=[{},{}] decision={:?}",
+                op.header.unique,
+                lock.pid,
+                lock.owner_id,
+                op.header.nodeid,
+                path,
+                lock.start,
+                lock.end,
+                blocker.pid,
+                blocker.owner_id,
+                blocker.start,
+                blocker.end,
+                decision
+            );
+            if decision.is_deadlock() {
                 // Cycle in the local wait graph. Re-sample Master once while
                 // keeping our edge published so a peer in a true multi-resource
                 // deadlock (LTP fcntl17) still observes the cycle. If the lock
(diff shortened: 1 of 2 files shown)


curvine-fuse/src/fs/plock_wait_registry.rs
```diff
@@ -31,10 +34,106 @@ impl LockOwner {
...
+struct WaitRecord {
+    blocked_by: LockOwner,
+    info: LockWaitInfo,
+}
+
 /// Tracks in-flight F_SETLKW waiters so circular wait chains return EDEADLK.
-#[derive(Default)]
 pub(crate) struct PlockWaitRegistry {
-    waiters: Mutex<HashMap<LockOwner, LockOwner>>,
+    waiters: Mutex<HashMap<LockOwner, WaitRecord>>,
```
szbr00: This map assumes only one in-flight wait per lock owner. Multiple concurrent `F_SETLKW` requests sharing an owner overwrite the previous edge, and either guard can then unregister the other request’s edge on drop. That can hide a real cycle indefinitely. Please key registrations by request identity, unregister only that request, and account for all owner edges during cycle detection.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -2171,20 +2198,50 @@ impl fs::FileSystem for CurvineFileSystem {
...
+                        cycle
+                    );
                     return err_fuse!(libc::EDEADLK);
                 }
             }
 
             ticks += 1;
             let sleep_ms = check_interval_max_ms.min(check_interval_min_ms.saturating_mul(ticks));
-            tokio::time::sleep(std::time::Duration::from_millis(sleep_ms)).await;
+            self.plock_waits
```
szbr00, line 2246: Moving retries to `wait_for_change` requires every successful unlock path to notify the registry. `flush` and `release` unlock through `fs_unlock_owner`, which currently sends no notification, so waiters can remain asleep for the full polling timeout after the lock becomes available. Please notify after those unlocks succeed.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -2149,13 +2160,29 @@ impl fs::FileSystem for CurvineFileSystem {
...
                 }
+                self.plock_waits.notify_waiters();
                 return Ok(());
             }
 
             let blocker = conflict.as_ref().expect("conflict lock");
-            if wait_guard
-                .register_blocked_by(LockOwner::new(blocker.client_id.clone(), blocker.owner_id))
-            {
+            let decision = wait_guard
```
szbr00, line 2173: If this registration closes a cycle but selects an already-waiting request as the victim, that request is not awakened. It can only return `EDEADLK` after its next timed retry, delaying deadlock resolution by up to the maximum interval. Please wake the selected victim when the cycle is created.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -2107,6 +2109,7 @@ impl fs::FileSystem for CurvineFileSystem {
             } else {
                 handle.add_lock(flag, owner_id);
             }
+            self.plock_waits.notify_waiters();
```
szbr00: This wakes every waiter for every successful operation, including lock acquisitions and unrelated files. Under contention, each success can trigger a filesystem-wide burst of Master `SetLock` RPCs. Please scope notifications by lock resource and emit them only for changes that may unblock a waiter; edge removal already provides the necessary graph-change notification.

curvine-fuse/src/fs/plock_wait_registry.rs
```diff
@@ -47,106 +146,141 @@ impl PlockWaitRegistry {
...
+        }
+    }
+
+    pub fn notify_waiters(&self) {
+        // Store one permit for a waiter that has observed the conflict but has
+        // not yet registered its `notified()` future. `notify_waiters()` alone
+        // only wakes already registered futures, which can lose the only unlock
+        // event needed to advance an OFD lock convoy.
+        self.change_notify.notify_one();
+        self.change_notify.notify_waiters();
```
szbr00: The `notify_one()` + `notify_waiters()` pair still loses wake-ups whenever more than one waiter sits between its Master sample and its `notified()` await. `notify_one()` stores at most one permit and `notify_waiters()` only wakes already-registered futures, so with N racing waiters a single unlock rescues one and the other N-1 sleep for the full backoff interval, which is the latency this change is meant to remove. The stored permit is also unscoped and never expires, so a later unrelated `wait_for_change` consumes it and returns immediately for a change that does not concern it, costing an extra Master `SetLock` round trip. Consider creating the `notified()` future before sampling Master and awaiting it afterwards, or replacing `Notify` with a `watch` channel carrying a monotonically increasing generation so each waiter can compare against the generation it observed. Note that `notify_waiters_wakes_blocked_retry` uses a single waiter and would not catch this.

curvine-fuse/src/fs/plock_wait_registry.rs
```diff
@@ -31,10 +34,106 @@ impl LockOwner {
...
+    pub victim_request_unique: u64,
+}
+
+impl PlockDeadlockCycle {
+    fn new(edges: Vec<LockWaitEdge>) -> Self {
+        debug_assert!(!edges.is_empty());
+
+        let mut victim_idx = 0;
+        for (idx, edge) in edges.iter().enumerate().skip(1) {
+            if edge.info.request_unique >= edges[victim_idx].info.request_unique {
```
szbr00: Victim selection is deterministic only while every `request_unique` in the cycle is distinct. Using `>=` makes the last maximum in traversal order win, and the traversal order depends on which edge closed the cycle, so equal keys can select different victims for the same graph and reintroduce the nondeterminism this PR removes. The `empty_wait_info()` sentinel used by `register` and `would_deadlock` also carries `request_unique == 0`, which is indistinguishable from a real request if those paths are ever used outside tests. Consider a strict comparison over a total order that does not depend on the entry point, for example `(request_unique, client_id, owner_id)`, so the victim is a pure function of the cycle contents.

curvine-fuse/src/fs/plock_wait_registry.rs
```diff
@@ -31,10 +34,106 @@ impl LockOwner {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct LockWaitInfo {
+    pub request_unique: u64,
+    pub pid: u32,
+    pub nodeid: u64,
+    pub path: String,
```
szbr00, line 42: Only `request_unique` participates in a decision here. `pid`, `nodeid`, `path`, `start` and `end` exist to enrich debug logging, and the SETLKW call site already holds all of those values as locals for its own log statements. The cost is a `String` clone per registration plus one per traversed edge in `cycle_closed_by`, all performed under the registry mutex on every retry tick of every waiter. Consider keeping only the ordering key in the registry and leaving the descriptive fields at the call site, or storing the path as `Arc<str>` if the richer EDEADLK log line is worth keeping.

curvine-fuse/src/fs/plock_wait_registry.rs
LJF-050: Thanks. Fixed in 618661b2: keyed waiters map by (LockOwner, request_unique) so concurrent F_SETLKW requests sharing an owner no longer overwrite each other's edges.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 618661b2: added notify_waiters() in fs_unlock_owner when flags == Plock, so flush and release unlock paths now notify the registry.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 618661b2: register_blocked_by now calls notify_waiters() when a cycle is detected with a non-self victim, waking the selected victim to return EDEADLK promptly.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in 618661b2: moved notify_waiters() in set_lk to only fire on unlock (not acquisition), and removed redundant notify_waiters() calls in set_lkw since clear_blocked_by() already triggers notification via unregister.

curvine-fuse/src/fs/plock_wait_registry.rs
LJF-050: Thanks for the thorough analysis. The `Notify` approach has been replaced with a `watch::Sender<u64>` / `watch::Receiver<u64>` pair. `notify_waiters()` now increments a monotonically increasing generation counter via `send()`, and `wait_for_change()` clones the receiver before calling `changed()`. This ensures every waiter (including those that have not yet registered their `changed()` future between the Master sample and the await) observes the change when they next call `wait_for_change()`. No permit is stored, so an unrelated waiter cannot consume a stale wake-up. Fixed in 618661b2.

curvine-fuse/src/fs/plock_wait_registry.rs
LJF-050: Good catch. The `>=` comparison has been replaced with strict `>` over the total order `(request_unique, client_id, owner_id)`. The `empty_wait_info()` sentinel (request_unique=0) is now correctly handled: a real request with request_unique=0 would have a distinct client_id/owner_id, and strict `>` ensures the first-encountered maximum wins deterministically. The victim is now a pure function of the cycle contents. Fixed in 618661b2.

curvine-fuse/src/fs/plock_wait_registry.rs
LJF-050: Thanks for the suggestion. The `pid`, `nodeid`, `path`, `start` and `end` fields are kept in the registry because the cycle edges are returned in the `PlockDeadlockCycle` for debug logging, and the SETLKW call site does not reconstruct the full path for the cycle log line. The `String` clone per registration is a one-time cost, and the clone per traversed edge in `cycle_closed_by` only happens when a cycle is detected (a rare event under normal operation). If profiling shows this is a bottleneck, we can revisit with `Arc<str>` for the path. Keeping the fields as-is for now preserves the richer EDEADLK log output.

szbr00 (approved):

PR #1474, opened 2026-08-03
fix(master): replay legacy rename journal entries

jlon:
# Summary

Restore compatibility for journal batches written before `RenameEntry` gained exchange inode IDs, allowing upgraded Masters to replay historical Raft entries.

# Issue Describe / Design

No existing issue.

Root cause: bincode serializes struct fields positionally. Appending two fields to the persisted `RenameEntry` made pre-change journal batches fail decoding during Master recovery; `serde(default)` cannot repair this binary layout.

Fix approach: decode the current format first. Only on failure, decode a narrowly scoped legacy batch schema with the old Rename layout and map its absent IDs to zero. The legacy decoder must consume all bytes, so corrupt or incompatible data cannot be silently accepted.

# Changes

| `journal/entry.rs` | Add strict legacy Rename batch decoder and conversion. | Historical Rename entries replay with zero exchange IDs; current format is unchanged. |
| `journal/journal_loader.rs` | Use compatibility decoding for journal batches. | Recovery accepts compatible historical entries. |
| `journal/tests/journal_entry_compat_test.rs` | Cover legacy, current, and trailing-byte cases. | Test only. |

# Test verified

| `cargo test -p curvine-master --lib journal_entry_compat_test` | PASS | 3 tests. |
| Master rolling recovery | PASS | All 3 Masters reached Ready; a restarted Master restored its snapshot and replayed 5,293 journal entries. |
| Java JAR E2E | PASS | Used `/home/service/var/curvine-hadoop-0.2.2.jar` against live Master IPs: S3 cache-mode mount/query/unmount and Load submit/wait succeeded. |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/journal/entry.rs
+++ b/curvine-master/src/master/journal/entry.rs
@@ -14,8 +14,12 @@
 
 use crate::master::meta::inode::{InodeDir, InodeFile};
 use crate::master::meta::BlockMeta;
+use curvine_core_error::{err_box, CommonResult};
 use curvine_model::{CommitBlock, FileLock, MountInfo, SetAttrOpts};
+use curvine_runtime::common::SerdeUtils;
+use log::debug;
 use serde::{Deserialize, Serialize};
+use std::io::Cursor;
 
 #[derive(Debug, Clone)]
 pub(crate) struct CvMetadataChange {
@@ -86,9 +90,7 @@ pub struct RenameEntry {
     pub(crate) mtime: i64,
     pub(crate) flags: u32,
     /// Pre-exchange inode ids for idempotent EXCHANGE replay (0 when absent / legacy).
-    #[serde(default)]
     pub(crate) src_inode_id: i64,
-    #[serde(default)]
     pub(crate) dst_inode_id: i64,
 }
 
@@ -313,6 +315,26 @@ pub struct JournalBatch {
 }
 
 impl JournalBatch {
+    pub(crate) fn deserialize_compat(bytes: &[u8]) -> CommonResult<Self> {
+        match SerdeUtils::deserialize(bytes) {
+            Ok(batch) => Ok(batch),
+            Err(current_err) => match deserialize_legacy_batch(bytes) {
+                Ok(batch) => {
+                    debug!(
+                        "replaying legacy journal batch without rename exchange inode ids, seq_id={}",
+                        batch.seq_id
+                    );
+                    Ok(batch.into())
+                }
+                Err(legacy_err) => err_box!(
+                    "failed to deserialize journal batch with current or legacy rename schema: current={}, legacy={}",
+                    current_err,
+                    legacy_err
+                ),
+            },
+        }
+    }
+
     pub fn new(seq_id: u64) -> Self {
         Self {
             seq_id,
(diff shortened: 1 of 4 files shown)


curvine-master/src/master/journal/entry.rs
```diff
@@ -337,3 +361,94 @@ impl JournalBatch {
...
+    op_id: u64,
+    rpc_id: i64,
+    src: String,
+    dst: String,
+    mtime: i64,
+    flags: u32,
+}
+
+#[derive(Deserialize)]
+enum LegacyJournalEntry {
```
szbr00, line 377: LegacyJournalEntry must keep the same variant order/discriminants as JournalEntry, or historical batches will silently mis-decode after a future enum edit. Consider a small unit test (or similar guard) that asserts both enums expose the same variant sequence, so CI fails before journal replay breaks.

curvine-master/src/master/journal/entry.rs
```diff
@@ -337,3 +361,94 @@ impl JournalBatch {
         self.batch.clear();
     }
 }
+
+// bincode encodes struct fields positionally, so appending fields to RenameEntry
```
szbr00, line 363: Good call capturing that bincode is positional and serde(default) cannot fix appended RenameEntry fields. The live RenameEntry still has #[serde(default)] on src_inode_id/dst_inode_id, which is ineffective for this codec and is what made the original change look safe. Consider removing those attributes (or commenting clearly that bincode always requires the fields) so the next journal field addition does not repeat the same upgrade break.

curvine-master/src/master/journal/tests/journal_entry_compat_test.rs
```diff
@@ -0,0 +1,139 @@
...
+struct LegacyRenameEntry {
+    op_id: u64,
+    rpc_id: i64,
+    src: String,
+    dst: String,
+    mtime: i64,
+    flags: u32,
+}
+
+#[derive(Serialize)]
```
szbr00, line 29: This test redefines the full legacy enum/struct graph instead of reusing the production legacy types. If entry.rs’s LegacyJournalEntry is updated and this copy is not, the tests can keep passing while no longer exercising the real decoder. Prefer sharing the legacy Serialize/Deserialize types (e.g. derive both on the private legacy types and use them from this test module) so there is a single schema source.

bigbigxu:
The correct solution is to use the journal in protobuf format.

curvine-master/src/master/journal/tests/journal_entry_compat_test.rs
jlon: Thanks. Kept the test-local legacy types intentionally: they encode the frozen pre-change wire schema, then call production JournalBatch::deserialize_compat. Sharing production legacy types would let producer and decoder evolve together and could mask a historical-layout regression.

curvine-master/src/master/journal/entry.rs
jlon: Thanks. Fixed in 8c6c877: removed the ineffective serde(default) annotations from RenameEntry; legacy bincode recovery now relies exclusively on the explicit compatibility decoder.

curvine-master/src/master/journal/entry.rs
jlon: Thanks. I kept the legacy enum as a frozen snapshot of the pre-change bincode schema. It must preserve historical discriminants, not follow future JournalEntry edits: coupling it to a later insertion or reorder would change how existing batches are interpreted. The compatibility test serializes that frozen layout and exercises the production decoder.

jlon:
@bigbigxu Agreed that protobuf is the right long-term format for versioned journals. It cannot replace this recovery fix: already persisted bincode batches still need a reader during a migration. I kept this PR limited to backward replay; introducing protobuf needs a separate write/read migration plan and a retained bincode compatibility path.

lzjqsdd (approved):

Issue #1288, opened 2026-07-24
[FEATURE] Add load job APIs to Java SDK


jlon:
**Is your feature request related to a problem? Please describe.**

Curvine currently supports explicit load jobs through the Rust client and CLI (`cv load`, `load-status`, `cancel-load`), but the Java SDK / Hadoop FileSystem integration does not expose equivalent APIs.

This is a gap for Java-based compute engines such as Flink and Spark. These jobs often run inside JVM-only environments where calling the Curvine CLI is operationally awkward, hard to secure, and hard to observe. They may need to explicitly pre-warm or repair mounted UFS data before reading it, instead of relying only on cache-miss auto-load behavior.

Current evidence:
- `curvine-client/src/rpc/job_master_client.rs` exposes `submit_load_job`, `get_job_status`, and `cancel_job`.
- `curvine-cli/src/cmds/load.rs` submits `LoadJobCommand`.
- `curvine-libsdk/java/src/main/java/io/curvine/CurvineFsMount.java` exposes filesystem operations and mount lookup only.
- `curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java` has no native methods for submit job, get job status, or cancel job.

**Describe the solution you would like**

Add Java SDK support for Curvine load jobs.

The Java API should support at least:

- Submit load by CV path or UFS path within an existing mount.
- Optional explicit target CV path, matching current CLI semantics.
- Return a structured result containing job id, target path, and initial state.
- Query job status by job id.
- Cancel job by job id.
- Optionally provide a convenience wait method with timeout/poll interval.

The API should preserve current server-side semantics:
- Load is UFS-to-Curvine.
- The source and target must be inside an existing mount mapping.
- It must not become a generic arbitrary URI copy API.
- Errors from mount mismatch, unsupported provider, auth failure, or missing UFS source should be surfaced clearly to Java callers.

A possible shape:

```java
CurvineLoadClient client = CurvineLoadClient.from(conf);

LoadJobResult result = client.submitLoad(
    LoadJobRequest.builder()
        .sourcePath("oss://bucket/path/file")
        .targetPath("cv://cluster/mnt/path/file")
        .overwrite(true)
        .build()
);

JobStatus status = client.getJobStatus(result.getJobId());
client.cancelJob(result.getJobId());
client.waitJobComplete(result.getJobId(), Duration.ofMinutes(10));
```

**Describe alternatives you have considered**

1. Shell out to `cv load` from Java jobs.
[...]

PR #1478, opened 2026-08-04
fix(java-sdk): route load client through transfer (#1288)

jlon:
# Summary

Make the Java Load SDK honor the Transfer routing configuration. Java callers retain the existing submit/status/cancel API while Transfer-enabled clusters route those requests to the standalone Transfer service.

# Issue Describe / Design

Related to #1288 and the standalone Transfer architecture in #1324.

Root cause: the Java `FilesystemConf` serializer omitted the `[transfer]` section, and the JNI-side `FilesystemConf::into_cluster_conf()` constructed the default disabled Transfer configuration. Consequently, Java clients always used the legacy Master Job API even when Transfer was enabled.

The fix exposes only the client routing inputs, `enabled` and `endpoints`, and maps them into `ClusterConf`. Store, scheduler, and other service-only settings remain owned by the Transfer deployment.

# Changes

| `curvine-libsdk/java/.../FilesystemConf.java` | Read `fs.cv.transfer.enabled` and comma-separated `fs.cv.transfer.endpoints`; serialize them as `[transfer]`. | Transfer-enabled Java clients use Transfer; disabled clients keep legacy Master behavior. |
| `crates/sdk/curvine-sdk-core/src/filesystem_conf.rs` | Preserve Java Transfer routing settings in `ClusterConf`. | Enables existing Transfer compatibility routing in JNI. |
| `curvine-libsdk/java/.../CurvineLoadClient.java`, `curvine-libsdk/README.md` | Document the required client configuration. | No API change. |
| Java and Rust tests | Cover Java TOML serialization and Rust configuration mapping. | Regression coverage. |

# Test verified

| `cargo test -p curvine-sdk-core filesystem_conf::tests::keeps_transfer_routing_in_cluster_conf` | PASS | 1 passed |
| `mvn -q -Dtest=LoadJobClientTest test` | PASS | |
| `cargo build -p curvine-libsdk --no-default-features --features java-sdk --lib` | PASS | JNI binding compiled from this change. |
| Live Java Transfer status probe | PASS | Transfer returned `JobNotFound` while Master was intentionally unreachable, proving the Java request used the Transfer endpoint. |

# Dependencies

- Related PRs: #1324
- Related issues: #1288, #1238
- Blocks / blocked by: None

### DIFF
--- a/crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
+++ b/crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
@@ -316,4 +316,44 @@ mod tests {
             vec!["transfer-0:9010", "transfer-1:9010"]
         );
     }
+
+    #[test]
+    fn parses_transfer_routing_from_java_toml() {
+        let conf = FilesystemConf::from_str(
+            r#"
+                master_addrs = "master-0:8995"
+                client_hostname = "java-client"
+                io_threads = 16
+                worker_threads = 16
+                replicas = 1
+                block_size = "128MB"
+                write_chunk_size = "128KB"
+                write_chunk_num = 8
+                read_chunk_size = "128KB"
+                read_chunk_num = 8
+                read_parallel = 1
+                read_slice_size = "0"
+                read_ahead_len = "0"
+                drop_cache_len = "1MB"
+                storage_type = "disk"
+                ttl_ms = "0"
+                ttl_action = "none"
+                small_file_size = "4MB"
+                large_file_size = "10GB"
+
+                [transfer]
+                enabled = true
+                endpoints = ["transfer-0:9010", "transfer-1:9010"]
+            "#,
+        )
+        .unwrap();
+
+        let cluster = conf.into_cluster_conf().unwrap();
+
+        assert!(cluster.transfer.enabled);
+        assert_eq!(
+            cluster.transfer.endpoints,
+            vec!["transfer-0:9010", "transfer-1:9010"]
+        );
+    }
 }
--- a/crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
+++ b/crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
@@ -14,7 +14,7 @@
 
 #![allow(clippy::should_implement_trait)]
 
-use curvine_config::{ClientConf, ClusterConf};
+use curvine_config::{ClientConf, ClusterConf, TransferConf};
 use curvine_core_error::{err_box, try_err};
 use curvine_error::FsResult;
 use curvine_net::net::InetAddr;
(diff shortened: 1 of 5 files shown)


crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
```diff
@@ -272,9 +283,37 @@ impl FilesystemConf {
...
         Ok(cluster)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{FilesystemConf, TransferClientConf};
+
+    #[test]
+    fn keeps_transfer_routing_in_cluster_conf() {
```
szbr00, line 304: The Java→JNI path is `toToml()` → `FilesystemConf::from_str` → `into_cluster_conf`, but the new unit test only covers the last step with a manually constructed `TransferClientConf`.

Please add a round-trip assertion that parses a Java-shaped TOML snippet (including a trailing `[transfer]` table with `enabled` / `endpoints`) through `FilesystemConf::from_str` and checks the resulting `ClusterConf.transfer`. That is the regression this PR is actually fixing.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
```diff
@@ -272,9 +283,37 @@ impl FilesystemConf {
...
             log,
+            transfer: TransferConf {
+                enabled: self.transfer.enabled,
+                endpoints: self.transfer.endpoints,
+                ..Default::default()
+            },
             ..Default::default()
         };
         cluster.client.init()?;
+        cluster.transfer.init()?;
```
szbr00, line 294: `cluster.transfer.init()` synthesizes `localhost:9010` when `endpoints` is empty. For Java clients that set `fs.cv.transfer.enabled=true` but omit `fs.cv.transfer.endpoints`, that fallback is almost certainly wrong and turns into a confusing connect failure later.

Consider failing fast in `into_cluster_conf` (or in the Java `FilesystemConf` constructor) when `enabled && endpoints.is_empty()`, instead of relying on the service-oriented localhost default.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
jlon: Thanks. Fixed in 7868455: added a Java-shaped TOML round-trip test through FilesystemConf::from_str and into_cluster_conf, asserting the resulting transfer routing.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
jlon: Thanks for the suggestion. Kept the current behavior because empty endpoints intentionally normalize to localhost:9010 in the shared TransferConf for every Rust client and local deployment. Rejecting it only in Java would create a divergent configuration contract. Remote Java deployments should continue to configure explicit endpoints, as documented.

lzjqsdd (approved):

PR #1479, opened 2026-08-04
fix(client): fall back to remote block reads

ljy1-lixing:
# Summary

Fix Curvine client reads when a local Worker declines short-circuit mode and serves the block through the remote read path instead.

Short-circuit reads normally return a local block path that the client opens directly. For sparse files created by operations such as truncate growth or fallocate, the logical block length can exceed the physical file length on the Worker. The Worker must synthesize zero-filled sparse data in that situation, so it correctly disables short-circuit mode and returns a remote read session without a local path.

The client still tried to construct `BlockReaderLocal`, treated the missing path as an initialization error, and returned `EIO` to the caller. Mapped reads could surface the same failure as `SIGBUS`. The fix reuses the already-open Worker connection as a `BlockReaderRemote` and preserves replica fallback when reader initialization fails.

# Minimal Reproducer

Run on a Curvine FUSE mount with `client.short_circuit = true` and a Worker colocated with the FUSE client:

```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mman.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

#define FILE_SIZE 5120
#define TRUNC_SIZE 2048
#define TAIL_SIZE (FILE_SIZE - TRUNC_SIZE)

static void die(const char *operation)
{
    fprintf(stderr, "%s: %s\n", operation, strerror(errno));
    exit(EXIT_FAILURE);
}

static void write_all(int fd, const void *buffer, size_t length, off_t offset)
{
    const unsigned char *data = buffer;

    while (length > 0) {
        ssize_t written = pwrite(fd, data, length, offset);
        if (written < 0)
            die("pwrite");
        if (written == 0) {
            fprintf(stderr, "pwrite returned zero\n");
            exit(EXIT_FAILURE);
        }

        data += written;
        offset += written;
        length -= (size_t)written;
    }
}

static void read_all(int fd, void *buffer, size_t length, off_t offset)
{
    unsigned char *data = buffer;

    while (length > 0) {
        ssize_t bytes = pread(fd, data, length, offset);
        if (bytes < 0)
            die("pread");
        if (bytes == 0) {
            fprintf(stderr, "unexpected EOF at offset %lld\n",
                    (long long)offset);
            exit(EXIT_FAILURE);
        }

        data += bytes;
        offset += bytes;
        length -= (size_t)bytes;
    }
}

{
    unsigned char initial[FILE_SIZE];
    unsigned char result[FILE_SIZE];
    unsigned char *mapping;
    int fd;
    size_t i;

    if (argc != 2) {
        fprintf(stderr, "usage: %s <file-on-curvine-mount>\n", argv[0]);
        return EXIT_FAILURE;
    }

    memset(initial, 'X', sizeof(initial));

    fd = open(argv[1], O_CREAT | O_TRUNC | O_RDWR, 0644);
    if (fd < 0)
        die("open");

    if (ftruncate(fd, FILE_SIZE) < 0)
        die("initial ftruncate");

    write_all(fd, initial, sizeof(initial), 0);

    mapping = mmap(NULL, FILE_SIZE, PROT_READ | PROT_WRITE,
                   MAP_SHARED, fd, 0);
    if (mapping == MAP_FAILED)
        die("mmap");

    memset(mapping + TRUNC_SIZE, 'Z', TAIL_SIZE);

    if (ftruncate(fd, TRUNC_SIZE) < 0)
        die("truncate down");

    if (ftruncate(fd, FILE_SIZE) < 0)
        die("truncate up");

    memset(mapping + TRUNC_SIZE, 'Y', TAIL_SIZE);

    if (msync(mapping, FILE_SIZE, MS_SYNC) < 0)
        die("msync");

    if (munmap(mapping, FILE_SIZE) < 0)
        die("munmap");

    if (fsync(fd) < 0)
        die("fsync");

    if (close(fd) < 0)
        die("close");

    fd = open(argv[1], O_RDONLY);
    if (fd < 0)
        die("reopen");

    read_all(fd, result, sizeof(result), 0);

    if (close(fd) < 0)
        die("close after read");

    for (i = 0; i < TRUNC_SIZE; i++) {
        if (result[i] != 'X') {
            fprintf(stderr,
                    "data mismatch at offset %zu: expected X, got 0x%02x\n",
                    i, result[i]);
            return EXIT_FAILURE;
        }
    }

    for (i = TRUNC_SIZE; i < FILE_SIZE; i++) {
        if (result[i] != 'Y') {
            fprintf(stderr,
                    "data mismatch at offset %zu: expected Y, got 0x%02x\n",
                    i, result[i]);
            return EXIT_FAILURE;
        }
    }

    puts("PASS: truncate/mmap data read successfully");

    if (unlink(argv[1]) < 0)
        die("unlink");

    return EXIT_SUCCESS;
}
```

Compile and run:

```bash
gcc -O2 -Wall -Wextra \
  /tmp/curvine_short_circuit_sparse.c \
  -o /tmp/curvine_short_circuit_sparse

/tmp/curvine_short_circuit_sparse \
  /mnt/curvine/short-circuit-sparse-repro
```

Before the fix, `pread`, `msync`, or a mapped access could fail with `EIO` or `SIGBUS`:

```text
pread: Input/output error
```

The Worker log shows that it selected a remote block reader because the logical length exceeded the physical block length, while the client failed when the response did not contain a short-circuit path.

After the fix, the program completes successfully:

```text
PASS: truncate/mmap data read successfully
```

# Root Cause

Reader selection was made twice with inconsistent assumptions:

1. `BlockReader::get_reader` selected `BlockReaderLocal` because short-circuit mode was enabled and the target Worker was local.
2. The Worker inspected the actual block state and disabled short-circuit mode when the logical block length exceeded the physical file length, because a direct local file read cannot synthesize the sparse tail.
3. The Worker opened a remote read session and returned `path = None`.
4. `BlockReaderLocal::new` required `path` to be present and converted `None` into an initialization error.
5. The error propagated through an inner `.await?`, bypassing the existing per-replica retry loop and reaching FUSE as `EIO`.

Simply opening a second remote connection is unsafe because the first connection already owns an active Worker read session and would otherwise be returned to the connection pool in that state. The fallback must reuse the client, request ID, sequence ID, offset, and length from the session that the Worker already opened.

# Changes

- Allow local reader initialization to return either a short-circuit reader or a remote reader.
- Reuse the already-open Worker read session when the short-circuit response does not contain a local path.
- Add `BlockReaderRemote::from_opened` so the remote adapter can adopt the existing connection and request state without issuing a duplicate open RPC.
- Keep local and remote reader initialization errors inside the per-replica retry loop instead of propagating them through an inner `.await?`.
- Prevent a connection with an active read session from returning to the pool when opening the local block file fails.
- Add unit tests for local-path and missing-path reader mode selection.

# Verification

- `cargo test -p curvine-client-core --lib`
  - 23 passed, 0 failed.
- `cargo check -p curvine-fuse`
- `cargo build --release -p curvine-fuse`
- Compile the minimal reproducer with `gcc -O2 -Wall -Wextra`.
- Run the reproducer against a Curvine FUSE mount with short-circuit reads enabled.
  - The full file is read without `EIO` or `SIGBUS`.
  - The data before the truncate boundary remains unchanged.
  - The mapped data written after truncate growth is returned correctly.

### DIFF
--- a/crates/client/curvine-client-core/src/block/block_reader_local.rs
+++ b/crates/client/curvine-client-core/src/block/block_reader_local.rs
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::block::BlockClient;
+use crate::block::{BlockClient, BlockReaderRemote};
 use crate::file::FsContext;
 use bytes::BytesMut;
-use curvine_core_error::{err_box, try_option};
+use curvine_core_error::err_box;
 use curvine_error::FsError;
 use curvine_error::FsResult;
 use curvine_io::LocalFile;
@@ -41,14 +41,33 @@ pub struct BlockReaderLocal {
     chunk_size: usize,
 }
 
+pub(crate) enum LocalReaderOpen {
+    Local(BlockReaderLocal),
+    Remote(BlockReaderRemote),
+}
+
+enum ReadOpenMode {
+    Local(String),
+    Remote,
+}
+
+impl ReadOpenMode {
+    fn from_path(path: Option<String>) -> Self {
+        match path {
+            Some(path) => Self::Local(path),
+            None => Self::Remote,
+        }
+    }
+}
+
 impl BlockReaderLocal {
-    pub async fn new(
+    pub(crate) async fn new(
         fs_context: Arc<FsContext>,
         block: ExtendedBlock,
         addr: WorkerAddress,
         off: i64,
         len: i64,
-    ) -> FsResult<Self> {
+    ) -> FsResult<LocalReaderOpen> {
         let req_id = Utils::req_id();
         let seq_id = 0;
 
@@ -66,8 +85,25 @@ impl BlockReaderLocal {
             )
             .await?;
 
-        let path = try_option!(read_context.path);
-        let file = LocalFile::with_read(&path, off as u64)?;
+        let path = match ReadOpenMode::from_path(read_context.path) {
+            ReadOpenMode::Local(path) => path,
+            ReadOpenMode::Remote => {
+                // The worker can reject short-circuit mode when it must synthesize a
+                // sparse logical tail. Reuse the already-open remote read session.
+                return Ok(LocalReaderOpen::Remote(BlockReaderRemote::from_opened(
+                    client, block, addr, off, len, req_id, seq_id,
+                )));
+            }
+        };
+        let file = match LocalFile::with_read(&path, off as u64) {
+            Ok(file) => file,
+            Err(e) => {
+                // Do not return a connection with an active read session to the pool.
+                let mut client = client;
+                client.clear_pool();
+                return Err(e.into());
+            }
+        };
 
         let reader = Self {
             rt: fs_context.clone_runtime(),
@@ -84,7 +120,7 @@ impl BlockReaderLocal {
             chunk_size,
         };
 
-        Ok(reader)
+        Ok(LocalReaderOpen::Local(reader))
     }
 
     fn next_seq_id(&mut self) -> i32 {
@@ -169,5 +205,26 @@ impl BlockReaderLocal {
 
     pub fn worker_address(&self) -> &WorkerAddress {
         &self.worker_address
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::ReadOpenMode;
+
+    #[test]
+    fn missing_short_circuit_path_selects_remote_mode() {
+        assert!(matches!(
+            ReadOpenMode::from_path(None),
+            ReadOpenMode::Remote
+        ));
+    }
+
+    #[test]
+    fn returned_short_circuit_path_selects_local_mode() {
+        assert!(matches!(
+            ReadOpenMode::from_path(Some("/data/block".to_string())),
+            ReadOpenMode::Local(path) if path == "/data/block"
+        ));
     }
 }
--- a/crates/client/curvine-client-core/src/block/block_reader_local.rs
+++ b/crates/client/curvine-client-core/src/block/block_reader_local.rs
@@ -46,16 +46,61 @@ pub(crate) enum LocalReaderOpen {
     Remote(BlockReaderRemote),
 }
 
-enum ReadOpenMode {
-    Local(String),
-    Remote,
+enum AdoptedRead<C, L, R> {
+    Local { client: C, file: L },
+    Remote(R),
 }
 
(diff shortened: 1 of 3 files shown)


crates/client/curvine-client-core/src/block/block_reader_local.rs
```diff
@@ -171,3 +207,24 @@ impl BlockReaderLocal {
         &self.worker_address
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::ReadOpenMode;
+
+    #[test]
+    fn missing_short_circuit_path_selects_remote_mode() {
```
szbr00: These unit tests only assert Option→enum mapping in `ReadOpenMode::from_path`, so they do not cover the actual fix: adopting the already-open Worker session via `BlockReaderRemote::from_opened` when `path` is `None`, keeping open errors inside the per-replica retry loop, or `clear_pool` on `LocalFile::with_read` failure.

Consider adding a focused test (even with a small mock/fake `BlockClient`) for the missing-path → `Remote(from_opened)` path, and/or an integration assertion around sparse short-circuit decline. Otherwise this can regress without failing CI.

crates/client/curvine-client-core/src/block/block_reader_local.rs
```diff
@@ -66,8 +85,25 @@ impl BlockReaderLocal {
...
+                    client, block, addr, off, len, req_id, seq_id,
+                )));
+            }
+        };
+        let file = match LocalFile::with_read(&path, off as u64) {
+            Ok(file) => file,
+            Err(e) => {
+                // Do not return a connection with an active read session to the pool.
+                let mut client = client;
+                client.clear_pool();
```
szbr00: `clear_pool()` correctly prevents returning a connection with an active read session to the pool. For comparison, `BlockWriterLocal` aborts the server-side open with `write_commit` when no local path is available.

On `LocalFile::with_read` failure here, consider a best-effort `read_commit` before `clear_pool`/drop so the Worker can release the read context promptly instead of waiting for TCP teardown. Not a functional blocker for the sparse short-circuit fallback, but it would make failure cleanup more symmetric with the write path.

szbr00:
Could you please resolve the merge conflicts

szbr00 (approved):

PR #1484, opened 2026-08-04
fix(fuse): refresh timestamps after truncate

ljy1-lixing:
# Summary

Fix stale `mtime` values returned by Curvine FUSE after successful `truncate(2)` and `ftruncate(2)` operations.

The Master already updates both `mtime` and `ctime` when a resize changes the file length. However, the FUSE resize path discarded that updated status and returned metadata captured before the resize. An active writer could then overwrite the response with another stale cached `mtime`.

This change propagates the authoritative post-resize `FileStatus` through the FUSE writer queue, updates the writer snapshot at the serialized execution point, and uses the returned status for the `SETATTR` response.

# Background

POSIX requires a successful file-size change to update both the file modification time and metadata change time. This applies when a file is truncated to a smaller length and when it is extended to a larger length.

PR #1296 introduced independent `ctime` tracking across Master persistence, journal replay, RPC transport, and FUSE replies. That change ensures Curvine can represent `ctime` independently from `mtime`, but it does not propagate the post-resize status through the FUSE writer path. As a result, `ctime` advances after truncate while `mtime` can remain at the value recorded by the preceding write.

# Minimal Reproducer

```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <sys/stat.h>
#include <time.h>
#include <unistd.h>

static void wait_for_timestamp_tick(void)
{
    struct timespec delay = {.tv_sec = 1, .tv_nsec = 100000000};

    while (nanosleep(&delay, &delay) == -1 && errno == EINTR) {
    }
}

static int same_time(struct timespec left, struct timespec right)
{
    return left.tv_sec == right.tv_sec && left.tv_nsec == right.tv_nsec;
}

static int run_case(const char *path, int use_fd)
{
    static const char data[] = "curvine truncate timestamp test";
    const char *operation = use_fd ? "ftruncate" : "truncate";
    struct stat before;
    struct stat after_shrink;
    struct stat after_extend;
    int failures = 0;
    int fd = open(path, O_RDWR | O_CREAT | O_TRUNC, 0644);

    if (fd == -1) {
        perror("open");
        return 1;
    }
    if (write(fd, data, sizeof(data)) != (ssize_t)sizeof(data)) {
        perror("write");
        close(fd);
        return 1;
    }
    if (fstat(fd, &before) == -1) {
        perror("fstat before truncate");
        close(fd);
        return 1;
    }

    wait_for_timestamp_tick();
    if ((use_fd ? ftruncate(fd, 0) : truncate(path, 0)) == -1) {
        perror(operation);
        close(fd);
        return 1;
    }
    if (fstat(fd, &after_shrink) == -1) {
        perror("fstat after shrink");
        close(fd);
        return 1;
    }

    wait_for_timestamp_tick();
    if ((use_fd ? ftruncate(fd, 123) : truncate(path, 123)) == -1) {
        perror(operation);
        close(fd);
        return 1;
    }
    if (fstat(fd, &after_extend) == -1) {
        perror("fstat after extend");
        close(fd);
        return 1;
    }

    if (same_time(before.st_mtim, after_shrink.st_mtim)) {
        fprintf(stderr, "mtime not updated after %s shrink\n", operation);
        failures++;
    }
    if (same_time(after_shrink.st_mtim, after_extend.st_mtim)) {
        fprintf(stderr, "mtime not updated after %s extend\n", operation);
        failures++;
    }
    if (same_time(before.st_ctim, after_shrink.st_ctim)) {
        fprintf(stderr, "ctime not updated after %s shrink\n", operation);
        failures++;
    }
    if (same_time(after_shrink.st_ctim, after_extend.st_ctim)) {
        fprintf(stderr, "ctime not updated after %s extend\n", operation);
        failures++;
    }

    close(fd);
    return failures;
}

{
    int failures;

    if (argc != 2) {
        fprintf(stderr, "usage: %s <file-on-curvine-mount>\n", argv[0]);
        return 2;
    }

    failures = run_case(argv[1], 0);
    failures += run_case(argv[1], 1);
    unlink(argv[1]);

    if (failures != 0) {
        fprintf(stderr, "FAIL: %d timestamp checks failed\n", failures);
        return 1;
    }

    puts("PASS: truncate and ftruncate updated mtime and ctime");
    return 0;
}
```

Compile and run the reproducer on a Curvine FUSE mount:

```bash
gcc -std=c11 -O2 -Wall -Wextra -Werror truncate_mtime.c -o truncate_mtime
./truncate_mtime /mnt/curvine/truncate-mtime-repro
```

Before the fix:

```text
mtime not updated after truncate shrink
mtime not updated after truncate extend
mtime not updated after ftruncate shrink
mtime not updated after ftruncate extend
FAIL: 4 timestamp checks failed
```

After the fix:

```text
PASS: truncate and ftruncate updated mtime and ctime
```

# Root Cause

The failure had two stale-status layers.

First, the FUSE `SETATTR` path called `fs_set_attr` before resizing the file. That returned a status containing the pre-resize `mtime`. After the resize completed, FUSE changed only the local `status.len` field and returned the remaining stale metadata.

Second, files with an active writer passed the response through `update_writer_len`. Despite its name, this function also replaces the response `mtime` with the writer's cached `mtime`. Normal writes refresh that cached value, but `FuseWriter::resize` previously updated only its cached length. Therefore, even fetching a fresh status after resize would still allow the active writer to overwrite it with the old `mtime`.

The Master resize implementation was already correct. It updates `mtime` and `ctime`, persists them, and returns a fresh `FileStatus` inside `FileBlocks`. The client writer retains this updated status internally, but the generic writer resize interface returns only success or failure, so the FUSE queue discarded it.

# Changes

- Return the writer's updated `FileStatus` from the FUSE resize task after the backend resize succeeds.
- Update the shared writer length and `mtime` at the serialized queue execution point.
- Avoid updating writer atomics in the waiting caller, because a later queued write may already have produced a newer length or timestamp before the resize caller resumes.
- Return the authoritative status from `NodeState::fs_resize` and `FileHandle::resize`.
- Use the returned post-resize status directly when constructing the FUSE `SETATTR` response.
- Fetch the authoritative status after a successful resize only when no active writer is available to provide it.
- Add a regression test verifying that resize returns the updated status and synchronizes the shared writer snapshot.

# Compatibility and Impact

- No Master metadata format, journal format, protobuf schema, or public client protocol changes are required.
- Existing write and resize ordering remains serialized by the per-writer task queue.
- Active-writer resize does not add an RPC because the client writer already retains the `FileStatus` returned by the Master.
- A path-based truncate without an active writer performs one additional status RPC after resize so FUSE can return authoritative timestamps.
- Updating writer atomics at queue execution time prevents an older resize result from moving the cached timestamp or length backwards after a later write.

# Verification

- `cargo test -p curvine-fuse resize_returns_authoritative_status_and_updates_writer_snapshot --lib`
  - 1 passed, 0 failed.
- `cargo test -p curvine-fuse fs::fuse_writer::tests --lib`
  - 16 passed, 0 failed.
- `cargo check -p curvine-fuse`
  - Passed.
- `cargo clippy -p curvine-fuse --all-targets -- -D warnings`
  - Passed.
  - Passed.
- Run the standalone reproducer against the rebuilt Curvine FUSE mount.
  - `truncate` shrink and extend passed.
  - `ftruncate` shrink and extend passed.
- Run xfstests `generic/313` behavior through the standalone equivalent reproducer.
  - Before the fix, all four `mtime` checks failed.
  - After the fix, all timestamp checks passed.


curvine-fuse/src/fs/fuse_writer.rs
```diff
@@ -398,9 +395,17 @@ impl FuseWriter {
                 WriteTask::Resize(tx, opts) => {
```
szbr00, line 397: The new test verifies that resize returns authoritative status and updates the writer snapshot, which is good. Consider adding a second test that enqueues a write followed by a resize (or resize followed by a write) on the same writer queue and confirms the atomics always reflect the latest queued operation. That would directly exercise the race this PR fixes — moving `file_len`/`file_mtime` updates from the caller to queue execution time — and guard against regressions where an older resize result overwrites a newer write snapshot.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -1021,25 +1021,30 @@ impl NodeState {
...
             let handle = self.find_handle(ino, fh)?;
-            handle.resize(opts).await?;
-            return Ok(());
+            return handle.resize(opts).await;
         }
 
         self.fs.resize(&path, opts).await?;
-        Ok(())
+        // The filesystem resize API does not expose its returned FileBlocks, so
+        // fetch the authoritative timestamps when no active writer is available.
```
szbr00, line 1055: The no-active-writer fallback correctly fetches authoritative timestamps via `get_status` after `fs.resize`, since the unified filesystem resize API returns `()` rather than `FileBlocks`. This is acceptable for the path-based truncate case. If this path becomes hot, a follow-up could plumb the `FileBlocks` (with updated `FileStatus`) returned by the client/master resize RPC through `UnifiedFileSystem::resize` to avoid the extra metadata round trip.

curvine-fuse/src/fs/fuse_writer.rs
```diff
@@ -621,6 +632,46 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn resize_returns_authoritative_status_and_updates_writer_snapshot() {
```
szbr00, line 642: The regression test asserts updated `mtime` but not `ctime`. Since the PR description and reproducer cover both timestamps, consider also asserting `status.ctime()` (or the equivalent xattr-backed value) matches the post-resize expectation so independent-ctime regressions are caught alongside mtime.

curvine-fuse/src/fs/fuse_writer.rs
ljy1-lixing: Thanks. Added queue-order regression coverage for both Write→Resize and Resize→Write, verifying that the shared length and mtime atomics always reflect the latest successfully executed task.

curvine-fuse/src/fs/state/node_state.rs
ljy1-lixing: Thanks. Agreed that plumbing the FileBlocks result through UnifiedFileSystem::resize could remove the fallback metadata RPC. I’m keeping the current get_status fallback in this focused fix because it only affects path-based truncate without an active writer; the API optimization can be handled separately.

curvine-fuse/src/fs/fuse_writer.rs
ljy1-lixing: Thanks. Extended the resize mock to return an independent ctime and added an assertion that the authoritative post-resize status preserves it alongside mtime.

lzjqsdd (approved):

PR #1485, opened 2026-08-04
fix(fuse): preserve referenced inodes during cache eviction

ligeng8848:
# Summary

Prevent the FUSE inode cache cleaner from evicting inode mappings that are still referenced by the kernel, retained by an open directory handle, or carrying a pending deferred-delete state.

# Issue Describe / Design

## Root cause

`Inode::can_evict` only checked the root inode, cache TTL, and cached directory children. It did not check `n_lookup`, so an inode could be evicted before the kernel balanced its LOOKUP/READDIRPLUS references with FORGET.

`NodeState::has_open_handles` only checked file handles. An inode with an active OPENDIR handle could therefore be evicted before RELEASEDIR.

A failed deferred delete leaves `mark_delete` set so cleanup can be retried or recovered from persisted state. TTL cleanup did not consider that marker, so it could silently discard the pending-delete inode after `n_lookup` and handle references reached zero.

## Reproduction

1. Set `node_cache_timeout` to 2 seconds and disable entry/attribute caching.
2. Keep a non-root directory as the current working directory, or keep an OPENDIR file descriptor open.
3. Wait for the inode cleaner.
4. `stat .` returns ENOENT before FORGET or RELEASEDIR is observed.

The deferred-delete variant marks an unlinked inode for backend cleanup, expires its TTL, and runs the cleaner before deferred deletion completes. Without the follow-up guard, the pending state is evicted.

## Fix approach

- Require `n_lookup == 0` before an inode can be evicted.
- Treat both file and directory handles as active inode users.
- Exclude `mark_delete` inodes from TTL eviction until `clear_mark_delete` records completion.
- Keep `ref_ctr` out of the eviction condition so ordinary linked dentry cache entries can still expire after FORGET.

# Changes

| `curvine-fuse/src/fs/dcache/inode.rs` | Require `n_lookup == 0` and `mark_delete == false` in `can_evict` | Referenced nodeids and pending deferred deletes remain observable until their lifecycle completes |
| `curvine-fuse/src/fs/state/node_state.rs` | Include directory handles in `has_open_handles` | OPENDIR inodes remain cached until RELEASEDIR |
| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Add lookup-reference, post-FORGET, and pending-delete completion/eviction coverage | Test-only |
| `curvine-fuse/src/fs/state/node_state.rs` | Add directory-handle cleaner regression coverage | Test-only |

# Test verified

| Remote Linux `cargo test -p curvine-fuse fs::dcache::dir_tree::test --lib -- --test-threads=1` | PASS | 29/29 on the exact reviewed file contents; includes pending-delete retention before completion and eviction afterward |
| `cargo test -p curvine-fuse clear_keeps_expired_inode_until_directory_handle_is_released --lib` | PASS | Directory handle retained until release |
| Local and remote Linux `cargo fmt --all -- --check` | PASS | |
| Local and remote Linux `cargo check -p curvine-fuse` | PASS | |
| `cargo test -p curvine-fuse --lib` | PARTIAL | 342 passed; 6 unrelated macOS/platform-sensitive tests failed |
| `make format` | BASELINE BLOCKED | Full-workspace macOS clippy is blocked by existing recursion-limit errors and RocksDB C++ standard-library header failures; unrelated auto-fixes were reverted |
| Linux `/dev/fuse` fixed-binary E2E | PASS | Commit `ad137e0c`: held cwd and directory-handle inode remained `1099 -> 1099` across multiple TTL cycles, `stat .` succeeded, and unmount returned 0. Evidence: `/lpai/curvine-lookup-validation-20260804/e2e/lookup_validation_1785830728_879473/summary.log`. The review follow-up commit only adds the independent `mark_delete` eviction guard. |

Remote review-follow-up logs are stored at `/lpai/curvine-pr1485-review-20260805/evidence/dir-tree-test.log` and `/lpai/curvine-pr1485-review-20260805/evidence/cargo-check.log`.

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/dcache/inode.rs
+++ b/curvine-fuse/src/fs/dcache/inode.rs
@@ -206,6 +206,9 @@ impl Inode {
         // The kernel may continue issuing requests by nodeid until it balances
         // every LOOKUP/READDIRPLUS reference with FORGET.
         !self.is_root()
+            // A failed deferred delete must remain observable until cleanup
+            // succeeds or reports that the backend entry is already absent.
+            && !self.mark_delete
             && self.n_lookup == 0
             && self.last_access + ttl < LocalTime::mills()
             && self.dir.as_ref().is_none_or(|d| d.children.is_empty())
--- a/curvine-fuse/src/fs/dcache/inode.rs
+++ b/curvine-fuse/src/fs/dcache/inode.rs
@@ -203,7 +203,10 @@ impl Inode {
     }
 
     pub fn can_evict(&self, ttl: u64) -> bool {
+        // The kernel may continue issuing requests by nodeid until it balances
+        // every LOOKUP/READDIRPLUS reference with FORGET.
         !self.is_root()
+            && self.n_lookup == 0
             && self.last_access + ttl < LocalTime::mills()
             && self.dir.as_ref().is_none_or(|d| d.children.is_empty())
     }
--- a/curvine-fuse/src/fs/dcache/dir_tree.rs
+++ b/curvine-fuse/src/fs/dcache/dir_tree.rs
@@ -1267,6 +1267,40 @@ mod test {
         );
     }
 
+    /// A failed deferred delete keeps `mark_delete` set for later retry or
+    /// persisted-state recovery. TTL cleanup must not silently discard it.
+    #[test]
+    fn clear_keeps_expired_pending_delete_until_completion() {
+        let mut tree = DirTree::default();
+        let ino = tree
+            .lookup(FUSE_ROOT_ID, "pending", file_st("pending", 900), true)
+            .unwrap()
+            .ino;
+        tree.forget(ino, 1).unwrap();
+        tree.unlink(FUSE_ROOT_ID, "pending", true).unwrap();
+        tree.get_inode_mut(ino, None).unwrap().last_access = 0;
+
+        tree.clear(|_| false);
+
+        assert!(
+            tree.pending_delete(ino),
+            "TTL cleanup must preserve pending deferred-delete state"
+        );
+        assert!(
+            tree.get_dir_check(FUSE_ROOT_ID)
+                .unwrap()
+                .is_deleted_child("pending"),
+            "the deleted-child marker must remain available for completion"
+        );
+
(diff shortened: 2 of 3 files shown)


curvine-fuse/src/fs/dcache/inode.rs
```diff
@@ -203,7 +203,10 @@ impl Inode {
     }
 
     pub fn can_evict(&self, ttl: u64) -> bool {
```
szbr00, line 205: `can_evict` still allows TTL eviction of `mark_delete` inodes once `n_lookup == 0` and there are no handles. That behavior predates this PR, but deferred-delete failure paths could lose the pending mark if the cleaner runs before a retry. Consider excluding `mark_delete` here or adding a follow-up regression test for that scenario.

szbr00 (commented):
The unit/regression coverage here looks solid. The remaining gap called out in the PR (Linux `/dev/fuse` fixed-binary E2E for the original `stat .` → ENOENT repro) would be good to run before merge if you can, since that is the exact failure mode this fixes.

curvine-fuse/src/fs/dcache/inode.rs
ligeng8848: Thanks for catching this. Fixed in cb51e0e7: pending-delete inodes are now excluded from TTL eviction while mark_delete remains set. I also added a regression test that verifies the state survives cleanup before completion and becomes evictable after clear_mark_delete(). The full dcache test group passes 29/29 on the remote Linux host.

ligeng8848:
Thanks for the review. The fixed-binary Linux /dev/fuse E2E is complete and the PR test table has been updated from PENDING to PASS. On commit ad137e0c, both a held cwd and an open directory handle kept inode 1099 stable across multiple TTL cycles; stat . succeeded and unmount returned 0. The follow-up commit cb51e0e7 only adds the independent pending-delete eviction guard, which was validated remotely by the 29/29 dcache test group and cargo check -p curvine-fuse.

szbr00 (approved):

PR #1487, opened 2026-08-04
fix(fuse): preserve timestamp semantics

ljy1-lixing:
# Summary

Fix Curvine FUSE timestamp handling for pre-epoch dates, extreme userspace timestamps, millisecond precision, and explicit timestamp updates that race with writer completion.

Curvine stores file timestamps as signed millisecond values, but the FUSE conversion path previously clamped negative values to zero, rejected FUSE timestamps outside the signed millisecond range, and advertised nanosecond precision. In addition, an asynchronous writer `RELEASE` could complete after `utimensat(2)` and overwrite the requested `mtime` and `ctime` with the current time.

This change preserves signed timestamps throughout FUSE and local metadata conversion, clamps only values that exceed Curvine's representable range, advertises the actual one-millisecond granularity, and serializes explicit timestamp updates with active writer completion.

# Background

Unix timestamps are signed values. Files can legitimately carry timestamps before January 1, 1970, and modern 64-bit userspace can request dates far beyond the traditional signed 32-bit range. FUSE transports seconds in an unsigned field, using the two's-complement representation for negative seconds, while Curvine stores timestamps as signed milliseconds.

The conversion must therefore preserve the signed seconds value and normalize fractional milliseconds correctly. For example, `-1ms` must be returned to FUSE as `-1s + 999000000ns`, not as zero or as an invalid negative nanosecond component.

FUSE also uses the filesystem's advertised `time_gran` value when normalizing timestamps. Advertising one-nanosecond precision while storing only milliseconds allows the kernel and applications to assume a precision that Curvine cannot persist.

There is also an ordering requirement around open writers. A FUSE `RELEASE` request does not have a kernel reply, so the userspace writer completion can still be running after `close(2)` returns. A following `utimensat(2)` can successfully persist an explicit timestamp and then have it overwritten by the delayed complete operation.

# Minimal Reproducer

```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>
#include <time.h>
#include <unistd.h>

static void die(const char *operation)
{
    fprintf(stderr, "%s: %s\n", operation, strerror(errno));
    exit(EXIT_FAILURE);
}

static void verify_timestamp(const char *path, time_t expected)
{
    struct stat status;

    if (stat(path, &status) == -1)
        die("stat");

    if (status.st_mtim.tv_sec != expected || status.st_mtim.tv_nsec != 0) {
        fprintf(stderr, "mtime mismatch: expected %lld.000000000, got %lld.%09ld\n", (long long)expected, (long long)status.st_mtim.tv_sec, status.st_mtim.tv_nsec);
        exit(EXIT_FAILURE);
    }
}

static void run_case(const char *path, time_t timestamp)
{
    static const char payload[] = "timestamp ordering test";
    struct timespec times[2] = {
        {.tv_sec = timestamp, .tv_nsec = 0},
        {.tv_sec = timestamp, .tv_nsec = 0},
    };
    int fd = open(path, O_CREAT | O_TRUNC | O_WRONLY, 0644);

    if (fd == -1)
        die("open");
    if (write(fd, payload, sizeof(payload)) != (ssize_t)sizeof(payload))
        die("write");
    if (close(fd) == -1)
        die("close");

    /* RELEASE may still be completing when this request reaches FUSE. */
    if (utimensat(AT_FDCWD, path, times, 0) == -1)
        die("utimensat");

    verify_timestamp(path, timestamp);
}

{
    char path[4096];

    if (argc != 2) {
        fprintf(stderr, "usage: %s <curvine-mount>\n", argv[0]);
        return 2;
    }

    for (int i = 0; i < 100; i++) {
        if (snprintf(path, sizeof(path), "%s/timestamp-repro-%d", argv[1], i) >= (int)sizeof(path)) {
            fprintf(stderr, "mount path is too long\n");
            return 2;
        }

        run_case(path, (time_t)-315593940);
        if (unlink(path) == -1)
            die("unlink");
    }

    puts("PASS: pre-epoch mtime survived writer completion");
    return 0;
}
```

Compile and run the reproducer on a Curvine FUSE mount:

```bash
gcc -std=c11 -O2 -Wall -Wextra -Werror timestamp_repro.c -o timestamp_repro
./timestamp_repro /mnt/curvine
```

Before the fix, the timestamp can be returned as the Unix epoch or replaced by the current time:

```text
mtime mismatch: expected -315593940.000000000, got 0.000000000
```

or:

```text
mtime mismatch: expected -315593940.000000000, got 1785831847.000000000
```

After the fix:

```text
PASS: pre-epoch mtime survived writer completion
```

# Root Cause

The failures came from four related assumptions in the timestamp path.

First, `status_to_attr` applied `max(0)` to `atime`, `mtime`, and `ctime`. Every valid pre-epoch timestamp was therefore exposed to the kernel as January 1, 1970.

Second, negative FUSE seconds were reinterpreted as signed values, but conversion back to FUSE did not use Euclidean division. Signed millisecond timestamps require a non-negative nanosecond remainder, so values such as `-1ms` need normalization to the preceding second.

Third, `timestamp_to_millis` used checked `i64` multiplication and returned `EINVAL` when a userspace timestamp exceeded Curvine's signed millisecond range. The VFS can parse and submit timestamps beyond that range, and filesystem timestamp limits should clamp such values instead of rejecting the operation.

Fourth, active writer completion and `SETATTR` were not serialized. The Master accepted the requested timestamp, but a delayed writer complete subsequently assigned the current time. The FUSE response could also overlay the requested negative value with a writer timestamp converted through the same zero-clamping logic.

The local filesystem metadata adapters had a matching issue: `SystemTime::duration_since(UNIX_EPOCH)` errors for pre-epoch values, and the adapters converted that error to zero.

# Changes

- Add signed millisecond-to-FUSE conversion using Euclidean division so negative timestamps retain a non-negative nanosecond component.
- Remove zero clamping from FUSE `atime`, `mtime`, and `ctime` responses.
- Compare FUSE timestamps using signed seconds so two's-complement negative values are not considered later than positive timestamps.
- Convert incoming FUSE timestamps through `i128` and clamp results to `i64::MIN..=i64::MAX` instead of returning `EINVAL` for representable VFS requests.
- Advertise `time_gran = 1000000` nanoseconds to match Curvine's millisecond timestamp storage.
- Add a shared `SystemTime` conversion that preserves pre-epoch values with floor-to-millisecond semantics and use it in both local metadata adapters.
- Serialize explicit writer timestamp updates with `RELEASE` cleanup through the per-inode shared writer entry.
- Queue the explicit `mtime` and `ctime` on the active writer and forward them with the final complete operation.
- Clear the queued explicit timestamp when a later successful write occurs so normal POSIX write-time updates still take precedence.
- Add regression coverage for negative timestamps, signed comparison, range clamping, pre-epoch `SystemTime`, writer completion ordering, and write-after-timestamp behavior.

# Compatibility and Impact

- No protobuf schema, Master inode layout, journal format, or public API changes are required.
- Existing non-negative millisecond timestamps retain the same representation.
- Pre-epoch timestamps become visible and persistent instead of being silently reported as the Unix epoch.
- Extreme timestamps are clamped to Curvine's signed millisecond range instead of failing with `EINVAL`.
- Timestamp precision is now reported accurately as one millisecond; applications may observe kernel-side normalization to that granularity.
- Explicit `mtime` and `ctime` updates wait for serialization with an active writer, preventing delayed internal completion from reversing a successful userspace operation.
- A real write queued after the explicit timestamp update continues to advance modification and change times normally.

# Verification

- `cargo test -p curvine-runtime system_time_millis_preserves_pre_epoch_values --lib`
  - Passed.
- `cargo test -p curvine-runtime async_map --lib`
  - 12 passed, 0 failed.
- `cargo test -p curvine-fs-api --lib`
  - 23 passed, 0 failed.
- `cargo test -p curvine-common --lib`
  - 12 passed, 0 failed.
- `cargo test -p curvine-fuse fuse_utils::tests --lib`
  - 39 passed, 0 failed.
- `cargo test -p curvine-fuse fuse_writer::tests --lib`
  - 17 passed, 0 failed.
- `cargo test -p curvine-fuse advertised_timestamp_granularity_matches_millisecond_storage --lib`
  - Passed.
- `cargo clippy -p curvine-runtime -p curvine-fs-api -p curvine-common -p curvine-fuse --all-targets -- -D warnings`
  - Passed.
  - Passed.
- `git diff --check`
  - Passed.
- Rebuild and remount Curvine FUSE, then verify that a pre-epoch timestamp remains negative after remount.
  - Passed for `atime = -315644340`.
- Verify the large-date operation used by the `generic/634` scenario no longer returns `EINVAL`.
  - `touch -d 'Dec 31 23:59:59 UTC 2147483647'` succeeded.

# Related Test Coverage

- `generic/258` exercises timestamps before the Unix epoch and persistence across remount.
- `generic/634` exercises timestamp extremes accepted by userspace and verifies that the values remain stable across remount.
- The truncate timestamp behavior covered by `generic/313` has a separate root cause and is handled by the dedicated `fix/fuse-truncate-mtime` change.
- The hardlink unlink `ctime` behavior covered by `generic/755` is a separate inode-cache consistency issue and is intentionally not included in this change.


curvine-fuse/src/fs/fuse_writer.rs
```diff
@@ -388,13 +405,30 @@ impl FuseWriter {
...
                     crate::fs::deliver_stream_result(res, tx, reply).await?;
                 }
 
+                WriteTask::SetTimes(tx, mtime, ctime) => {
+                    file_mtime.set(mtime);
+                    pending_times = Some((mtime, ctime));
+                    crate::fs::deliver_stream_result(Ok(()), tx, None).await?;
+                }
+
                 WriteTask::Resize(tx, opts) => {
```
szbr00, line 431: `pending_times` is cleared on a successful Write (line 386) but not on a successful Resize. For the sequence utimensat(T) -> ftruncate -> close, the resize already stamps mtime/ctime on the master (`fs_dir::resize` completes the file), but the following Complete then forwards the queued (T, ctime) and overwrites them with the pre-truncate values. POSIX requires truncate to update mtime/ctime, and a truncate is a data modification just like a write. Consider clearing `pending_times` when the resize succeeds, mirroring the Write arm.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -1011,10 +1013,37 @@ impl NodeState {
...
+            match self
+                .writers
+                .with_resource_result(&ino, |writer| async move {
+                    // Serialize with RELEASE cleanup. Otherwise complete can
+                    // overwrite an mtime that a following utimensat just set.
+                    let status = match fs.fuse_set_attr(&writer_path, writer_opts).await? {
+                        Some(status) => status,
+                        None => fs.get_status(&writer_path).await?,
+                    };
+                    writer.set_times(mtime, status.ctime()).await?;
```
szbr00: This await runs while `with_resource_result` holds the per-inode writer entry lock, and `set_times` waits for the writer loop to FIFO-process the SetTimes task — i.e. it waits for every queued Write's backend IO to finish. During that window the lock also stalls RELEASE (`release_with_cleanup`), OPEN (`get_or_create`), and read-only getattr/stat (`find_writer` -> `get`) for this inode. Correctness only requires the SetTimes enqueue to be ordered before a cleanup Complete, so the lock only needs to cover the master setattr plus the enqueue. Consider returning the CallReceiver from the closure and awaiting it after the map lock is released, so the queue drain happens outside the critical section.

curvine-fuse/src/fs/fuse_writer.rs
```diff
@@ -388,13 +405,30 @@ impl FuseWriter {
                     crate::fs::deliver_stream_result(res, tx, reply).await?;
                 }
 
-                WriteTask::Complete(tx, reply, opts) => {
+                WriteTask::Complete(tx, reply, mut opts) => {
                     *preserve_on_exit = true;
+                    if let Some((mtime, ctime)) = pending_times {
+                        let opts = opts.get_or_insert_with(SetAttrOpts::default);
+                        opts.mtime = Some(mtime);
+                        opts.add_x_attr.insert(
```
szbr00: Forwarding the queued ctime can move ctime backwards. If another metadata operation lands between the utimensat and the final complete (e.g. utimensat at t1 queues ctime=t1, chmod at t2 sets master ctime=t2), this complete rewrites ctime to t1 even though t2 is newer. `InodeFile::complete` already stamps mtime/ctime=now before the opts are applied, and `update_ctime` is monotonic, so either omitting the forwarded ctime (letting the complete-time stamp stand) or clamping it to now would keep ctime non-decreasing in all interleavings.

curvine-fuse/src/fs/fuse_writer.rs
ljy1-lixing: Thanks. Fixed by clearing the pending explicit mtime after a successful resize and updating the cached length/mtime from the writer's authoritative status at queue execution time. Added SetMtime→Resize, Write→Resize, and Resize→Write ordering tests.

curvine-fuse/src/fs/state/node_state.rs
ljy1-lixing: Thanks. Fixed by splitting mtime enqueue from completion waiting. The per-inode writer lock now covers the master setattr and SetMtime enqueue only; queue draining is awaited after the lock is released.

curvine-fuse/src/fs/fuse_writer.rs
ljy1-lixing: Thanks. Fixed by no longer forwarding the captured ctime through the final Complete. ctime remains master-owned, preventing an older queued writer snapshot from overwriting a newer metadata update.

lzjqsdd (approved):

PR #1490, opened 2026-08-04
fix(ci): enforce minimal RDMA guardrails (#1243)

### LINKED ISSUE #1243: [BUG]: curvine-cli should not require RDMA runtime libraries on non-RDMA clients

lzjqsdd:
# Summary

This PR tightens the #1243 dependency guardrails so server-side SPDK/RDMA features cannot silently pollute the default CLI artifact, while still allowing explicitly RDMA/SPDK-enabled client or FUSE artifacts. It also updates the Chinese README to use the current split crate APIs.

# Issue Describe / Design

Closes #1243

Root cause:
- The `mixed-spdk-feature-risk` check only selected `curvine-cli` while passing `curvine-server/spdk-rdma`, so Cargo rejected the feature set and the script reported `SKIP`.
- The minimal artifact guard wording looked like a blanket client-side RDMA ban, even though explicit RDMA client/FUSE artifacts are valid.
- `README_zh.md` still referenced the removed `curvine_common::*` API.

Reproduction steps:
- Run `scripts/check-deps.sh --mode ci` before this change and observe `SKIP [mixed-spdk-feature-risk]`.

Fix approach:
- Restore mixed `curvine-cli + curvine-server` feature resolution and check `curvine-storage-spdk` reverse dependencies.
- Fail CI if the mixed check command fails instead of silently skipping.
- Add an explicit `--allow-rdma-spdk` mode for RDMA/SPDK-enabled artifact checks; it requires explicit artifacts and the implicit default set stays strict.
- Update Chinese README examples and build notes.

# Changes

| `scripts/check-deps.sh` | Uses reverse-dependency checking for `curvine-storage-spdk` in mixed CLI/server SPDK resolution | CI now catches real CLI pollution and no longer silently skips |
| `scripts/check-minimal-artifact-deps.sh` | Adds `--allow-rdma-spdk`, rejects it for the implicit default artifact set, separates RDMA/SPDK, native UFS, and native storage checks, and uses mode-agnostic failure wording | Default minimal checks remain strict; explicit RDMA artifacts can be checked without false failure |
| `build/build.sh` | Narrows error wording to minimal client artifacts and includes `libdpdk` in the SPDK native pattern | No build behavior change except clearer diagnostics |
| `README_zh.md` | Updates Rust API imports and SPDK/RDMA build notes | Documentation now matches split crates and current build profiles |
| `scripts/README.md` | Documents explicit RDMA/SPDK artifact check mode and the explicit-artifact requirement | Clarifies intended guard usage |

# Test verified

| `bash -n scripts/check-deps.sh scripts/check-minimal-artifact-deps.sh build/build.sh` | PASS | Shell syntax |
| `scripts/check-deps.sh --mode ci` | PASS | `mixed-spdk-feature-risk` reports `OK`, not `SKIP` |
| `cargo tree --no-default-features --features curvine-server/spdk-rdma,curvine-alloc/system -p curvine-cli -p curvine-server -e normal -i curvine-storage-spdk --prefix none -f '{p}'` | PASS | Reverse deps do not include `curvine-cli` |
| `scripts/check-minimal-artifact-deps.sh --allow-missing` | PASS | Minimal/non-RDMA artifact guard |
| `scripts/check-minimal-artifact-deps.sh --allow-missing --allow-rdma-spdk --artifact smoke=/bin/ls` | PASS | Explicit allow mode smoke test |
| `scripts/check-minimal-artifact-deps.sh --allow-rdma-spdk` | PASS | Rejected with exit 2 unless explicit artifacts are supplied |
| `scripts/check-minimal-artifact-deps.sh --allow-rdma-spdk --artifact rdma=<tmp ELF>` | PASS | RDMA-only fixture passes in explicit allow mode |
| `scripts/check-minimal-artifact-deps.sh --allow-rdma-spdk --artifact storage=<tmp ELF>` | PASS | Native storage dependency still fails with "client artifact" wording |
| `scripts/check-minimal-artifact-deps.sh --allow-rdma-spdk --artifact ufs=<tmp ELF>` | PASS | Native UFS dependency still fails with "client artifact" wording |
| `rg "curvine_common::|curvine_common" README_zh.md` | PASS | No stale Chinese README API reference |
| `make format` | PASS | Project formatting/check workflow |
| `git diff --check` | PASS | Whitespace check |

# Dependencies

- Related PRs: None
- Related issues: #1243
- Blocks / blocked by: None

### DIFF
--- a/scripts/check-minimal-artifact-deps.sh
+++ b/scripts/check-minimal-artifact-deps.sh
@@ -6,14 +6,19 @@ set -euo pipefail
 
 usage() {
   cat <<'EOF'
-Usage: scripts/check-minimal-artifact-deps.sh [--allow-missing] [--artifact LABEL=PATH] [PATH ...]
+Usage: scripts/check-minimal-artifact-deps.sh [--allow-missing] [--allow-rdma-spdk] [--artifact LABEL=PATH] [PATH ...]
 
-With no artifacts, the script checks known Curvine outputs in target/{debug,release}
-and build/dist/lib. Explicit artifacts must exist unless --allow-missing is set.
+By default this enforces minimal/non-RDMA client artifacts. With no artifacts,
+the script checks known Curvine outputs in target/{debug,release} and
+build/dist/lib. Explicit artifacts must exist unless --allow-missing is set.
+
+Use --allow-rdma-spdk only for explicitly RDMA/SPDK-enabled client or FUSE
+artifacts; native UFS and storage-library checks remain enforced.
 EOF
 }
 
 allow_missing=0
+allow_rdma_spdk=0
 declare -a artifact_specs=()
 using_defaults=0
 
@@ -21,6 +26,10 @@ while [[ $# -gt 0 ]]; do
   case "$1" in
     --allow-missing)
       allow_missing=1
+      shift
+      ;;
+    --allow-rdma-spdk)
+      allow_rdma_spdk=1
       shift
       ;;
     --artifact)
@@ -105,7 +114,9 @@ artifact_inspector() {
   fi
 }
 
-forbidden_pattern='libibverbs\.so|librdmacm\.so|libspdk|libdpdk|librte_|libjindosdk|libhdfs|libjvm|libjli|librocksdb'
+rdma_spdk_pattern='libibverbs\.so|librdmacm\.so|libspdk|libdpdk|librte_'
+native_ufs_pattern='libjindosdk|libhdfs|libjvm|libjli'
+native_storage_pattern='librocksdb'
 checked=0
 failures=0
 
@@ -137,13 +148,28 @@ for spec in "${artifact_specs[@]}"; do
 
   checked=$((checked + 1))
   needed="$(artifact_dynamic_entries "$inspector" "$artifact")"
-  if grep -E "$forbidden_pattern" <<<"$needed" >/dev/null; then
-    echo "FAIL [$label] forbidden native runtime dependency found in $artifact" >&2
+  if [[ "$allow_rdma_spdk" -eq 0 ]] && grep -E "$rdma_spdk_pattern" <<<"$needed" >/dev/null; then
+    echo "FAIL [$label] RDMA/SPDK runtime dependency found in minimal/non-RDMA artifact: $artifact" >&2
     echo "$needed" | sed 's/^/  /' >&2
     failures=$((failures + 1))
-  else
-    echo "OK   [$label] runtime dependencies"
+    continue
   fi
+
+  if grep -E "$native_storage_pattern" <<<"$needed" >/dev/null; then
+    echo "FAIL [$label] native storage runtime dependency found in minimal client artifact: $artifact" >&2
+    echo "$needed" | sed 's/^/  /' >&2
+    failures=$((failures + 1))
+    continue
+  fi
+
+  if grep -E "$native_ufs_pattern" <<<"$needed" >/dev/null; then
+    echo "FAIL [$label] native UFS runtime dependency found in minimal client artifact: $artifact" >&2
+    echo "$needed" | sed 's/^/  /' >&2
+    failures=$((failures + 1))
+    continue
+  fi
+
+  echo "OK   [$label] runtime dependencies"
 done
 
 if ((failures > 0)); then
@@ -156,4 +182,8 @@ if ((checked == 0)); then
   exit 2
 fi
 
-echo "Minimal artifact runtime dependency check passed."
+if [[ "$allow_rdma_spdk" -eq 1 ]]; then
+  echo "Runtime dependency check passed with explicit RDMA/SPDK dependencies allowed."
+else
+  echo "Minimal/non-RDMA artifact runtime dependency check passed."
+fi
--- a/scripts/check-minimal-artifact-deps.sh
+++ b/scripts/check-minimal-artifact-deps.sh
@@ -13,7 +13,9 @@ the script checks known Curvine outputs in target/{debug,release} and
 build/dist/lib. Explicit artifacts must exist unless --allow-missing is set.
 
 Use --allow-rdma-spdk only for explicitly RDMA/SPDK-enabled client or FUSE
-artifacts; native UFS and storage-library checks remain enforced.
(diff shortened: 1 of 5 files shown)


scripts/check-minimal-artifact-deps.sh
```diff
@@ -23,6 +28,10 @@ while [[ $# -gt 0 ]]; do
       allow_missing=1
       shift
       ;;
+    --allow-rdma-spdk)
```
szbr00, line 33: The allowance is global, so running this flag without `--artifact` checks the implicit default set while permitting RDMA/SPDK dependencies in `curvine-cli`, SDKs, and FUSE alike. This can let the minimal guard pass when CLI pollution regresses—the scenario this issue is intended to catch. Please reject `--allow-rdma-spdk` unless explicit artifacts are supplied, or scope the opt-out per artifact so default outputs remain strict.

scripts/check-minimal-artifact-deps.sh
```diff
@@ -137,13 +148,28 @@ for spec in "${artifact_specs[@]}"; do
...
-    echo "OK   [$label] runtime dependencies"
+    continue
   fi
+
+  if grep -E "$native_storage_pattern" <<<"$needed" >/dev/null; then
+    echo "FAIL [$label] native storage runtime dependency found in minimal client artifact: $artifact" >&2
+    echo "$needed" | sed 's/^/  /' >&2
+    failures=$((failures + 1))
+    continue
+  fi
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 5b13600d: native storage and native UFS failure messages now use mode-agnostic "client artifact" wording.

scripts/check-minimal-artifact-deps.sh
lzjqsdd: Thanks. Fixed in 5b13600d: --allow-rdma-spdk now requires explicit --artifact or PATH arguments and is rejected for the implicit default artifact set; usage and scripts/README.md updated accordingly.

szbr00 (approved):

PR #1495, opened 2026-08-05
fix(fuse): preserve hard links and stabilize file locking (CUR-230)

### LINKED ISSUE #1486: [BUG]: FUSE hard link count incorrect after unlink and file handle stale after drop_caches

When using Curvine FUSE, two distinct bugs are present: (1) After creating a hard link and deleting the original file, the link count reported by stat is incorrect; (2) After the kernel drops caches, file handles become stale and return ESTALE (errno 116) instead of being recoverable.

(1) After deleting the original file of a hard link pair, the stat link count is wrong. (2) After kernel drop_caches, FUSE file handle operations return ESTALE.

Reproduction:

1. 1. Create a Curvine FUSE mount at a mount point
2. 2. Create a file and a hard link to it
3. 3. Delete the original file and run `stat` on the remaining hard link — observe incorrect link count
4. 4. Open a file handle on the FUSE mount, run `echo 3 > /proc/sys/vm/drop_caches`, then perform a file handle operation (e.g., fstat, read) — observe ESTALE error

**Impact**

Users relying on hard links in Curvine FUSE mounts may experience incorrect file metadata. File handle operations after cache drops may fail with stale handle errors, making applications that rely on long-lived file handles unreliable.

(1) The link count should correctly reflect the remaining hard link (1) after the original file is deleted. (2) File handles should remain valid or be recoverable after kernel cache drop.

Investigate the FUSE implementation: (1) Fix link count reporting after unlink of a hard-linked file; (2) Fix file handle recovery after kernel cache drop to return valid handles instead of ESTALE. Relevant code areas: FUSE mount handler and file handle management.

**Additional context**

The FIO stress test (5/5) passes, indicating the basic FUSE read/write path works. The two failures are specific to hard link metadata and file handle lifecycle management.

Root-cause hypothesis: The FUSE test suite fails with two issues: (1) Hard link count after deleting the original file is incorrect — the FUSE layer does not properly report link count changes after unlink; (2) File handle at operations after kernel drop_caches return ESTALE (errno 116), indicating the FUSE file handle recovery mechanism is broken.

Evidence summary: FUSE test log shows: at line 441 "Hard link behavior incorrect after original deletion", at line 443 "Link count incorrect after deletion (expected: 1, got: ...)", and at line 687 "FAIL: drop_caches round-trip — unexpected OSError errno=116 (ESTALE): [Errno 116] Stale file handle". The FIO test passes (5/5), but the FUSE-specific tests fail with 2/4 failures.

Acceptance / test gaps:

- Hard link stat shows correct link count (1) after the original file is deleted
- File handle operations succeed after kernel drop_caches without returning ESTALE
- All FUSE-specific tests pass (4/4, not 2/4)

### LINKED ISSUE #1492: [BUG]: FUSE blocking read returns EIO on OFD lock test (fcntl36) under concurrent dirty-read

On Curvine FUSE, LTP fcntl36 fails during OFD read lock vs OFD write lock synchronization because blocking read() returns EIO (errno 5). The test calls read(49, buf, 4096) on a file descriptor opened on a curvine-fuse mount, and the FUSE layer returns EIO. POSIX blocking reads on a regular file must not return EIO unless the filesystem encounters a genuine I/O error. The 64-bit variant (fcntl36_64) passes all subtests, which suggests the issue is specific to the 32-bit read path or the specific lock contention scenario. Previous issue #1429 fixed a similar failure where read() returned EAGAIN; the current failure is a different symptom (EIO) on the same test.

LTP fcntl36 test fails with TBROK at fcntl36.c:171: read(49, 0xffff715147c0, 4096) returns -1 with errno 5 (EIO). The test summary shows: passed=0, failed=0, broken=1, skipped=0. The fcntl36_64 variant passes with 7/7 subtests passed. The tst_rmdir cleanup also warns with ENOTEMPTY (errno 39), indicating the test directory was not properly cleaned up.

Reproduction:

1. Build Curvine at upstream/main (commit d134f5bf) and start a single-node cluster with FUSE support.
2. Mount curvine-fuse to a local directory (e.g., /curvine-fuse).
3. Run the LTP 20210524 syscalls suite: runltp -f syscalls -s fcntl36
4. Observe TBROK: read(49, 0xffff715147c0, 4096) failed, returned -1: EIO (5)
5. Confirm fcntl36_64 passes: runltp -f syscalls -s fcntl36_64

**Impact**

LTP fcntl36 test failure indicates that Curvine FUSE blocking read semantics are broken under concurrent OFD lock traffic. The previous fix for EAGAIN (issue #1429) addressed one symptom, but the same test now fails with EIO, suggesting the root cause was not fully resolved. Applications that use blocking reads with file locking on FUSE mounts may encounter unexpected I/O errors.

Blocking read() on a regular file descriptor opened on a curvine-fuse mount should either return data or block until data is available. It must not return EIO for a legitimate read operation during OFD lock contention. The fcntl36 test should PASS with all subtests completing successfully, as fcntl36_64 does.

Investigate the FUSE read path in the Curvine product code to determine why blocking read() returns EIO under OFD lock contention. The 64-bit variant passes but 32-bit fails, which may indicate a platform-specific issue in the read path, snapshot logic, or error translation. Check if the dirty-read snapshot fix from the PR associated with #1429 correctly handles all cases or if EIO is being returned from a different code path (e.g., backend read error, worker communication error, or snapshot version mismatch).

**Additional context**

The previous issue #1429 (CLOSED) fixed EAGAIN on the same fcntl36 test by publishing a sampled snapshot when no enqueue is in flight. The current failure mode is EIO instead of EAGAIN, which suggests either a regression from the fix or a different code path that was not covered. The 64-bit variant passes, which is unexpected and warrants investigation into whether the 32-bit and 64-bit read paths differ in the FUSE implementation.

Root-cause hypothesis: On Curvine FUSE, LTP fcntl36 fails during OFD read lock vs OFD write lock synchronization because blocking read() returns EIO (errno 5). The Curvine FUSE layer returns EIO for a legitimate read() operation on a regular file descriptor during OFD lock contention. The fcntl36_64 variant passes, suggesting the 32-bit read path or specific lock contention handling is defective. Previous issue #1429 fixed EAGAIN for the same test, but the current failure is EIO, indicating the root cause was not fully resolved or a new code path is affected.

Evidence summary: LTP console log shows fcntl36.c:171: TBROK: read(49,0xffff715147c0,4096) failed, returned -1: EIO (5). The fcntl36_64 variant passes all 7 subtests. The test runs on the Curvine FUSE mount /curvine-fuse.

Related issues: https://github.com/CurvineIO/curvine/issues/1429

Acceptance / test gaps:

- LTP fcntl36 test passes on curvine-fuse mount with all subtests completed.
- Blocking read() on a regular file descriptor on curvine-fuse does not return EIO under OFD lock contention.
- The 64-bit variant (fcntl36_64) continues to pass.
- No new regressions in other LTP fcntl or lock-related tests.

LJF-050:
# Summary

Fix FUSE hard-link unlink semantics and export-support handle reconstruction, and stabilize concurrent advisory/OFD lock I/O so LTP fcntl36 and related lock tests pass without false EIO or EDEADLK.

Related to CUR-230. Closes #1486. Closes #1492.

# Issue Describe / Design

## Root cause

- Unlinking one name of a multi-link inode could leave stale canonical paths, incorrect nlink handling, and broken `name_to_handle_at` reconstruction after dentry cache eviction.
- Concurrent writes plus `fsync` could flush before all writes were queued, and advisory lock paths unnecessarily flushed file data.
- OFD locks (pid `0` / `-1`) were modeled as POSIX process wait edges, producing false deadlock detection; local short-circuit block reads could surface EIO when a local worker was temporarily unavailable.

## Fix approach

- Repoint surviving hard-link dentries, notify master on non-final unlinks, and enable `FUSE_EXPORT_SUPPORT` with working root `.`/`..` lookup.
- Serialize `FuseWriter::flush` behind the same enqueue gate as writes; remove data flush from lock RPC handlers.
- Exclude OFD waiters from the POSIX deadlock graph and only return `EDEADLK` for cross-process cycles.
- Fall back to remote block reads when local short-circuit reader creation fails.

# Changes

| `curvine-fuse/src/fs/dcache/dir_tree.rs` | Repoint inode after canonical hard-link unlink; keep `#1489` negative inode validation | Hard-link unlink correctness only |
| `curvine-fuse/src/fs/state/node_state.rs` | Master unlink for non-final links; root `.`/`..` lookup for export support | Hard-link + handle export paths |
| `curvine-fuse/src/lib.rs` | Advertise `FUSE_EXPORT_SUPPORT` | Enables kernel handle reconstruction after cache drop |
| `curvine-fuse/src/fs/fuse_writer.rs` | Gate flush behind write enqueue order | Fixes concurrent write/fsync ordering |
| `curvine-fuse/src/fs/curvine_file_system.rs` | Lock path flush removal; OFD/POSIX deadlock rules | Lock semantics only; no data-path regression |
| `curvine-fuse/src/fs/plock_wait_registry.rs` | Process-scoped deadlock detection | Fewer false `EDEADLK` on OFD tests |
| `crates/client/curvine-client-core/src/block/block_reader.rs` | Local→remote read fallback | Error-path only; avoids spurious EIO |
| `curvine-tests/regression/utils/cluster_utils.py` | Data-path readiness probe before FUSE tests | Test harness only |

# Test verified

| Harness `fast` profile | PASS | No new failures |
| Harness `integration` profile | PASS* | Two known baseline failures only |
| Harness `daily` profile | PASS* | Same two known baseline failures only |
| Harness `fuse` profile | PASS | No new failures |
| Harness `ltp` profile | PASS | `fcntl36` / `fcntl36_64` pass |
| Harness `csi` profile | PASS | No new failures |
| `cargo clippy -p curvine-fuse -p curvine-client-core` | PASS | Pre-PR check |
| Intermittent raft failure | N/A | Re-run cleared; treated as flake |

# Dependencies

- Related issues: #1486, #1492, CUR-230
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -56,6 +56,11 @@ pub struct CurvineFileSystem {
 }
 
 impl CurvineFileSystem {
+    fn is_ofd_lock_pid(pid: u32) -> bool {
+        // FUSE kernels have used both 0 and -1 for the pid of an OFD lock.
+        pid == 0 || pid == u32::MAX
+    }
+
     pub fn new(conf: ClusterConf, rt: Arc<Runtime>) -> FuseResult<Self> {
         FuseMetrics::ensure_init()?;
 
@@ -278,15 +283,11 @@ impl CurvineFileSystem {
         Ok(RenameFlags::from_bits(flags).unwrap_or(RenameFlags::empty()))
     }
 
-    fn to_file_lock(&self, arg: &fuse_lk_in, header_pid: u32) -> FileLock {
+    fn to_file_lock(&self, arg: &fuse_lk_in, _header_pid: u32) -> FileLock {
         let client_id = self.fs.cv().fs_context().clone_client_name();
-        // Prefer the flock pid from the kernel request; fall back to the FUSE
-        // header pid when the kernel leaves lk.pid unset (common on some paths).
-        let pid = if arg.lk.pid != 0 {
-            arg.lk.pid
-        } else {
-            header_pid
-        };
+        // Keep the kernel pid intact. OFD locks use either 0 or -1 depending on
+        // the kernel and are owned by the open file description, not a process.
+        let pid = arg.lk.pid;
         FileLock {
             client_id,
             owner_id: arg.owner,
@@ -2075,8 +2076,6 @@ impl fs::FileSystem for CurvineFileSystem {
         let path = self.state.get_path(op.header.nodeid)?;
         let lock = self.to_file_lock(op.arg, op.header.pid);
 
-        self.state.fs_fsync(op.header.nodeid, None).await?;
-
         let conflict = self.fs.get_lock(&path, lock).await?;
         let lk = match conflict {
             Some(lk) => fuse_file_lock {
@@ -2099,8 +2098,6 @@ impl fs::FileSystem for CurvineFileSystem {
         let path = self.state.get_path(op.header.nodeid)?;
         self.ensure_writable_path(&path, RpcCode::SetLock).await?;
         let handle = self.state.find_handle(op.header.nodeid, op.arg.fh)?;
-
-        self.state.fs_fsync(op.header.nodeid, None).await?;
 
         let mut lock = self.to_file_lock(op.arg, op.header.pid);
         let (flag, owner_id) = (lock.lock_flags, lock.owner_id);
@@ -2134,8 +2131,6 @@ impl fs::FileSystem for CurvineFileSystem {
         self.ensure_writable_path(&path, RpcCode::SetLock).await?;
         let handle = self.state.find_handle(op.header.nodeid, op.arg.fh)?;
 
-        self.state.fs_fsync(op.header.nodeid, None).await?;
-
         let conf = &self.fs.conf().client;
         let check_interval_min_ms = conf.sync_check_interval_min_ms;
         let check_interval_max_ms = conf.sync_check_interval_max_ms;
@@ -2144,6 +2139,9 @@ impl fs::FileSystem for CurvineFileSystem {
         let mut ticks: u64 = 0;
         let time = TimeSpent::new();
 
+        // Linux reports OFD locks with lk.pid == 0 or -1 and does not perform
+        // deadlock detection for F_OFD_SETLKW; keep them out of the POSIX wait graph.
+        let detect_deadlock = !Self::is_ofd_lock_pid(op.arg.lk.pid);
         let mut lock = self.to_file_lock(op.arg, op.header.pid);
         let is_unlock = lock.lock_type == LockType::UnLock;
         let full_range_unlock = Self::is_full_range_unlock(&lock);
@@ -2178,8 +2176,14 @@ impl fs::FileSystem for CurvineFileSystem {
             }
 
             let blocker = conflict.as_ref().expect("conflict lock");
-            let decision = wait_guard
-                .register_blocked_by(LockOwner::new(blocker.client_id.clone(), blocker.owner_id));
+            // An OFD blocker (pid == 0) does not represent a process wait edge,
+            // so a POSIX waiter blocked by it cannot close a POSIX deadlock cycle.
+            let decision = (detect_deadlock && !Self::is_ofd_lock_pid(blocker.pid)).then(|| {
+                wait_guard.register_blocked_by(LockOwner::new(
+                    blocker.client_id.clone(),
+                    blocker.owner_id,
+                ))
+            });
             debug!(
                 "plock SETLKW wait decision unique={} pid={} owner_id={} nodeid={} path={} range=[{},{}] blocker_pid={} blocker_owner_id={} blocker_range=[{},{}] decision={:?}",
                 op.header.unique,
@@ -2195,7 +2199,10 @@ impl fs::FileSystem for CurvineFileSystem {
                 blocker.end,
                 decision
             );
-            if decision.is_deadlock() {
+            if decision
+                .as_ref()
+                .is_some_and(PlockWaitDecision::is_process_deadlock)
+            {
                 // Cycle in the local wait graph. Re-sample Master once while
                 // keeping our edge published so a peer in a true multi-resource
                 // deadlock (LTP fcntl17) still observes the cycle. If the lock
@@ -2215,10 +2222,15 @@ impl fs::FileSystem for CurvineFileSystem {
                     return Ok(());
                 }
                 let blocker2 = conflict2.as_ref().expect("conflict lock");
-                let decision2 = wait_guard.register_blocked_by(LockOwner::new(
-                    blocker2.client_id.clone(),
-                    blocker2.owner_id,
-                ));
+                let decision2 = if !Self::is_ofd_lock_pid(blocker2.pid) {
+                    Some(wait_guard.register_blocked_by(LockOwner::new(
+                        blocker2.client_id.clone(),
+                        blocker2.owner_id,
+                    )))
+                } else {
+                    wait_guard.clear_blocked_by();
+                    None
+                };
                 debug!(
                     "plock SETLKW deadlock recheck unique={} pid={} owner_id={} nodeid={} path={} range=[{},{}] blocker_pid={} blocker_owner_id={} blocker_range=[{},{}] decision={:?}",
                     op.header.unique,
@@ -2234,19 +2246,21 @@ impl fs::FileSystem for CurvineFileSystem {
                     blocker2.end,
                     decision2
                 );
-                if let PlockWaitDecision::Deadlock { cycle, .. } = decision2 {
-                    debug!(
-                        "plock SETLKW returning EDEADLK unique={} pid={} owner_id={} nodeid={} path={} range=[{},{}] cycle={:?}",
-                        op.header.unique,
-                        lock.pid,
-                        lock.owner_id,
-                        op.header.nodeid,
-                        path,
-                        lock.start,
-                        lock.end,
-                        cycle
-                    );
-                    return err_fuse!(libc::EDEADLK);
+                if let Some(PlockWaitDecision::Deadlock { cycle, .. }) = decision2 {
+                    if cycle.spans_multiple_processes() {
+                        debug!(
+                            "plock SETLKW returning EDEADLK unique={} pid={} owner_id={} nodeid={} path={} range=[{},{}] cycle={:?}",
+                            op.header.unique,
+                            lock.pid,
+                            lock.owner_id,
+                            op.header.nodeid,
+                            path,
+                            lock.start,
+                            lock.end,
+                            cycle
+                        );
+                        return err_fuse!(libc::EDEADLK);
+                    }
                 }
             }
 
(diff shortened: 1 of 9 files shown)


curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -852,17 +868,26 @@ impl NodeState {
...
 
     pub async fn fs_lookup(&self, ino: u64, name: &str) -> FuseResult<fuse_attr> {
-        // NOTE: the `.`/`..` branches below are BROKEN wherever they resolve through the root, because
-        // root's `parent` is the `0` sentinel and inode 0 does not exist.
         let (ino, cow_name) = if name == FUSE_CURRENT_DIR {
             let dir = self.dir_read();
             let inode = dir.get_inode_check(ino, None)?;
-            (inode.parent, Cow::Owned(inode.name.to_owned()))
+            if inode.is_root() {
+                (FUSE_ROOT_ID, Cow::Owned(FUSE_PATH_SEPARATOR.to_owned()))
```
szbr00: Mapping root `.`/`..` to `(FUSE_ROOT_ID, "/")` reuses the generic child-name lookup path. `"/"` is not a dentry under root: this can return ENOENT with a valid directory scan or insert root as its own child through `DirTree::lookup`. Please return root attributes directly without passing a synthetic child name through the normal lookup pipeline, and add a root `.`/`..` test.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -852,17 +868,26 @@ impl NodeState {
...
-            let parent_inode = dir.get_inode_check(ino, None)?;
-            let inode = dir.get_inode_check(parent_inode.parent, None)?;
-            (inode.parent, Cow::Owned(inode.name.to_owned()))
+            let inode = dir.get_inode_check(ino, None)?;
+            if inode.is_root() {
+                (FUSE_ROOT_ID, Cow::Owned(FUSE_PATH_SEPARATOR.to_owned()))
+            } else {
+                let parent_inode = dir.get_inode_check(inode.parent, None)?;
+                (
+                    parent_inode.parent,
```
szbr00: `LOOKUP(<child-of-root>, "..")` resolves to `(parent_inode.parent, parent_inode.name) = (0, "/")`. Inode 0 is not present in the dcache, so exportfs parent reconstruction fails with ENOENT. Please handle a root `parent_inode` explicitly and test this path before advertising `FUSE_EXPORT_SUPPORT`.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -2178,8 +2176,14 @@ impl fs::FileSystem for CurvineFileSystem {
             }
 
             let blocker = conflict.as_ref().expect("conflict lock");
-            let decision = wait_guard
-                .register_blocked_by(LockOwner::new(blocker.client_id.clone(), blocker.owner_id));
+            // An OFD blocker (pid == 0) does not represent a process wait edge,
+            // so a POSIX waiter blocked by it cannot close a POSIX deadlock cycle.
+            let decision = (detect_deadlock && !Self::is_ofd_lock_pid(blocker.pid)).then(|| {
```
szbr00: When the blocker changes from POSIX to OFD, this branch leaves the previous wait-graph edge registered because neither registration nor clearing occurs. Other waiters can then observe a false cycle and return `EDEADLK`. Please call `wait_guard.clear_blocked_by()` when the blocker is OFD or deadlock detection is disabled, matching the recheck path below.

curvine-fuse/src/fs/state/node_state.rs
```diff
@@ -662,12 +662,28 @@ impl NodeState {
...
 
+        // Removing one of several hard links must reach the master immediately:
+        // the master owns the authoritative nlink count and removes only this
+        // directory entry. Open handles do not require deferral while another
+        // link still keeps the inode alive.
+        if !last_link {
+            match self.fs.delete(&path, false).await {
+                Ok(()) | Err(FsError::FileNotFound(_)) => (),
+                Err(e) => {
+                    self.clear_unlink_state(ino, parent, name)?;
```
szbr00: On master delete failure, `clear_unlink_state` does not undo the dentry removal, link/ref counter changes, or canonical-path repointing performed by `DirTree::unlink`. This leaves the local cache inconsistent with the master after a transient error. Please implement full rollback or defer the local mutation until deletion succeeds.

curvine-fuse/src/fs/dcache/dir_tree.rs
```diff
@@ -294,11 +294,28 @@ impl DirTree {
             dir.mark_deleted_child(name);
         }
 
+        // get_path(ino) is used for subsequent opens. If the canonical dentry
+        // was removed, repoint it at one of the surviving hard-link names.
+        if !last_link {
+            let replacement = self.inodes.iter().find_map(|(parent_ino, parent_inode)| {
```
szbr00, line 325: This replacement scan only sees cached hardlink dentries. If surviving links exist on the master but were never cached, unlinking the canonical dentry leaves `parent`/`name` pointing to a removed path and breaks later path-based metadata operations. Please provide a master-backed fallback when no cached replacement exists.

szbr00, line 325: This scans every cached inode and child while `dir_write` is held, blocking unrelated metadata operations as the cache grows. Consider maintaining a per-inode reverse index of cached link names.

curvine-tests/regression/utils/cluster_utils.py
```diff
@@ -103,11 +103,33 @@ def prepare_curvine_cluster(project_path, test_path='/curvine-fuse'):
...
+                    print("Cluster data path is ready")
+                    return True, ""
+                except OSError as exc:
+                    last_error = str(exc)
+                    try:
+                        os.unlink(probe_path)
+                    except OSError:
+                        pass
+                    time.sleep(2)
+            return False, f"Cluster mounted but data path was not ready: {last_error}"
```
szbr00: The new probe gates fresh startup, but `ensure_cluster_ready()` still returns success for any existing mount without probing the data path. Since regression entry points use that function, a stale mount can still bypass this protection. Consider extracting the probe and invoking it from both paths.

curvine-fuse/src/fs/state/node_state.rs
LJF-050: Thanks. Fixed in e789631d: `fs_lookup` now returns root attributes directly for root `.`/`..` without routing a synthetic `"/"` child through dentry lookup. Added unit tests `fs_lookup_root_dot_and_dotdot_returns_root_attr` and `fs_lookup_child_of_root_dotdot_returns_root_attr`.

curvine-fuse/src/fs/state/node_state.rs
LJF-050: Thanks. Fixed in e789631d: when `LOOKUP(<child-of-root>, "..")` resolves to the mount root, we return the root inode attributes directly (ino=e789631d sets `attr.ino = FUSE_ROOT_ID`) instead of reconstructing `(0, "/")`. Covered by `fs_lookup_child_of_root_dotdot_returns_root_attr`.

curvine-fuse/src/fs/curvine_file_system.rs
LJF-050: Thanks. Fixed in e789631d: the initial SETLKW wait path now calls `wait_guard.clear_blocked_by()` when deadlock detection is disabled or the blocker is OFD, matching the recheck branch and preventing stale wait-graph edges.

curvine-fuse/src/fs/state/node_state.rs
LJF-050: Thanks. Fixed in e789631d: `DirTree::unlink` now captures an `UnlinkRollback` snapshot and `fs_unlink` restores dentry removal, link/ref counters, canonical repointing, and `mark_delete` state via `restore_unlink` when master delete fails. Added `unlink_rollback_restores_dentry_and_link_count`.

curvine-fuse/src/fs/dcache/dir_tree.rs
LJF-050: Thanks. Fixed in e789631d: when unlink removes the canonical dentry and no cached hardlink replacement exists, `repoint_canonical_from_master` lists master directory entries for cached parent directories and repoints the inode to a surviving link with the same master file id.

LJF-050: Thanks for the suggestion. I kept the current cached scan for the common case and added a master-backed fallback for uncached surviving links (see e789631d). A per-inode reverse index would further reduce work under `dir_write`; I will track that as a follow-up optimization rather than expanding this PR's scope.

curvine-tests/regression/utils/cluster_utils.py
LJF-050: Thanks. Fixed in e789631d: extracted `probe_data_path_ready()` and `ensure_cluster_ready()` now runs the same read/write/fsync probe for already-mounted clusters instead of returning immediately after `check_mount_point`.

LJF-050:
Merged \`upstream/main\` into \`fix/fuse-hardlink-fhandle\` in ebe6f9ff to clear the merge conflict.

Conflict resolution kept the PR's hardlink/unlink rollback, root \`.\`/\`..\` export support, and OFD lock wait-graph fixes, while integrating upstream changes (e.g. \`BlockReaderLocal::open\` with remote fallback, expanded FUSE init flags, dcache TTL/FORGET semantics, and lock-path \`fs_fsync\`).

Verification: \`make format\` passed; \`cargo test -p curvine-fuse --lib\` 380/380 passed.

szbr00 (approved):

PR #1497, opened 2026-08-05
feat(unified): add cache-mode write cache mirror

lzjqsdd:
# Summary

This PR adds a mount-level `write_cache` option for cache-mode read-write mounts. When enabled, sequential create/overwrite writes are committed to UFS first and mirrored to Curvine as a best-effort write cache. The feature targets SDK-mode usage; read-miss `auto_cache` remains a separate mechanism and is not coordinated with the mirror.

# Issue Describe / Design

No linked issue.

This feature keeps UFS as the source of truth while allowing selected cache-mode mounts to proactively populate Curvine on writes. The mirror path is intentionally best-effort: UFS write/complete failures fail the operation, while Curvine mirror failures, queue backpressure timeouts, and completion timeouts fall back to UFS success.

- `write_cache` is mount-scoped for v1 and only valid with `cache_mode + read_write`.
- Append remains UFS-only.
- Non-sequential writes disable the mirror and continue through the existing UFS writer behavior.
- Mirror shutdown is cooperative: dropping the channel sender lets the worker `cancel()` the Curvine writer and clean up the partial file, avoiding leaked writers from task `abort()`.
- Mirror enqueues use a bounded 5s send timeout so fast UFS writers get backpressure instead of silently abandoning the mirror after a few chunks.
- Curvine mirror completion waits up to 5 seconds after UFS completes; on timeout the worker keeps draining and completes/cancels in the background.
- The mirrored CV file forces `storage_policy.ufs_mtime=0` until completion so a partially mirrored file can never pass `cv_valid`.
- An initial process-local `write_cache_pending` guard was removed per review: the client is stateless, and for SDK-mode usage read-miss auto_cache does not overlap with an in-flight mirror.

# Changes

| `crates/common/curvine-model` | Add `write_cache` to mount model, builder, proto conversion, and persisted decode compatibility; add `MountOptionsBuilder::update_only_explicit`. | Existing mounts decode with `write_cache=false`; `mount --update` preserves omitted toggles. |
| `crates/common/curvine-proto` | Add `write_cache` to mount proto and mount table display. | CLI/API mount display includes the new column. |
| `curvine-master` | Validate `write_cache` only for `cache_mode + read_write`. | Invalid mount configs are rejected. |
| `crates/client/curvine-unified-fs` | Add `WriteCacheWriter` with UFS-first mirroring, cooperative shutdown, bounded enqueue backpressure, 5s completion timeout, and forced `ufs_mtime=0` invariant. | Sequential create/overwrite can populate Curvine; append/random/resize preserve existing behavior. |
| `curvine-cli` | Add `--write-cache` mount option; update only sends explicitly provided optional toggle fields. | Users can enable write cache at mount creation/update without resetting omitted options. |
| `curvine-libsdk/java` | Add `MountRequest.Builder.writeCache`; writeCache is encoded only when explicitly provided. | Java clients can set the new mount option; updates preserve existing values when omitted. |
| Tests | Add/extend mount model, compatibility, Java builder, master validation, and mirror invariant tests. | Improves coverage for new mount option. |

# Test verified

| `cargo clippy --release -p curvine-unified-fs -p curvine-model -p curvine-cli --all-targets -- --deny warnings --allow clippy::uninlined-format-args` | PASS | Mirrors pre-commit gate for touched crates. |
| `cargo test -p curvine-model --lib mount::` | PASS | Mount model/builder/merge coverage incl. `update_only_explicit`. |
| `cargo test -p curvine-unified-fs` | PASS | Includes new `mirror_create_opts` invariant tests. |
| `cargo test -p curvine-cli` | PASS | CLI mount parsing coverage. |
| `mvn -q -o test -Dtest=MountRequestTest` | PASS | Java builder coverage. |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{FallbackFsReader, MountCache, MountValue, UnifiedReader, UnifiedWriter};
+use crate::{
+    FallbackFsReader, MountCache, MountValue, UnifiedReader, UnifiedWriter, WriteCacheWriter,
+};
 use bytes::BytesMut;
 use curvine_client_core::file::{CurvineFileSystem, FsClient, FsContext, FsReader};
 use curvine_client_core::ClientMetrics;
@@ -55,6 +57,7 @@ pub struct UnifiedFileSystem {
     enable_read_ufs: bool,
     audit_logging_enabled: bool,
     async_cache_pending: Arc<DashSet<String>>,
+    write_cache_pending: Arc<DashSet<String>>,
     metrics: &'static ClientMetrics,
 }
 
@@ -74,6 +77,7 @@ impl UnifiedFileSystem {
             enable_read_ufs,
             audit_logging_enabled,
             async_cache_pending: Arc::new(DashSet::new()),
+            write_cache_pending: Arc::new(DashSet::new()),
             metrics: FsContext::get_metrics(),
         };
 
@@ -677,9 +681,35 @@ impl UnifiedFileSystem {
                     write_path = ufs_path.full_path().to_owned();
                     let ufs = mount.ufs()?;
                     if flags.append() {
-                        ufs.append(&ufs_path).await
+                        return ufs.append(&ufs_path).await;
+                    }
+
+                    let writer = ufs.create_ufs_writer(&ufs_path, flags.overwrite()).await?;
+
+                    if mount.info.write_cache_enabled() {
+                        let mirror_opts = mount.info.merge_create_opts(opts);
+                        match WriteCacheWriter::new(
+                            writer,
+                            self.cv.clone(),
+                            ufs,
+                            path.clone(),
+                            ufs_path.clone(),
+                            mirror_opts,
+                            self.write_cache_pending.clone(),
+                        )
+                        .await
+                        {
+                            Ok(writer) => Ok(UnifiedWriter::WriteCache(Box::new(writer))),
+                            Err((writer, e)) => {
+                                warn!(
+                                    "failed to open write cache mirror for cv_path={}, ufs_path={}: {}",
+                                    path, ufs_path, e
+                                );
+                                Ok(writer.into_unified())
+                            }
+                        }
                     } else {
-                        ufs.create(&ufs_path, flags.overwrite()).await
+                        Ok(writer.into_unified())
                     }
                 }
             }
@@ -988,7 +1018,15 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
                     .inc();
 
                 if mount.info.auto_cache() {
-                    self.async_cache(&ufs_path)?;
+                    let ufs_key = ufs_path.clone_uri();
+                    if self.write_cache_pending.contains(&ufs_key) {
+                        debug!(
+                            "skip async cache for {} while write cache mirror is pending",
+                            ufs_path
+                        );
+                    } else {
+                        self.async_cache(&ufs_path)?;
+                    }
                 }
 
                 read_path = ufs_path.full_path().to_owned();
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -57,7 +57,6 @@ pub struct UnifiedFileSystem {
     enable_read_ufs: bool,
     audit_logging_enabled: bool,
     async_cache_pending: Arc<DashSet<String>>,
-    write_cache_pending: Arc<DashSet<String>>,
     metrics: &'static ClientMetrics,
 }
 
@@ -77,7 +76,6 @@ impl UnifiedFileSystem {
             enable_read_ufs,
             audit_logging_enabled,
             async_cache_pending: Arc::new(DashSet::new()),
-            write_cache_pending: Arc::new(DashSet::new()),
             metrics: FsContext::get_metrics(),
         };
 
@@ -695,7 +693,6 @@ impl UnifiedFileSystem {
                             path.clone(),
                             ufs_path.clone(),
                             mirror_opts,
-                            self.write_cache_pending.clone(),
                         )
                         .await
                         {
@@ -1018,15 +1015,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
                     .inc();
 
                 if mount.info.auto_cache() {
-                    let ufs_key = ufs_path.clone_uri();
-                    if self.write_cache_pending.contains(&ufs_key) {
-                        debug!(
-                            "skip async cache for {} while write cache mirror is pending",
-                            ufs_path
-                        );
-                    } else {
-                        self.async_cache(&ufs_path)?;
-                    }
+                    self.async_cache(&ufs_path)?;
                 }
 
                 read_path = ufs_path.full_path().to_owned();
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -684,7 +684,7 @@ impl UnifiedFileSystem {
                         return ufs.append(&ufs_path).await;
                     }
 
-                    let writer = ufs.create_ufs_writer(&ufs_path, flags.overwrite()).await?;
+                    let writer = ufs.create(&ufs_path, flags.overwrite()).await?;
 
                     if mount.info.write_cache_enabled() {
                         let mirror_opts = mount.info.merge_create_opts(opts);
@@ -705,11 +705,11 @@ impl UnifiedFileSystem {
                                     "failed to open write cache mirror for cv_path={}, ufs_path={}: {}",
                                     path, ufs_path, e
                                 );
-                                Ok(writer.into_unified())
+                                Ok(writer)
                             }
                         }
                     } else {
-                        Ok(writer.into_unified())
+                        Ok(writer)
                     }
                 }
             }
--- a/crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
+++ b/crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{UfsFileSystem, UfsWriter};
+use crate::{UfsFileSystem, UfsWriter, UnifiedWriter};
 use bytes::{Bytes, BytesMut};
 use curvine_client_core::file::{CurvineFileSystem, FsWriter};
 use curvine_error::{FsError, FsResult};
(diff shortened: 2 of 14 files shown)


crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
```diff
@@ -0,0 +1,401 @@
...
+            }
+        }
+    }
+
+    async fn enqueue_mirror_write(&mut self, pos: i64, data: Bytes) {
+        let Some(mirror) = self.mirror.as_ref() else {
+            return;
+        };
+
+        if let Err(e) = mirror.tx.try_send(MirrorTask::Write { pos, data }) {
```
bigbigxu: The mirror queue is a bounded mpsc channel with capacity `write_chunk_num` (default 8), and writes are enqueued with `try_send`. When UFS write throughput outpaces the Curvine mirror consumer (common for local/high-speed UFS, which is exactly the workload this feature targets), the channel fills after a few chunks (~1MB) and the mirror is silently abandoned via `abandon_mirror`, which also deletes all data mirrored so far. In practice the mirror almost always fails mid-file for fast sequential writers, so the feature rarely takes effect.

Suggest one of: (a) await the enqueue with a bounded timeout (e.g. `time::timeout(5s, tx.send(..)).await`) to provide backpressure, (b) a much larger channel, or (c) on mirror failure keep the partially mirrored data instead of deleting it (it is already unreadable until complete because `ufs_mtime=0`).

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
```diff
@@ -0,0 +1,401 @@
...
+                "failed to enqueue write cache complete for cv_path={}, ufs_path={}: {}",
+                self.cv_path, self.ufs_path, e
+            );
+            mirror.handle.abort();
+            self.pending.remove(&self.pending_key);
+            self.spawn_cleanup_cv();
+            return;
+        }
+
+        match time::timeout(WRITE_CACHE_COMPLETE_TIMEOUT, rx).await {
```
bigbigxu, line 178: `complete_with_attr` waits up to 30s for the mirror worker to drain queued writes and complete the CV file, blocking the FUSE close path. With a slow or jittery Curvine, every file close can stall for seconds (up to 30s) even though the UFS write already succeeded. For workloads with many small files this is a significant latency regression on a synchronous path.

Suggest a much shorter timeout (2-5s), or make completion fully asynchronous (close returns immediately and the mirror worker completes/cleans up on its own).

curvine-cli/src/cmds/mount.rs
```diff
@@ -565,6 +571,10 @@ impl MountCommand {
             opts = opts.access_mode(access_mode);
         }
 
+        if let Some(write_cache) = self.write_cache {
```
bigbigxu, line 581: `MountOptionsBuilder::new()` defaults `write_cache` to `Some(false)`, and `to_mnt_opts` always builds from that builder, so a `mount --update` without an explicit `--write-cache` silently resets an enabled write_cache to false (via `merge_with` on the master). `auto_cache`/`access_mode` follow the same pattern, but for write_cache the default direction is "off", which silently disables a previously enabled feature. Consider sending only explicitly provided fields on update, or at least documenting this behavior.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
```diff
@@ -0,0 +1,401 @@
...
+        };
+        let chunk_size = primary.chunk_size();
+        let pos = primary.pos();
+        let path = primary.path().clone();
+        let status = primary.status().clone();
+
+        let cv_flags = OpenFlags::new_write_only()
+            .set_create(true)
+            .set_overwrite(true);
+        let cv_writer = match cv.open_with_opts(&cv_path, opts, cv_flags).await {
```
bigbigxu: Correctness of concurrent reads while a mirror is in-flight relies on an implicit contract: the mirrored CV file has `ufs_mtime=0` until `complete_mirror` sets it, so `cv_valid` rejects the partial file. This is fragile - if `merge_create_opts` ever passes a non-zero `ufs_mtime` (e.g. from the sync path), readers could hit a partially mirrored file. Consider explicitly forcing `storage_policy.ufs_mtime=0` on the mirror open and adding a comment + test for this invariant.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
```diff
@@ -0,0 +1,401 @@
...
+                    WRITE_CACHE_COMPLETE_TIMEOUT, self.cv_path, self.ufs_path
+                );
+                mirror.handle.abort();
+                self.pending.remove(&self.pending_key);
+                self.spawn_cleanup_cv();
+            }
+        }
+    }
+
+    async fn enqueue_mirror_write(&mut self, pos: i64, data: Bytes) {
```
bigbigxu, line 210: `enqueue_mirror_write` and `enqueue_mirror_flush` are `async fn` without any `.await` inside (they only `try_send`); consider making them sync, which also avoids clippy `unused_async`.

bigbigxu (commented):
`write_cache_writer.rs` (392 lines) implements the core mirroring logic - worker loop, abandon paths, 30s timeout, seek/resize fallback - with no unit tests. The PR tests cover mount model/proto compatibility, CLI parsing, and master validation only. Suggest at least a few unit tests with mock UFS/CV writers covering: successful mirror, mirror write failure fallback, enqueue overflow -> abandon + cleanup, seek -> abandon, and complete timeout.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -988,7 +1018,15 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
                     .inc();
 
                 if mount.info.auto_cache() {
-                    self.async_cache(&ufs_path)?;
+                    let ufs_key = ufs_path.clone_uri();
```
bigbigxu: `write_cache_pending` is process-local. While client A is mirroring file X (CV file exists with partial data, `ufs_mtime=0`), another client B can read-miss X and submit an auto_cache transfer job that re-creates/overwrites the same CV path concurrently with A's mirror writer. After A completes and sets `ufs_mtime`, the CV file may hold interleaved/corrupt data which passes `cv_valid(None)` when `read_verify_ufs=false`, so readers can hit corrupted data. Please confirm the transfer path is mutually exclusive with client-side writes, or make the mirror marker visible cluster-wide.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -74,6 +77,7 @@ impl UnifiedFileSystem {
             enable_read_ufs,
             audit_logging_enabled,
             async_cache_pending: Arc::new(DashSet::new()),
+            write_cache_pending: Arc::new(DashSet::new()),
```
bigbigxu: The client is stateless and should not retain these properties.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
```diff
@@ -0,0 +1,401 @@
...
+        if let Some(mirror) = self.mirror.take() {
+            warn!(
+                "disable write cache mirror for cv_path={}, ufs_path={}: {}",
+                self.cv_path,
+                self.ufs_path,
+                reason.as_ref()
+            );
+            mirror.handle.abort();
+            self.pending.remove(&self.pending_key);
+            self.spawn_cleanup_cv();
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in ae1f7a9e: `abandon_mirror` and `complete_mirror` no longer call `task::abort()`. Dropping the channel sender makes the worker exit through its normal path, where it calls `writer.cancel().await` and cleans up the partial CV file; on completion timeout the worker keeps draining queued writes and completes/cancels in the background, so `FsWriter` is never dropped unclosed.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
lzjqsdd: Thanks. Fixed in ae1f7a9e: mirror enqueue now uses a bounded `time::timeout(5s, tx.send(..))` instead of `try_send`, providing backpressure so fast UFS writers are not silently abandoned after a few chunks; the mirror is only abandoned on a real timeout or a closed channel.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
lzjqsdd: Thanks. Fixed in ae1f7a9e: `WRITE_CACHE_COMPLETE_TIMEOUT` is reduced from 30s to 5s. On timeout the caller returns immediately and the worker cooperatively finishes draining and completes (or cancels + cleans up) in the background, so close latency is bounded.

curvine-cli/src/cmds/mount.rs
lzjqsdd: Thanks. Fixed in 86b0afcc: `mount --update` now sends only explicitly provided optional toggle fields (auto_cache/access_mode/write_cache) via `MountOptionsBuilder::update_only_explicit()`, so an omitted `--write-cache` preserves the existing value instead of resetting it to false. The Java SDK `MountRequest` likewise encodes `writeCache` only when explicitly set, and the `--write-cache` help text documents the behavior.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
lzjqsdd: Thanks. Fixed in ae1f7a9e: the mirror open path now forces `storage_policy.ufs_mtime=0` via `mirror_create_opts()` before opening the CV target, with a comment documenting the invariant and unit tests covering both the zeroing and field preservation.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
lzjqsdd: Thanks. Fixed in ae1f7a9e: `enqueue_mirror_write`/`enqueue_mirror_flush` are now genuinely async (bounded `tx.send` await), so the `unused_async` concern no longer applies.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
lzjqsdd: Thanks. The process-local guard was removed in ae1f7a9e together with the `write_cache_pending` set (see the statelessness thread). Write cache targets SDK-mode usage where read-miss auto_cache is a separate feature and does not overlap with an in-flight mirror, so the cross-client race is out of scope; the mirror remains best-effort with UFS as the source of truth.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
lzjqsdd: Thanks. Fixed in ae1f7a9e: `write_cache_pending` and its read-path check are removed; the client retains no per-file state for the mirror. `async_cache_pending` is kept as-is since it predates this PR and only throttles job submissions.

lzjqsdd:
Thanks for the review. Added unit tests for the new `mirror_create_opts` invariant in ae1f7a9e (forced zero `ufs_mtime` + field preservation). Full mirror lifecycle tests (success path, enqueue overflow, seek/resize fallback, complete timeout) would need a Curvine test double for `FsWriter`/`CurvineFileSystem`, which does not exist in this crate; those paths are best covered E2E with a real cluster. Happy to add a mock layer in a follow-up if you'd like.

bigbigxu (approved):

PR #1499, opened 2026-08-05
fix(test): restore parallel load fault hooks and mount test configs

LJF-050:
# Summary

Restore parallel UFS-to-Curvine load execution and the fault-injection hooks required by regression tests, and fix mount integration tests that reused HDFS mount options for S3 remount scenarios.

# Issue Describe / Design

This is a test-infrastructure repair for two daily/integration regressions:

- `canceled_final_parallel_stream_does_not_mark_cache_valid` failed because parallel load fault points were removed from the authoritative data-transfer runner while the test remained.
- `test_mount_umount_s3_hdfs_and_path_mapping` failed because `mnt_opt` was shadowed with HDFS properties and later S3 mounts reused the wrong configuration after S3 endpoint validation became mandatory.

**Root cause**
- Load-task refactor dropped parallel fan-out and fault catalog entries still referenced by `load_task_runner_fault_test`.
- Mount test used one `mnt_opt` variable for both backends.

**Fix approach**
- Reintroduce parallel load in `curvine-data-transfer` with cancellation guards that avoid stamping partial files as cache-valid.
- Wire fault injection through `curvine-data-transfer` and `curvine-worker` feature flags.
- Split mount options into `s3_mnt_opt` and `hdfs_mnt_opt`.

# Changes

| `crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs` | Restore parallel UFS→CV load and fault hooks | Large-file UFS→CV loads can fan out again; fault tests can inject delays |
| `crates/server/curvine-data-transfer/Cargo.toml`, `src/lib.rs` | Add fault-injection feature wiring | No runtime change unless fault injection is enabled |
| `curvine-server/Cargo.toml`, `curvine-worker/Cargo.toml` | Propagate fault-injection feature | Enables worker fault catalog registration |
| `curvine-tests/tests/mount_test.rs` | Use separate S3/HDFS mount option builders | Test-only fix; product mount validation unchanged |

# Test verified

| `make format` | PASS | |
| `cargo test -p curvine-server --features fault-injection --test load_task_runner_fault_test canceled_final_parallel_stream_does_not_mark_cache_valid` | PASS | Exact HEAD `fd9ab2d2` |
| Profile `integration` after run `20260805T041242Z-482d4afa` | PASS | Original daily/integration failures absent |
| Profile `daily` after run `20260805T041938Z-ed72b5d2` | PASS | Original daily/integration failures absent |
| Profile `fast` after run `20260805T041805Z-5c7516c2` | PASS | |
| Profile `csi` after run `20260805T051927Z-2d3472da` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs
+++ b/crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs
@@ -216,9 +216,11 @@ impl LoadTaskRunner {
             && target_path.is_cv()
             && self.max_parallel_streams() > 1
         {
-            let src_len = self.get_ufs()?.get_status(&source_path).await?.len;
-            if self.effective_streams(src_len) > 1 {
-                return self.run_parallel(&source_path, &target_path, src_len).await;
+            let initial_source = self.get_ufs()?.get_status(&source_path).await?;
+            if self.effective_streams(initial_source.len) > 1 {
+                return self
+                    .run_parallel(&source_path, &target_path, &initial_source)
+                    .await;
             }
         }
 
@@ -848,8 +850,9 @@ impl LoadTaskRunner {
         &self,
         source_path: &Path,
         target_path: &Path,
-        src_len: i64,
+        initial_source: &FileStatus,
     ) -> FsResult<bool> {
+        let src_len = initial_source.len;
         let streams = self.effective_streams(src_len);
         let block_size = self.task.info.job.block_size.max(1);
         let seg = Self::segment_len(src_len, streams, block_size);
@@ -992,7 +995,18 @@ impl LoadTaskRunner {
             );
         }
 
-        let ufs_mtime = self.get_ufs()?.get_status(source_path).await?.mtime;
+        let final_source = self.get_ufs()?.get_status(source_path).await?;
+        if final_source.len != initial_source.len || final_source.mtime != initial_source.mtime {
+            return err_box!(format!(
+                "Task {} parallel load source changed during transfer (initial len={}, initial mtime={}, final len={}, final mtime={}); refusing to mark cache valid",
+                self.task.info.task_id,
+                initial_source.len,
+                initial_source.mtime,
+                final_source.len,
+                final_source.mtime,
+            ));
+        }
+        let ufs_mtime = initial_source.mtime;
         let attr_opts = SetAttrOptsBuilder::new().ufs_mtime(ufs_mtime).build();
         self.fs.set_attr(target_path, attr_opts).await?;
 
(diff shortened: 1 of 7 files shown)


crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs
```diff
@@ -767,6 +792,239 @@ impl LoadTaskRunner {
...
+        if written != src_len {
+            return err_box!(
+                "Task {} parallel load incomplete: wrote {} of {} bytes; refusing to mark cache valid",
+                self.task.info.task_id,
+                written,
+                src_len
+            );
+        }
+
+        let ufs_mtime = self.get_ufs()?.get_status(source_path).await?.mtime;
```
szbr00: The parallel readers are opened independently, but this stamps the target with a fresh post-copy mtime. If the source object is overwritten during the transfer with the same length, different segments can come from different versions and the final stat can return the new mtime, causing the mixed data to pass cache validation. Please capture the initial source status and verify that its length/mtime are unchanged before setting that original mtime, or use version/ETag-constrained reads; otherwise fail and leave the cache invalid.

crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs
```diff
@@ -767,6 +792,239 @@ impl LoadTaskRunner {
...
+    }
+
+    fn stream_count(src_len: i64, min_bytes: i64, cap: usize) -> usize {
+        let min_bytes = min_bytes.max(1);
+        let cap = cap.max(1);
+        let by_size = (src_len / min_bytes).max(0) as usize;
+        by_size.clamp(1, cap)
+    }
+
+    fn segment_len(src_len: i64, streams: usize, block_size: i64) -> i64 {
```
szbr00, line 839: The refactor removed the unit coverage for this range-planning logic, but this restore does not bring it back. Since block alignment and complete, non-overlapping range coverage are correctness-critical for parallel writes, please restore tests covering contiguous full-file coverage, block-aligned boundaries, small files, and stream-count bounds.

crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs
LJF-050: Thanks. Fixed in 5507735f: `run_parallel` now takes the initial UFS `FileStatus` captured before fan-out, re-stats the source after all segments complete, and refuses to stamp `ufs_mtime` when length or mtime changed during transfer (leaving the cache invalid).

crates/server/curvine-data-transfer/src/worker/task/load_task_runner.rs
LJF-050: Thanks. Fixed in 5507735f: restored unit tests for `stream_count`, `segment_len`, and the `plan()` helper covering contiguous full-file coverage, block-aligned boundaries, small files, and stream-count bounds (`cargo test -p curvine-data-transfer load_task_runner` — 9 passed).

szbr00 (approved):

PR #1500, opened 2026-08-05
feat(fuse): improve setattr

201341:


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -790,37 +790,59 @@ impl CurvineFileSystem {
...
-            return err_fuse!(
-                libc::EPERM,
-                "setattr uid change requires privilege for uid {}",
-                caller_uid
-            );
-        }
-
-        if (valid & FATTR_GID) != 0 {
+            let no_op = matches!(target_uid, Some(uid) if uid == file_uid)
+                || matches!(target_uid, Some(uid) if uid == u32::MAX)
```
szbr00: Treating `u32::MAX` as a no-op only in this permission check is unsafe. The `FATTR_UID`/`FATTR_GID` bits remain set, so `fuse_setattr_to_opts` converts the sentinel to `"4294967295"` and `fs_set_attr` persists it as the actual owner/group (while also applying chown mode-bit clearing). Please normalize these sentinels before conversion by omitting the corresponding attribute, or reject them, and add a test that verifies the generated `SetAttrOpts`.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -790,37 +790,59 @@ impl CurvineFileSystem {
...
             }
-            if let Some(gid) = target_gid {
-                if !FuseUtils::caller_in_file_group(caller_gid, gid, caller_pid) {
+        }
+
+        if (valid & FATTR_GID) != 0 {
+            let gid_no_op = matches!(target_gid, Some(gid) if gid == file_gid)
+                || matches!(target_gid, Some(gid) if gid == u32::MAX)
+                || target_gid.is_none();
+            if !gid_no_op {
```
szbr00: This branch skips the inode-owner check when the requested GID equals the current GID. Linux `setattr_prepare` still requires an unprivileged caller to own the inode for an `ATTR_GID` request, even when the value is unchanged. Keep the `caller_uid != file_uid` check outside the `!gid_no_op` branch; only the target-group membership check should be skipped for an unchanged GID.

szbr00 (approved):

PR #1501, opened 2026-08-05
fix(master): clean up legacy path terminal load jobs (#1028)

### LINKED ISSUE #1028: [BUG]: Master OOM — completed load jobs never released until 24h TTL [LiAuto]
### Describe the bug
Under auto_cache read-through, master creates a JobContext (kept in an in-memory DashMap `self.jobs`) plus a tokio task for every unique cache-miss file. The cleanup task only evicts by `create_time + job_life_ttl` (default 24h) and never checks completion state. Completed jobs are never released early, so `self.jobs` grows monotonically with unique miss files → RSS climbs linearly to OOM.

### To Reproduce
1. Mount a UFS with auto_cache enabled.
2. Run high-QPS reads over a large set of unique small files (all cache-miss).
3. Watch master RSS grow linearly; no jobs are evicted within the 24h TTL.

### Expected behavior
Terminal (Completed/Failed/Canceled) jobs should be reclaimed promptly, or `self.jobs` should be capacity-bounded, so master memory stays bounded under sustained cache-miss load.

### OS Version
- Curvine branch: main
- Relevant code:
  - curvine-server/src/master/job/job_manager.rs:281 (time-only cleanup)
  - curvine-server/src/master/job/job_store.rs (no removal on completion)
  - curvine-client/src/unified/unified_filesystem.rs:692 (per-miss async_cache)
  - curvine-common/src/conf/job_conf.rs:60 (default job_life_ttl = 24h)

### Additional context
Observed: ~2M queued jobs after ~2.1h runtime, 0 evicted (2.1h ≪ 24h TTL); heap roughly doubled as job count doubled. Suggested fixes: evict on `state.is_terminal() && now > completed_time + short_ttl`; and/or cap `self.jobs` with LRU.

RPG-Nia:
# Summary

Bounds terminal job retention in the legacy Master LoadJob path to prevent completed, failed, or canceled jobs from staying in Master memory until the full job life TTL.

# Issue Describe / Design

Closes #1028

Root cause:

- Legacy Master LoadJobs are stored in Master-side `JobStore`.
- The cleanup task previously used `job_life_ttl` for all jobs, including terminal states.
- With the default `job_life_ttl` of 24h, many auto-cache or load jobs can remain in Master memory long after completion, causing memory growth under high-volume cache-miss workloads.

Relation to #1040:

- #1040 introduces the Transfer direction, where Load/Export execution is moved out of Master when `transfer.enabled = true`.
- This PR does not change the Transfer path.
- **This PR fixes the legacy Master LoadJob path that is still used when** `transfer.enabled = false`.
- Since the existing LoadJob API surface is still present, terminal job retention should remain bounded for clusters that have not enabled Transfer yet.

Fix approach:

- Add `job.terminal_retention` for terminal legacy LoadJobs, defaulting to `10m`.
- Keep running/non-terminal jobs protected by the existing `job_life_ttl`.
- Use the job progress update time as the terminal timestamp, with create time as a defensive fallback.
- Remove registered callbacks when an expired job is removed from `JobStore`.

# Changes

| `crates/common/curvine-config/src/job_conf.rs` | Add `terminal_retention` config for legacy Master LoadJobs. | Terminal job retention can be configured independently from `job_life_ttl`; existing configs keep working through defaults. |
| `curvine-master/src/master/job/job_manager.rs` | Clean terminal jobs using `terminal_retention`, while preserving `job_life_ttl` for running jobs. | Completed/Failed/Canceled legacy LoadJobs are reclaimed earlier; running jobs keep the previous behavior. |
| `curvine-master/src/master/job/job_store.rs` | Add removal path that clears both job entries and callbacks. | Avoids retaining callbacks after expired jobs are cleaned. |
| `crates/common/curvine-config/tests/job_conf_test.rs` / `curvine-master/src/master/job/job_manager.rs` | Add regression coverage for config parsing, terminal cleanup, running-job retention, and callback cleanup. | Protects the cleanup behavior from regressions. |

# Test verified

| Local Rust validation | PASS | Ran formatting, targeted unit tests, package check, clippy, and diff whitespace check successfully. |
| Local legacy LoadJob cleanup verification | PASS | Started local Master and Worker with `transfer.enabled = false`, shortened cleanup intervals only for manual testing, submitted a legacy `cv load` job from a mounted `file://` path, and verified the terminal job is removed from `load-status` after the retention window. |
image

# Dependencies

- Related PRs: None
- Related issues: #1028, #1040
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/job/job_manager.rs
+++ b/curvine-master/src/master/job/job_manager.rs
@@ -14,7 +14,7 @@
 
 use crate::common::UfsFactory;
 use crate::master::fs::MasterFilesystem;
-use crate::master::{JobStore, LoadJobRunner, MountManager};
+use crate::master::{JobContext, JobStore, LoadJobRunner, MountManager};
 use core::time::Duration;
 use curvine_config::ClusterConf;
 use curvine_core_error::{err_box, err_ext, CommonResult};
@@ -297,6 +297,40 @@ struct JobCleanupTask {
     terminal_retention_ms: i64,
 }
 
+struct ExpiredJob {
+    job_id: String,
+    run_id: u64,
+}
+
+impl JobCleanupTask {
+    fn is_expired(&self, job: &JobContext, now: i64) -> bool {
+        let state: JobTaskState = job.state.state();
+        let expired_at = if state.is_finish() {
+            let terminal_at = if job.progress.update_time > 0 {
+                job.progress.update_time
+            } else {
+                job.info.create_time
+            };
+            terminal_at.saturating_add(self.terminal_retention_ms)
+        } else {
+            job.info.create_time.saturating_add(self.running_ttl_ms)
+        };
+
+        now > expired_at
+    }
+
+    fn remove_expired_job(&self, expired_job: ExpiredJob) -> FsResult<()> {
+        let now = LocalTime::mills() as i64;
+        if let Some(v) = self.jobs.remove_job_if(&expired_job.job_id, |job| {
+            job.run_id == expired_job.run_id && self.is_expired(job, now)
+        })? {
+            debug!("Removing expired job: {:?}", v.1.info);
+        }
+
+        Ok(())
+    }
+}
+
 impl LoopTask for JobCleanupTask {
     type Error = FsError;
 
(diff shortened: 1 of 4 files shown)


curvine-master/src/master/job/job_manager.rs
```diff
@@ -301,13 +306,25 @@ impl LoopTask for JobCleanupTask {
...
+            };
+
+            if now > expired_at {
                 jobs_to_remove.push(job.info.job_id.clone());
             }
         }
 
         for job_id in jobs_to_remove {
-            if let Some(v) = self.jobs.remove(&job_id) {
+            if let Some(v) = self.jobs.remove_job(&job_id)? {
```
szbr00: Cleanup records only the `job_id`, but job IDs are deterministic and a terminal entry can be replaced before `remove_job` executes. In that race, cleanup removes the fresh running job and its callbacks. Please carry the scanned `run_id` and remove the entry only if the same run is still expired, with callback cleanup tied to that conditional removal.

RPG-Nia: Thanks for the feedback. Fixed by carrying the scanned `run_id` and removing the job only when the current entry still matches that run and remains expired. Callback cleanup now only happens after that conditional removal succeeds. I also added a regression test for same-`job_id` replacement.

szbr00 (approved):

PR #1507, opened 2026-08-05
fix(cli): format fs stat permissions from mode

### LINKED ISSUE #1191: [BUG]: `cv fs stat` reports hard-coded permissions instead of `FileStatus.mode`
**Describe the bug**

`bin/cv fs stat` fetches a `FileStatus`, but formats the `Permission` line with a hard-coded `rwxr-xr-x` suffix rather than using `status.mode`.

As a result, after a successful chmod operation, the FUSE mount can report the updated mode while `cv fs stat` still reports `0755`.

**To Reproduce**

1. Create a file in Curvine.
2. Set its mode:

   ```sh
   bin/cv fs chmod 777 /path/to/file
   ```

3. Inspect the file:

   ```sh
   bin/cv fs stat /path/to/file
   ```

4. Observe:

   ```text
   Permission: -rwxr-xr-x
   ```

**Expected behavior**

The permission string should be formatted from the fetched `FileStatus.mode`. For the example above, it should be:

```text
Permission: -rwxrwxrwx
```

- OS: Linux
- Curvine: current mainline source build
- Relevant component: `curvine-cli/src/cmds/fs/stat.rs`

**Additional context**

The command already fetches `FileStatus` and prints its other fields, so using `status.mode` only requires local formatting and does not add an RPC or metadata lookup.

Please confirm whether the hard-coded `0755` output was an intentional performance or compatibility tradeoff.

Related: #812 tracks broader FUSE permission and metadata compatibility.

sf1999817:
## Summary

`cv fs stat` previously printed a hard-coded `rwxr-xr-x` permission string, so chmod changes were not reflected in stat output.

This patch adds a shared fs CLI permission formatter based on `FileStatus.mode` and uses it in both `stat` and `ls` output.

Fixes #1191.

Thanks @zztaki for reporting the issue.

## Scope

This only changes CLI-side permission formatting. It does not change chmod, FUSE, or metadata semantics.

## Tests

- `cargo fmt`
- `cargo test -p curvine-cli test_format_permission`
- `cargo check -p curvine-cli`

Manual verification on a dev Curvine cluster:
- regular bits: `777`, `755`, `000`

- `cv fs chmod 777 /tmp/stat-mode-dir`
  - `cv fs stat /tmp/stat-mode-dir` prints `Permission: drwxrwxrwx`
  - `cv fs ls -l /tmp` prints `drwxrwxrwx`

- `cv fs chmod 755 /tmp/stat-mode-dir`
  - `cv fs stat /tmp/stat-mode-dir` prints `Permission: drwxr-xr-x`
  - `cv fs ls -l /tmp` prints `drwxr-xr-x`
  

- `cv fs chmod 000 /tmp/stat-mode-dir`
  - `cv fs stat /tmp/stat-mode-dir` prints `Permission: d---------`
  - `cv fs ls -l /tmp` prints `d---------`

- special bits:
  - `2775` -> `drwxrwsr-x`
  - `1755` -> `drwxr-xr-t`
  - `2664` -> `drw-rwSr--`
  - `4755` -> `drwsr-xr-x`
  - `1666` -> `drw-rw-rwT`

### DIFF
--- a/curvine-cli/src/cmds/fs/common.rs
+++ b/curvine-cli/src/cmds/fs/common.rs
@@ -66,6 +66,52 @@ pub fn format_size(size: u64) -> String {
     }
 }
 
+/// Formats Unix permission bits to a rwx string.
+pub fn format_permission(mode: u32) -> String {
+    let mode = mode & 0o7777;
+
+    let user_read = if mode & 0o400 != 0 { 'r' } else { '-' };
+    let user_write = if mode & 0o200 != 0 { 'w' } else { '-' };
+    let user_exec = match (mode & 0o4000 != 0, mode & 0o100 != 0) {
+        (true, true) => 's',
+        (true, false) => 'S',
+        (false, true) => 'x',
+        (false, false) => '-',
+    };
+
+    let group_read = if mode & 0o040 != 0 { 'r' } else { '-' };
+    let group_write = if mode & 0o020 != 0 { 'w' } else { '-' };
+    let group_exec = match (mode & 0o2000 != 0, mode & 0o010 != 0) {
+        (true, true) => 's',
+        (true, false) => 'S',
+        (false, true) => 'x',
+        (false, false) => '-',
+    };
+
+    let other_read = if mode & 0o004 != 0 { 'r' } else { '-' };
+    let other_write = if mode & 0o002 != 0 { 'w' } else { '-' };
+    let other_exec = match (mode & 0o1000 != 0, mode & 0o001 != 0) {
+        (true, true) => 't',
+        (true, false) => 'T',
+        (false, true) => 'x',
+        (false, false) => '-',
+    };
+
+    [
+        user_read,
+        user_write,
+        user_exec,
+        group_read,
+        group_write,
+        group_exec,
+        other_read,
+        other_write,
+        other_exec,
+    ]
+    .iter()
+    .collect()
+}
+
 /// Formats a Unix epoch timestamp in milliseconds using the local timezone.
 pub fn format_epoch_ms_local(timestamp_ms: i64, fmt: &str) -> String {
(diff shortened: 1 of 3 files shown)


curvine-cli/src/cmds/fs/common.rs
```diff
@@ -66,6 +66,27 @@ pub fn format_size(size: u64) -> String {
     }
 }
 
+/// Formats Unix permission bits to a rwx string.
+pub fn format_permission(mode: u32) -> String {
+    let mode = mode & 0o777;
```
szbr00: Masking the mode to `0o777` drops setuid, setgid, and sticky bits, although `FileStatus.mode` retains `0o7777` and Curvine relies on these bits for setgid inheritance and sticky-directory behavior. For example, mode `0o2775` will be displayed as `rwxrwxr-x` instead of `rwxrwsr-x`. Please preserve the special bits and render `s`/`S`/`t`/`T` in the execute positions; the `0o1755` test should be updated accordingly.

curvine-cli/src/cmds/fs/ls.rs
```diff
@@ -322,24 +322,7 @@ async fn print_file_entry(
...
-            permissions.insert(0, '-');
-        }
-        if mode & 0o4 != 0 {
-            permissions.insert(0, 'r');
-        } else {
-            permissions.insert(0, '-');
-        }
-        mode >>= 3;
-    }
+    let permissions = crate::cmds::fs::common::format_permission(file.mode as u32);
```
szbr00, line 323: This refactor leaves the previous `mode` and initial `permissions` bindings unused before `permissions` is shadowed here, introducing compiler warnings. Please remove those two obsolete declarations.

szbr00 (approved):

PR #1526, opened 2026-08-07
fix(master): recover cache-mode files after worker loss (#1519)

### LINKED ISSUE #1519: [BUG]: cache_mode retains valid cache metadata after last worker replica is lost, preventing auto-cache re-admission
**Describe the bug**

On a `cache_mode` UFS mount with `auto_cache=true` and a single cache replica, losing a Curvine worker can leave files marked as valid in Curvine metadata even when no readable cache block remains.

The client initially behaves well: a read directed to the unavailable cache block falls back to the authoritative UFS/S3 object. However, that fallback does not invalidate the stale Curvine-valid state or submit a replacement cache-load job. Each later read repeats the same failed Curvine attempt followed by an S3 fallback, so the cache does not recover through normal workload reads.

This was reproduced through a direct SDK integration using `UnifiedFileSystem`, not through a FUSE/S3 mount.

**To Reproduce**

1. Create a read-only `cache_mode` S3/UFS mount with `auto_cache=true` and one replica.
2. Read a set of UFS files until they are cached on a worker.
3. Permanently lose or replace the worker that holds the only replica, for example by replacing a Kubernetes worker with node-local cache storage.
4. Wait for the Master to mark the old worker lost.
5. Read one of the affected files through `UnifiedFileSystem`.

**Actual behavior**

- The file's cached metadata still passes `FileStatus::cv_valid()`.
- `UnifiedFileSystem::get_cv_reader()` chooses the Curvine reader rather than the normal cache-miss path.
- The Curvine block read fails because there is no available worker for the block.
- `FallbackFsReader` successfully serves the read from UFS/S3.
- `FallbackFsReader` does not invalidate the cached state or call `async_cache`.
- Subsequent reads repeat the S3 fallback indefinitely.

As a workaround, we manually ran `cv fs free` for the affected dataset paths. The next workload reads then used the normal cache-miss path and automatically rebuilt the cache: 3,338 training shards / 3.96 TB were re-admitted in roughly 45 minutes.

**Expected behavior**

When a cache-mode file has no live, readable Curvine block locations after a worker is declared lost:

1. Curvine should preserve the UFS mapping and source metadata.
2. It should mark only the Curvine cache state as invalid.
3. The next read should follow the existing cache-miss path: serve from UFS/S3 and submit a deduplicated background LoadJob.
4. Later reads should use the new cache copy.

A single worker loss should not require an operator to discover and manually free every affected cache prefix.

**Root cause**

`FileStatus::cv_valid()` checks cached-state flags, expiry, and UFS metadata, but does not verify that the file still has a live/readable block location. After lost-worker cleanup removes the worker's locations, the cached file can remain valid enough to enter `FallbackFsReader`.

The fallback reader is deliberately pass-through: it continues the current read from UFS, but does not promote the data back to Curvine.

**Suggested fix**

Prefer a Master-side repair when a worker is marked lost:

- After removing that worker's block locations, identify cache-mode files whose blocks have no remaining live locations.
- Clear only their Curvine-valid/cache-present state; retain the UFS mapping and metadata.
- Let the existing `UnifiedFileSystem` cache-miss path submit the normal, deduplicated auto-cache LoadJob.

A client-side fallback-triggered invalidation/reload could be a secondary safeguard, but it needs deduplication and should not create one LoadJob per reader.

**Related issues**

- #950 identifies that `FallbackFsReader` is pass-through and does not promote reads into cache. This issue is narrower: stale cache validity after worker loss prevents the normal cache-miss path from running at all.
- #869 identifies node-local worker storage loss during Kubernetes rescheduling.

**Environment**

- Curvine image: `v0.3.6-alpha`
- UFS: S3
- Mount mode: read-only `cache_mode`, `auto_cache=true`
- Client: direct SDK / `UnifiedFileSystem`
- Cache worker storage: node-local NVMe

HarryMWinters:
# Summary

Recover cache-mode files when their last Curvine block replica is lost.
The master preserves authoritative UFS metadata, so the next read falls back
to UFS and automatically reloads the cache.

# Issue Describe / Design

Closes #1519

Root cause: worker-loss cleanup deleted location records but left UFS-backed
cache files marked as valid Curvine copies. Clients therefore fell back to UFS
without scheduling a replacement load job.

Fix approach: invalidate only cache-mode files with an unreadable block,
preserve their UFS metadata, clear their cache blocks, and avoid normal
replica recovery for those discarded cache blocks.

The integration test also exposed a stale round-robin worker-selection cursor
after worker removal; normalize that cursor before selecting the replacement
load target.

# Changes

| Master metadata | Invalidate unreadable cache-mode files after worker loss | Re-admits cache from UFS |
| Worker-loss handling | Exclude invalidated cache blocks from replica recovery | FS-mode and Curvine-only recovery unchanged |
| Worker policy | Wrap round-robin cursor after membership shrink | Prevents reload-job placement failure |
| Tests | Add metadata, policy, and local-cluster integration coverage | Regression protection |

# Test verified

| `cargo test -p curvine-server --test master_fs_test lost_worker_` | PASS | 4 tests |
| `cargo test -p curvine-master choose_wraps_cursor_after_worker_removal` | PASS | Cursor regression |
| `cargo test -p curvine-tests --test write_cache_test test_cache_mode_recaches_after_lost_worker_cleanup -- --exact` | PASS | Master + 2 workers + local UFS + load job |

# Dependencies

- Related PRs: None
- Related issues: #1519
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -80,6 +80,12 @@ pub struct BlockReportResult {
     pub delete_blocks: Vec<i64>,
 }
 
+#[derive(Default)]
+pub struct LostWorkerLocationCleanup {
+    pub removed_block_ids: Vec<i64>,
+    pub replication_block_ids: Vec<i64>,
+}
+
 pub(crate) enum BlockInodeState {
     File,
     Missing,
@@ -1557,9 +1563,27 @@ impl MasterFilesystem {
             .unwrap_or(false)
     }
 
-    pub fn delete_locations(&self, worker_id: u32) -> FsResult<Vec<i64>> {
-        let fs_dir = self.fs_dir.write();
-        fs_dir.delete_locations(worker_id)
+    pub fn delete_locations(&self, worker_id: u32) -> FsResult<LostWorkerLocationCleanup> {
+        let mut fs_dir = self.fs_dir.write();
+        let removed_block_ids = fs_dir.delete_locations(worker_id)?;
+        let invalidated = fs_dir.invalidate_lost_cache_files(&removed_block_ids)?;
+        let replication_block_ids = removed_block_ids
+            .iter()
+            .copied()
+            .filter(|block_id| !invalidated.invalidated_block_ids.contains(block_id))
+            .collect();
+        drop(fs_dir);
+
+        if !invalidated.delete_result.blocks.is_empty() {
+            self.worker_manager
+                .write()
+                .remove_blocks(&invalidated.delete_result);
+        }
+
+        Ok(LostWorkerLocationCleanup {
+            removed_block_ids,
+            replication_block_ids,
+        })
     }
 
     pub fn set_attr<T: AsRef<str>>(&self, path: T, opts: SetAttrOpts) -> FsResult<FileStatus> {
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -16,7 +16,7 @@ use crate::master::fs::context::ValidateAddBlock;
 use crate::master::fs::policy::ChooseContext;
 use crate::master::journal::JournalSystem;
 use crate::master::meta::inode::{InodeFile, InodePath, InodePtr, InodeView, PATH_SEPARATOR};
-use crate::master::meta::FsDir;
+use crate::master::meta::{CacheInvalidationResult, FsDir};
 
 use crate::master::fs::DeleteResult;
 use crate::master::meta::parse_glob_pattern;
@@ -134,6 +134,8 @@ const FULL_BLOCK_RECONCILE_QUEUE_SIZE: usize = 128;
 impl MasterFilesystem {
     // Max block-report location updates applied under a single fs_dir write lock.
     const BLOCK_REPORT_WRITE_CHUNK: usize = 4096;
+    // Max lost-worker block ids inspected under a single fs_dir write lock.
+    const LOST_WORKER_INVALIDATION_CHUNK: usize = Self::BLOCK_REPORT_WRITE_CHUNK;
 
     fn validate_alloc_capacity(
         current_len: i64,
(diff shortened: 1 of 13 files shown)


curvine-master/src/master/meta/store/inode_store.rs
```diff
@@ -202,6 +202,15 @@ impl InodeStore {
         Ok(())
     }
 
+    pub fn apply_cache_invalidations(&self, inodes: Vec<InodeView>) -> CommonResult<()> {
```
lzjqsdd, line 205: This invalidation is written directly to RocksDB and never goes through `JournalWriter`, so it is not replayed on Raft followers. A leader failover after worker-loss cleanup but before a client reloads can leave the new leader with `StorageState::Both` and stale blocks, reproducing #1519. Please add a journaled cache-invalidation entry with a replay handler and cover the leader/follower path with a test.

curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1557,9 +1563,27 @@ impl MasterFilesystem {
             .unwrap_or(false)
     }
 
-    pub fn delete_locations(&self, worker_id: u32) -> FsResult<Vec<i64>> {
-        let fs_dir = self.fs_dir.write();
-        fs_dir.delete_locations(worker_id)
+    pub fn delete_locations(&self, worker_id: u32) -> FsResult<LostWorkerLocationCleanup> {
+        let mut fs_dir = self.fs_dir.write();
+        let removed_block_ids = fs_dir.delete_locations(worker_id)?;
+        let invalidated = fs_dir.invalidate_lost_cache_files(&removed_block_ids)?;
```
lzjqsdd: If `invalidate_lost_cache_files` fails after `fs_dir.delete_locations` has already removed the worker's location records, this function returns `Err`. `HeartbeatChecker` then uses `Default::default()` and drops `removed_block_ids`, so under-replicated blocks are not reported at all. Make invalidation best-effort, or return the removed block IDs even when invalidation fails, so normal replica recovery still runs.

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -950,6 +956,63 @@ impl FsDir {
...
+    /// be scheduled for deletion from any surviving workers.
+    pub(crate) fn invalidate_lost_cache_files(
+        &mut self,
+        block_ids: &[i64],
+    ) -> FsResult<CacheInvalidationResult> {
+        let inode_ids: HashSet<_> = block_ids.iter().map(|id| InodeId::get_id(*id)).collect();
+        let mut result = CacheInvalidationResult::default();
+        let mut changed_inodes = Vec::new();
+
+        for inode_id in inode_ids {
```
lzjqsdd, line 988: This loop performs a RocksDB inode lookup per unique inode and a block-location lookup per block while holding the global `fs_dir` write lock. A lost worker can host a very large cache prefix, so this can stall all metadata operations for the duration of the scan. Consider collecting the affected inode/block set before taking the lock, or processing invalidation in bounded chunks like `BLOCK_REPORT_WRITE_CHUNK` so the lock is held only briefly.

lzjqsdd (commented):
Reviewed the worker-loss cache invalidation path. The core approach is sound: preserve UFS metadata, clear unusable cache blocks, and exclude discarded cache blocks from normal replica recovery. I left three comments on Raft durability, replication reporting when invalidation fails, and unbounded metadata work under the global lock. I would address at least the journaling and error-path findings before merge.

HarryMWinters:
Thanks for the speedy feedback @lzjqsdd! I'll address the comments and self-review more before marking ready-for-review.

curvine-master/src/master/meta/store/inode_store.rs
HarryMWinters: Good call! Addressed in f7c1ab01. Cache invalidations now have a journal entry containing the updated inode views, a writer path, and a JournalLoader replay handler, so followers persist the same UFS-only state after replay. Added `lost_worker_cache_invalidation_replays_on_follower` to cover the leader/follower path.

curvine-master/src/master/fs/master_filesystem.rs
HarryMWinters: Agreed. Addressed in f7c1ab01. Once worker locations have been removed, cache invalidation is now best-effort: a failed chunk is logged and cleanup continues. Blocks not successfully marked as intentionally invalidated remain in `replication_block_ids` and are reported to normal replica recovery

curvine-master/src/master/meta/fs_dir.rs
HarryMWinters: Addressed in f7c1ab01 using the suggested chunking approach. Lost-worker block IDs are processed in 4,096-ID chunks, and the `fs_dir` write guard is scoped to one chunk and dropped before the next. `CacheInvalidationResult::extend` combines the per-chunk cleanup results. The inode/location checks remain within each chunk to retain a consistent metadata view.

HarryMWinters:
Comments addressed and self review complete. Ready for final review and CI imo. Thanks for your time and the project @lzjqsdd.

lzjqsdd (approved):

Issue #1527, opened 2026-08-08
[FEATURE] Component version, release, and wire-protocol compatibility management


lzjqsdd:
**Is your feature request related to a problem? Please describe.**
Curvine components do not have a single, traceable version contract:

- Cargo workspace version stays at `0.2.0` while the repository advances past `v0.4.0-alpha`, so local builds cannot be mapped back to a commit or tag.
- Worker heartbeat only reports a display string (`software_version`), which is not reliable enough for compatibility checks.
- Client-master has no handshake that returns master version, protocol version, minimum supported version, or capabilities.
- Client-worker data-plane requests do not carry component or protocol version information.
- Known incompatible versions have no code-default policy, compatibility matrix, or blocklist; operators are forced to manage compatibility manually.
- Release artifacts (binaries, images, Helm, Java/Python SDKs) derive versions from different places and drift from the release tag.

**Describe the solution you'd like**
Implement a version and compatibility design that covers all component versions, the three main protocol planes, and the release pipeline.

### Version model

Keep four independent version dimensions:

```text
release_version   X.Y.Z[-pre]        release/package version
git_version       commit/tag/branch  build provenance
protocol_version  u32                wire protocol version
schema_version    u64                storage/local state version
```

Git tag is the release anchor:

```text
Git tag:            vX.Y.Z[-alpha|-beta|-rc]
Cargo workspace:    equals X.Y.Z at release; next-dev after release
Local build:        derived from git describe; CI overrides with BUILD_VERSION
Images/Helm/SDK:    version from release tag with v removed
```

All components expose structured version info through `--version`, `--version-json`, and startup logs.

### Public version protocol

Add common protobuf messages:

```proto
message ComponentInfoProto {
  optional string component = 1 [default = ""];
  optional string release_version = 2 [default = ""];
  optional string git_commit = 3 [default = ""];
  optional string git_tag = 4 [default = ""];
  optional string git_branch = 5 [default = ""];
  optional uint32 protocol_version = 6 [default = 1];
  optional uint32 min_protocol_version = 7 [default = 1];
  repeated string capabilities = 8;
}

enum CompatibilityModeProto {
  DIAGNOSE = 1;
  ENFORCE = 2;
}

message ServerCompatibilityInfoProto {
  required ComponentInfoProto server = 1;
  required string min_worker_version = 2 [default = ""];
  required string min_client_version = 3 [default = ""];
[...]

PR #1529, opened 2026-08-09
build(build): resolve workspace version and support BUILD_VERSION override (#1527)

lzjqsdd:
# Summary

Resolve the Cargo workspace version during packaging and allow builds to override the artifact version with `BUILD_VERSION`. Release builds now derive `BUILD_VERSION` from the pushed tag, keeping packaged artifacts and generated Rust version constants aligned with the release version.

# Issue Describe / Design

Related to #1527.

This PR implements the first version-management step:

- The previous `grep '^version ='` did not match the indented `[workspace.package] version`, so `build/build.sh` could resolve the wrong package version.
- `BUILD_VERSION` is now the explicit release/CI override for packaging and generated version constants.
- Release workflow derives `BUILD_VERSION` from the tag name and strips the leading `v` before `make dist`.
- Missing or unparseable `[workspace.package].version` now fails instead of publishing artifacts with an `unknown` version.
- When `BUILD_VERSION` is not set, behavior falls back to the Cargo workspace version, so existing local builds remain stable.

# Changes

| `.github/workflows/release.yml` | Export `BUILD_VERSION` from the release tag before `make dist` | Release artifacts and generated binary version output follow the tag version |
| `build/version.sh` | Parse workspace version, apply non-empty `BUILD_VERSION`, and fail when workspace version cannot be resolved | Invalid workspace metadata now fails the build instead of producing `unknown` artifacts |
| `build/build.sh` | Use `get_build_version` instead of root-level `grep`; document `BUILD_VERSION` in help | Local build version now resolves from `[workspace.package] version` |
| `crates/core/curvine-sys/build.rs` | Rebuild when `BUILD_VERSION` changes; override `PKG_VERSION` and `VERSION` | Generated version constants honor release override |
| `crates/core/curvine-sys/src/version.rs` | Add version constant tests and ignore empty `BUILD_VERSION` in override assertions | None |
| `build/tests/test_workspace_version.sh` | Add workspace version, override, empty override fallback, and missing version tests | None |

# Test verified

| `bash build/tests/test_workspace_version.sh` | PASS | workspace version, `BUILD_VERSION` override, empty override fallback, missing version failure |
| `cargo test -p curvine-sys` | PASS | 14 passed |
| `make format` | PASS | pre-commit and clippy checks clean |
| `git diff --check origin/main...HEAD` | PASS | no whitespace errors |

# Dependencies

- Related PRs: None
- Related issues: #1527
- Blocks / blocked by: None

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -23,6 +23,9 @@ set -e
 # ./build release, release mode.
 FS_HOME="$(cd "`dirname "$0"`/.."; pwd)"
 
+# Shared version helpers used by build packaging and by version tests.
+source "$FS_HOME/build/version.sh"
+
 # Check if cargo is available
 if ! command -v cargo &> /dev/null; then
     echo "Error: cargo is not installed or not in PATH" >&2
@@ -136,6 +139,7 @@ print_help() {
   echo "  --skip-java-sdk        Skip Java SDK compilation (useful for Docker builds)"
   echo "  --skip-python-sdk      Skip Python SDK compilation (useful for Docker builds)"
   echo "  -h, --help            Show this help message"
+  echo "  BUILD_VERSION=<ver>   Override the package version used for artifacts and build-version"
   echo
   echo "Examples:"
   echo "  $0                                      # Build all packages in release mode with opendal-s3"
@@ -164,7 +168,7 @@ fi
 ARCH_NAME=$(get_arch_name)
 OS_VERSION=$(get_os_version)
 FUSE_VERSION=$(get_fuse_version)
-CURVINE_VERSION=$(grep '^version =' "$FS_HOME/Cargo.toml" | sed 's/^version = "\(.*\)"/\1/')
+CURVINE_VERSION=$(get_build_version "$FS_HOME/Cargo.toml")
 
 # Package Directory
 DIST_DIR="$FS_HOME/build/dist"
--- a/build/version.sh
+++ b/build/version.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+#
+# Copyright 2025 OPPO.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


build/build.sh
```diff
@@ -164,7 +168,7 @@ fi
 ARCH_NAME=$(get_arch_name)
```
szbr00, line 168: The release workflow exports only `RELEASE_VERSION`, while this path and `curvine-sys/build.rs` read `BUILD_VERSION`. Therefore a tagged release can retain `0.2.0` in `build-version` and binary version output while only the outer archive uses the tag. Please derive `BUILD_VERSION` from the release tag, stripping the `v` prefix, before invoking `make dist`.

build/version.sh
```diff
@@ -0,0 +1,46 @@
...
+get_build_version() {
+  local cargo_toml="${1:?Cargo.toml path required}"
+  if [ -n "${BUILD_VERSION:-}" ]; then
+    printf '%s\n' "$BUILD_VERSION"
+    return 0
+  fi
+  local version
+  version="$(get_workspace_version "$cargo_toml")"
+  printf '%s\n' "${version:-unknown}"
+}
```
szbr00, line 53: A missing or unparseable workspace version currently succeeds as `unknown`, allowing incorrectly versioned artifacts to be published. Please return a non-zero status when `[workspace.package].version` cannot be resolved and add a negative test for that case.

crates/core/curvine-sys/src/version.rs
```diff
@@ -14,3 +14,23 @@
...
+        assert!(!VERSION.is_empty());
+    }
+
+    #[test]
+    fn build_version_override_is_honored_when_set() {
+        if let Ok(build_version) = std::env::var("BUILD_VERSION") {
+            assert_eq!(PKG_VERSION, build_version);
+            assert_eq!(VERSION, build_version);
+        }
+    }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 9f5d436c: the test now skips empty BUILD_VERSION values, matching build.rs behavior.

build/version.sh
lzjqsdd: Thanks. Fixed in 9f5d436c: get_build_version now returns non-zero when workspace package version cannot be resolved, with a negative shell test.

build/build.sh
lzjqsdd: Thanks. Fixed in 9f5d436c: release.yml now derives BUILD_VERSION from RELEASE_VERSION with the leading v stripped and exports it before make dist.

szbr00 (approved):

PR #1533, opened 2026-08-10
fix(fuse): reply EPROTO for malformed request payloads (#1523)

### LINKED ISSUE #1523: [BUG]: FUSE parse failures can hang callers because finish_early sends no reply
## Describe the bug

When a FUSE request has a valid header and `unique` value but `FuseRequest::parse_operator()` fails, both metadata and stream dispatch call:

```rust
reply.finish_early(err.errno(), "other");
return Err(err);
```

`finish_early()` only finalizes metrics bookkeeping. It does not enqueue a FUSE response:

- with metrics disabled, it is effectively a no-op;
- with metrics enabled, it marks the response slot as `finished` and drops the active guard;
- because the slot is already finished, a later `send_rep()` on the same `FuseResponse` is treated as a duplicate reply and skipped.

The receiver caller only logs the dispatch error and continues serving. It does not close the FUSE connection or send a fallback error response. For reply-expecting opcodes, this leaves the kernel request pending and the calling thread can remain blocked in `request_wait_answer()`.

## Impact

Any body-level parse failure on a reply-expecting request can orphan a FUSE request and hang the caller. Examples include:

- a LOOKUP request without a NUL-terminated name;
- a truncated ACCESS request;
- a truncated READ or WRITE request;
- a future kernel/userspace ABI mismatch that reaches `parse_operator()`.

Fixing individual decoder incompatibilities removes particular triggers, but does not enforce the protocol-level invariant that every reply-expecting request must receive a response.

## Expected behavior

Once Curvine has successfully read a FUSE request header and has a valid `unique` value:

- every reply-expecting opcode must enqueue exactly one success or error response;
- protocol no-reply opcodes must continue to enqueue no response;
- metrics cleanup must not prevent the required wire response.

A malformed request body should return an error such as `EIO` or `EINVAL` instead of leaving the kernel waiting indefinitely.

## Proposed change

Add a dedicated helper such as `send_parse_error()` that atomically handles both responsibilities:

1. record the parse/decode error and finish the metrics context;
2. enqueue an error response for reply-expecting opcodes using the original `unique`.

Do not call `finish_early()` and then `send_rep()`. When metrics are enabled, `finish_early()` has already marked the slot as finished, so the later response is rejected as a duplicate.

Keep a metrics-only/no-wire path for opcodes that are intentionally no-reply. If `finish_early()` remains, consider renaming it to make its metrics-only semantics explicit, for example `finish_metrics_on_parse_error()`.

## Evidence

The existing unit test confirms that `finish_early()` enqueues no task:

```bash
cargo test -p curvine-fuse t8_finish_early_drops_guard_no_task -- --nocapture
```

Observed result on current `main` (`c8e98fc4`):

```text
test session::fuse_response::tests::t8_finish_early_drops_guard_no_task ... ok
test result: ok. 1 passed
```

The test explicitly asserts:

```rust
assert!(rx.try_recv().unwrap().is_none(), "no request task enqueued");
```

## Acceptance criteria

- [ ] A malformed reply-expecting metadata request with a valid header/unique enqueues exactly one error response.
- [ ] A malformed reply-expecting stream request with a valid header/unique enqueues exactly one error response.
- [ ] The response carries the original request `unique`.
- [ ] Protocol no-reply opcodes continue to enqueue no response.
- [ ] Behavior is correct with metrics both enabled and disabled.
- [ ] Parse/decode metrics are recorded and the active guard is released exactly once.
- [ ] No double-reply warning or assertion is triggered.
- [ ] A real `/dev/fuse` regression test verifies that the calling syscall returns an error within a bounded timeout instead of hanging.

### LINKED ISSUE #1532: [BUG]: FUSE requests may hang when malformed opcode payloads receive no reply
**Describe the bug**

After a FUSE request header has been decoded successfully, an opcode payload decoding failure is recorded through `finish_early()`, but no response is queued for operations that require a reply. The kernel caller may therefore wait indefinitely for the request's `unique`.

**To Reproduce**

1. Build Curvine FUSE from `main`.
2. Construct a request with a valid `fuse_in_header`, known opcode, and trusted `unique`.
3. Use a malformed opcode payload, such as:
   - `LOOKUP` without a terminating NUL.
   - `READ` with a body shorter than `fuse_read_in`.
4. Dispatch the request.
5. Observe that payload parsing fails without producing a `FuseTask` response.

**Expected behavior**

When the header, opcode, and `unique` are trusted, malformed payloads for reply-requiring operations should receive exactly one `-EPROTO` response.

Requests whose headers cannot be trusted should still terminate the receiver session. `FORGET`, `BATCH_FORGET`, and `INTERRUPT` should retain their protocol no-reply behavior.

**Screenshots**

Not applicable.

- Linux x86_64 remote development environment
- Curvine base commit: `6c85f575`

**Additional context**

The affected paths are metadata and stream dispatch. The issue is distinct from response-construction fallback handling added in PR #1369.

ligeng8848:
# Summary

Reply with `EPROTO` when a trusted FUSE request header is followed by a malformed opcode payload, preventing reply-requiring requests from waiting indefinitely. Preserve protocol no-reply behavior for `FORGET`, `BATCH_FORGET`, `INTERRUPT`, and `FUSE_NOTIFY_REPLY`.

# Issue Describe / Design

Closes #1523
Closes #1532

Root cause:

- Metadata and stream dispatch call `parse_operator()` after the request header has already been validated.
- On payload decoding failure, the previous implementation called `finish_early()` and returned the parsing error.
- `finish_early()` only completed metrics bookkeeping and did not enqueue a response for the kernel.

Reproduction:

1. Submit a request with a valid header, opcode, and `unique`.
2. Use a malformed payload such as a non-NUL-terminated `LOOKUP` name or a short `READ` argument.
3. Observe that no response task is queued.

Fix approach:

- Add a parse-error response path that records the decoding error and sends exactly one `-EPROTO` response.
- Classify every protocol no-reply operation explicitly and keep its existing behavior.
- Leave untrusted header/frame failures session-fatal.
- Preserve pending-request cleanup and lifecycle behavior for interruptible `SETLKW`.
- Complement #1512: that PR accepts a valid LOOKUP name followed by rename-race remnants, while this PR guarantees an error response when an opcode payload is genuinely malformed.

Real `/dev/fuse` testing boundary:

- A normal Linux syscall cannot make the kernel emit a deliberately malformed FUSE opcode payload, so the receiver-level tests use exact raw frames to cover the response invariant.
- A real mount test requires kernel fault injection or a feature-gated receiver corruption hook. That broader test harness is tracked separately in #1537 so this bug fix does not introduce a production fault-injection surface.

# Changes

| `session/fuse_request.rs` | Classify requests that expect a reply, including all protocol no-reply opcodes | No change for valid reply-requiring requests |
| `session/fuse_response.rs` | Add the `EPROTO` payload parse-error response path | Malformed reply-requiring requests now complete |
| `session/channel/fuse_receiver.rs` | Use the response path in metadata and stream dispatch and add regression coverage | Prevents request hangs while preserving no-reply semantics |

# Test verified

| `cargo test -p curvine-fuse --lib expects_reply_classifies_every_protocol_no_reply_opcode -- --test-threads=1` | PASS | 1/1; review follow-up |
| `cargo test -p curvine-fuse --lib malformed_ -- --test-threads=1` | PASS | 5/5 |
| `cargo test -p curvine-fuse --lib --offline -- --test-threads=1` | PASS | 395/395 on remote Linux before the review-only classification addition |
| `cargo fmt --all -- --check` | PASS | Review follow-up formatting check |

# Dependencies

- Related PRs: #1369, #1512
- Related issues: #1523, #1532, #1537
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/session/fuse_request.rs
+++ b/curvine-fuse/src/session/fuse_request.rs
@@ -103,7 +103,7 @@ impl FuseRequest {
     pub fn expects_reply(&self) -> bool {
         !matches!(
             self.opcode,
-            FUSE_FORGET | FUSE_BATCH_FORGET | FUSE_INTERRUPT
+            FUSE_FORGET | FUSE_BATCH_FORGET | FUSE_INTERRUPT | FUSE_NOTIFY_REPLY
         )
     }
 
@@ -432,6 +432,27 @@ mod tests {
             Err(err) => err,
         };
         assert!(err.to_string().contains("length mismatch"));
+    }
+
+    #[test]
+    fn expects_reply_classifies_every_protocol_no_reply_opcode() {
+        for opcode in [
+            FUSE_FORGET,
+            FUSE_BATCH_FORGET,
+            FUSE_INTERRUPT,
+            FUSE_NOTIFY_REPLY,
+        ] {
+            let header = header(opcode, FUSE_IN_HEADER_LEN);
+            let request = FuseRequest::from_bytes(request_bytes(&header, &[])).unwrap();
+            assert!(
+                !request.expects_reply(),
+                "{opcode:?} must not produce a response frame"
+            );
+        }
+
+        let header = header(FUSE_LOOKUP, FUSE_IN_HEADER_LEN);
+        let request = FuseRequest::from_bytes(request_bytes(&header, &[])).unwrap();
+        assert!(request.expects_reply(), "LOOKUP requires a response frame");
     }
 
     #[test]
--- a/curvine-fuse/src/session/fuse_request.rs
+++ b/curvine-fuse/src/session/fuse_request.rs
@@ -100,6 +100,13 @@ impl FuseRequest {
         )
     }
 
+    pub fn expects_reply(&self) -> bool {
+        !matches!(
+            self.opcode,
+            FUSE_FORGET | FUSE_BATCH_FORGET | FUSE_INTERRUPT
+        )
+    }
+
     pub fn should_audit(&self) -> bool {
         !matches!(self.opcode, FUSE_READ | FUSE_WRITE)
     }
--- a/curvine-fuse/src/session/channel/fuse_receiver.rs
+++ b/curvine-fuse/src/session/channel/fuse_receiver.rs
@@ -293,13 +293,15 @@ impl<T: FileSystem> FuseReceiver<T> {
         rep: FuseResponse,
     ) -> FuseResult<()> {
         let metrics_enabled = rep.metrics.is_some();
-        // Parse failure here is after the ctx exists: finish it early (no reply).
-        // No stable errno for a parse error, so tag the catch-all "other".
+        // The frame header is already trusted here, so complete a malformed payload
+        // with EPROTO instead of leaving the kernel waiting for this unique.
         let operator = match req.parse_operator() {
             Ok(op) => op,
             Err(err) => {
-                rep.finish_early(err.errno(), "other");
-                return Err(err);
+                return rep
+                    .send_parse_error("other", err.to_string())
+                    .await
+                    .map_err(Into::into);
             }
         };
 
@@ -561,10 +563,17 @@ impl<T: FileSystem> FuseReceiver<T> {
         req: &FuseRequest,
         reply: &FuseResponse,
     ) -> FuseResult<()> {
-        // Parse failure here is after the ctx exists: finish it early (no reply).
+        // The frame header is already trusted here, so complete a malformed payload
+        // with EPROTO instead of leaving the kernel waiting for this unique.
         let operator = match req.parse_operator() {
             Ok(op) => op,
             Err(err) => {
+                if req.expects_reply() {
+                    return reply
+                        .send_parse_error("other", err.to_string())
+                        .await
+                        .map_err(Into::into);
+                }
                 reply.finish_early(err.errno(), "other");
                 return Err(err);
             }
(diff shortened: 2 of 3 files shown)


curvine-fuse/src/session/fuse_request.rs
```diff
@@ -100,6 +100,13 @@ impl FuseRequest {
         )
     }
 
+    pub fn expects_reply(&self) -> bool {
+        !matches!(
+            self.opcode,
+            FUSE_FORGET | FUSE_BATCH_FORGET | FUSE_INTERRUPT
```
szbr00: `FUSE_NOTIFY_REPLY` is also a protocol no-reply operation. Please include it in this match and add a classification test covering every no-reply opcode, so malformed notification replies cannot enqueue a response.

szbr00 (commented):
Issue #1523’s acceptance criteria require a real `/dev/fuse` regression test proving that the originating syscall returns within a bounded timeout. The current tests only validate the in-memory reply queue; please add the end-to-end case or document why it must be tracked separately.

curvine-fuse/src/session/fuse_request.rs
ligeng8848: Thanks. Fixed in e2515b41: `FUSE_NOTIFY_REPLY` is now classified as protocol no-reply, and the new classification test covers `FORGET`, `BATCH_FORGET`, `INTERRUPT`, and `FUSE_NOTIFY_REPLY`.

ligeng8848:
Thanks for flagging the real `/dev/fuse` acceptance criterion. A normal Linux syscall cannot make the kernel emit a deliberately malformed opcode payload, so the PR now documents why the current raw-frame receiver tests stop at the reply queue. The real-mount, bounded-syscall case requires a feature-gated corruption hook or kernel fault injection and is tracked separately in #1537. This keeps the production fix small while preserving explicit follow-up ownership.

szbr00 (approved):

Issue #924, opened 2026-06-22
Raft snapshot checkpoint can accumulate without bound when pre-created checkpoints are not applied


jlon:
## Issue

Curvine master can create thousands of local RocksDB checkpoint directories and fill the master disk. This is not just a bad `snapshot_entries` setting. The current code has a failure window where a checkpoint is created before the Raft snapshot entry is applied, but cleanup only runs after that entry is applied.

In the observed cluster:

- `cv-curvine-master-0` and `cv-curvine-master-2` are in `CrashLoopBackOff`.
- Both fail with `No space left on device` while opening `/opt/curvine/data/journal/db/data/LOG`.
- Their host `/data` disks are full.
- Each node has more than 5,000 checkpoint directories:
  - `master-0`: 5,253 `ck-*` directories
  - `master-2`: 5,347 `ck-*` directories
- The directory names are timestamp-style, for example `ck-1781754652318`, which matches `JournalWriter::maybe_emit_snapshot()` pre-created checkpoints.
- The healthy `master-1` has only 3 checkpoint directories, which matches the default retained checkpoint count.

## Relevant Code

`JournalWriter::maybe_emit_snapshot()` creates the checkpoint first:

```rust
let dir = fs_dir.store.create_checkpoint(now)?;
self.send_inner(JournalEntry::Snapshot(SnapshotEntry { dir, ... }))?;
```

Cleanup happens later, only after the snapshot entry is applied by the leader:

```rust
let snap_data = self.create_snapshot0(Some(e.dir.to_string()))?;
```

`create_snapshot0()` calls `purge_checkpoint()`.

So the lifecycle is currently:

```text
create checkpoint
  -> enqueue/send SnapshotEntry
  -> wait for Raft apply
  -> create raft snapshot metadata
  -> compact raft log
  -> purge old checkpoints
```

If the node crashes, Raft apply stalls, the queue backs up, disk fills, or `SnapshotEntry` never reaches the cleanup path, the checkpoint already exists but no cleanup is guaranteed.

## Why changing parameters is not enough

Increasing `snapshot_entries` only reduces how often Curvine creates checkpoints. It does not close the lifecycle gap.

For example:

- With `snapshot_entries = 100`, Curvine creates checkpoints very frequently, so the bug appears quickly.
- With a larger value, Curvine creates fewer checkpoints, so disk growth is slower.
- But if a checkpoint is created and the process fails before the snapshot entry is applied, that checkpoint can still become orphaned.

[...]

PR #1535, opened 2026-08-10
fix(master): clean checkpoint on snapshot enqueue failure

sf1999817:
## Problem

`JournalWriter::maybe_emit_snapshot()` creates a RocksDB checkpoint before enqueueing the corresponding `SnapshotEntry`.

If enqueueing fails because the journal receiver is disconnected, ownership is not transferred to the downstream snapshot path, but the newly created checkpoint is currently left on disk.

## Change

Remove only the checkpoint created by the current invocation when enqueueing the `SnapshotEntry` fails, then return the original send error.

A successfully enqueued checkpoint remains available to the downstream snapshot path.

## Tests

- Added a regression test verifying that an enqueue failure removes the newly created checkpoint.
- Added a success-path test verifying that a successfully enqueued checkpoint remains available.
- Confirmed that the failure-path test fails without the cleanup and passes with this patch.

## Scope

Related to #924.

This patch addresses only the definite enqueue-failure window. It does not handle process crashes, post-enqueue failures, apply stalls, or startup reconciliation of orphan checkpoints.

For the remaining cases, would maintainers prefer an identified single in-flight checkpoint guard, startup reconciliation based on persisted checkpoint ownership, or another existing snapshot lifecycle design?

### DIFF
--- a/curvine-master/src/master/journal/journal_writer.rs
+++ b/curvine-master/src/master/journal/journal_writer.rs
@@ -24,10 +24,10 @@ use curvine_error::FsResult;
 use curvine_model::{CommitBlock, FileLock, MountInfo, RenameFlags, SetAttrOpts};
 use curvine_raft::conf::JournalConfExt;
 use curvine_raft::raft::RaftClient;
-use curvine_runtime::common::LocalTime;
+use curvine_runtime::common::{FileUtils, LocalTime};
 use curvine_runtime::sync::channel::{BlockingChannel, BlockingReceiver, BlockingSender};
 use curvine_runtime::sync::AtomicCounter;
-use log::{debug, info};
+use log::{debug, info, warn};
 use std::collections::VecDeque;
 use std::sync::Mutex;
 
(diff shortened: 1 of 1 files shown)


curvine-master/src/master/journal/journal_writer.rs
```diff
@@ -129,12 +129,23 @@ impl JournalWriter {
...
             op_id: fs_dir.next_op_id(),
             rpc_id: 0,
             node_id: self.node_id,
-            dir,
-        }))?;
+            dir: dir.clone(),
+        });
+
+        if let Err(send_error) = self.send_inner(snapshot_entry) {
+            if let Err(cleanup_error) = FileUtils::delete_path(&dir, true) {
```
szbr00, line 140: This cleanup path has no regression coverage in the PR, although the description says failure- and success-path tests were added. Please include those tests to verify that a failed enqueue removes the checkpoint while a successful enqueue preserves it.

szbr00 (approved):

PR #1536, opened 2026-08-10
feat: improve fuse_pure_mount

### LINKED ISSUE #1394: [BUG]: Harden pure FUSE mount and unmount error handling [LiAuto]
**Describe the bug**

The pure mount implementation assumes mount paths are UTF-8 and calls `unwrap`, replaces specific mount errors with a generic `-1`, and treats a spawned `fusermount` process as success without checking its exit status.

**To Reproduce**

1. Use a valid Unix mount path containing non-UTF-8 bytes.
2. Trigger a mount failure such as `EACCES`, `ENOENT`, or `EBUSY`.
3. Make direct `umount2` fail and make fallback `fusermount` exit nonzero.

**Expected behavior**

Valid non-UTF-8 paths must not panic. The original mount errno and fallback command failure must be propagated with useful diagnostics.

- Linux
- Component: `curvine-fuse/src/raw/fuse_pure.rs`

**Additional context**

Use `OsStrExt::as_bytes`, return a normal error for embedded NUL, preserve the original mount error, and check `output.status.success()` plus stderr for fallback unmount.

201341:
close https://github.com/CurvineIO/curvine/issues/1394

### DIFF
--- a/curvine-fuse/src/raw/fuse_pure.rs
+++ b/curvine-fuse/src/raw/fuse_pure.rs
@@ -12,6 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Linux-only pure FUSE mount/unmount implementation.
+//!
+//! This module talks to the kernel directly via `libc::mount` / `libc::umount2`
+//! (with a `fusermount` fallback for unmount). It has no macOS/BSD code path and
+//! is not designed to compile off Linux, so the whole module is gated to
+//! `target_os = "linux"` in `raw/mod.rs`. Because of that outer gate, the code
+//! below can use Linux-specific APIs unconditionally without per-item `#[cfg]`.
+
 use curvine_config::FuseConf;
 use curvine_io::{IOError, IOResult};
 use curvine_sys::RawIO;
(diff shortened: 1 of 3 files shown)


curvine-fuse/src/raw/fuse_pure.rs
```diff
@@ -158,7 +155,7 @@ fn fuse_mount_sys(mnt: &Path, conf: &FuseConf) -> IOResult<RawIO> {
 
     // Default name is "/dev/fuse", then use the subtype, and lastly prefer the name
     let c_source = CString::new("curvinefs").unwrap();
-    let c_mountpoint = CString::new(mnt.to_str().unwrap()).unwrap();
+    let c_mountpoint = path_to_cstring(mnt)?;
```
szbr00, line 160: The macOS implementations of `options_to_flag`, mount, and unmount are removed here, but `fuse_pure` is still compiled unconditionally. On macOS, `options_to_flag` is now undefined and the `result` blocks contain no value-producing branch, so `curvine-fuse` will not compile. Please restore the macOS branches, or explicitly gate this module and its callers to Linux if macOS support is intentionally being dropped.

201341: fix

szbr00 (approved):

PR #1538, opened 2026-08-11
fix(fuse): make FUSE startup failures actionable

jlon:
# Summary

Make FUSE startup failures actionable while preserving mount, splice, and runtime data-path behavior. Preserve the original OS errno through daemon startup so callers and scripts no longer see only generic EIO.

# Issue Describe / Design

Related issue: None.

Root cause: several startup failures either lost their original errno during conversion to FuseError or reported low-level context without actionable remediation. This included kernel ABI mismatch, mount syscall and device failures, restricted-container splice-pipe setup, and reload FD restoration.

Reproduction: a kernel advertising FUSE ABI 7.22 fails the protocol handshake; a restricted container can mount FUSE successfully but fail F_SETPIPE_SZ with EPERM while creating Curvine splice pipes; a container without CAP_SYS_ADMIN fails the mount syscall.

Fix approach: add diagnostics and preserve errno only on daemon startup and restore paths. Do not change enable_splice defaults, mount options, request-path EIO normalization, or introduce automatic fallback behavior.

# Changes

| curvine-fuse startup and raw mount | Add ABI, mountpoint, /dev/fuse, mount syscall, unmount, splice-pipe, and FD-stage diagnostics | Failure output only; raw errno is preserved for startup |
| Fuse error/session boundary | Keep startup and reload errno without changing request-path EIO normalization | Runtime request behavior unchanged |
| FUSE launch scripts | Preflight mount targets, decode Linux mountinfo path escapes before duplicate-mount checks, keep version passthrough, avoid duplicate config injection, and surface daemon startup logs | Startup failures are returned to callers |
| FUSE E2E tests | Add opt-in privileged and restricted-runtime coverage | No default test-environment requirement |

# Test verified

| make format | PASS | Includes workspace release Clippy and formatting |
| cargo test -p curvine-fuse -- --test-threads=1 | PASS | 414 passed, 3 opt-in tests ignored |
| raw_mount_unmount_round_trip | PASS | Executed with CAP_SYS_ADMIN and /dev/fuse |
| unprivileged_mount_reports_permission_remediation | PASS | Executed without CAP_SYS_ADMIN |
| restricted_splice_pipe_growth_returns_eperm | PASS | Executed without CAP_SYS_RESOURCE |
| bash -n build/bin/curvine-fuse.sh build/bin/launch-process.sh build/tests/test_curvine_fuse.sh | PASS | Script syntax |
| bash build/tests/test_curvine_fuse.sh | PASS | mountinfo space, tab, newline, and backslash escaping |
| Current-branch runtime image in isolated Kubernetes pod | NOT RUN | Current kube context was unreachable; no cluster mutation performed |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/build/bin/curvine-fuse.sh
+++ b/build/bin/curvine-fuse.sh
@@ -14,40 +14,97 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Execute fuse mount with parameter passthrough
-
-# Default mount point
+BIN_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)"
 MNT_PATH=/curvine-fuse
 
-# Process arguments in a single loop
-for arg in "$@"; do
-    if [[ "$arg" == "--help" ]] || [[ "$arg" == "-h" ]]; then
-        . "$(cd "`dirname "$0"`"; pwd)"/../conf/curvine-env.sh
-        ${CURVINE_HOME}/lib/curvine-fuse --help
-        exit 0
+is_mountpoint() {
+    local path="$1"
+    local mountinfo="${2:-/proc/self/mountinfo}"
+    local mount_id parent major_minor root mount_path mount_options
+
+    while read -r mount_id parent major_minor root mount_path mount_options; do
+        # Linux mountinfo escapes path whitespace and backslashes as octal
+        # sequences. Decode in-place so trailing newlines remain lossless.
+        mount_path=${mount_path//\\040/ }
+        mount_path=${mount_path//\\011/$'\t'}
+        mount_path=${mount_path//\\012/$'\n'}
+        mount_path=${mount_path//\\134/$'\\'}
+        if [[ "$mount_path" == "$path" ]]; then
+            return 0
+        fi
+    done < "$mountinfo"
+
+    return 1
+}
+
+prepare_mount_path() {
+    if is_mountpoint "$MNT_PATH"; then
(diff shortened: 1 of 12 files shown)


build/bin/curvine-fuse.sh
```diff
@@ -14,40 +14,72 @@
...
-    elif [[ "$next_is_path" == true ]]; then
-        MNT_PATH="$arg"
-        next_is_path=false
+    elif [[ "$arg" == "--mnt-path" ]] && (( i + 1 < ${#PARAMS[@]} )); then
+        MNT_PATH=${PARAMS[$((i + 1)] )}
     fi
 done
 
-mkdir -p "$MNT_PATH"
+is_mountpoint() {
```
szbr00: `/proc/self/mountinfo` escapes spaces, tabs, newlines, and backslashes in path fields (for example, `/mnt/fuse\040data`), so comparing field 5 directly with the unescaped `MNT_PATH` fails for valid mounted paths containing these characters. `prepare_mount_path` can then proceed and start another mount over the existing one, leaving stacked mounts. Please decode/escape the mountinfo field before comparison (or use a mount-table-aware helper with safe argument handling) and add a regression case with a mount path containing a space.

jlon: Thanks. Fixed in 43a21a4: decode Linux mountinfo octal escapes before comparison and add a regression covering a mount path with a space, plus tab, newline, and backslash.

szbr00 (approved):

PR #1539, opened 2026-08-11
fix(io): keep LocalFile position and read/write consistent

bigbigxu:
﻿# Summary

Keep `LocalFile` position and length tracking consistent across read/write paths, and open write handles with read access so the same handle can read back.

# Issue Describe / Design

No linked issue.

Root cause:
- `read_full` began advancing `pos` via `seek`/`read_all`, but `read_region` still advanced `pos` again on the Buffer path.
- Write helpers updated `pos` without extending `len`, so read-back and bounds checks used a stale size.
- `Write::write` previously left `pos` unchanged.

Fix approach:
- Buffer path relies on `read_full` for `pos`; `IOSlice` path still advances explicitly.
- All write helpers call `advance_after_write` so `len = max(len, pos)`.
- `with_write` keeps `.read(true)` by design (writable implies readable on the same handle).
- Drop unused `is_tmpfs` tracking; kernel already decides whether read-ahead is useful.

# Changes

| `crates/core/curvine-io/src/local_file.rs` | Fix Buffer/IOSlice `pos` accounting; sync `len` on write; keep write+read open; drop `is_tmpfs`; add regression tests | Write-handle read-back works; multi-chunk `read_region` no longer double-advances |

# Test verified

| `cargo test -p curvine-io --lib read_region_multi_chunk_keeps_pos_in_sync` | PASS | Multi-chunk Buffer path |
| `cargo test -p curvine-io --lib write_extends_len_and_allows_read_back` | PASS | Write extends `len` + same-handle read-back |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/core/curvine-io/src/local_file.rs
+++ b/crates/core/curvine-io/src/local_file.rs
@@ -54,7 +54,6 @@ macro_rules! try_err {
 pub struct LocalFile {
     inner: fs::File,
     path: String,
-    is_tmpfs: bool,
     len: i64,
     pos: i64,
     buf: BytesMut,
@@ -62,14 +61,11 @@ pub struct LocalFile {
 
 impl LocalFile {
     pub fn new<T: AsRef<str>>(path: T, mut inner: fs::File) -> IOResult<Self> {
-        let is_tmpfs = curvine_sys::is_tmpfs(path.as_ref())?;
-
         let len = inner.metadata()?.len() as i64;
         let pos = inner.stream_position()? as i64;
         let file = Self {
             inner,
             path: path.as_ref().to_string(),
-            is_tmpfs,
             len,
             pos,
             buf: BytesMut::new(),
@@ -93,9 +89,10 @@ impl LocalFile {
         Self::new(path.as_ref(), file)
     }
 
-    // Open a file for writing.
+    // Open a file for writing (also requests read so the same handle can read back).
     pub fn with_write<T: AsRef<str>>(path: T, overwrite: bool) -> IOResult<Self> {
         let file = OpenOptions::new()
+            .read(true)
             .write(true)
             .create(true)
             .truncate(overwrite)
@@ -122,6 +119,13 @@ impl LocalFile {
 
     pub fn from_file(path: &str, inner: fs::File) -> IOResult<Self> {
         Self::new(path, inner)
+    }
+
+    fn advance_after_write(&mut self, n: i64) {
+        self.pos += n;
+        if self.pos > self.len {
+            self.len = self.pos;
+        }
     }
 
     pub fn read_region(&mut self, enable_send_file: bool, len: i32) -> IOResult<DataSlice> {
@@ -174,13 +178,13 @@ impl LocalFile {
             }
         };
 
-        self.pos += len as i64;
+        self.advance_after_write(len as i64);
         Ok(())
     }
 
     pub fn write_all(&mut self, buf: &[u8]) -> IOResult<()> {
         try_err!(self.inner.write_all(buf));
-        self.pos += buf.len() as i64;
+        self.advance_after_write(buf.len() as i64);
         Ok(())
     }
 
@@ -231,33 +235,25 @@ impl LocalFile {
         self.path.as_str()
     }
 
-    pub fn is_tmpfs(&self) -> bool {
-        self.is_tmpfs
-    }
-
     pub fn read_ahead(
         &mut self,
         os_cache: &CacheManager,
         last_task: Option<ReadAheadTask>,
     ) -> Option<ReadAheadTask> {
-        // Memory files do not use pre-reading.
-        if self.is_tmpfs {
-            None
-        } else {
-            os_cache.read_ahead(&self.inner, self.pos, self.len, last_task)
-        }
+        os_cache.read_ahead(&self.inner, self.pos, self.len, last_task)
     }
 
     pub fn read_full(&mut self, off: Option<i64>, len: usize) -> IOResult<BytesMut> {
         if let Some(v) = off {
-            self.inner.seek(SeekFrom::Start(v as u64))?;
+            self.seek(v)?;
         }
 
         self.buf.reserve(len);
         unsafe { self.buf.set_len(len) }
         let mut buf = self.buf.split();
 
-        self.inner.read_exact(&mut buf)?;
+        self.read_all(&mut buf)?;
+
         Ok(buf)
     }
 
@@ -333,7 +329,9 @@ impl Display for LocalFile {
 
 impl Write for LocalFile {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        self.inner.write(buf)
+        let len = self.inner.write(buf)?;
+        self.advance_after_write(len as i64);
+        Ok(len)
     }
 
     fn flush(&mut self) -> std::io::Result<()> {


crates/core/curvine-io/src/local_file.rs
```diff
@@ -231,33 +228,25 @@ impl LocalFile {
...
-            self.inner.seek(SeekFrom::Start(v as u64))?;
+            self.seek(v)?;
         }
 
         self.buf.reserve(len);
         unsafe { self.buf.set_len(len) }
         let mut buf = self.buf.split();
 
-        self.inner.read_exact(&mut buf)?;
+        self.read_all(&mut buf)?;
```
szbr00, line 255: `read_full` now advances `self.pos` through `read_all`, but `read_region` still advances it again. Buffered reads therefore report twice the consumed bytes, diverge from the OS cursor, and can reach logical EOF after reading only half the file. Please ensure only one layer advances the buffered path while retaining the explicit advance for `IOSlice`, and add a multi-chunk regression test.

crates/core/curvine-io/src/local_file.rs
```diff
@@ -333,7 +322,9 @@ impl Display for LocalFile {
 
 impl Write for LocalFile {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        self.inner.write(buf)
+        let len = self.inner.write(buf)?;
+        self.pos += len as i64;
```
szbr00: `Write::write` advances `pos` but leaves `len` unchanged. After extending a file, `len()`, `read_region`, and read-ahead still use the open-time length, so read-back through the newly readable handle can reject or truncate appended data. Please update `len` after successful writes and align the other write helpers with the same invariant.

crates/core/curvine-io/src/local_file.rs
```diff
@@ -97,6 +93,7 @@ impl LocalFile {
     pub fn with_write<T: AsRef<str>>(path: T, overwrite: bool) -> IOResult<Self> {
         let file = OpenOptions::new()
             .write(true)
+            .read(true)
```
szbr00, line 95: Adding read access changes `with_write` from write-only to read/write for every existing caller. Opening a writable file without read permission will now fail, although current callers do not read through these handles. Consider preserving the existing constructor semantics and introducing an explicit read/write constructor for read-back use.

crates/core/curvine-io/src/local_file.rs
```diff
@@ -231,33 +228,25 @@ impl LocalFile {
...
         os_cache: &CacheManager,
         last_task: Option<ReadAheadTask>,
     ) -> Option<ReadAheadTask> {
-        // Memory files do not use pre-reading.
-        if self.is_tmpfs {
-            None
-        } else {
-            os_cache.read_ahead(&self.inner, self.pos, self.len, last_task)
-        }
+        os_cache.read_ahead(&self.inner, self.pos, self.len, last_task)
```
szbr00, line 243: The tmpfs state was used by `read_ahead`; it was not unused. Removing the guard issues `POSIX_FADV_WILLNEED` for large memory-backed reads, adding hot-path syscalls without storage latency to hide. Please retain the tmpfs guard or provide benchmark evidence supporting this behavior change.

crates/core/curvine-io/src/local_file.rs
```diff
@@ -231,33 +228,25 @@ impl LocalFile {
...
+            self.seek(v)?;
         }
 
         self.buf.reserve(len);
         unsafe { self.buf.set_len(len) }
         let mut buf = self.buf.split();
 
-        self.inner.read_exact(&mut buf)?;
+        self.read_all(&mut buf)?;
+
```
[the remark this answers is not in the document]
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
```diff
@@ -333,7 +322,9 @@ impl Display for LocalFile {
 
 impl Write for LocalFile {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
-        self.inner.write(buf)
+        let len = self.inner.write(buf)?;
+        self.pos += len as i64;
+        Ok(len)
     }
```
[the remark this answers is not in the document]
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
```diff
@@ -97,6 +93,7 @@ impl LocalFile {
     pub fn with_write<T: AsRef<str>>(path: T, overwrite: bool) -> IOResult<Self> {
         let file = OpenOptions::new()
             .write(true)
+            .read(true)
             .create(true)
             .truncate(overwrite)
             .open(path.as_ref());
```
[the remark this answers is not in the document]
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
bigbigxu: System.Collections.Hashtable.body

crates/core/curvine-io/src/local_file.rs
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 1dc32bbfc: Buffer path now relies on read_full to advance pos; IOSlice path still advances explicitly.

crates/core/curvine-io/src/local_file.rs
bigbigxu: Thanks. Fixed in 1dc32bbfc: only one layer advances the Buffer path, IOSlice keeps the explicit advance, and multi-chunk regression coverage was added.

crates/core/curvine-io/src/local_file.rs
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 1dc32bbfc: write paths now call advance_after_write so len is extended when pos grows.

crates/core/curvine-io/src/local_file.rs
bigbigxu: Thanks. Fixed in 1dc32bbfc: Write::write, write_all, and write_region all share advance_after_write to keep len in sync.

crates/core/curvine-io/src/local_file.rs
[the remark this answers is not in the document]
bigbigxu: Thanks for the suggestion. Kept with_write as read+write by design: a writable LocalFile handle should also be readable for same-handle read-back.

crates/core/curvine-io/src/local_file.rs
bigbigxu: Thanks for the suggestion. Kept current approach: with_write requests read+write intentionally so callers can read back through the same handle without a separate constructor.

crates/core/curvine-io/src/local_file.rs
bigbigxu: Thanks for the suggestion. Kept removing the tmpfs guard: the kernel already decides whether read-ahead is useful, so the LocalFile-level check is redundant.

szbr00 (approved):

PR #1541, opened 2026-08-11
fix(master): return file_not_found for missing inode

bigbigxu:
﻿# Summary

Return `FsError::file_not_found` when master inode lookup fails, instead of a generic `err_box` message. Keep missing `inode_id` in error context for debugging.

# Issue Describe / Design

No linked issue. Align missing-inode / missing-path failures with the structured filesystem error used elsewhere so clients get a consistent not-found signal.

# Changes

| `curvine-master/src/master/fs/master_filesystem.rs` | Use `err_ext!(FsError::file_not_found(path).ctx(...))` for missing inode id; `file_not_found(path)` for unresolved path; add unit tests with encode/decode | Error type becomes `FileNotFound`; inode id retained in context |

# Test verified

| Diff review | PASS | Call sites + context update |
| `resolve_file_inode_missing_inode_id_returns_file_not_found` | SKIPPED locally | Added; not run in this push |
| `resolve_file_inode_unresolved_path_returns_file_not_found` | SKIPPED locally | Added; not run in this push |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -612,14 +612,14 @@ impl MasterFilesystem {
         match inode_id {
             Some(v) if v > 0 => match fs_dir.store.get_inode(v, None)? {
                 Some(view) => Ok(InodePtr::from_owned(view)),
-                None => err_box!("File inode {} not exists", v),
+                None => err_ext!(FsError::file_not_found(path)),
             },
 
             _ => {
                 let inp = Self::resolve_path(fs_dir, path)?;
                 match inp.task_last() {
                     Some(ptr) => Ok(ptr),
-                    None => err_box!("File {} not exists", path),
+                    None => err_ext!(FsError::file_not_found(path)),
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -612,14 +612,14 @@ impl MasterFilesystem {
         match inode_id {
             Some(v) if v > 0 => match fs_dir.store.get_inode(v, None)? {
                 Some(view) => Ok(InodePtr::from_owned(view)),
-                None => err_box!("File inode {} not exists", v),
+                None => err_ext!(FsError::file_not_found(path)),
```
szbr00: This change’s intended behavior is the new structured error kind, but it is not covered by a test. Could we add cases for both a missing positive `inode_id` and an unresolved path, asserting `FsError::FileNotFound` (ideally after RPC encoding/decoding)? This would prevent these paths from silently reverting to `Common`.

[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 71b14ad7e: keep `FsError::file_not_found(path)` and attach `inode_id={}` via `.ctx(...)` so the structured kind is preserved while the missing inode ID remains visible for debugging.

bigbigxu: Thanks. Fixed in 71b14ad7e: added unit tests for both a missing positive `inode_id` and an unresolved path. Each asserts `ErrorKind::FileNotFound` / `FsError::FileNotFound` before and after encode/decode so these paths cannot silently regress to `Common`.

szbr00 (approved):

PR #1544, opened 2026-08-11
fix(unified-fs): use DeleteResult for delete stats (#1543)

### LINKED ISSUE #1543: [BUG]: cache-mode free reports zero released inodes and bytes
**Describe the bug**

When `free` is called for a path in cache mode, the cache metadata and blocks are removed, but the CLI reports `inodes: 0` and `space: 0`. The command should report the actual number of inodes deleted and bytes released.

**To Reproduce**

1. Mount a UFS directory in cache mode and write at least one file so Curvine caches it.
2. Run `curvine free <path>` against the cached path.
3. Observe that the command output reports zero inodes and zero space even though the cache is removed.

**Expected behavior**

`free` in cache mode should report the number of deleted inodes and released bytes, matching the behavior of `free` outside cache mode.

**Screenshots**

N/A

 - OS: Linux
 - Curvine version/branch: main

**Additional context**

Cache-mode `free` removes Curvine metadata through delete but does not call `free` to collect the `FreeResult` statistics. The cache-mode path therefore returns a default `FreeResult` and the CLI displays zeros.

lzjqsdd:
# Summary

Use a dedicated `DeleteResult` for the delete API so callers can report inode and byte statistics without conflating delete semantics with `free`. The delete RPC wire format stays compatible with existing clients: `DeleteResponse` continues to carry `FreeResultProto`, and new code converts it to/from `DeleteResult` at the protocol boundary.

# Issue Describe / Design

Closes #1543

- Root cause: cache-mode `free` called `cv.delete` and returned `FreeResult::default()`, so the CLI displayed zero inode and space statistics even though cache metadata and blocks were removed.
- Reproduction steps: mount a UFS directory in cache mode, cache at least one file, run `free`, and observe zero inode/space output.
- Fix approach: `delete` returns a dedicated `DeleteResult` across the master/client APIs. The SDK dispatch layer converts `DeleteResult` into `FreeResult` only for the free/CLI reporting path. `DeleteResponse.res` remains `FreeResultProto` on the wire, and the field is optional so a new client talking to an old master degrades to zero stats instead of failing.
- Stats semantics: `DeleteResult.inodes` counts file inodes only, matching `FreeResult` semantics from regular `free`.

# Changes

| `crates/common/curvine-model/src/result.rs` | Add shared `DeleteResult` with checked conversions to/from `FreeResult` | Delete API uses delete-specific result type |
| `crates/common/curvine-model/src/proto_utils.rs` | Convert `DeleteResult` to/from the existing `FreeResultProto` | Delete RPC wire format remains compatible with old clients |
| `crates/common/curvine-proto/proto/master.proto` | Keep `FreeResultProto` and make `DeleteResponse.res` optional | New clients can tolerate old masters with empty delete responses |
| `crates/common/curvine-fs-api/src/filesystem.rs` and local/UFS adapters | `delete` returns `DeleteResult` | All FS implementations align with the new result type |
| `curvine-master/src/master/fs/*` and `master_handler.rs` | Master delete returns `DeleteResult` and serializes it to the existing response type | CLI can receive non-zero delete stats without breaking old clients |
| `curvine-master/src/master/meta/store/inode_store.rs` | Count file inodes only and compute bytes from block locations with checked conversions | Delete stats match free semantics |
| `crates/client/curvine-client-core/*` and `curvine-unified-fs/*` | Delete APIs return `DeleteResult`; cache-mode `free` converts it to `FreeResult` | No `free + delete`, no compatibility wrapper |
| `crates/sdk/*` | Delete returns `DeleteResult`; language wrappers map to their existing external API | SDK users keep stable wrapper behavior |
| `curvine-server/tests/master_fs_test.rs` | Extend delete retry regression with a cached child file and recursive delete | Delete result stats are covered by tests |

# Test verified

| `make format` | PASS | Formatting and clippy checks passed |
| `cargo check --workspace --all-targets` | PASS | Full workspace compile check |
| `cargo +1.92.0 test -p curvine-model result:: --lib` | PASS | New conversion tests passed |
| `cargo +1.92.0 test -p curvine-tests --test write_cache_test test_cache_mode_free -- --test-threads=1` | PASS | Cache-mode free regression passed |
| `cargo +1.92.0 test -p curvine-server --test master_fs_test test_rpc_retry_cache_for_idempotent_operations -- --test-threads=1` | PASS | Delete retry regression passed |

# Dependencies

- Related PRs: None
- Related issues: #1543
- Blocks / blocked by: None

### DIFF
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -309,7 +309,7 @@ impl UnifiedFileSystem {
         if path.path() == mount.cv_path && recursive {
             for status in self.cv.list_status(path).await? {
                 let child = Path::from_str(status.path)?;
-                match self.cv.delete_with_stats(&child, true).await {
+                match self.cv.delete(&child, true).await {
                     Ok(res) => {
                         total.inodes += res.inodes;
                         total.bytes += res.bytes;
@@ -321,7 +321,7 @@ impl UnifiedFileSystem {
                 }
             }
         } else {
-            total = self.cv.delete_with_stats(path, recursive).await?;
+            total = self.cv.delete(path, recursive).await?;
         }
 
         Ok(total)
@@ -1064,7 +1064,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
         self.rename_with_flags(src, dst, RenameFlags::empty()).await
     }
 
-    async fn delete(&self, path: &Path, recursive: bool) -> FsResult<()> {
+    async fn delete(&self, path: &Path, recursive: bool) -> FsResult<FreeResult> {
         let fut = async {
             match self.get_mount_checked(path, RpcCode::Delete).await? {
                 None => self.cv.delete(path, recursive).await,
@@ -1077,16 +1077,21 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
                         );
                     }
 
-                    mount.ufs()?.delete(&ufs_path, recursive).await?;
+                    let mut free_res = mount.ufs()?.delete(&ufs_path, recursive).await?;
 
                     // delete cache
-                    if let Err(e) = self.cv.delete(path, recursive).await {
-                        if !matches!(e, FsError::FileNotFound(_)) {
+                    match self.cv.delete(path, recursive).await {
+                        Ok(res) => {
+                            free_res.inodes += res.inodes;
+                            free_res.bytes += res.bytes;
+                        }
+                        Err(FsError::FileNotFound(_)) => {}
+                        Err(e) => {
                             warn!("failed to delete cache for {}: {}", path, e);
                         }
-                    };
+                    }
 
-                    Ok(())
+                    Ok(free_res)
                 }
             }
         };
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -286,9 +286,9 @@ impl UnifiedFileSystem {
                 None => err_box!(
                     "the current path is not mounted to ufs, so the `free` command cannot be executed."
                 ),
-                // Cache mode: drop Curvine metadata without touching UFS.
-                // cv.free computes the released inode/byte stats and clears blocks,
-                // while cv.delete then removes the remaining Curvine inode.
+                // Cache mode: drop Curvine metadata and blocks via cv.delete.
+                // UFS is untouched; the returned DeleteResult is converted to
+                // FreeResult so the CLI can report inode/byte stats.
                 Some(mount) if mount.is_cache_mode() => {
                     self.free_cache_mode(path, &mount, recursive).await
                 }
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -304,22 +304,27 @@ impl UnifiedFileSystem {
         mount: &MountInfo,
         recursive: bool,
     ) -> FsResult<FreeResult> {
-        let free_res = self.cv.free(path, recursive).await?;
+        let mut total = FreeResult::default();
 
         if path.path() == mount.cv_path && recursive {
             for status in self.cv.list_status(path).await? {
                 let child = Path::from_str(status.path)?;
-                if let Err(e) = self.cv.delete(&child, true).await {
-                    if !matches!(e, FsError::FileNotFound(_)) {
+                match self.cv.delete_with_stats(&child, true).await {
+                    Ok(res) => {
+                        total.inodes += res.inodes;
+                        total.bytes += res.bytes;
+                    }
+                    Err(FsError::FileNotFound(_)) => {}
+                    Err(e) => {
                         return Err(e);
                     }
                 }
             }
         } else {
-            self.cv.delete(path, recursive).await?;
+            total = self.cv.delete_with_stats(path, recursive).await?;
         }
 
-        Ok(free_res)
+        Ok(total)
     }
 
     pub async fn symlink(&self, target: &str, link: &Path, force: bool) -> FsResult<()> {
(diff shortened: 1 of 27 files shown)


crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -286,10 +286,9 @@ impl UnifiedFileSystem {
                 None => err_box!(
                     "the current path is not mounted to ufs, so the `free` command cannot be executed."
                 ),
-                // Cache mode: drop Curvine metadata (and its blocks) via delete.
-                // Master delete already releases worker blocks; calling free first
-                // would only clear locs then delete the inode — redundant.
-                // Use cv.delete (not UnifiedFileSystem::delete) so UFS is untouched.
+                // Cache mode: drop Curvine metadata without touching UFS.
+                // cv.free computes the released inode/byte stats and clears blocks,
+                // while cv.delete then removes the remaining Curvine inode.
```
szbr00: The new comment states "cv.free computes the released inode/byte stats and clears blocks, while cv.delete then removes the remaining Curvine inode", but this path only invokes cv.delete (never cv.free). The previous comment accurately described the single delete call. Consider updating the comment to match the implementation, e.g. "Cache mode: drop Curvine metadata and blocks via cv.delete; UFS is untouched. The returned DeleteResult is converted to FreeResult so the CLI can report inode/byte stats."

crates/common/curvine-proto/proto/master.proto
```diff
@@ -89,7 +89,9 @@ message DeleteRequest {
     required bool recursive = 2;
 }
 
-message DeleteResponse {}
+message DeleteResponse {
+    required FreeResultProto res = 1;
```
szbr00: DeleteResponse changed from an empty message to carrying `required FreeResultProto res = 1`. The PR description notes old clients can keep parsing delete responses, which is true (the field is retained as an unknown field by old clients). However, the reverse direction is not safe: during a rolling master upgrade, a new client built against this proto that sends Delete to an old master will receive an empty DeleteResponse and fail required-field validation.

If rolling upgrades across this version are supported, consider making `res` optional (or relying on proto3 default semantics) so a new client degrades gracefully to 0/0 stats when talking to an old master instead of erroring. If rolling upgrades are not a supported scenario, this is fine as-is.

curvine-master/src/master/meta/store/inode_store.rs
```diff
@@ -176,6 +176,7 @@ impl InodeStore {
                 _ => {
                     deleted_files += 1;
                     let res = self.decrement_inode_nlink(inode.id(), ctime, &mut batch)?;
+                    del_res.bytes += res.bytes;
```
szbr00, line 179: apply_delete increments del_res.inodes for every popped inode (line 160), so DeleteResult.inodes counts both directories and files. The regular free path (unprotected_free) only increments inodes inside the File branch via free_res.add, so FreeResult.inodes counts files only.

Because free_cache_mode builds its FreeResult from cv.delete's DeleteResult, a cache-mode free of a tree will report more inodes than a regular cv.free of the same tree (dirs are included on the cache-mode path but not on the regular path). If callers compare stats across the two free modes this divergence will be surprising.

Consider either excluding directories from DeleteResult.inodes (move the increment into the file branch) to align with free semantics, or document that delete-side inodes include directories while free-side inodes do not.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
lzjqsdd: Thanks. Fixed in a72b3dc: the comment now describes the single cv.delete path and DeleteResult to FreeResult conversion.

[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in a72b3dc: the comment now matches the actual delete-only implementation.

crates/common/curvine-proto/proto/master.proto
```diff
@@ -89,7 +89,9 @@ message DeleteRequest {
...
+}
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in a72b3dc: DeleteResponse.res is now optional and the client uses unwrap_or_default() when talking to an old master.

crates/common/curvine-proto/proto/master.proto
lzjqsdd: Thanks. Fixed in a72b3dc: the field is optional, so new clients degrade to zero stats instead of failing when an old master returns an empty response.

curvine-master/src/master/meta/store/inode_store.rs
```diff
@@ -486,7 +487,9 @@ impl InodeStore {
                         batch.delete_inode(inode_id)?;
 
                         // Collect block info
-                        del_res.blocks.extend(f.get_locs(self)?);
+                        let locs = f.get_locs(self)?;
+                        del_res.bytes = f.get_locs_bytes(&locs) as u64;
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in a72b3dc: get_locs_bytes() is converted with u64::try_from().unwrap_or(0) instead of an unchecked cast.

crates/common/curvine-model/src/result.rs
```diff
@@ -29,3 +29,40 @@ impl FreeResult {
...
+    }
+}
+
+impl From<DeleteResult> for FreeResult {
+    fn from(value: DeleteResult) -> Self {
+        Self {
+            inodes: value.inodes as i64,
+            bytes: value.bytes as i64,
+            blocks: value.blocks,
+        }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in a72b3dc: u64 to i64 conversions now clamp to i64::MAX, and i64 to u64 conversions ignore negative values.

curvine-master/src/master/meta/store/inode_store.rs
lzjqsdd: Thanks. Fixed in a72b3dc: DeleteResult.inodes now counts file inodes only, matching FreeResult semantics; unit and retry tests cover the behavior.

szbr00 (approved):

PR #1546, opened 2026-08-11
fix(fuse): require file owner for setattr gid changes (#1548)

### LINKED ISSUE #1548: [BUG]: FUSE setattr allows non-owner chown(group) after #1500 (ACL bypass / setuid strip)
## Describe the bug

With `fuse.check_permission = true` (default), `check_setattr_permission` no longer requires the caller to own the inode before accepting an effective `FATTR_GID` change. Any non-root caller who is a member of the *target* group can change another user's file group.

**Impact**
1. **ACL bypass**: move a `0640`/`0660` file into a group the attacker belongs to and gain read/write.
2. **Privilege stripping**: effective chown clears `S_ISUID` (and group-exec `S_ISGID`), so a non-owner can strip setuid via `chgrp`.

Master `SetAttr` does not re-check caller credentials; FUSE is the only gate.

## Root cause

Introduced by `6c298c7` (#1500). While fixing pjdfstest `chown(path, -1, gid)` for owners, the FATTR_GID owner gate was dropped and only group membership remained.

## Reproduction

Unit: non-owner (`uid=2000`) + `target_gid` equal to caller gid → currently `Ok(())`, expected `EPERM`.

FUSE E2E (without kernel `default_permissions`): alice owns `0640` `alice:staff`; bob in `bobsgroup` runs `os.chown(path, -1, bob_gid)` → succeeds (bug).

## Expected

POSIX `chown(2)`: only owner (or privileged) may change group, and only to a group they belong to.

## Upstream tracking

Originally filed: https://github.com/gongxun0928/curvine/issues/18  
Fix PR: https://github.com/CurvineIO/curvine/pull/1546

gongxun0928:
# Summary

Restore the POSIX owner gate for effective `FATTR_GID` changes in `check_setattr_permission` (owner ∧ in-target-group). After #1500, membership alone was sufficient, allowing non-owners to `chgrp` into their group (ACL bypass / setuid strip).

Closes #1548

# Issue Describe / Design

Related fork issue: https://github.com/gongxun0928/curvine/issues/18

## Root cause

Introduced by `6c298c7` (#1500). While fixing pjdfstest owner `chown(path, -1, gid)`, the FATTR_GID **owner gate** was removed and only group-membership remained.

**Before (#1500):** owner check then membership.  
**After (#1500, buggy):** membership only → non-owner in target group allowed.

## Reproduction

**Unit:** `check_setattr_permission(true, hdr(2000,3000), file_uid=1000, file_gid=100, FATTR_GID, None, Some(3000))` → was `Ok(())`, expected `EPERM`.

**FUSE E2E:** with userspace permission path (no kernel `default_permissions`), alice `0640` `alice:staff`; bob in `bobsgroup` runs `os.chown(path, -1, bob_gid)` → succeeded before fix.

Note: default mounts force `default_permissions`, so the kernel may EPERM first and hide the userspace bug.

## Fix approach

Require `caller_uid == file_uid` before allowing an effective gid change; keep membership check; preserve root / `check_permission=false` short-circuit and #1500 owner uid-noop + gid paths.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | Restore owner gate for effective FATTR_GID; add regression test | Behavior change: non-owner gid change → `EPERM` even if in target group |

# Test verified

| `cargo test -p curvine-fuse setattr_permission` | PASS | Includes `denies_non_owner_gid_change_even_if_in_target_group` |
| E2E non-owner `os.chown(-1,gid)` without `default_permissions` | PASS | EPERM after fix |
| E2E owner supplementary-group chown (#1500) | PASS | Still OK |
| E2E `chown(-1,-1)` preserves S_ISUID | PASS | |

# Dependencies

- Related PRs: #1547 (independent same-gid S_ISUID fix)
- Related issues: #1548; fork https://github.com/gongxun0928/curvine/issues/18
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -822,7 +822,20 @@ impl CurvineFileSystem {
             }
         }
 
+        // Changing the group is also privileged for non-root: caller must own the
+        // file *and* belong to the target group (POSIX chown). Membership alone is
+        // not sufficient — #1500 dropped the owner gate and allowed non-owners who
+        // share the target group to chgrp (ACL bypass / setuid strip).
+        // See CurvineIO/curvine#1548 (originally gongxun0928/curvine#18).
         if let Some(gid) = target_gid {
+            if gid != file_gid && caller_uid != file_uid {
+                return err_fuse!(
+                    libc::EPERM,
+                    "setattr gid change denied: caller uid {} is not file owner (uid {})",
+                    caller_uid,
+                    file_uid
+                );
+            }
             if gid != file_gid && !FuseUtils::caller_in_file_group(caller_gid, gid, caller_pid) {
                 return err_fuse!(
                     libc::EPERM,
@@ -2682,6 +2695,23 @@ mod tests {
             Some(100),
         )
         .unwrap();
+    }
+
+    /// Issue #18 / gongxun0928#18: non-owner who is a member of the *target* group must
+    /// still get EPERM (owner gate ∧ group membership).
+    #[test]
+    fn setattr_permission_denies_non_owner_gid_change_even_if_in_target_group() {
+        let err = super::CurvineFileSystem::check_setattr_permission(
+            true,
+            &hdr(2000, 3000, 0),
+            1000, // file owned by another user
+            100,  // current file gid
+            FATTR_GID,
+            None,
+            Some(3000), // caller primary gid — membership OK, ownership not
+        )
+        .unwrap_err();
+        assert_eq!(err.errno(), libc::EPERM);
     }
 
     #[test]


szbr00 (commented):
Reviewed the diff in `curvine-fuse/src/fs/curvine_file_system.rs`. The added owner gate (`caller_uid != file_uid` before the existing `caller_in_file_group` membership check) correctly restores POSIX `chown(2)` semantics for effective `FATTR_GID` changes, closing the #1500 regression where target-group membership alone let a non-owner `chgrp` (ACL bypass / setuid strip).

- Correctness: all four (owner × in-target-group) quadrants are covered by unit tests, including the new `setattr_permission_denies_non_owner_gid_change_even_if_in_target_group` regression. Sentinels (`chown(-1,-1)`, `chown(-1, file_gid)` no-op), root, and `check_permission=false` short-circuits are preserved; `fchown` (FATTR_FH) is gated by the same path.
- Safety: purely additive guard, no new locks/panics/unsafe.
- Performance: metadata path only; the owner check short-circuits before the `caller_in_file_group` proc lookup, so non-owner `chgrp` attempts are not slower.

No must-fix or suggestion-level issues found. The only observation is a low-value diagnostic-string nit on the new EPERM message wording, which is consistent with the existing `setattr uid change requires privilege for uid {}` pattern and not worth changing.

szbr00 (approved):

Issue #1551, opened 2026-08-11
[BUG]: master crash-loops with "hs.commit … out of range [0, 0]" after a wiped follower rejoins via snapshot


papapumpkin:
**Describe the bug**

When a master's raft storage (journal and meta) is cleared and it rejoins a healthy quorum, it installs a snapshot from the leader successfully on its first boot, but then crash-loops permanently on every subsequent restart with:

```
panic occurred at raft-0.7.0/src/raft.rs:2786: hs.commit <N> is out of range [0, 0], raft_id: <id>
  raft::raft::Raft::<T>::new
  raft::raw_node::RawNode::<T>::new
  curvine_server::master::journal::journal_system::JournalSystem::start
```

The panic is deterministic and reads from persisted state, so the master cannot self-heal. It stays in `CrashLoopBackOff`. Replacing a single master this way permanently removes it from the quorum. Doing it to a second master before the first is repaired takes the cluster below quorum with no automatic path back.

This was reproduced by replacing one of three Kubernetes master pods with fresh node-local and EBS-backed journal and meta volumes, leaving the other two masters healthy.

**To Reproduce**

1. Run a 3-master raft cluster and write enough metadata that the leader has taken at least one snapshot and compacted its log, so the leader `commit` index is well above 0.
2. Stop one **follower** and clear its journal and meta storage (fresh, empty volumes), leaving the other two masters healthy and holding quorum.
3. Start the follower. On this first boot it receives and applies a snapshot from the leader (logs `apply_snapshot`), then the process restarts, for example when a liveness check restarts it or on any pod reschedule.
4. On the next boot it panics with `hs.commit <N> is out of range [0, 0]` and crash-loops indefinitely.

**Actual behavior**

The follower's boot sequence:

```
init raft state: RaftState { hard_state: HardState { term: T, vote: 1, commit: <N> }, conf_state: ... voters: [] }
apply_snapshot: fs_dir_restore=… total=…
switched to configuration, raft_id: <id>, voters: {1, 2, 3}
panic: hs.commit <N> is out of range [0, 0], raft_id: <id>
```

- The first boot persists `HardState.commit = <N>` (advanced by the leader) and installs the snapshot blob.
- Every later boot reconstructs storage with `snapshot_metadata.index = 0` and an empty log, so `last_index()` returns `0` while `HardState.commit` is `<N>`.
- `RawNode::new()` asserts `commit <= last_index` (`<N> <= 0`) and panics before the master can serve.

[...]

PR #1554, opened 2026-08-12
fix(raft): restore durable snapshot recovery

jlon:
## Summary
Persist and restore the full local Raft recovery state so a voter restarted after snapshot compaction does not lose its durable progress.

## Issue/Design
This is one bounded part of #1551 and supersedes the RocksStorageCore portion of #1552. A snapshot previously updated in-memory progress while the restart path restored only the log range and hard state. A snapshot-only or compacted-log restart could therefore reconstruct an inconsistent durable range.

## Changes
| Area | Change |
| --- | --- |
| Raft storage | Load the persisted snapshot and membership configuration before serving the Raft state. |
| Atomicity | Persist snapshot, compacted range, hard state, and conf state in one RocksDB write batch, then update memory. |
| Recovery validation | During storage initialization, reject corrupt records and hard-state commits outside the durable range. Do not perform that range check while persisting a Ready, because its HardState can legitimately precede entries in the same Ready. |
| Tests | Add restart, compaction, corruption, range-validation, and HardState-before-append recovery coverage in the curvine-raft test module. |

## Test verified
- `cargo fmt --check -- crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs crates/metadata/curvine-raft/tests/rocks_storage_recovery_test.rs`
- `cargo test -q -p curvine-raft --test rocks_storage_recovery_test`
- `cargo clippy -q -p curvine-raft --all-targets -- --deny=warnings --allow clippy::uninlined-format-args`
- `git diff --check`

## Dependencies

### DIFF
--- a/crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
+++ b/crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
@@ -150,8 +150,6 @@ impl RocksStorageCore {
     }
 
     pub fn set_hard_state(&mut self, hs: HardState) -> RaftResult<()> {
-        self.validate_hard_state_commit(hs.commit)?;
-
         let mut batch = StoreWriteBatch::new(&self.db);
         batch.set_state(&hs)?;
         batch.commit()?;
@@ -161,8 +159,6 @@ impl RocksStorageCore {
     }
 
     pub fn set_hard_state_commit(&mut self, commit: u64) -> RaftResult<()> {
-        self.validate_hard_state_commit(commit)?;
-
         let mut hard_state = self.raft_state.hard_state.clone();
         hard_state.set_commit(commit);
 
--- a/crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
+++ b/crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
@@ -15,7 +15,7 @@
 use crate::proto::raft::SnapshotData;
 use crate::raft::{RaftError, RaftResult, LOG_START_INDEX};
 use crate::rocksdb::{DBConf, DBEngine, RocksUtils, WriteBatch};
-use curvine_core_error::{err_box, err_ext, CommonResult};
+use curvine_core_error::{err_box, err_ext, CommonResult, ErrorExt};
 use log::warn;
 use prost::Message;
 use raft::eraftpb::{ConfState, Entry, HardState, Snapshot, SnapshotMetadata};
@@ -28,6 +28,7 @@ pub struct RocksStorageCore {
     db: DBEngine,
     first_index: Option<u64>,
     last_index: Option<u64>,
+    init_error: Option<String>,
 
     pub(crate) trigger_snap_unavailable: bool,
     pub(crate) trigger_log_unavailable: bool,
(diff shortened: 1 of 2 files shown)


crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
```diff
@@ -104,22 +150,45 @@ impl RocksStorageCore {
...
         let mut batch = StoreWriteBatch::new(&self.db);
-        batch.set_state(&self.raft_state.hard_state)?;
+        batch.set_state(&hard_state)?;
         batch.commit()?;
 
+        self.raft_state.hard_state = hard_state;
+        Ok(())
+    }
+
+    fn validate_hard_state_commit(&self, commit: u64) -> RaftResult<()> {
```
szbr00, line 173: The runtime safety of `validate_hard_state_commit` depends on a subtle raft-rs 0.7.0 ordering property: that `ready.hs().commit` never exceeds the durable `last_index` at the moment `set_hard_state` is called. This holds for the leader propose path (commit is delivered via `light_rd.commit_index` after `advance`/`append`) but is fragile and not asserted anywhere. A future raft-rs upgrade or a reordering of `on_ready` could silently turn this into a cluster-wide outage. Consider adding a multi-node RocksLogStorage restart/recovery integration test (e.g. un-skipping `rocks_snap_test` or a new follower-receives-entries-then-restarts scenario) so the invariant is exercised end-to-end.

crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
```diff
@@ -104,22 +150,45 @@ impl RocksStorageCore {
     }
 
     pub fn set_hard_state(&mut self, hs: HardState) -> RaftResult<()> {
-        self.raft_state.hard_state = hs.clone();
+        self.validate_hard_state_commit(hs.commit)?;
```
szbr00: `validate_hard_state_commit` here rejects `commit > last_index`, but `on_ready` calls `set_hard_state(hs)` before `append(entries)` (raft_node.rs:461 vs 485). For a follower, raft-rs advances `committed` during `step` via `RaftLog::maybe_append` -> `commit_to` (raft_log.rs:275), so `ready.hs().commit` can reference entries that are still in the same Ready and not yet appended to storage. The validation then fails on a legitimate commit, `on_ready` returns Err, and the raft node stops. This breaks the follower append path in any multi-node cluster.

The single-node test passes only because the leader path advances commit during `advance` via `on_persist_entries` (raft.rs:1056) and persists it through `set_hard_state_commit` after `append`.

Suggest keeping the range check only at startup in `init_state` (which is the corruption case this PR targets) and removing it from `set_hard_state` / `set_hard_state_commit`, since raft-rs already guarantees the commit is valid for the in-memory log. A multi-node integration test (the existing `rocks_snap_test` is commented out) would lock in this invariant.

crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
jlon: Thanks. Fixed in 118cacc7: the runtime range check is removed, and a persistent HardState-before-append recovery test covers the exact storage ordering. Corruption validation remains during initialization, where the durable range invariant applies.

crates/metadata/curvine-raft/src/raft/storage/rocks_storage_core.rs
jlon: Thanks. Fixed in 118cacc7. set_hard_state and set_hard_state_commit no longer reject commit values before entries in the same Ready are appended; the range check remains only during storage recovery.

szbr00 (approved):

PR #1555, opened 2026-08-12
fix(rocksdb): flush all column families before checkpoint

jlon:
## Summary
Flush every configured column family before creating a RocksDB checkpoint.

## Issue/Design
This is an independent durability correction extracted from #1552. `DB::flush_opt` only flushes the default column family, while Curvine metadata is stored in named column families. A checkpoint created with WAL disabled could therefore omit recently written metadata.

## Changes
| Area | Change |
| --- | --- |
| Checkpoint preparation | Resolve every configured column family and invoke RocksDB `flush_cfs_opt`. |
| Test | Write to two named column families with WAL disabled, checkpoint, restore, and verify both values remain. |

## Test verified
- `cargo fmt --check -- crates/adapters/curvine-rocksdb/src/rocksdb/db_engine.rs crates/adapters/curvine-rocksdb/tests/rocks_db_test.rs`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/rocksdb-checkpoint cargo test -q -p curvine-rocksdb --test rocks_db_test`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/rocksdb-checkpoint cargo clippy -q -p curvine-rocksdb --test rocks_db_test -- -D warnings`
- `git diff --check`

## Dependencies

### DIFF
--- a/crates/adapters/curvine-rocksdb/src/rocksdb/db_engine.rs
+++ b/crates/adapters/curvine-rocksdb/src/rocksdb/db_engine.rs
@@ -278,7 +278,13 @@ impl DBEngine {
     pub fn flush_mem(&self, sync: bool) -> CommonResult<()> {
         let mut opts = FlushOptions::default();
         opts.set_wait(sync);
-        self.db.flush_opt(&opts)?;
+        let cfs = self
+            .conf
+            .family_list
+            .iter()
+            .map(|cf| self.cf(cf))
+            .collect::<CommonResult<Vec<_>>>()?;
+        self.db.flush_cfs_opt(&cfs, &opts)?;
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


szbr00 (approved):
Verified the fix and the regression test on the PR branch (`cargo test -p curvine-rocksdb --test rocks_db_test`, 5/5 passed).

- Correct: `flush_opt` only flushes the default CF; `flush_cfs_opt` over `conf.family_list` covers named CFs, which is the durability gap when WAL is disabled.
- No regression for the default CF: `family_list` is initialized with `["default"]` (`db_conf.rs:272`), so it is still flushed.
- No perf concern: `flush_mem` is only reached via `create_checkpoint` (raft snapshot path, periodic); `commit_and_flush_mem` has no callers.
- Test is a valid regression test (writes to two named CFs with WAL disabled, checkpoints, restores, asserts both values survive).

LGTM.

PR #1556, opened 2026-08-12
fix(master): skip empty placeholder snapshots

jlon:
## Summary
Do not install a zero-index, empty placeholder Raft snapshot over initialized Master metadata.

## Issue/Design
This independent guard is extracted from #1552. The placeholder snapshot carries neither metadata state nor a meaningful Raft index. Applying it would replace an initialized journal state with an empty payload.

## Changes
| Area | Change |
| --- | --- |
| Journal restore | Skip only snapshots with index `0` and empty payload. |
| Test | Verify that the existing journal state is preserved when this placeholder is received. |

## Test verified
- `cargo fmt --check -- curvine-master/src/master/journal/journal_loader.rs curvine-server/tests/journal_test.rs`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/empty-snapshot cargo test -q -p curvine-server --test journal_test`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/empty-snapshot cargo clippy -q -p curvine-server --test journal_test -- -D warnings`
- `git diff --check`

## Dependencies

### DIFF
--- a/curvine-master/src/master/journal/journal_loader.rs
+++ b/curvine-master/src/master/journal/journal_loader.rs
@@ -28,7 +28,7 @@ use curvine_raft::conf::JournalConfExt;
 use curvine_raft::proto::raft::{AppliedIndex, FsmState, SnapshotData};
 use curvine_raft::raft::storage::{AppStorage, ApplyMsg, LogStorage, RocksLogStorage};
 use curvine_raft::raft::{RaftClient, RaftResult, RaftUtils};
-use curvine_runtime::common::{FileUtils, LocalTime, TimeSpent};
+use curvine_runtime::common::{FileUtils, TimeSpent};
 use curvine_runtime::runtime::{RpcRuntime, Runtime};
 use curvine_runtime::sync::channel::{AsyncChannel, AsyncReceiver, AsyncSender, CallChannel};
 use log::{debug, error, info, warn};
@@ -482,16 +482,23 @@ impl JournalLoader {
     fn apply_snapshot0(&self, snapshot: SnapshotData) -> RaftResult<()> {
         let mut spend = TimeSpent::new();
 
-        // Resolve restore path first. Always measure dir size for safety checks;
-        // the logged checkpoint_size may still be 0 when info logging is disabled.
-        let is_empty_placeholder_snapshot =
-            snapshot.files_data.is_none() && snapshot.bytes_data.is_none();
+        // Raft uses the default value as a no-snapshot placeholder. It carries no
+        // filesystem state and must not create a restore directory or replace metadata.
+        let is_empty_placeholder_snapshot = snapshot.snapshot_id == 0
+            && snapshot.files_data.is_none()
+            && snapshot.bytes_data.is_none();
+        if is_empty_placeholder_snapshot {
+            warn!("skip zero-index empty placeholder snapshot");
+            return Ok(());
+        }
+
         let restore_path = match &snapshot.files_data {
             Some(data) => data.dir.clone(),
             None => {
-                let dir = self.fs_dir.read().get_checkpoint_path(LocalTime::mills());
-                FileUtils::create_dir(&dir, true)?;
-                dir
+                return err_box!(
+                    "refusing to apply empty snapshot {} without checkpoint files",
+                    snapshot.snapshot_id
+                )
             }
         };
         let actual_size = FileUtils::dir_size(&restore_path).unwrap_or_else(|e| {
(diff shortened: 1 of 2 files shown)


curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -484,6 +484,8 @@ impl JournalLoader {
 
         // Resolve restore path first. Always measure dir size for safety checks;
         // the logged checkpoint_size may still be 0 when info logging is disabled.
+        let is_empty_placeholder_snapshot =
+            snapshot.files_data.is_none() && snapshot.bytes_data.is_none();
```
szbr00: The PR description states the guard should skip only snapshots with "index 0 and empty payload", but `is_empty_placeholder_snapshot` only checks the payload and ignores `snapshot_id`. A snapshot with a non-zero `snapshot_id` but empty payload (which would indicate a real restore bug) would be silently skipped here instead of triggering the existing refusal, masking the problem.

Tighten the guard to match the intent, e.g.:
    let is_empty_placeholder_snapshot =
        snapshot.snapshot_id == 0
            && snapshot.files_data.is_none()
            && snapshot.bytes_data.is_none();

and include `snapshot_id` in the warn log so the placeholder vs. real-snapshot distinction is visible in operations.

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -513,6 +515,13 @@ impl JournalLoader {
             let fs_dir = self.fs_dir.read();
             let (dir_count, file_count) = fs_dir.get_file_counts();
             if actual_size == 0 && file_count > 0 {
+                if is_empty_placeholder_snapshot {
+                    warn!(
+                        "skip empty placeholder snapshot at {} over filesystem with {} files and {} dirs",
+                        restore_path, file_count, dir_count
```
szbr00: When `files_data` is `None`, `restore_path` is already created via `FileUtils::create_dir` above before this placeholder check runs. Returning early here leaves that empty checkpoint directory behind, and it accumulates across repeated placeholder installs (only `create_snapshot0` calls `purge_checkpoint`).

Consider checking `is_empty_placeholder_snapshot` before creating the directory and returning early, or removing `restore_path` best-effort on this skip path.

curvine-server/tests/journal_test.rs
```diff
@@ -544,6 +544,12 @@ fn test_apply_snapshot_refuses_empty_over_populated_files() -> CommonResult<()>
     let (dir_count, file_count) = fs.get_file_counts();
     assert!(file_count > 0, "expected files after create");
 
+    rt.block_on(loader.apply_snapshot(SnapshotData::default()))?;
```
szbr00, line 542: This new assertion only covers the placeholder case (`SnapshotData::default()` has `snapshot_id == 0`). Add a negative case verifying that a non-placeholder empty snapshot (e.g. `snapshot_id > 0` with both `files_data` and `bytes_data` set to `None`) is still refused with "refusing to apply empty snapshot", so the skip guard does not regress to swallowing real snapshots that happen to have an empty payload.

curvine-master/src/master/journal/journal_loader.rs
jlon: Thanks. Fixed in 2183f4c2. The predicate now requires index 0 as well as no file or byte payload, and the log includes the snapshot index.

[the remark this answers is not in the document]
jlon: Thanks. Fixed in 2183f4c2. Only a zero-index snapshot with neither file nor byte payload is skipped. A nonzero empty snapshot is rejected and covered by a regression test.

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -513,6 +515,13 @@ impl JournalLoader {
...
+                    );
+                    return Ok(());
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in 2183f4c2. The zero-index placeholder is recognized before any restore directory is created, so repeated placeholders leave no checkpoint directories behind.

curvine-server/tests/journal_test.rs
jlon: Thanks. Added the non-placeholder empty-snapshot rejection case in 2183f4c2.

curvine-master/src/master/journal/journal_loader.rs
jlon: Thanks. Fixed in 2183f4c2. The zero-index placeholder check now runs before restore-path creation.

szbr00 (approved):

PR #1557, opened 2026-08-12
fix(raft): fence leader promotion after metadata apply

jlon:
## Summary
Make Raft durability, metadata application, and Leader publication occur in one strict order.

## Issue/Design
This independent protocol fix is extracted from #1552 and addresses #1551. A voter could become externally active before all committed journal entries had advanced the local metadata ID high-water marks. The next mutation could then reuse an inode ID.

## Changes
| Area | Change |
| --- | --- |
| Raft Ready | Persist entries before the HardState that commits them, and send messages only after persistence. |
| Proposal response | Wait for apply-worker acknowledgement before returning a Leader proposal response. |
| Leader promotion | Make `role_change(Leader)` wait until committed metadata is replayed; only then enable Leader UFS replay. |
| Invariant | Reject replay that allocates an inode ID at or below the local high-water mark. |
| Tests | Cover promotion with an unapplied committed entry and duplicate allocated inode detection. |

## Test verified
- `cargo fmt --check -- crates/metadata/curvine-raft/src/raft/raft_node.rs crates/metadata/curvine-raft/src/raft/storage/apply_msg.rs curvine-master/src/master/journal/entry.rs curvine-master/src/master/journal/journal_loader.rs curvine-master/src/master/journal/journal_system.rs crates/metadata/curvine-raft/tests/raft_node_test.rs`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/ordered-apply cargo test -q -p curvine-raft --test raft_node_test -p curvine-server --test journal_test`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/ordered-apply cargo clippy -q -p curvine-raft -p curvine-server --tests -- -D warnings`
- `git diff --check`

## Dependencies

### DIFF
--- a/crates/metadata/curvine-raft/src/raft/raft_node.rs
+++ b/crates/metadata/curvine-raft/src/raft/raft_node.rs
@@ -463,22 +463,10 @@ where
         // Get the ready structure.
         let mut ready = self.raw.ready();
 
-        // If hard state changes, it needs to be saved.
-        if let Some(hs) = ready.hs() {
-            let store = self.raw.mut_store();
-            store.set_hard_state(hs)?;
-        }
-
         let soft_state = ready.ss().map(|ss| SoftState {
             leader_id: ss.leader_id,
             raft_state: ss.raft_state,
         });
-
-        // Get the message that needs to be sent to other nodes.
-        // Only the leader call returns true.
-        if !ready.messages().is_empty() {
-            self.send_messages(ready.take_messages()).await?;
-        }
 
         // Process snapshots.
         if *ready.snapshot() != Snapshot::default() {
@@ -486,11 +474,20 @@ where
                 .gen_apply_snapshot_job(ready.snapshot().clone())?;
         }
 
-        // Persist new log entries first so followers can be notified immediately
-        // via persisted_messages.  FSM apply of already-committed entries can
-        // happen afterwards without blocking the persistence pipeline.
+        // Persist entries before the HardState that may commit them. A crash may
+        // leave extra uncommitted entries, but must never leave commit past tail.
         if !ready.entries().is_empty() {
             self.storage.append(&ready.entries()[..])?;
+        }
+
+        if let Some(hs) = ready.hs() {
+            let store = self.raw.mut_store();
+            store.set_hard_state(hs)?;
+        }
+
+        // Raft messages may be sent only after their Ready state is durable.
+        if !ready.messages().is_empty() {
+            self.send_messages(ready.take_messages()).await?;
         }
 
         // Get the message that the leader has fallen into the disk and send these messages to other nodes.
--- a/crates/metadata/curvine-raft/src/raft/raft_node.rs
+++ b/crates/metadata/curvine-raft/src/raft/raft_node.rs
@@ -647,20 +647,13 @@ where
         req_id: i64,
         sender: Callback,
         response: ProposeResponse,
-        apply_done: Option<CallReceiver<RaftResult<()>>>,
+        apply_done: CallReceiver<RaftResult<()>>,
     ) {
         let response_msg = Builder::new_rpc(RaftCode::Propose)
             .response(ResponseStatus::Success)
             .proto_header(response)
             .req_id(req_id)
             .build();
-
-        let Some(apply_done) = apply_done else {
-            if sender.send(Ok(response_msg)).is_err() {
-                warn!("The client connection has been closed, req {}", req_id)
-            }
-            return;
-        };
 
         rt.spawn(async move {
             let result = match apply_done.receive().await {
@@ -731,6 +724,17 @@ where
                 let apply = self.apply_propose(entry, should_respond).await?;
                 if should_respond {
                     let Some(entry_context) = entry_context else {
+                        warn!(
+                            "leader committed entry has no response context, index {}, term {}",
+                            entry_index, entry_term
+                        );
+                        continue;
+                    };
+                    let Some(apply_done) = apply.apply_done else {
+                        warn!(
+                            "leader committed entry has no apply acknowledgement, index {}, term {}",
+                            entry_index, entry_term
+                        );
                         continue;
                     };
                     let req_id: i64 = match SerdeUtils::deserialize(&entry_context) {
@@ -749,7 +753,7 @@ where
                             req_id,
                             sender,
                             apply.response,
-                            apply.apply_done,
+                            apply_done,
                         ),
 
                         None => {
--- a/crates/metadata/curvine-raft/src/raft/raft_node.rs
+++ b/crates/metadata/curvine-raft/src/raft/raft_node.rs
@@ -20,11 +20,14 @@ use crate::raft::raft_error::RaftError;
 use crate::raft::storage::{AppStorage, ApplyMsg, LogStorage, PeerStorage};
 use crate::raft::*;
 use crate::utils::SerdeUtils;
+use curvine_core_error::ErrorExt;
+use curvine_io::DataSlice;
 use curvine_net::net::InetAddr;
 use curvine_rpc::client::dispatch::{Callback, Envelope};
 use curvine_rpc::message::{Builder, RefMessage, ResponseStatus};
 use curvine_runtime::common::{DurationUnit, LocalTime, TimeSpent};
 use curvine_runtime::runtime::{RpcRuntime, Runtime};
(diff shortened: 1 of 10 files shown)


curvine-master/src/master/journal/entry.rs
```diff
@@ -303,6 +303,15 @@ impl JournalEntry {
...
     }
+
+    pub fn allocated_inode_id(&self) -> Option<i64> {
+        match self {
+            JournalEntry::Mkdir(e) => Some(e.dir.id),
+            JournalEntry::CreateFile(e) => Some(e.file.id),
+            JournalEntry::Symlink(e) => Some(e.new_inode.id),
+            _ => None,
+        }
+    }
```
szbr00, line 314: `allocated_inode_id` is added here but is never called anywhere in this PR. The PR description lists an "Invariant" change to "Reject replay that allocates an inode ID at or below the local high-water mark", but I cannot find any rejection logic in the diff. `fs_dir.update_last_inode_id` (journal_loader.rs:261) still silently skips when `new_value <= old_value` rather than rejecting, and no call site uses `allocated_inode_id` to enforce the invariant.

Could you clarify whether the rejection enforcement is missing, planned for a follow-up, or whether the description should be updated? As-is, the claimed invariant is not enforced by this change.

crates/metadata/curvine-raft/src/raft/storage/apply_msg.rs
```diff
@@ -32,13 +33,26 @@ impl ApplyMsg {
...
         ApplyMsg::Scan(applied_index)
     }
 
     pub fn take_entry(self) -> Entry {
         match self {
             ApplyMsg::Entry(entry) => entry,
+            ApplyMsg::EntryWithAck((entry, _)) => entry,
+            _ => panic!("invalid entry"),
+        }
+    }
```
szbr00: `take_entry` is no longer used after both `HashAppStorage` and `RocksAppStorage` switched to `take_entry_with_ack`. Consider removing it to avoid a panic-prone API that is now dead code.

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -350,6 +350,30 @@ impl JournalLoader {
...
+        let committed = self.log_store.hard_state().commit;
+        if metadata_applied < committed {
+            return err_box!(
+                "metadata catch-up stopped before committed raft index: applied={}, commit={}",
+                metadata_applied,
+                committed
+            );
+        }
+        Ok(())
+    }
```
szbr00, line 389: `catch_up_committed_metadata` reads `hard_state().commit` inside `apply_msg`'s Scan loop and then re-reads it for the final comparison. This is currently safe only because the raft-node thread is blocked awaiting `role_change` inside `on_ready`, so no concurrent `set_hard_state_commit` can advance commit between the two reads. That invariant is implicit and fragile — a short comment noting "safe because role_change is awaited synchronously by the raft node" would help prevent a future caller from invoking this outside that blocking context and hitting a spurious "metadata catch-up stopped before committed raft index" error.

crates/metadata/curvine-raft/src/raft/raft_node.rs
```diff
@@ -610,11 +615,74 @@ where
...
+                        .req_id(req_id)
+                        .data(DataSlice::Buffer(error.encode()))
+                        .build())
+                }
+            };
+            if sender.send(result).is_err() {
+                warn!("The client connection has been closed, req {}", req_id)
+            }
+        });
     }
```
szbr00, line 679: `send_propose_response_after_apply` always reaches this function with `apply_done = Some` when `should_respond` is true (the only caller passes `wait_for_response = should_respond`), so the `None` early-return branch is unreachable in practice. Additionally, the success `response_msg` is built up front and then discarded/rebuilt on both error arms. It's minor, but either documenting the `None` branch as a defensive fallback or building the success message only in the `Ok(Ok(()))` arm would make the intent clearer.

crates/metadata/curvine-raft/src/raft/storage/hash_app_storage.rs
```diff
@@ -86,19 +86,28 @@ where
...
-            index: entry.index,
-            op_id: 0,
-            rpc_id: 0,
-        };
-
-        Ok(())
+        let (entry, ack) = msg.take_entry_with_ack();
+        let result = (|| -> RaftResult<()> {
+            let mut map = self.write()?;
+            let pairs: (K, V) = SerdeUtils::deserialize(&entry.data)?;
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in b059c2fb. HashAppStorage now accepts ApplyMsg::Scan as a no-op and returns structured errors for other non-entry control messages.

crates/metadata/curvine-raft/src/raft/storage/hash_app_storage.rs
```diff
@@ -188,21 +197,30 @@ where
...
-            index: entry.index,
-            op_id: 0,
-            rpc_id: 0,
-        };
-
-        Ok(())
+        let (entry, ack) = msg.take_entry_with_ack();
+        let result = (|| -> RaftResult<()> {
+            let db = self.lock()?;
+            let pairs: (K, V) = SerdeUtils::deserialize(&entry.data)?;
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in b059c2fb. RocksAppStorage now accepts ApplyMsg::Scan as a no-op and returns structured errors for other non-entry control messages.

curvine-master/src/master/journal/journal_loader.rs
jlon: Thanks. Fixed in b059c2fb. catch_up_committed_metadata snapshots hard_state.commit before scanning and compares against that fixed target, so it no longer relies on a moving value.

curvine-master/src/master/journal/entry.rs
```diff
@@ -303,6 +303,15 @@ impl JournalEntry {
             | JournalEntry::Snapshot(_) => Vec::new(),
         }
     }
+
+    pub fn allocated_inode_id(&self) -> Option<i64> {
+        match self {
+            JournalEntry::Mkdir(e) => Some(e.dir.id),
+            JournalEntry::CreateFile(e) => Some(e.file.id),
+            JournalEntry::Symlink(e) => Some(e.new_inode.id),
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in b059c2fb. JournalLoader::apply0 now rejects a follower replay whose allocated inode ID is at or below the local high-water mark, with a duplicate-allocation regression test.

curvine-master/src/master/journal/entry.rs
jlon: Thanks. Fixed in b059c2fb. The high-water check is now enforced in JournalLoader::apply0 before replay applies the command.

crates/metadata/curvine-raft/src/raft/raft_node.rs
jlon: Thanks. Fixed in b059c2fb. The response helper now receives a non-optional acknowledgement only on the response path; the unreachable fallback was removed.

crates/metadata/curvine-raft/tests/raft_node_test.rs
```diff
@@ -166,20 +167,127 @@ impl TestKvAppStorage {
...
+        self.map.read().unwrap().get(key).cloned()
+    }
+
+    async fn apply_entry(&self, entry: Entry, wait: bool) -> RaftResult<()> {
+        let pair: (String, String) = SerdeUtils::deserialize(&entry.data)?;
+        self.started.store(true, Ordering::SeqCst);
+        self.started_notify.notify_one();
+        if wait {
+            self.release.notified().await;
+        }
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in b059c2fb. The test now uses a zero-permit Semaphore; add_permits persists the release and cannot be lost before apply waits.

crates/metadata/curvine-raft/src/raft/storage/apply_msg.rs
jlon: Thanks. Fixed in b059c2fb. The unused panic-prone take_entry API was removed; the remaining extraction API returns RaftResult.

szbr00 (approved):

PR #1558, opened 2026-08-12
feat(master): version committed journal commands

jlon:
## Summary
Add a versioned journal envelope while retaining read compatibility with existing legacy journal batches.

## Issue/Design
This is the protocol-layer extraction from #1552. Commit-before-apply metadata commands require an explicit journal wire format. The envelope has a stable magic prefix and version number; legacy serialized batches remain decodable.

## Changes
| Area | Change |
| --- | --- |
| Wire format | Add `JournalEnvelope`, `JournalCommandBatch`, and versioned command variants. |
| Compatibility | Decode legacy batches when the magic prefix is absent. |
| Safety | Reject unknown envelope versions and unsupported metadata commands until their execution PRs are merged. |
| Test | Replay a versioned envelope carrying an unchanged legacy command. |

## Test verified
- `cargo fmt --check -- curvine-master/src/master/journal/entry.rs curvine-master/src/master/journal/journal_loader.rs curvine-master/src/master/journal/journal_system.rs`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/journal-versioning cargo test -q -p curvine-server --test journal_test`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/journal-versioning cargo clippy -q -p curvine-server --test journal_test -- -D warnings`
- `git diff --check`

## Dependencies
Stacked on #1557 for the Leader promotion fence. Merge after #1557.

### DIFF
--- a/curvine-master/src/master/journal/entry.rs
+++ b/curvine-master/src/master/journal/entry.rs
@@ -28,7 +28,7 @@ pub(crate) struct CvMetadataChange {
     pub(crate) include_subtree: bool,
 }
 
-pub const JOURNAL_ENVELOPE_MAGIC: &[u8; 8] = b"CVJNL001";
+pub const JOURNAL_ENVELOPE_MAGIC: &[u8; 8] = b"CVJNLHDR";
 pub const JOURNAL_ENVELOPE_VERSION: u16 = 1;
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
@@ -548,14 +548,41 @@ impl DecodedJournalBatch {
         self.len() == 0
     }
 
-    pub fn into_commands(self) -> Vec<JournalCommand> {
+    pub fn contains_metadata_commands(&self) -> bool {
+        matches!(
+            self,
+            DecodedJournalBatch::Versioned(batch)
+                if batch
+                    .commands
+                    .iter()
+                    .any(|command| matches!(command, JournalCommand::Metadata(_)))
+        )
+    }
+
+    pub fn into_commands(self) -> JournalCommandIter {
         match self {
-            DecodedJournalBatch::Legacy(batch) => batch
-                .batch
-                .into_iter()
-                .map(JournalCommand::Legacy)
-                .collect(),
-            DecodedJournalBatch::Versioned(batch) => batch.commands,
+            DecodedJournalBatch::Legacy(batch) => {
+                JournalCommandIter::Legacy(batch.batch.into_iter())
+            }
+            DecodedJournalBatch::Versioned(batch) => {
+                JournalCommandIter::Versioned(batch.commands.into_iter())
+            }
+        }
+    }
+}
+
+pub enum JournalCommandIter {
+    Legacy(std::vec::IntoIter<JournalEntry>),
+    Versioned(std::vec::IntoIter<JournalCommand>),
+}
+
+impl Iterator for JournalCommandIter {
+    type Item = JournalCommand;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            JournalCommandIter::Legacy(entries) => entries.next().map(JournalCommand::Legacy),
+            JournalCommandIter::Versioned(commands) => commands.next(),
         }
     }
 }
--- a/curvine-master/src/master/journal/entry.rs
+++ b/curvine-master/src/master/journal/entry.rs
@@ -544,6 +544,10 @@ impl DecodedJournalBatch {
         }
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
     pub fn into_commands(self) -> Vec<JournalCommand> {
         match self {
             DecodedJournalBatch::Legacy(batch) => batch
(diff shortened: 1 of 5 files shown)


curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -230,54 +230,81 @@ impl JournalLoader {
...
-        for (seq, op_entry) in batch.batch.into_iter().enumerate() {
-            applied.op_id = op_entry.op_id();
-            applied.rpc_id = op_entry.rpc_id();
+        for (seq, command) in batch.into_commands().into_iter().enumerate() {
+            applied.op_id = command.op_id();
+            applied.rpc_id = command.rpc_id();
 
-            match op_entry {
-                JournalEntry::Snapshot(e) if is_leader && e.node_id == self.node_id => {
+            if matches!(&command, JournalCommand::Metadata(_)) {
```
szbr00: This early return fires after commands earlier in the same batch have already been applied (UFS side effects on the leader, fs_dir updates on the follower) but before `set_applied`, so the applied index is not advanced. On retry the whole batch is re-applied from the first command, re-running UFS work for the already-applied prefix. If a Metadata command can ever share a batch with Legacy commands, consider rejecting the whole batch up front (pre-scan) so the batch stays atomic. If batches are guaranteed homogeneous by construction, a one-line comment stating that assumption would help.

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -230,54 +230,81 @@ impl JournalLoader {
...
+                            );
+                            continue;
+                        }
+
+                        return err_box!("failed to apply journal: {:?}: {}", op_entry, e);
+                    }
                 }
 
-                return err_box!("failed to apply journal: {:?}: {}", op_entry, e);
+                JournalCommand::Metadata(command) => {
```
szbr00, line 304: The `JournalCommand::Metadata(command)` arm in this `match command` block is unreachable: the `matches!(&command, JournalCommand::Metadata(_))` guard above already returns before this point. The redundant arm makes the intent ambiguous (reader has to re-derive that it is dead). Consider dropping this arm and letting the early guard be the single rejection point, or replace it with `unreachable!()` to make the invariant explicit.

curvine-master/src/master/journal/entry.rs
```diff
@@ -28,6 +28,9 @@ pub(crate) struct CvMetadataChange {
     pub(crate) include_subtree: bool,
 }
 
+pub const JOURNAL_ENVELOPE_MAGIC: &[u8; 8] = b"CVJNL001";
```
szbr00: The magic `CVJNL001` already embeds a version-like suffix `001`, and `JOURNAL_ENVELOPE_VERSION` carries the same value as a separate `u16`. There are now two sources of truth for the envelope version that must be kept in sync across bumps. Either drop the version suffix from the magic (e.g. `CVJNLVER`) so the `version` field is the sole source, or remove the explicit `version` field and derive it from the magic. As-is it is fine for v1, but worth a note for whoever bumps v2.

curvine-master/src/master/journal/entry.rs
```diff
@@ -28,6 +28,9 @@ pub(crate) struct CvMetadataChange {
...
+pub const JOURNAL_ENVELOPE_VERSION: u16 = 1;
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in 98bca088. The magic is now the stable CVJNLHDR prefix; JOURNAL_ENVELOPE_VERSION is the sole version source.

curvine-master/src/master/journal/journal_loader.rs
jlon: Thanks. Fixed in 98bca088. Pre-validation is now the only metadata-command rejection point, and the dead match arm was removed.

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -230,54 +230,81 @@ impl JournalLoader {
...
+        let batch = JournalEnvelope::decode(&entry.data)?;
         let batch_len = batch.len();
         let mut snapshot = None;
         let mut applied = Self::build_applied(entry);
         let mut has_ufs_affecting = false;
 
-        for (seq, op_entry) in batch.batch.into_iter().enumerate() {
-            applied.op_id = op_entry.op_id();
-            applied.rpc_id = op_entry.rpc_id();
+        for (seq, command) in batch.into_commands().into_iter().enumerate() {
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in 98bca088. DecodedJournalBatch::into_commands now yields a variant iterator, so legacy batches wrap entries lazily without a second Vec allocation.

curvine-master/src/master/journal/journal_loader.rs
jlon: Thanks. Fixed in 98bca088. The whole batch is validated before command execution, so an unsupported metadata command cannot leave a legacy prefix partially applied.

curvine-master/src/master/journal/entry.rs
jlon: Thanks. Fixed in 98bca088. The magic is now version-agnostic CVJNLHDR; the explicit version field controls compatibility.

szbr00 (approved):

PR #1559, opened 2026-08-12
fix(master): commit namespace creation before metadata apply

jlon:
## Summary
Commit active-Master `mkdir` and new-file creation through Raft before changing local namespace metadata.

## Issue/Design
This is the namespace-creation slice of #1552. Local-first namespace mutations can allocate an inode before the committed log has ordered the mutation across voters. This PR prepares deterministic commands, commits them, then lets ordered journal apply mutate the namespace.

## Changes
| Area | Change |
| --- | --- |
| Active namespace creation | Route `mkdir` and new file creation through committed metadata commands. |
| Command preparation | Allocate inode/op IDs once while resolving the namespace, without locally applying the mutation. |
| Journal execution | Apply committed `Mkdir` and `CreateFile` commands on followers and the active Master. |
| Test | Exercise a two-Master election and verify namespace creation replicates without legacy local-first entries. |

## Test verified
- `cargo fmt --check -- curvine-master/src/master/fs/master_filesystem.rs curvine-master/src/master/journal/journal_loader.rs curvine-master/src/master/journal/journal_writer.rs curvine-master/src/master/meta/fs_dir.rs curvine-server/tests/journal_test.rs`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/namespace-creation cargo test -q -p curvine-server --test journal_test`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/namespace-creation cargo clippy -q -p curvine-server --test journal_test -- -D warnings`
- `git diff --check`

## Dependencies
Stacked on #1558, which introduces the versioned committed-command envelope. Merge after #1558.

### DIFF
--- a/curvine-master/src/master/journal/journal_writer.rs
+++ b/curvine-master/src/master/journal/journal_writer.rs
@@ -121,25 +121,63 @@ impl JournalWriter {
     }
 
     fn record_metadata_delta(&self, entry: &JournalEntry) {
-        let changes = entry.cv_metadata_changes();
+        self.record_metadata_changes(entry.cv_metadata_changes());
+    }
+
+    fn record_metadata_changes(&self, changes: Vec<CvMetadataChange>) {
         if changes.is_empty() {
             return;
         }
         self.metadata_delta_log.lock().unwrap().push(changes);
     }
 
+    pub(crate) fn on_metadata_commands_applied(
+        &self,
+        fs_dir: &FsDir,
+        commands: &[MetadataCommand],
+    ) {
+        if !self.enable || commands.is_empty() {
+            return;
+        }
+
+        for command in commands {
+            self.record_metadata_changes(command.cv_metadata_changes());
+        }
+        if let Err(e) = self.maybe_emit_snapshot_after(fs_dir, commands.len() as u64) {
+            warn!(
+                "failed to emit snapshot after committed metadata batch: {}",
+                e
+            );
+        }
+    }
+
     fn maybe_emit_snapshot(&self, fs_dir: &FsDir) -> FsResult<()> {
+        self.maybe_emit_snapshot_after(fs_dir, 1)
+    }
+
+    fn maybe_emit_snapshot_after(&self, fs_dir: &FsDir, applied_entries: u64) -> FsResult<()> {
         if self.snapshot_entries == 0 {
             return Ok(());
         }
 
-        let entries = self.entries_since_snapshot.add_and_get(1);
+        let mut previous = self.entries_since_snapshot.get();
+        let entries = loop {
+            let current = previous.saturating_add(applied_entries);
+            let next = if current < self.snapshot_entries {
+                current
+            } else {
+                current % self.snapshot_entries
+            };
+            if self.entries_since_snapshot.compare_and_set(previous, next) {
+                break current;
+            }
+            previous = self.entries_since_snapshot.get();
+        };
         if entries < self.snapshot_entries {
             return Ok(());
         }
 
         let now = LocalTime::mills();
-        self.entries_since_snapshot.set(0);
         let dir = match fs_dir.store.create_checkpoint(now) {
             Ok(d) => d,
             Err(e) => {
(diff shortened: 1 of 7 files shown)


curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -200,6 +210,91 @@ impl FsDir {
...
+            while inp.existing_len() + 1 < inp.len() {
+                commands.push(MetadataCommand::Mkdir(
+                    self.prepare_mkdir_command(&mut inp, dir_opts.clone())?,
+                ));
+            }
+        }
+
+        let name = inp.name().to_string();
+        let path = inp.path().to_string();
+        let file = InodeFile::with_opts(self.next_inode_id()?, LocalTime::mills() as i64, opts);
```
szbr00, line 287: `prepare_create_file_commands` builds the `InodeFile` from the caller's `opts` unchanged, but the legacy `create_file` path (fs_dir.rs:568-570) applies `setgid_parent_group` so a new file inherits its parent directory's group when the parent has MODE_SETGID. The committed mkdir path was fixed to preserve setgid inheritance (`apply_setgid_directory_inheritance`), but the create-file path here was not. Files created on a setgid directory via the active master will therefore get the caller's group instead of the parent's group — a POSIX setgid semantic regression vs legacy behavior. Apply the same `apply_setgid_directory_inheritance(&inp, &mut opts)` (files inherit the parent group; no MODE_SETGID bit on files) before constructing the `InodeFile`.

curvine-master/src/master/journal/journal_writer.rs
```diff
@@ -85,6 +87,30 @@ impl JournalWriter {
...
+        }
+
+        let spend = TimeSpent::new();
+        let seq_id = commands[0].op_id();
+        let mut batch = JournalCommandBatch::new(seq_id);
+        for command in commands {
+            batch.push_metadata(command);
+        }
+        let bytes = JournalEnvelope::encode(batch)?;
+        self.client.block_on_send_propose(bytes)?;
```
szbr00, line 105: `commit_metadata_commands` proposes directly via `client.block_on_send_propose` and, unlike the legacy `send()` path, does not call `record_metadata_delta` or `maybe_emit_snapshot`. As a result `cv_metadata_changes_since` consumers no longer see namespace-creation events, and the leader-snapshot cadence (`entries_since_snapshot` -> `JournalEntry::Snapshot`) is no longer advanced by mkdir/create (other ops still advance it via `send`). Since this is a slice of #1552, confirm whether metadata-delta recording and snapshot-cadence accounting are wired up in a later slice, or need to be handled here so delta consumers and standby catch-up stay consistent.

curvine-master/src/master/meta/fs_dir.rs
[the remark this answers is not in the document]
jlon: Thanks. Fixed in aa98cdff. Committed file creation now applies the parent setgid group inheritance before constructing the file inode.

jlon: Thanks. Fixed in aa98cdff. The committed create path now applies the same parent setgid group inheritance as the legacy path.

curvine-master/src/master/journal/journal_writer.rs
```diff
@@ -85,6 +87,30 @@ impl JournalWriter {
...
+
+        self.metrics.journal_flush_count.inc();
+        self.metrics
+            .journal_flush_time
+            .inc_by(spend.used_us() as i64);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
jlon: Thanks. Fixed in aa98cdff. Committed metadata commands now record metadata deltas and advance snapshot accounting, emitting at most one checkpoint after an atomic batch.

curvine-master/src/master/journal/journal_writer.rs
jlon: Thanks. Fixed in aa98cdff. Committed metadata commands now record metadata deltas and run the existing snapshot cadence accounting.

szbr00 (approved):

PR #1561, opened 2026-08-12
fix(master): commit attrs and empty deletes before apply

jlon:
## Summary
Commit active-Master attribute changes and empty-directory deletes before applying local metadata.

## Issue/Design
This is the small metadata-mutation slice of #1552. `set_attr` and an empty directory delete still changed active metadata before their operation had a committed Raft order. Both now use the existing versioned metadata-command path.

## Changes
| Area | Change |
| --- | --- |
| Attribute changes | Normalize and prepare `SetAttr` commands, then commit before apply. |
| Empty directory delete | Commit only the safe empty-directory delete case; recursive/non-empty deletion retains its existing path. |
| Replay | Apply commands on every voter and perform UFS side effects only on the active Master. |
| Test | Extend the two-Master integration test with owner update and empty-directory removal. |

## Test verified
- `cargo fmt --check -- curvine-master/src/master/fs/master_filesystem.rs curvine-master/src/master/journal/entry.rs curvine-master/src/master/journal/journal_loader.rs curvine-master/src/master/meta/fs_dir.rs curvine-server/tests/journal_test.rs`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/attrs-delete cargo test -q -p curvine-server --test journal_test`
- `CARGO_TARGET_DIR=/mnt/data/curvine/.build/pr-1552-split/attrs-delete cargo clippy -q -p curvine-server --test journal_test -- -D warnings`
- `git diff --check`

## Dependencies
Stacked on #1560. Merge after #1560.

### DIFF
--- a/curvine-server/tests/journal_test.rs
+++ b/curvine-server/tests/journal_test.rs
@@ -289,6 +289,7 @@ fn active_namespace_changes_replicate_without_legacy_writer_queue() -> CommonRes
     };
 
     active.mkdir("/committed-dir", false)?;
+    active.mkdir("/deleted-dir", false)?;
     active.create("/committed-file", false)?;
     active.rename("/committed-file", "/renamed-file", RenameFlags::empty())?;
     let parent_opts = MkdirOptsBuilder::new()
@@ -320,6 +321,7 @@ fn active_namespace_changes_replicate_without_legacy_writer_queue() -> CommonRes
         SetAttrOptsBuilder::new().owner("committed-owner").build(),
     )?;
     assert_eq!(status.owner, "committed-owner");
+    active.delete("/deleted-dir", false)?;
     let legacy_entries = active.fs_dir.read().take_entries();
     assert!(
         !legacy_entries.iter().any(|entry| matches!(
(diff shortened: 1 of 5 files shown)


curvine-server/tests/journal_test.rs
```diff
@@ -237,24 +237,38 @@ fn active_namespace_changes_replicate_without_legacy_writer_queue() -> CommonRes
...
+            JournalEntry::Mkdir(_)
+                | JournalEntry::CreateFile(_)
+                | JournalEntry::Rename(_)
+                | JournalEntry::SetAttr(_)
+                | JournalEntry::Delete(_)
         )),
         "active namespace changes must not emit legacy local-first namespace journal entries: {legacy_entries:?}"
     );
 
     let deadline = std::time::Instant::now() + Duration::from_secs(60);
```
szbr00, line 394: The test asserts the standby eventually sees `owner == "committed-owner"` but discards the `FileStatus` returned by `active.set_attr(...)`. Asserting that the returned status already has `owner == "committed-owner"` would verify the active master reflects the committed SetAttr immediately after `commit_metadata_commands` returns (which, with #1557 in the stack, blocks until the local apply completes). It's a small, direct check of the commit-then-apply ordering this PR relies on.

jlon: Thanks. Fixed in ebf80597. The integration test now asserts that active.set_attr returns the committed owner immediately, in addition to verifying follower replay.

PR #1565, opened 2026-08-12
fix(transfer): isolate standalone service configuration

jlon:
## Summary

Complete standalone Transfer service isolation after #1562 separated the default web ports.

## Root Cause

Transfer reused Master logging, loaded unrelated Master/FUSE/Job configuration, printed the full cluster configuration, and applied `net_interface` to the Master address used for metadata refresh. A Transfer-only deployment could therefore fail on unused configuration or connect to its own IP instead of the real Master.

## Design

Transfer now has its own startup boundary:

- `curvine-server --service transfer` uses `ClusterConf::from_transfer()` and validates only `[client]` and `[transfer]`.
- Transfer uses `[transfer].log` with default file name `transfer.log`.
- Startup logs a Transfer-scoped structured configuration line before connecting to Master, rather than dumping the full cluster configuration.
- `CURVINE_TRANSFER_HOSTNAME` controls the Transfer bind hostname.
- With `net_interface`, Transfer changes only its local Transfer/Client hostnames. It never rewrites Master/Journal addresses used to discover the cluster.

Transfer still uses `client.master_addrs` to fetch mounts and live Workers. This is required control-plane communication, not process or configuration reuse. Master-side rejection of legacy Load/Export APIs remains intentional when Transfer is enabled.

## Example

```toml
[client]
master_addrs = [{ hostname = "cv-master", port = 8995 }]

[transfer]
enabled = true
hostname = "cv-transfer"
rpc_port = 9010
web_port = 9011
log = { level = "info", log_dir = "stdout", file_name = "transfer.log" }
```

## Changes

| Module | Change |
| --- | --- |
| `curvine-config` | Dedicated Transfer initialization, hostname override, and regression tests. |
| `curvine-data-transfer` | Dedicated logger and Transfer-scoped startup configuration log. |
| `curvine-server` | Select Transfer-specific configuration initialization. |
| Example configuration | Document explicit Transfer ports and logging. |

## Verification

- `cargo fmt --check`
- `cargo test -p curvine-config cluster_conf --lib` (7 passed)
- `cargo test -p curvine-config transfer_conf --lib` (5 passed)
- `cargo check -p curvine-data-transfer`
- `cargo check -p curvine-server`
- `git diff --check`

## Dependencies

- Related PR: #1562


crates/server/curvine-data-transfer/src/transfer/transfer_server.rs
```diff
@@ -94,11 +94,10 @@ impl TransferServer {
         if !conf.transfer.enabled {
             return Err(FsError::common("curvine-transfer requires transfer.enabled=true").into());
         }
-        Logger::init(conf.master.log.clone());
+        Logger::init(conf.transfer.log.clone());
         let _ = TransferMetrics::get()?;
         info!("allocator: {}", curvine_alloc::allocator_type_name());
         info!("git version: {}", curvine_sys::version::GIT_VERSION);
```
szbr00, line 100: Removing `conf.print()` here means the standalone Transfer service no longer logs its loaded configuration at startup, while Master (`master_server.rs:177`) and Worker (`worker_server.rs:202`) still do. That hurts post-mortem debugging for Transfer operators.

I understand the motivation — `print()` dumps the whole `ClusterConf`, and with `from_transfer` skipping `master/fuse/job.init()` those sections would be misleading. But please add a Transfer-scoped startup log (e.g. `info!("transfer conf: ...", transfer.hostname, rpc_port, web_port, store_backend, max_running, ...)` — several of those fields are already logged a few lines below at line 148, so consolidating into a single structured line would be enough).

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -503,4 +529,26 @@ mod tests {
...
+                [master]
+                conn_limit = 0
+
+                [transfer]
+                enabled = true
+            "#,
+        )
+        .unwrap();
+
+        let conf = ClusterConf::from_transfer(path.to_str().unwrap()).unwrap();
```
szbr00, line 578: Using `.unwrap()` on the cleanup `remove_file` turns a benign cleanup failure (e.g. file already gone, transient permission issue) into a test failure that hides the actual behavior under test. Prefer `let _ = std::fs::remove_file(&path);` so cleanup errors are ignored. The temp-file name also only uses the PID; if parallel test processes share a temp dir this can collide — a random suffix (e.g. `Utils::rand_str`) would make it robust.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -88,18 +88,43 @@ impl ClusterConf {
...
+        conf.job.init()?;
+        conf.transfer.init()?;
+        conf.resolve_master_addrs();
+        Ok(conf)
+    }
+
+    /// Load only the configuration needed by the standalone Transfer service.
+    pub fn from_transfer<T: AsRef<str>>(path: T) -> CommonResult<Self> {
+        let mut conf = Self::read(path)?;
+        conf.apply_hostname_overrides()?;
```
szbr00: `from_transfer` reuses `apply_hostname_overrides`, so when `net_interface` is set it writes the local interface IP into `master.hostname` (and `journal.hostname`). `resolve_master_addrs` then derives `client.master_addrs` from `master.hostname` / `journal.journal_addrs`, so a standalone Transfer process that relies on the default master-address resolution ends up trying to reach the master at the Transfer node's own IP:8995 instead of the real master host. The service then fails cluster metadata refresh on startup.

For the standalone Transfer path, `net_interface` should only influence `transfer.hostname` (what Transfer binds to), not the master/journal addresses it connects to. Consider either:
- having `from_transfer` skip the master/journal/worker hostname writes (only apply net_interface / `ENV_TRANSFER_HOSTNAME` to `transfer.hostname` and `client.hostname`), or
- documenting that standalone Transfer with `net_interface` requires an explicit `client.master_addrs` and enforcing it (e.g. error when both are set and `client.master_addrs` is empty).

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -503,4 +529,26 @@ mod tests {
         assert_eq!(conf.master.rpc_port, ClusterConf::DEFAULT_MASTER_PORT);
         assert!(!conf.client.master_addrs.is_empty());
     }
+
+    #[test]
+    fn transfer_init_skips_unused_master_validation() {
+        let path =
+            std::env::temp_dir().join(format!("curvine-transfer-conf-{}.toml", std::process::id()));
```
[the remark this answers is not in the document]
jlon: Fixed in 20d70a85. The test path now includes `Utils::rand_str(6)` in addition to the PID.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -503,4 +529,26 @@ mod tests {
...
+
+                [transfer]
+                enabled = true
+            "#,
+        )
+        .unwrap();
+
+        let conf = ClusterConf::from_transfer(path.to_str().unwrap()).unwrap();
+        std::fs::remove_file(&path).unwrap();
+
```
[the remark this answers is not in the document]
jlon: Fixed in 20d70a85. Cleanup is now best-effort and cannot mask the assertion result.

crates/server/curvine-data-transfer/src/transfer/transfer_server.rs
jlon: Fixed in 20d70a85. Transfer keeps its structured startup configuration log, moved before the Master client is created; it no longer dumps the full ClusterConf.

crates/common/curvine-config/src/cluster_conf.rs
jlon: Fixed in 20d70a85. The temporary name has a random suffix and cleanup is best-effort.

crates/common/curvine-config/src/cluster_conf.rs
jlon: Fixed in 20d70a85. `from_transfer` now uses a dedicated hostname override path: `net_interface` affects only the Transfer and client local hostnames, never Master/Journal addresses. A regression test verifies default Master address resolution remains `cv-master`.

szbr00 (approved):

PR #1566, opened 2026-08-12
fix(fuse): synchronize concurrent reader refresh

### LINKED ISSUE #1498: [BUG]: Concurrent reader refresh can panic Curvine FUSE during direct writes
**Describe the bug**
Curvine FUSE can panic and return a short buffered read when multiple reads on the same file handle race with direct writes that trigger a dirty-reader refresh. The read path stores `FuseReader` behind an unsynchronized `RawPtr`, then replaces that reader in place after publishing a writer snapshot. Concurrent FUSE read requests can use or replace the same reader at the same time, corrupting the reader channel lifecycle and triggering a Tokio mpsc internal assertion.

This is reproducible through xfstests `generic/451`, which mixes frequent buffered reads with asynchronous direct writes. The failure is intermittent: a repeated run against the same Curvine build produced one pass followed by failures after 12 seconds and 6 seconds.

**To Reproduce**

1. Start Curvine and mount Curvine FUSE at `/mnt/curvine` with read-write access.
2. Save the following code as `curvine_fuse_concurrent_read_refresh_repro.c`:

```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <linux/aio_abi.h>
#include <pthread.h>
#include <stdatomic.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <time.h>
#include <unistd.h>

#define FILE_SIZE (640 * 1024)
#define ALIGNMENT 4096
#define DEFAULT_READERS 20
#define DEFAULT_SECONDS 60

struct shared_state {
    int read_fd;
    atomic_int stop;
    atomic_int failure;
    atomic_int first_errno;
    atomic_ulong reads;
    atomic_ulong writes;
};

static int io_setup(unsigned int nr_events, aio_context_t *context)
{
    return syscall(SYS_io_setup, nr_events, context);
}

static int io_destroy(aio_context_t context)
{
    return syscall(SYS_io_destroy, context);
}

static int io_submit(aio_context_t context, long nr, struct iocb **iocbs)
{
    return syscall(SYS_io_submit, context, nr, iocbs);
}

static int io_getevents(aio_context_t context, long min_nr, long nr,
                        struct io_event *events)
{
    return syscall(SYS_io_getevents, context, min_nr, nr, events, NULL);
}

static void record_failure(struct shared_state *state, int error,
                           const char *operation)
{
    int expected = 0;

    if (atomic_compare_exchange_strong(&state->failure, &expected, 1)) {
        atomic_store(&state->first_errno, error);
        fprintf(stderr, "%s failed: %s\n", operation, strerror(error));
    }
    atomic_store(&state->stop, 1);
}

static void *reader_thread(void *argument)
{
    struct shared_state *state = argument;
    char *buffer;

    if (posix_memalign((void **)&buffer, ALIGNMENT, FILE_SIZE) != 0) {
        record_failure(state, ENOMEM, "allocate read buffer");
        return NULL;
    }

    while (!atomic_load(&state->stop)) {
        ssize_t result = pread(state->read_fd, buffer, FILE_SIZE, 0);

        if (result == -1) {
            record_failure(state, errno, "buffered pread");
            break;
        }
        if (result != FILE_SIZE) {
            fprintf(stderr, "buffered pread returned %zd, expected %d\n",
                    result, FILE_SIZE);
            record_failure(state, EIO, "short buffered pread");
            break;
        }
        atomic_fetch_add(&state->reads, 1);
    }

    free(buffer);
    return NULL;
}

static int run_aio_writer(struct shared_state *state, int write_fd,
                          char *buffer)
{
    aio_context_t context = 0;
    unsigned int pattern = 0;

    if (io_setup(1, &context) == -1) {
        perror("io_setup");
        return -1;
    }

    while (!atomic_load(&state->stop)) {
        struct iocb request = {0};
        struct iocb *requests[] = {&request};
        struct io_event event;
        int result;

        memset(buffer, pattern++ & 1 ? 0x55 : 0xaa, FILE_SIZE);
        request.aio_fildes = write_fd;
        request.aio_lio_opcode = IOCB_CMD_PWRITE;
        request.aio_buf = (uint64_t)buffer;
        request.aio_nbytes = FILE_SIZE;
        request.aio_offset = 0;

        result = io_submit(context, 1, requests);
        if (result != 1) {
            record_failure(state, result == -1 ? errno : EIO,
                           "io_submit direct write");
            break;
        }

        result = io_getevents(context, 1, 1, &event);
        if (result != 1) {
            record_failure(state, result == -1 ? errno : EIO,
                           "io_getevents direct write");
            break;
        }
        if ((int64_t)event.res < 0) {
            record_failure(state, -(int)event.res, "AIO direct write");
            break;
        }
        if (event.res != FILE_SIZE) {
            fprintf(stderr, "AIO direct write returned %lld, expected %d\n",
                    (long long)event.res, FILE_SIZE);
            record_failure(state, EIO, "short AIO direct write");
            break;
        }
        atomic_fetch_add(&state->writes, 1);
    }

    if (io_destroy(context) == -1) {
        perror("io_destroy");
        return -1;
    }
    return atomic_load(&state->failure) ? -1 : 0;
}

static long parse_positive(const char *value, const char *name)
{
    char *end;
    long result;

    errno = 0;
    result = strtol(value, &end, 10);
    if (errno != 0 || *value == '\0' || *end != '\0' || result <= 0) {
        fprintf(stderr, "invalid %s: %s\n", name, value);
        exit(2);
    }
    return result;
}

{
    struct shared_state state = {0};
    int reader_count = DEFAULT_READERS;
    int seconds = DEFAULT_SECONDS;
    pthread_t *threads;
    char *write_buffer;
    struct timespec deadline;
    int write_fd;
    int created = 0;
    int result = 0;

    if (argc < 2 || argc > 4) {
        fprintf(stderr, "Usage: %s <file-on-curvine-fuse> [seconds] [readers]\n",
                argv[0]);
        return 2;
    }
    if (argc >= 3)
        seconds = (int)parse_positive(argv[2], "duration");
    if (argc == 4)
        reader_count = (int)parse_positive(argv[3], "reader count");

    write_fd = open(argv[1], O_WRONLY | O_CREAT | O_TRUNC | O_DIRECT, 0666);
    if (write_fd == -1) {
        perror("open direct writer");
        return 1;
    }
    if (posix_memalign((void **)&write_buffer, ALIGNMENT, FILE_SIZE) != 0) {
        fprintf(stderr, "cannot allocate aligned write buffer\n");
        close(write_fd);
        return 1;
    }
    memset(write_buffer, 0xaa, FILE_SIZE);
    if (pwrite(write_fd, write_buffer, FILE_SIZE, 0) != FILE_SIZE) {
        perror("initialize test file");
        free(write_buffer);
        close(write_fd);
        return 1;
    }

    state.read_fd = open(argv[1], O_RDONLY);
    if (state.read_fd == -1) {
        perror("open buffered reader");
        free(write_buffer);
        close(write_fd);
        return 1;
    }

    threads = calloc(reader_count, sizeof(*threads));
    if (threads == NULL) {
        perror("calloc reader threads");
        result = 1;
        goto out;
    }

    for (int i = 0; i < reader_count; i++) {
        int error = pthread_create(&threads[i], NULL, reader_thread, &state);

        if (error != 0) {
            record_failure(&state, error, "pthread_create");
            break;
        }
        created++;
    }

    if (clock_gettime(CLOCK_MONOTONIC, &deadline) == -1) {
        perror("clock_gettime");
        atomic_store(&state.stop, 1);
    } else {
        deadline.tv_sec += seconds;
    }

    printf("Running %d concurrent buffered readers and one AIO direct writer "
           "for %d seconds\n", created, seconds);
    fflush(stdout);

    while (!atomic_load(&state.stop)) {
        struct timespec now;

        if (clock_gettime(CLOCK_MONOTONIC, &now) == -1 ||
            now.tv_sec > deadline.tv_sec ||
            (now.tv_sec == deadline.tv_sec && now.tv_nsec >= deadline.tv_nsec)) {
            atomic_store(&state.stop, 1);
            break;
        }
        if (run_aio_writer(&state, write_fd, write_buffer) == -1)
            break;
    }

    atomic_store(&state.stop, 1);
    for (int i = 0; i < created; i++)
        pthread_join(threads[i], NULL);

    if (atomic_load(&state.failure)) {
        fprintf(stderr, "FAIL: filesystem I/O failed after %lu reads and %lu writes "
                        "(errno=%d)\n",
                atomic_load(&state.reads), atomic_load(&state.writes),
                atomic_load(&state.first_errno));
        result = 1;
    } else {
        printf("PASS: completed %lu reads and %lu writes without disconnect\n",
               atomic_load(&state.reads), atomic_load(&state.writes));
    }

    free(threads);
out:
    close(state.read_fd);
    close(write_fd);
    free(write_buffer);
    if (unlink(argv[1]) == -1 && errno != ENOENT)
        perror("unlink test file");
    return result;
}
```

3. Compile the reproducer with Linux AIO and pthread support:

```bash
cc -std=c11 -O2 -Wall -Wextra -Werror -pthread \
  curvine_fuse_concurrent_read_refresh_repro.c \
  -o /tmp/curvine_fuse_concurrent_read_refresh_repro
```

4. Run it against a file on the Curvine FUSE mount. Increasing the duration or reader count makes the race easier to trigger:

```bash
/tmp/curvine_fuse_concurrent_read_refresh_repro \
  /mnt/curvine/concurrent-read-refresh-repro 60 20
```

5. Alternatively, run xfstests `generic/451` repeatedly. The observed sequence with Curvine commit `d134f5bf` was:

```text
generic/451      PASS     rc=0   duration=31s
generic/451      FAIL     rc=1   duration=12s
generic/451      FAIL     rc=1   duration=6s
```

6. Observe a short buffered read. The xfstests helper prints `pread: Success` because `pread()` returned a length different from the requested 655360 bytes while `errno` remained zero:

```text
QA output created by 451
pread: Success
Silence is golden
```

7. Check the Curvine FUSE log and observe the reader task panic:

```text
thread 'curvine-fuse' panicked at tokio-1.48.0/src/sync/mpsc/chan.rs:324:13:
assertion failed: self.inner.semaphore.is_idle()

stack backtrace:
   3: tokio::sync::mpsc::chan::Rx<T,S>::recv
   4: curvine_fuse::fs::fuse_reader::FuseReader::new::{{closure}}
```

**Expected behavior**
Concurrent buffered reads and direct writes must not panic the FUSE daemon, close or corrupt the per-handle reader channel, or return an unexpected short read. Every completed `pread()` in this test should return the full 655360-byte file, containing one complete writer pattern. The standalone reproducer should run for the requested duration and finish with output similar to:

```text
Running 20 concurrent buffered readers and one AIO direct writer for 60 seconds
PASS: completed <reads> reads and <writes> writes without disconnect
```

Repeated xfstests `generic/451` runs should consistently produce only:

```text
QA output created by 451
Silence is golden
```

 - OS: Ubuntu Linux
 - Kernel: 5.15.0-72-generic
 - Architecture: x86_64
 - Curvine version: `0.2.0`, commit `d134f5bf`, branch `main`
 - Filesystem client: Curvine FUSE
 - Tokio version: 1.48.0

**Additional context**
`BackendHandle.reader` is currently an `Option<RawPtr<FuseReader>>`. When a read detects that the inode's shared writer has advanced, `BackendHandle::read()` publishes a dirty-read snapshot, opens a new reader, and calls `reader.replace(new_reader)` before sending the read request. `RawPtr::replace()` uses `std::ptr::replace` without synchronization, while `RawPtr<T>` is unconditionally marked `Send` and `Sync`.

Multiple FUSE read requests for the same file handle can therefore race through the refresh path. One request can replace and drop the old `FuseReader` while another request still accesses it or sends a `ReadTask` through its Tokio mpsc sender. More than one request can also reopen and replace the reader for different sampled writer versions. This is a data race and undefined behavior; the observed Tokio semaphore assertion is a consequence of the corrupted sender/receiver lifecycle.

The standalone reproducer deliberately shares one buffered-read file descriptor across multiple threads to concentrate requests on one FUSE file handle. xfstests `generic/451` creates many buffered reader processes, and each 640 KiB read can also generate overlapping FUSE read requests while the asynchronous direct writer continuously advances the writer version.

The writer-side `publish_dirty_read_snapshot()` serialization introduced by issue/PR `#1434` prevents a Flush from overtaking an in-progress write enqueue, but it does not serialize the per-handle reader reopen, replacement, and `read_ver` update. A complete fix should replace the unsynchronized `RawPtr<FuseReader>` ownership with a safe lifetime-managed reader reference, serialize refresh publication per handle, and allow in-flight reads to retain the previous reader until their task has been safely queued or completed.

Suggested regression coverage:

- Run the standalone reproducer repeatedly with 20 or more reader threads.
- Run xfstests `generic/451` for at least 20 consecutive iterations.
- Add a Rust concurrency test in which multiple reads on one `BackendHandle` race with repeated writer-version changes and reader refreshes.
- Verify that no short reads, reader-channel errors, Tokio panics, or FUSE daemon exits occur.

ljy1-lixing:
**Describe the bug**
Curvine FUSE can panic or return a short buffered read when concurrent reads on the same file handle race with direct writes that trigger a dirty-reader refresh. `BackendHandle` stores its `FuseReader` behind an unsynchronized `RawPtr`, then replaces the pointed-to reader after publishing a writer snapshot. Multiple read requests can access, replace, or drop the same reader concurrently, corrupting the Tokio mpsc channel lifecycle.

The observed FUSE process failure is:

```text
thread 'curvine-fuse' panicked at tokio-1.48.0/src/sync/mpsc/chan.rs:324:13:
assertion failed: self.inner.semaphore.is_idle()
```

Once the daemon exits, outstanding or subsequent reads may return a short result, `EIO`, or `ENOTCONN`. This PR fixes the unsafe reader ownership and refresh race reported in #1498.

Closes #1498.

**To Reproduce**
Use the standalone reproducer and instructions attached to #1498. It shares one buffered-read file descriptor across multiple threads while an AIO direct writer repeatedly updates the same file, concentrating concurrent requests on one FUSE handle.

On an affected build, the test may report a short read or a disconnected mount:

```text
buffered pread returned <short-length>, expected 655360
```

or:

```text
buffered pread failed: Transport endpoint is not connected
```

The Curvine FUSE log then contains the Tokio mpsc assertion shown above. The reproducer source is intentionally not duplicated here because #1498 already contains the complete program, compilation command, and execution instructions.

**Expected behavior**
Concurrent buffered reads and direct writes must not race the lifetime of the per-handle reader. Every read request must retain a valid `FuseReader` until its task has entered that reader's channel, even when another request refreshes the handle to a newer writer snapshot.

The standalone reproducer should run for the requested duration without short reads, channel errors, FUSE panics, or mount disconnection:

```text
Running 20 concurrent buffered readers and one AIO direct writer for 60 seconds
PASS: completed <reads> reads and <writes> writes without disconnect
```

**Root cause**
The previous `BackendHandle` representation kept the reader and its snapshot version as independent, unsynchronized fields:

```rust
pub reader: Option<RawPtr<FuseReader>>,
read_ver: AtomicCounter,
```

When a read observed a newer shared writer version, it performed this sequence:

```text
publish dirty writer snapshot
open a new FuseReader
replace the value behind RawPtr
update read_ver
send the read through the selected reader
```

`RawPtr::replace()` uses `std::ptr::replace` without synchronization, while `RawPtr<T>` is unconditionally marked `Send` and `Sync`. Concurrent reads on one handle can therefore interleave as follows:

```text
Read A obtains the old raw reader pointer
Read B replaces the pointed-to FuseReader
Read B drops the old reader and its channel receiver
Read A accesses the replaced value or sends through a corrupted channel
```

Two refresh requests can also publish and install different snapshots while independently updating `read_ver`. The pointer mutation is a Rust data race and undefined behavior; the Tokio semaphore assertion is one observed consequence rather than the underlying defect.

Serializing only the writer-side snapshot publication is insufficient. The per-handle reader selection, reopen, installation, and snapshot-version update must form one synchronized state transition, and an in-flight request must retain ownership of the selected reader after leaving that critical section.

**Changes**

- Replace `RawPtr<FuseReader>` with reference-counted `Arc<FuseReader>` ownership.
- Store the current reader and `read_ver` together in a per-handle `ReaderState`.
- Guard `ReaderState` with an asynchronous mutex so snapshot publication, reader reopen, reader installation, and version advancement are serialized for one file handle.
- Clone the selected reader's `Arc` before releasing the refresh lock, allowing the request to safely enqueue work after a later refresh replaces the handle's current reader.
- Release the refresh lock before calling `FuseReader::read()`, preserving concurrent read submission instead of serializing backend reads.
- Use the same safe reader representation when opening and restoring file handles.
- Add a reader-lifetime regression test proving that replacing the handle's current reader does not drop the old reader while an in-flight request still owns an `Arc`.

The resulting read path is:

```text
lock per-handle ReaderState
check and publish the writer snapshot
reopen and install a reader when the version changed
clone Arc<FuseReader> for this request
unlock ReaderState
enqueue the read through the retained reader
```

**Concurrency and performance impact**
The new mutex protects only refresh bookkeeping and reader selection. It is not held while the backend read executes or while the queued request waits for its FUSE reply.

- Reads using the same handle may briefly serialize while checking or refreshing the writer snapshot.
- Normal read tasks remain concurrently enqueueable after cloning the selected reader.
- Different file handles use independent locks.
- Replaced readers stay alive only while existing requests retain them and are released automatically afterward.
- No global filesystem or path lock is introduced.

 - OS: Ubuntu Linux
 - Kernel: 5.15.0-72-generic
 - Architecture: x86_64
 - Filesystem client: Curvine FUSE
 - Tokio version: 1.48.0

**Verification**

- Run the standalone concurrent-read reproducer from #1498 and verify that it completes without a short read, channel error, or FUSE disconnect.
- Run the affected concurrent buffered-read/direct-write scenario repeatedly. An isolated Curvine cluster produced 10 consecutive successful runs, each completing in approximately 31-32 seconds:

```text
TOTAL:    10
PASS:     10
FAIL:     0
TIMEOUT:  0
```

- Run the reader replacement lifetime unit test.
- Run `cargo test -p curvine-fuse --lib -- --test-threads=1`.
- Run `cargo check -p curvine-fuse`.
- Run `cargo clippy -p curvine-fuse --all-targets -- -D warnings`.
- Run `cargo fmt --all -- --check`.
- Run `git diff --check`.
- Run `make all`.

**Additional context**
The earlier writer-side `publish_dirty_read_snapshot()` serialization prevents a flush from overtaking an in-progress write enqueue, but it does not protect mutable reader ownership inside `BackendHandle`. This PR keeps that writer ordering and adds the missing per-handle reader-state synchronization.

The fix deliberately avoids modifying the generic `RawPtr` implementation. Other uses may have different ownership constraints, while `BackendHandle` specifically needs shared lifetime management across asynchronous read requests. Using `Arc<FuseReader>` at this boundary makes that lifetime explicit and removes the unsafe pointer operation from the affected path.

### DIFF
--- a/curvine-fuse/src/fs/state/backend_handle.rs
+++ b/curvine-fuse/src/fs/state/backend_handle.rs
@@ -22,8 +22,7 @@ use curvine_core_error::err_box;
 use curvine_fs_api::Path;
 use curvine_fs_api::{StateReader, StateWriter};
 use curvine_model::{CreateFileOptsBuilder, FileStatus, LockFlags, OpenFlags};
-use curvine_runtime::sync::AtomicCounter;
-use curvine_sys::RawPtr;
+use curvine_runtime::sync::AsyncMutex;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 use std::sync::Arc;
@@ -56,33 +55,44 @@ pub struct BackendHandle {
     pub ino: u64,
     pub fh: u64,
 
-    pub reader: Option<RawPtr<FuseReader>>,
+    reader: Option<AsyncMutex<ReaderState>>,
     pub writer: Option<Arc<FuseWriter>>, // Writer uses Arc for global sharing
     /// Open-time file status snapshot. Guarded by a lock because the read path
     /// refreshes it in place after a dirty-read reopen (`read()` takes `&self`).
     status: std::sync::RwLock<FileStatus>,
 
     fh_locks: std::sync::Mutex<HandleLock>,
+}
 
-    read_ver: AtomicCounter,
+struct ReaderState {
+    reader: Arc<FuseReader>,
+    read_ver: u64,
+}
+
+impl ReaderState {
+    fn new(reader: Arc<FuseReader>) -> Self {
+        Self {
+            reader,
+            read_ver: 0,
+        }
+    }
 }
 
 impl BackendHandle {
     pub fn new(
         ino: u64,
         fh: u64,
-        reader: Option<RawPtr<FuseReader>>,
+        reader: Option<Arc<FuseReader>>,
         writer: Option<Arc<FuseWriter>>,
         status: FileStatus,
     ) -> Self {
         Self {
             ino,
             fh,
-            reader,
+            reader: reader.map(|reader| AsyncMutex::new(ReaderState::new(reader))),
             writer,
             status: std::sync::RwLock::new(status),
             fh_locks: std::sync::Mutex::new(HandleLock::default()),
-            read_ver: AtomicCounter::new(0),
         }
     }
 
@@ -125,10 +135,15 @@ impl BackendHandle {
             );
         }
 
-        let reader = match &self.reader {
-            Some(v) => v,
+        let reader_state = match &self.reader {
+            Some(reader) => reader,
             None => return err_fuse!(libc::EIO),
         };
+
+        // Serialize refresh state per file handle. The Arc cloned below keeps the
+        // selected reader alive until this request has entered its channel, even
+        // if a later request replaces the handle's current reader.
+        let mut reader_state = reader_state.lock().await;
 
         if let Some(writer) = state.find_writer(self.ino).await {
             // `write_ver` advances only after a write/resize is successfully queued.
@@ -143,18 +158,20 @@ impl BackendHandle {
             // Pin read_ver to the flushed snapshot; later writes leave write_ver
             // ahead so the next read publishes again.
             if let Some(target_ver) = writer
-                .publish_dirty_read_snapshot(self.read_ver.get())
+                .publish_dirty_read_snapshot(reader_state.read_ver)
                 .await?
             {
-                let path = reader.path().clone();
+                let path = reader_state.reader.path().clone();
                 let new_reader = state.new_reader(&path).await?;
                 // Refresh status from the reopened reader before installing it.
                 self.refresh_status(new_reader.status().clone());
-                reader.replace(new_reader);
-                self.read_ver.set(target_ver);
+                reader_state.reader = Arc::new(new_reader);
+                reader_state.read_ver = target_ver;
             }
         }
 
+        let reader = reader_state.reader.clone();
+        drop(reader_state);
         reader.read(op, reply).await?;
         Ok(())
     }
@@ -312,7 +329,7 @@ impl BackendHandle {
 
         let reader = if has_reader {
             let reader = state.new_reader(&path).await?;
-            Some(RawPtr::from_owned(reader))
+            Some(Arc::new(reader))
         } else {
             None
         };
@@ -320,12 +337,11 @@ impl BackendHandle {
         let handle = Self {
             ino,
             fh,
-            reader,
+            reader: reader.map(|reader| AsyncMutex::new(ReaderState::new(reader))),
             writer,
             status: std::sync::RwLock::new(status),
 
             fh_locks: std::sync::Mutex::new(locks),
-            read_ver: AtomicCounter::new(0),
         };
         Ok(handle)
     }
@@ -334,6 +350,71 @@ impl BackendHandle {
 #[cfg(test)]
 mod tests {
     use super::BackendHandle;
+
+    #[test]
+    fn replaced_reader_stays_alive_while_request_holds_arc() {
+        use crate::fs::FuseReader;
+        use curvine_client::unified::UnifiedReader;
+        use curvine_config::FuseConf;
+        use curvine_fs_api::local::LocalReader;
+        use curvine_fs_api::Path;
+        use curvine_model::FileStatus;
+        use curvine_runtime::runtime::{AsyncRuntime, RpcRuntime};
+        use std::io::Write as _;
+        use std::sync::Arc;
+
+        let path_buf = std::env::temp_dir().join(format!(
+            "backend_handle_reader_arc_{}_{}",
+            std::process::id(),
+            std::thread::current().name().unwrap_or("test")
+        ));
+        {
+            let mut file = std::fs::File::create(&path_buf).unwrap();
+            file.write_all(b"reader lifetime").unwrap();
+        }
+        let path = Path::from_str(path_buf.to_str().unwrap()).unwrap();
+        let runtime = Arc::new(AsyncRuntime::single());
+        let make_reader = || {
+            Arc::new(FuseReader::new(
+                &FuseConf::default(),
+                runtime.clone(),
+                UnifiedReader::Local(LocalReader::new(&path, 4096).unwrap()),
+            ))
+        };
+
+        let original = make_reader();
+        let original_weak = Arc::downgrade(&original);
+        let handle = BackendHandle::new(
+            1,
+            10,
+            Some(original.clone()),
+            None,
+            FileStatus::with_name(1, "f".into(), false),
+        );
+        drop(original);
+
+        runtime.block_on(async {
+            let mut state = handle.reader.as_ref().unwrap().lock().await;
+            let in_flight = state.reader.clone();
+            state.reader = make_reader();
+            drop(state);
+
+            assert!(
+                original_weak.upgrade().is_some(),
+                "an in-flight request must retain the reader replaced on the handle"
+            );
+            drop(in_flight);
+            assert!(
+                original_weak.upgrade().is_none(),
+                "the replaced reader is released after the in-flight request drops its Arc"
+            );
+        });
+
+        // FuseReader workers retain the runtime handle; avoid dropping a Tokio
+        // runtime from a test async context while those tasks wind down.
+        std::mem::forget(runtime);
+        let _ = std::fs::remove_file(path_buf);
+    }
 
     // Reject offsets that would wrap negative when cast to i64.
     #[test]
--- a/curvine-fuse/src/fs/state/backend_handle.rs
+++ b/curvine-fuse/src/fs/state/backend_handle.rs
@@ -143,6 +143,9 @@ impl BackendHandle {
         // Serialize refresh state per file handle. The Arc cloned below keeps the
         // selected reader alive until this request has entered its channel, even
         // if a later request replaces the handle's current reader.
+        // Keep this lock across snapshot publication and reader reopen/install;
+        // releasing it between those awaits lets concurrent refreshes install
+        // readers for different snapshots and reintroduces the lifetime race.
         let mut reader_state = reader_state.lock().await;
 
         if let Some(writer) = state.find_writer(self.ino).await {
@@ -364,9 +367,9 @@ mod tests {
         use std::sync::Arc;
 
         let path_buf = std::env::temp_dir().join(format!(
-            "backend_handle_reader_arc_{}_{}",
+            "backend_handle_reader_arc_{}_{:?}",
             std::process::id(),
-            std::thread::current().name().unwrap_or("test")
+            std::thread::current().id()
         ));
         {
             let mut file = std::fs::File::create(&path_buf).unwrap();
--- a/curvine-fuse/src/fs/state/file_handle.rs
+++ b/curvine-fuse/src/fs/state/file_handle.rs
@@ -21,7 +21,6 @@ use crate::{err_fuse, FuseResult};
 use curvine_core_error::err_box;
 use curvine_fs_api::{StateReader, StateWriter};
 use curvine_model::{FileAllocOpts, FileStatus, LockFlags};
-use curvine_sys::RawPtr;
 use std::sync::Arc;
 
 pub enum FileHandle {
@@ -34,7 +33,7 @@ impl FileHandle {
     pub fn new_backend(
         ino: u64,
         fh: u64,
-        reader: Option<RawPtr<FuseReader>>,
+        reader: Option<Arc<FuseReader>>,
         writer: Option<Arc<FuseWriter>>,
         status: FileStatus,
     ) -> Self {
--- a/curvine-fuse/src/fs/state/node_state.rs
+++ b/curvine-fuse/src/fs/state/node_state.rs
@@ -39,7 +39,6 @@ use curvine_model::{
 };
 use curvine_runtime::common::FastHashMap;
 use curvine_runtime::sync::{AsyncMutex, AsyncSharedMap, AtomicCounter, RwLockHashMap};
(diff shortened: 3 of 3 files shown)


curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -125,11 +135,16 @@ impl BackendHandle {
...
-            Some(v) => v,
+        let reader_state = match &self.reader {
+            Some(reader) => reader,
             None => return err_fuse!(libc::EIO),
         };
 
+        // Serialize refresh state per file handle. The Arc cloned below keeps the
+        // selected reader alive until this request has entered its channel, even
+        // if a later request replaces the handle's current reader.
+        let mut reader_state = reader_state.lock().await;
```
szbr00, line 149: The per-handle mutex is held across `publish_dirty_read_snapshot()` (which awaits a flush) and `state.new_reader()` (filesystem open). This is intentional: releasing the lock between the version check and the reader install would let two refreshes reopen and install different readers and re-introduce the race this PR fixes. Consider adding a one-line note here so a future maintainer does not move the blocking flush/reopen outside the lock to reduce lock hold time and silently restore the UB.

curvine-fuse/src/fs/state/backend_handle.rs
```diff
@@ -335,6 +351,71 @@ impl BackendHandle {
...
+        use curvine_model::FileStatus;
+        use curvine_runtime::runtime::{AsyncRuntime, RpcRuntime};
+        use std::io::Write as _;
+        use std::sync::Arc;
+
+        let path_buf = std::env::temp_dir().join(format!(
+            "backend_handle_reader_arc_{}_{}",
+            std::process::id(),
+            std::thread::current().name().unwrap_or("test")
+        ));
```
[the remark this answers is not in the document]
ljy1-lixing: Fixed in 3f06e9f4. The temp-file suffix now uses `std::thread::current().id()` together with the process ID, avoiding collisions between parallel test threads.

curvine-fuse/src/fs/state/backend_handle.rs
ljy1-lixing: Added the requested invariant comment in 3f06e9f4. It now explicitly states that the lock must remain held across snapshot publication and reader reopen/install to prevent concurrent refreshes from installing different snapshots and restoring the lifetime race.

szbr00 (approved):

PR #1570, opened 2026-08-13
fix(fuse): order OFD lock cleanup with queries

### LINKED ISSUE #1502: [BUG]: Curvine FUSE intermittently exposes stale OFD locks after the final close
**Describe the bug**
Curvine FUSE intermittently reports a stale open-file-description (OFD) lock after the last reference to the locked file description has been closed. A concurrent `F_OFD_GETLK` can still observe the old lock even though Linux OFD lock semantics require it to be released when the final reference to that open file description is closed.

This causes `xfstests generic/478` to fail intermittently. On Curvine commit `d134f5bf`, 10 repeated runs produced 6 passes and 4 failures. A minimized reproducer triggered the stale lock twice in 500 runs, while the same reproducer completed 500 runs on local ext4 without a failure.

**To Reproduce**

1. Start Curvine and mount Curvine FUSE at `/mnt/curvine` with read-write access.
2. Save the following code as `curvine_fuse_ofd_release_getlk_repro.c`:

```c
#define _GNU_SOURCE

#include <errno.h>
#include <fcntl.h>
#include <sched.h>
#include <signal.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/wait.h>
#include <unistd.h>

#ifndef F_OFD_GETLK
#define F_OFD_GETLK 36
#define F_OFD_SETLK 37
#define F_OFD_SETLKW 38
#endif

static char stack[1024 * 1024] __attribute__((aligned(16)));

struct close_args {
    int fd;
    int notify;
};

static void die(const char *what)
{
    fprintf(stderr, "%s: %s\n", what, strerror(errno));
    exit(2);
}

static int close_shared_fd(void *opaque)
{
    struct close_args *args = opaque;

    if (close(args->fd) || write(args->notify, "x", 1) != 1)
        _exit(2);
    _exit(0);
}

{
    struct flock lock = {
        .l_type = F_WRLCK,
        .l_whence = SEEK_SET,
        .l_start = 0,
        .l_len = 10,
    };
    struct close_args args;
    int ready[2], notify[2], fd, status;
    pid_t observer, closer;
    char byte;

    if (argc != 2) {
        fprintf(stderr, "usage: %s TEST_FILE\n", argv[0]);
        return 2;
    }

    fd = open(argv[1], O_CREAT | O_TRUNC | O_RDWR, 0600);
    if (fd < 0 || write(fd, "0123456789", 10) != 10 || close(fd))
        die("create test file");
    if (pipe(ready) || pipe(notify))
        die("pipe");

    /* Fork before opening the lock fd so the observer gets a separate OFD. */
    observer = fork();
    if (observer < 0)
        die("fork");
    if (observer == 0) {
        fd = open(argv[1], O_RDWR);
        if (fd < 0 || write(ready[1], "r", 1) != 1 ||
            read(notify[0], &byte, 1) != 1 ||
            fcntl(fd, F_OFD_GETLK, &lock) < 0)
            _exit(2);
        if (lock.l_type == F_UNLCK)
            _exit(0);
        _exit(1);
    }

    if (read(ready[0], &byte, 1) != 1)
        die("observer ready");
    fd = open(argv[1], O_RDWR);
    if (fd < 0 || fcntl(fd, F_OFD_SETLKW, &lock) < 0)
        die("F_OFD_SETLKW");

    args.fd = fd;
    args.notify = notify[1];
    closer = clone(close_shared_fd, stack + sizeof(stack),
                   CLONE_FILES | SIGCHLD, &args);
    if (closer < 0 || waitpid(closer, NULL, 0) < 0)
        die("clone/waitpid");
    if (waitpid(observer, &status, 0) < 0)
        die("wait observer");

    unlink(argv[1]);
    if (!WIFEXITED(status) || WEXITSTATUS(status) == 2) {
        fprintf(stderr, "observer failed\n");
        return 2;
    }
    if (WEXITSTATUS(status) == 1) {
        puts("FAIL: stale OFD lock observed after close");
        return 1;
    }
    puts("PASS: OFD lock released before GETLK");
    return 0;
}
```

3. Compile the reproducer:

```bash
cc -std=c11 -O2 -Wall -Wextra -Werror \
  curvine_fuse_ofd_release_getlk_repro.c \
  -o /tmp/curvine_fuse_ofd_release_getlk_repro
```

4. Run it repeatedly against the Curvine FUSE mount. The race has a low probability in the minimized program, so several hundred iterations may be required:

```bash
pass=0
fail=0
errors=0

for i in $(seq 1 500); do
  output=$(/tmp/curvine_fuse_ofd_release_getlk_repro \
    /mnt/curvine/ofd-release-getlk-test 2>&1)
  rc=$?
  printf 'run=%03d rc=%d %s\n' "$i" "$rc" "$output"

  if [ "$rc" -eq 0 ]; then
    pass=$((pass + 1))
  elif [ "$rc" -eq 1 ]; then
    fail=$((fail + 1))
  else
    errors=$((errors + 1))
  fi
done

printf 'pass=%d fail=%d errors=%d\n' "$pass" "$fail" "$errors"
```

5. Observe intermittent stale-lock failures. In one run, iterations 87 and 186 failed:

```text
run=087 rc=1 FAIL: stale OFD lock observed after close
run=186 rc=1 FAIL: stale OFD lock observed after close
pass=498 fail=2 errors=0
```

6. Run the same binary against a local filesystem as a control:

```bash
for i in $(seq 1 500); do
  /tmp/curvine_fuse_ofd_release_getlk_repro \
    /tmp/ofd-release-getlk-local || break
done
```

The local ext4 control completed 500 runs without a stale lock:

```text
local_ext4: pass=500 fail=0 errors=0
```

The reproducer opens the observer before creating the locked open file description, acquires an OFD write lock with `F_OFD_SETLKW`, and then uses `clone(CLONE_FILES)` to close the final shared descriptor. The observer executes `F_OFD_GETLK` immediately after the close and fails if it still sees the write lock.

The same race is visible in `xfstests generic/478`. In the failing `CLONE_FILES` case, the expected result is that the lock can be placed, but Curvine returns the stale write lock:

```diff
 get wrlck
-lock could be placed
+get wrlck
 get wrlck
```

**Expected behavior**
After the final reference to an OFD-locked open file description is closed, a subsequent `F_OFD_GETLK` from another open file description must return `F_UNLCK`. Curvine must preserve the kernel-visible ordering between the close/release operation and later lock queries.

```text
PASS: OFD lock released before GETLK
```

Repeated executions on Curvine FUSE should complete without a stale-lock result, matching the local ext4 behavior and allowing `xfstests generic/478` to pass consistently.

 - OS: Ubuntu Linux
 - Kernel: 5.15.0-72-generic
 - Architecture: x86_64
 - Filesystem client: Curvine FUSE
 - Curvine commit: d134f5bfa839a520747f5a3f5a656566782f7c19

**Additional context**
Curvine FUSE dispatches metadata requests as independent asynchronous tasks in `curvine-fuse/src/session/channel/fuse_receiver.rs`. The `RELEASE` path removes the handle and then sends full-range unlock requests to the Master in `curvine-fuse/src/fs/curvine_file_system.rs`, while `GETLK` independently queries the Master lock state.

Consequently, a later `GETLK` task can complete its Master query before the earlier `RELEASE` task has removed the OFD lock. The process issuing `F_OFD_GETLK` then observes a stale lock even though `close()` has already completed and the test synchronization explicitly orders the query after the close.

The fix should preserve per-inode ordering between lock cleanup performed by `RELEASE`/`FLUSH` and subsequent `GETLK`/`SETLK` operations. Care is required for `SETLKW`: a serialization guard must not be held for the entire blocking wait, because doing so could prevent the unlock operation needed to satisfy the waiter. Serializing each individual Master lock-state read or update, or adding an explicit pending-release barrier, would avoid that deadlock risk.

ljy1-lixing:
**Describe the bug**
Curvine FUSE can intermittently expose an open-file-description (OFD) lock after the final reference to the locked file description has been closed. A concurrent `F_OFD_GETLK` may still observe the old lock even though Linux requires OFD locks to be released when the final reference is closed.

This PR fixes the close-versus-lock-query ordering race reported in #1502.

Closes #1502.

**To Reproduce**
Use the standalone reproducer and instructions attached to #1502. The reproducer acquires an OFD write lock, closes the final shared descriptor through a `clone(CLONE_FILES)` child, and immediately queries the lock from a separate open file description.

On an affected build, repeated executions occasionally observe the stale write lock after `close(2)` has completed:

```text
run=087 rc=1 FAIL: stale OFD lock observed after close
run=186 rc=1 FAIL: stale OFD lock observed after close
pass=498 fail=2 errors=0
```

The reproducer source is intentionally not duplicated here because #1502 already contains the complete program, compilation command, execution loop, and local-filesystem control result.

**Expected behavior**
After the final reference to an OFD-locked open file description is closed, a later lock query from another open file description must observe `F_UNLCK`:

```text
PASS: OFD lock released before GETLK
```

Curvine must preserve the request ordering between close-time lock cleanup and subsequent Master lock-state reads or updates for the same file.

**Root cause**
FUSE `RELEASE` and metadata lock requests are dispatched independently. `RELEASE` already acquires Curvine's striped path lock and keeps it held while it removes the file handle and sends the final flock/plock unlock operations to the Master:

```text
RELEASE
  acquire path lock
  remove/close handle
  clear lock owners in Master
  release path lock
  reply to kernel
```

`GETLK`, `SETLK`, and `SETLKW` previously called the Master without acquiring that path lock. Their tasks could therefore interleave with an earlier `RELEASE` task:

```text
RELEASE task                    GETLK task
------------                    ----------
waiting to run/unlock           query Master
                                observe stale OFD lock
clear lock in Master
reply to close
```

The process-level synchronization in the reproducer orders the userspace lock query after `close(2)`, but Curvine's independent asynchronous request tasks did not preserve that ordering at the Master lock state.

The same ordering requirement applies to `FLUSH` when it performs POSIX lock-owner cleanup. Without serialization, a later lock query or update can overtake the cleanup RPC.

Holding one serialization guard throughout a blocking `SETLKW` wait would be unsafe. The unlock request needed to satisfy the waiter could require the same guard, causing a deadlock. The guard must cover only each individual Master lock-state attempt and be released before the retry wait.

**Changes**

- Add ordered Master lock helpers that acquire the existing striped path lock around one `get_lock` or `set_lock` RPC.
- Serialize `GETLK` with close-time lock cleanup for the same path.
- Serialize nonblocking `SETLK` operations with close-time lock cleanup.
- Acquire the path lock for each individual `SETLKW` Master attempt, then release it before waiting or retrying.
- Apply the same path ordering to `FLUSH` when it clears a recorded POSIX lock owner.
- Keep `RELEASE`'s existing path-lock scope across file-handle cleanup and final flock/plock unlock RPCs.
- Preserve existing lock-owner bookkeeping, deadlock detection, retry notification, and full-range unlock normalization.

The resulting blocking lock flow is:

```text
acquire path lock
attempt Master SETLKW operation
release path lock

if blocked:
  register/recheck deadlock state
  wait without the path lock
  retry
```

**Concurrency and performance impact**
The change introduces no global lock. Curvine continues to use its existing striped path-lock table.

- Lock-state RPCs for the same path are serialized with `RELEASE` and lock-cleaning `FLUSH` requests.
- Lock operations on different stripes remain concurrent.
- `SETLKW` does not hold the path lock while sleeping or waiting for another owner.
- The guard covers only one Master lock read or update at a time.
- Normal reads, writes, metadata lookups, and files without lock operations are unaffected.
- No RPC schema, stored metadata, or configuration change is required.

 - OS: Ubuntu Linux
 - Kernel: 5.15.0-72-generic
 - Architecture: x86_64
 - Filesystem client: Curvine FUSE

**Verification**

- Build the updated Curvine FUSE binary and mount it at an isolated temporary mount point connected to the existing Curvine cluster.
- Run the standalone reproducer from #1502 for 500 iterations. The fixed build completed without stale locks or execution errors:

```text
pass=500 fail=0 errors=0
```

- Run `cargo test -p curvine-fuse --lib -- --test-threads=1`:

```text
test result: ok. 417 passed; 0 failed; 0 ignored
```

- Run `cargo check -p curvine-fuse`.
- Run `cargo fmt --all -- --check`.
- Run `git diff --check`.

**Additional context**
The fix uses the same path lock that `RELEASE` already holds, so it directly orders the Master lock operation that removes the OFD owner against later queries and updates. It does not rely on sleeps, retries after a stale result, or changing the Master lock model.

The lock is striped rather than one mutex per inode, so unrelated paths can collide on a stripe and briefly serialize. That behavior already exists for release and other protected operations; this PR extends the same ordering boundary only to remote lock-state RPCs.

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -369,6 +369,18 @@ impl CurvineFileSystem {
         Ok(())
     }
 
+    /// Serialize one Master lock-state operation with close-time lock cleanup.
+    /// Blocking SETLKW retries acquire this guard per attempt, never while waiting.
+    async fn get_lock_ordered(&self, path: &Path, lock: FileLock) -> FuseResult<Option<FileLock>> {
+        let _guard = self.state.lock_path(path).await;
+        Ok(self.fs.get_lock(path, lock).await?)
+    }
+
+    async fn set_lock_ordered(&self, path: &Path, lock: FileLock) -> FuseResult<Option<FileLock>> {
+        let _guard = self.state.lock_path(path).await;
+        Ok(self.fs.set_lock(path, lock).await?)
+    }
+
     fn record_negative_entry(&self) {
         if self.conf.metrics_enabled {
             FuseMetrics::with(|m| m.record_negative_entry());
@@ -1731,6 +1743,8 @@ impl fs::FileSystem for CurvineFileSystem {
         // lock_owner on almost every FUSE_FLUSH/close; unconditional unlock
         // caused a Master SetLock+journal storm for Spark local-dirs (#1227).
         if op.arg.lock_owner != 0 && handle.take_plock_if_owner(op.arg.lock_owner).is_some() {
+            let path = Path::from_str(&handle.status().path)?;
+            let _guard = self.state.lock_path(&path).await;
             if let Err(e) = self
                 .fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
                 .await
@@ -2115,7 +2129,7 @@ impl fs::FileSystem for CurvineFileSystem {
 
         self.state.fs_fsync(op.header.nodeid, None).await?;
 
-        let conflict = self.fs.get_lock(&path, lock).await?;
+        let conflict = self.get_lock_ordered(&path, lock).await?;
         let lk = match conflict {
             Some(lk) => fuse_file_lock {
                 start: lk.start,
@@ -2149,7 +2163,7 @@ impl fs::FileSystem for CurvineFileSystem {
             lock.end = u64::MAX;
         }
 
-        let conflict = self.fs.set_lock(&path, lock).await?;
+        let conflict = self.set_lock_ordered(&path, lock).await?;
         if conflict.is_none() {
             if is_unlock {
                 // Full-range unlock drops this owner from handle bookkeeping so a
@@ -2204,7 +2218,7 @@ impl fs::FileSystem for CurvineFileSystem {
             ),
         );
         loop {
-            let conflict = self.fs.set_lock(&path, lock.clone()).await?;
+            let conflict = self.set_lock_ordered(&path, lock.clone()).await?;
             if conflict.is_none() {
                 wait_guard.clear_blocked_by();
                 if is_unlock {
@@ -2254,7 +2268,7 @@ impl fs::FileSystem for CurvineFileSystem {
                 // deadlock (LTP fcntl17) still observes the cycle. If the lock
                 // is free now (OFD unlock/re-lock race, LTP fcntl34), acquire
                 // instead of returning a false EDEADLK.
-                let conflict2 = self.fs.set_lock(&path, lock.clone()).await?;
+                let conflict2 = self.set_lock_ordered(&path, lock.clone()).await?;
                 if conflict2.is_none() {
                     wait_guard.clear_blocked_by();
                     if is_unlock {
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1743,7 +1743,13 @@ impl fs::FileSystem for CurvineFileSystem {
         // lock_owner on almost every FUSE_FLUSH/close; unconditional unlock
         // caused a Master SetLock+journal storm for Spark local-dirs (#1227).
         if op.arg.lock_owner != 0 && handle.take_plock_if_owner(op.arg.lock_owner).is_some() {
-            let path = Path::from_str(&handle.status().path)?;
+            let path = match Path::from_str(&handle.status().path) {
+                Ok(path) => path,
+                Err(e) => {
+                    handle.add_lock(LockFlags::Plock, op.arg.lock_owner);
+                    return Err(e.into());
+                }
+            };
             let _guard = self.state.lock_path(&path).await;
             if let Err(e) = self
                 .fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)


szbr00 (commented):
Reviewed the OFD lock cleanup ordering change. The core serialization design is correct: `*_ordered` helpers acquire the striped path lock and call the raw `self.fs.*` methods, `fs_unlock_owner` uses the raw `set_lock` (safe under an already-held path lock), and `SETLKW` releases the guard before waiting so the satisfying unlock RPC cannot self-deadlock. One minor suggestion below on the `flush` error-path symmetry.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1731,6 +1743,8 @@ impl fs::FileSystem for CurvineFileSystem {
         // lock_owner on almost every FUSE_FLUSH/close; unconditional unlock
         // caused a Master SetLock+journal storm for Spark local-dirs (#1227).
         if op.arg.lock_owner != 0 && handle.take_plock_if_owner(op.arg.lock_owner).is_some() {
+            let path = Path::from_str(&handle.status().path)?;
```
szbr00: Before this change, `Path::from_str` lived inside `fs_unlock_owner`, so a path-construction error returned `Err` from `fs_unlock_owner` and was caught by the surrounding `if let Err(e) { handle.add_lock(...); return Err(e); }`, restoring the plock owner that `take_plock_if_owner` had just removed.

Hoisting `Path::from_str` above that block means its `?` now propagates out of `flush` directly. The owner has already been taken from handle bookkeeping, but on this error path it is neither unlocked at the Master nor re-added with `handle.add_lock`, breaking the take/unlock-or-restore invariant used by the rest of the flush/release paths.

`handle.status().path` was valid at open time and should remain valid, so this path is effectively unreachable today. Still, to preserve the invariant (and guard against future changes to `Path::from_str` validation), consider restoring the owner before propagating:

```rust
if op.arg.lock_owner != 0 && handle.take_plock_if_owner(op.arg.lock_owner).is_some() {
    let path = match Path::from_str(&handle.status().path) {
        Ok(p) => p,
        Err(e) => {
            handle.add_lock(LockFlags::Plock, op.arg.lock_owner);
            return Err(e.into());
        }
    };
    let _guard = self.state.lock_path(&path).await;
    if let Err(e) = self
        .fs_unlock_owner(&handle, LockFlags::Plock, op.arg.lock_owner)
        .await
    {
        handle.add_lock(LockFlags::Plock, op.arg.lock_owner);
        return Err(e);
    }
}
```

ljy1-lixing: Fixed in 5ac230f7. `flush` now restores the plock owner before propagating a `Path::from_str` error, preserving the take/unlock-or-restore invariant. Verified with `cargo fmt --all -- --check`, `cargo check -p curvine-fuse`, and all 417 `curvine-fuse` library tests.

szbr00 (approved):

PR #1577, opened 2026-08-14
fix(fuse): owner gate for same-gid chown and killpriv setid clear

gongxun0928:
## Summary
Follow-up to #1574 / Copilot review: align non-owner `chown` permission with Linux.

- **GID owner gate**: any present `FATTR_GID` (including same-gid) requires file owner; membership checked only when gid value changes.
- **`valid==0` / `chown(-1,-1)`**: if the file has setuid/setgid and caller is non-owner non-root → `EPERM`; if no setid bits → allow no-op (matches Linux).

## Test plan
- [x] `cargo test -p curvine-fuse --lib setattr_permission` (incl. new `denies_non_owner_same_gid_chown`)
- [x] `cargo test -p curvine-fuse --lib chown_`
- [ ] FUSE remount: non-owner `chown(-1,-1)` on `04755` → EPERM; on `0755` → OK; non-owner `chown(-1,same_gid)` → EPERM

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -890,13 +890,11 @@ impl CurvineFileSystem {
             }
         }
 
-        // Changing the group is also privileged for non-root: caller must own the
-        // file *and* belong to the target group (POSIX chown). Membership alone is
-        // not sufficient — #1500 dropped the owner gate and allowed non-owners who
-        // share the target group to chgrp (ACL bypass / setuid strip).
+        // Non-root chown requires ownership whenever FATTR_GID is present (even same-gid).
+        // Membership is only checked when the gid value actually changes.
         // See CurvineIO/curvine#1548 (originally gongxun0928/curvine#18).
         if let Some(gid) = target_gid {
-            if gid != file_gid && caller_uid != file_uid {
+            if caller_uid != file_uid {
                 return err_fuse!(
                     libc::EPERM,
                     "setattr gid change denied: caller uid {} is not file owner (uid {})",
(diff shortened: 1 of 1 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1488,6 +1486,20 @@ impl fs::FileSystem for CurvineFileSystem {
         // Clear setuid/setgid on chown (FATTR_UID/GID, or valid==0 for chown(-1,-1)).
         let chown_effective = Self::chown_should_clear_setid_bits(op.arg.valid);
         if chown_effective && cur_status.file_type == FileType::File {
+            // killpriv chown(-1,-1) → valid==0: Linux EPERM if setid present and non-owner;
+            // allow no-op when the file has no setid bits.
+            if op.arg.valid == 0
```
szbr00, line 1491: The new valid==0 EPERM path in set_attr is only covered indirectly by the
check_setattr_permission unit tests; the added
setattr_permission_denies_non_owner_same_gid_chown test exercises the GID
owner gate, not this killpriv block (valid==0 skips the uid/gid branches in
check_setattr_permission and returns Ok, so the gate is enforced solely
here). The PR test plan lists the FUSE remount assertion as unchecked.

Consider adding a test that drives set_attr with valid==0 for a non-owner on a
setid file (expect EPERM) and on a non-setid file (expect success / no-op), so
this gate is regression-protected. The chown_should_clear_setid_bits(valid==0)
branch is easy to regress in a future refactor since nothing asserts its
permission behavior.

curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -1488,6 +1486,20 @@ impl fs::FileSystem for CurvineFileSystem {
...
+                && self.conf.check_permission
+                && op.header.uid != 0
+                && op.header.uid != file_uid
+                && (cur_status.mode & (libc::S_ISUID as u32 | libc::S_ISGID as u32)) != 0
```
szbr00, line 1495: This EPERM gate fires whenever S_ISUID or S_ISGID is set, but for a setgid bit
without S_IXGRP (mandatory-locking mode) Linux does not always reject a
non-owner chown(-1,-1). Per setattr_should_drop_sgid() in fs/attr.c, the
kernel only adds ATTR_KILL_SGID for the no-exec-setgid case when the caller is
not in the file's group; if a non-owner caller IS in the file's group, no
ATTR_KILL_SGID is emitted, so notify_change never synthesizes ATTR_MODE and
setattr_prepare skips the ownership check — the call succeeds and the setgid
bit is preserved. This block returns EPERM in that case, which is stricter
than Linux.

It is a conservative divergence and an obscure edge (setgid-no-exec +
non-owner in-group + write perm + explicit chown(-1,-1)), so it may be
acceptable as-is. If exact Linux parity is wanted, consider narrowing the
condition to the bits that would actually be cleared: S_ISUID, or
(S_ISGID & S_IXGRP), or (S_ISGID & !S_IXGRP & !caller_in_file_group).

PR #1578, opened 2026-08-14
fix(client): degrade auto-cache overload to UFS

ljy1-lixing:
# Summary

Prevent auto-cache admission overload from failing foreground FUSE reads, make the client pending capacity configurable, and make concurrent admission and same-path deduplication race-free.

On a cache miss, the unified filesystem schedules an asynchronous load from UFS into Curvine and immediately falls back to reading the file from UFS. The cache load is an optimization and is not required to serve the current open. Curvine previously returned `TransferOverloaded` when 1024 distinct cache submissions were pending. That error escaped from `open()`, was mapped to the default FUSE `EIO`, and failed an otherwise valid UFS read.

The pending limit is now configured with `transfer.client_pending_queue_size`, defaults to 1024, and must be greater than zero. Concurrent client submissions are independently limited by `transfer.client_submit_concurrency`, which defaults to 64. When the pending limit is reached, Curvine skips the new cache request while the foreground open continues through UFS.

# Minimal Reproducer

Configure a small client-side pending capacity and an unreachable Transfer endpoint so that the first asynchronous submission remains pending during its retry backoff:

```toml
[transfer]
enabled = true
endpoints = ["127.0.0.1:19010"]
client_pending_queue_size = 1
client_submit_concurrency = 1
```

Mount a UFS directory with auto-cache enabled, create several files directly in the UFS source, and open them concurrently through the Curvine FUSE mount:

```bash
for i in $(seq 1 32); do
    printf 'file-%s\n' "$i" > "/path/to/ufs/source/file-$i"
done

seq 1 32 | xargs -P32 -I{} sh -c \
    'cat "/mnt/curvine/file-{}" >/dev/null'
```

Before the fix, the first miss occupied the only pending slot. Other misses returned `TransferOverloaded` from the client admission check. Because `UnifiedFileSystem::open()` propagated that result with `?`, FUSE converted the unmapped filesystem error to `EIO`, and some `cat` processes failed even though all source files were readable from UFS.

After the fix, the extra cache requests are skipped. All opens continue to UFS and complete successfully; a later cache miss can retry admission after a slot is released.

# Root Cause

Auto-cache scheduling was coupled to foreground open success:

```text
cache miss
  -> async_cache()
  -> pending count reaches 1024
  -> TransferOverloaded
  -> open() returns the error
  -> FUSE maps the unrecognized error to EIO
```

This violates the intended cache-mode behavior. The current request does not wait for the asynchronous load job and reads from UFS regardless, so failure to schedule the optional cache population must not fail the current open.

The old admission logic also used three separate concurrent operations:

```text
contains(path) -> len() < capacity -> insert(path)
```

Those operations were not atomic. Concurrent requests for the same path could both pass `contains()`, while concurrent requests for different paths could all pass the length check and exceed the configured limit. Server-side Transfer request idempotency can avoid creating duplicate jobs, but it does not prevent duplicate client RPC submissions or enforce the client pending bound.

Finally, the 1024 limit was exposed only through a getter that always returned a constant, so deployments could not tune admission capacity for their file count, Transfer latency, or expected cache-miss concurrency.

# Changes

- Add `transfer.client_pending_queue_size` with a default of 1024 and reject a zero value during configuration initialization.
- Add `transfer.client_submit_concurrency` with a default of 64 to bound simultaneous client submission and retry attempts independently from the pending queue.
- Propagate the setting through the SDK/Java filesystem configuration path.
- Propagate `fs.cv.transfer.client_pending_queue_size` and `fs.cv.transfer.client_submit_concurrency` through the Hadoop/Java `FilesystemConf` TOML generator.
- Add the setting to the standard, Docker example, and SPDK configuration templates.
- Replace the racy multi-step admission check with a short `FastMutex<HashSet>` critical section that atomically deduplicates paths, checks capacity, and records accepted requests.
- Hold an RAII admission permit for the full submission attempt, including retry backoff, and release the path marker on success, failure, cancellation, or task drop.
- Treat an already-pending or overloaded cache request as a successful best-effort outcome instead of returning `TransferOverloaded`.
- Count skipped auto-cache admissions by reason and keep per-path overload messages at debug level to avoid warning storms during scans.
- Isolate `open()` from any current or future auto-cache scheduling error so foreground reads still fall back to UFS.
- Add tests for configurable capacity and submit concurrency, zero-value validation, capacity enforcement, permit release, submit concurrency enforcement, and concurrent same-path deduplication.

# Deployment Note

Existing configurations remain compatible and continue to use the default capacity of 1024 when `client_pending_queue_size` is omitted.

`client_pending_queue_size` limits distinct auto-cache requests waiting for or performing client-side submission. `client_submit_concurrency` limits how many of those requests may concurrently perform UFS status lookup, Transfer RPC submission, and retry backoff. Neither setting controls the number of Transfer jobs executing on the server; `transfer.max_running_transfers` remains the server-side execution limit.

The admission tracker is created when the unified filesystem client starts. Changing either client-side setting therefore requires restarting the affected FUSE, Rust, or Java client process. The Master and Worker do not need to restart for this client-side behavior change. A standalone Transfer service only needs a restart if its shared configuration file is changed as part of the deployment.

When the queue is full, new cache population requests are deliberately dropped instead of being buffered without a bound. The foreground read is still served from UFS, and a later cache miss can submit the load after pending capacity is available.

# Verification

- `cargo test -p curvine-config -p curvine-client-core -p curvine-unified-fs -p curvine-sdk-core` passes.
- Auto-cache capacity, submit concurrency, and concurrent same-path deduplication tests pass.
- Transfer configuration parsing, defaulting, validation, and SDK propagation tests pass.
- `mvn -f curvine-libsdk/java/pom.xml -Dtest=LoadJobClientTest test` passes: `8 passed`.
- `cargo clippy -p curvine-config -p curvine-unified-fs -p curvine-sdk-core --tests -- -D warnings` passes.
- `git diff --check` passes.

### DIFF
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -32,13 +32,13 @@ use curvine_model::{
 use curvine_runtime::common::TimeSpent;
 use curvine_runtime::common::Utils;
 use curvine_runtime::runtime::{RpcRuntime, Runtime};
-use dashmap::DashSet;
+use curvine_runtime::sync::FastMutex;
 use log::{debug, error, info, warn};
 use std::borrow::Cow;
+use std::collections::HashSet;
 use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tokio::time;
 
 const TRANSFER_SUBMIT_MAX_ATTEMPTS: usize = 3;
@@ -52,8 +52,7 @@ enum CacheValidity {
 
 #[derive(Clone)]
 struct AsyncCachePending {
-    paths: Arc<DashSet<String>>,
-    slots: Arc<Semaphore>,
+    paths: Arc<FastMutex<HashSet<String>>>,
     capacity: usize,
 }
 
@@ -65,43 +64,37 @@ enum AsyncCacheAdmission {
 
 struct AsyncCachePermit {
     path: String,
-    paths: Arc<DashSet<String>>,
-    _slot: OwnedSemaphorePermit,
+    paths: Arc<FastMutex<HashSet<String>>>,
 }
 
 impl AsyncCachePending {
     fn new(capacity: usize) -> Self {
         Self {
-            paths: Arc::new(DashSet::new()),
-            slots: Arc::new(Semaphore::new(capacity)),
+            paths: Arc::new(FastMutex::new(HashSet::new())),
             capacity,
         }
     }
 
     fn try_admit(&self, path: String) -> AsyncCacheAdmission {
-        if self.paths.contains(&path) {
+        let mut paths = self.paths.lock();
+        if paths.contains(&path) {
             return AsyncCacheAdmission::AlreadyPending;
         }
-
-        let slot = match self.slots.clone().try_acquire_owned() {
-            Ok(slot) => slot,
-            Err(_) => return AsyncCacheAdmission::Overloaded,
-        };
-        if !self.paths.insert(path.clone()) {
-            return AsyncCacheAdmission::AlreadyPending;
+        if paths.len() >= self.capacity {
+            return AsyncCacheAdmission::Overloaded;
         }
+        paths.insert(path.clone());
 
         AsyncCacheAdmission::Accepted(AsyncCachePermit {
             path,
             paths: self.paths.clone(),
-            _slot: slot,
         })
     }
 }
 
 impl Drop for AsyncCachePermit {
     fn drop(&mut self) {
-        self.paths.remove(&self.path);
+        self.paths.lock().remove(&self.path);
     }
 }
 
@@ -524,7 +517,7 @@ impl UnifiedFileSystem {
 
     pub fn async_cache(&self, source_path: &Path) -> FsResult<()> {
         let source_path = source_path.clone_uri();
-        let _pending = match self.async_cache_pending.try_admit(source_path.clone()) {
+        let pending_permit = match self.async_cache_pending.try_admit(source_path.clone()) {
             AsyncCacheAdmission::Accepted(pending) => pending,
             AsyncCacheAdmission::AlreadyPending => {
                 debug!("async cache request already pending for {}", source_path);
@@ -543,6 +536,7 @@ impl UnifiedFileSystem {
         let metrics = self.metrics;
 
         self.fs_context().rt().spawn(async move {
+            let _pending_permit = pending_permit;
             let time = TimeSpent::new();
             let res = fs.submit_async_cache(&source_path).await;
 
@@ -573,7 +567,6 @@ impl UnifiedFileSystem {
                     debug!("submitted async cache job {} for {}", job_id, source_path);
                 }
             }
-            drop(_pending);
         });
 
         Ok(())
@@ -1262,6 +1255,7 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
 #[cfg(test)]
 mod tests {
     use super::{AsyncCacheAdmission, AsyncCachePending};
+    use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::{Arc, Mutex};
 
     #[test]
@@ -1292,16 +1286,26 @@ mod tests {
     fn async_cache_pending_deduplicates_concurrent_requests() {
         let pending = Arc::new(AsyncCachePending::new(32));
         let accepted = Arc::new(Mutex::new(Vec::new()));
+        let already_pending = Arc::new(AtomicUsize::new(0));
+        let overloaded = Arc::new(AtomicUsize::new(0));
         let mut threads = Vec::new();
 
         for _ in 0..32 {
             let pending = pending.clone();
             let accepted = accepted.clone();
+            let already_pending = already_pending.clone();
+            let overloaded = overloaded.clone();
             threads.push(std::thread::spawn(move || {
-                if let AsyncCacheAdmission::Accepted(permit) =
-                    pending.try_admit("ufs://bucket/same".to_string())
-                {
-                    accepted.lock().unwrap().push(permit);
+                match pending.try_admit("ufs://bucket/same".to_string()) {
+                    AsyncCacheAdmission::Accepted(permit) => {
+                        accepted.lock().unwrap().push(permit);
+                    }
+                    AsyncCacheAdmission::AlreadyPending => {
+                        already_pending.fetch_add(1, Ordering::Relaxed);
+                    }
+                    AsyncCacheAdmission::Overloaded => {
+                        overloaded.fetch_add(1, Ordering::Relaxed);
+                    }
                 }
             }));
         }
@@ -1310,5 +1314,7 @@ mod tests {
         }
 
         assert_eq!(accepted.lock().unwrap().len(), 1);
+        assert_eq!(already_pending.load(Ordering::Relaxed), 31);
+        assert_eq!(overloaded.load(Ordering::Relaxed), 0);
     }
 }
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -38,6 +38,7 @@ use std::borrow::Cow;
 use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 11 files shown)


crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -514,7 +573,7 @@ impl UnifiedFileSystem {
                     debug!("submitted async cache job {} for {}", job_id, source_path);
                 }
             }
-            fs.async_cache_pending.remove(&source_path);
+            drop(_pending);
```
szbr00: The explicit `drop(_pending)` at the end of the spawned task is redundant: the permit would already be dropped at the end of the async block's scope, releasing both the path marker and the semaphore slot in the correct order (the `Drop` impl runs before field drops, so the path is removed before the slot is released). Removing the explicit drop would make the intent clearer that the permit lives for the entire submission attempt including retry backoff, with no risk of early release. If the explicit drop is kept for readability, a short comment explaining the RAII guarantee would help future readers.

ljy1-lixing: Thanks. Fixed in 7374124a: the permit is now moved into a task-local RAII binding and is released automatically when the async block exits.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -49,14 +50,69 @@ enum CacheValidity {
...
+        if !self.paths.insert(path.clone()) {
+            return AsyncCacheAdmission::AlreadyPending;
+        }
+
+        AsyncCacheAdmission::Accepted(AsyncCachePermit {
+            path,
+            paths: self.paths.clone(),
+            _slot: slot,
+        })
+    }
```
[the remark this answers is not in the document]
ljy1-lixing: Thanks. Fixed in 7374124a: admission is now serialized in a short FastMutex<HashSet> critical section, so same-path deduplication and capacity accounting are atomic and duplicates cannot transiently consume capacity.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -467,23 +517,26 @@ impl UnifiedFileSystem {
...
-        }
-        self.async_cache_pending.insert(source_path.clone());
+        let pending_permit = match self.async_cache_pending.try_admit(source_path.clone()) {
+            AsyncCacheAdmission::Accepted(pending) => pending,
+            AsyncCacheAdmission::AlreadyPending => {
+                debug!("async cache request already pending for {}", source_path);
+                return Ok(());
+            }
+            AsyncCacheAdmission::Overloaded => {
+                warn!(
```
szbr00: This warn runs on the foreground cache-miss open path for every skipped path. When Transfer is slow or unreachable, the 1024 admitted tasks hold slots across UFS get_status plus submit retry backoff, so a directory walk or Spark scan can emit a warning per extra file. That can stall FUSE workers on log I/O and hide other signals. Consider a skipped-admission counter (next to mount_cache_misses) and a rate-limited or debug log for individual paths.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
```diff
@@ -122,11 +122,22 @@ pub struct FilesystemConf {
     pub transfer: TransferClientConf,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(default)]
 pub struct TransferClientConf {
     pub enabled: bool,
     pub endpoints: Vec<String>,
+    pub client_pending_queue_size: usize,
```
szbr00, line 130: Rust FilesystemConf can parse client_pending_queue_size, but io.curvine.FilesystemConf still only maps fs.cv.transfer.enabled and fs.cv.transfer.endpoints into the generated TOML. A Hadoop/Spark client that sets fs.cv.transfer.client_pending_queue_size is silently ignored and always gets 1024. If Java clients are expected to tune this, wire the property through FilesystemConf.toToml() the same way as endpoints; otherwise the Java-path test should not imply the field is produced by Java.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
ljy1-lixing: Thanks. Fixed in a334a1f8: overloaded admissions now increment client_async_cache_admission_skipped{reason="overloaded"}, and individual path messages are debug-level. Already-pending skips are counted separately.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
ljy1-lixing: Thanks. Fixed in a334a1f8: io.curvine.FilesystemConf now reads fs.cv.transfer.client_pending_queue_size, emits it under [transfer], and the Java regression test verifies a configured value of 2048.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -467,23 +524,47 @@ impl UnifiedFileSystem {
...
         let log = self.audit_logging_enabled;
         let metrics = self.metrics;
 
         self.fs_context().rt().spawn(async move {
+            let _pending_permit = pending_permit;
+            let _submit_permit = match fs
+                .async_cache_pending
+                .submit_slots
+                .clone()
+                .acquire_owned()
```
szbr00, line 559: `try_admit()` reserves a pending slot before this task blocks on `acquire_owned()`. Spawn is cheap, so a cache-miss burst will immediately occupy `client_pending_queue_size` slots while only `client_submit_concurrency` tasks actually run `submit_async_cache`. The remaining unique paths are skipped as overloaded.

Because the submit permit is also held across UFS `get_status`, RPC, and retry backoff, that skip window is much longer than the old "up to 1024 in-flight submits" model. Lowering `client_submit_concurrency` to protect Transfer therefore increases `reason="overloaded"` skips. This also applies when `transfer.enabled = false`, because the JobMaster `submit_load_job` path uses the same semaphore.

If idle waiters should not starve new paths, use `try_acquire_owned()` here and treat a full submit limiter as a skip (drop the pending permit, count a `submit_busy` reason). If blocking wait is intended, please say so in the deployment note: pending capacity includes submit waiters, and the two knobs are coupled.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -1195,3 +1279,82 @@ impl FileSystem<UnifiedWriter, UnifiedReader> for UnifiedFileSystem {
...
+            thread.join().unwrap();
+        }
+
+        assert_eq!(accepted.lock().unwrap().len(), 1);
+        assert_eq!(already_pending.load(Ordering::Relaxed), 31);
+        assert_eq!(overloaded.load(Ordering::Relaxed), 0);
+    }
+
+    #[test]
+    fn async_cache_submit_slots_enforce_concurrency() {
```
szbr00: This test only exercises `Semaphore::try_acquire_owned` and never calls `try_admit()`. Production behavior is the opposite order: admit first (counts against pending), then wait for a submit slot. Please add a case with `pending=2, submit_concurrency=1` that both paths are accepted, the second task cannot obtain a submit permit until the first is dropped, and dropping only the submit permit does not remove the path from the pending set. Without that, a future change can re-break the two-limiter coupling while these tests still pass.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
ljy1-lixing: The blocking waiter behavior is intentional: `client_pending_queue_size` is the bounded total of distinct requests waiting for or performing submission, while `client_submit_concurrency` is the worker-style execution limit. This prevents cache-miss bursts from bypassing the bounded queue when submit concurrency is low. The PR deployment note already states that pending capacity includes requests waiting for submission; 1a389159 now makes that coupling explicit in the configuration field documentation and adds a lifecycle regression test.

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
ljy1-lixing: Addressed in 1a389159. The test now uses `pending=2` and `submit_concurrency=1`, admits two distinct paths, verifies the second submit slot is unavailable until the first is dropped, and confirms that dropping a submit permit does not release either pending path. Each path is removed only when its own pending RAII permit is dropped.

szbr00 (approved):

szbr00 (approved):

PR #1579, opened 2026-08-14
fix(raft): advance applied index for leader no-op

ljy1-lixing:
# Summary

Prevent a Master from stopping during leader promotion when the committed Raft log contains entries without metadata mutations, including empty leader no-ops and Raft configuration changes.

Raft appends an empty no-op entry whenever a node becomes leader. The metadata apply path correctly skipped the entry because it contains no namespace mutation, but it also left `fsm_state.applied` unchanged. The leader-promotion fence introduced by #1557 requires the metadata applied index to reach the committed Raft index before publishing leadership. On a fresh single-node cluster, the first no-op can be committed at index 1 while metadata remains applied at index 0, so promotion fails and the Raft node exits.

Non-data entries now advance the metadata apply high-water mark without attempting to deserialize or execute a journal batch. They are also dispatched through the normal committed-entry path, so the application state remains current when a no-op commits after the initial promotion scan. Leader promotion can therefore distinguish "no metadata mutation" from "not applied" and complete safely.

# Minimal Reproducer

Start a fresh single-node Master with an empty journal and wait for its first election. A single-node quorum can commit the leader no-op immediately:

```text
initial state: committed=0, metadata_applied=0
first election: append and commit empty no-op at index=1
```

Before the fix, committed-entry processing skipped the empty entry, and the promotion fence later observed:

```text
metadata catch-up stopped before committed raft index: applied=0, commit=1
raft node stop: metadata catch-up stopped before committed raft index: applied=0, commit=1
```

The Raft task then advanced its role state to `Exit`. `MasterServer::start()` could not observe a Leader or Follower state and returned an error after the role listener closed, causing the Master process to exit.

The same condition can be reproduced directly by appending an empty committed entry to a fresh `JournalSystem` and invoking `role_change(StateRole::Leader)`. Before the fix, role change returns the metadata catch-up error; after the fix, it succeeds with `fsm_state.applied.index == 1`.

# Root Cause

Leader promotion was fenced on the committed Raft index:

```text
Raft commits leader no-op at index 1
  -> apply_committed_entries() skips empty data
  -> metadata applied index remains 0
  -> role_change(Leader)
  -> catch_up_committed_metadata()
  -> scan sees the no-op but apply0() returns without updating applied
  -> metadata_applied(0) < committed(1)
  -> promotion fails and the Raft node exits
```

An empty Raft entry is a no-op only at the application-data layer. It still occupies a real committed log index and must advance the application state machine's applied high-water mark. Treating it as completely invisible breaks the invariant checked by the leader-promotion fence.

Raft configuration entries have the same application-layer property: their protobuf payload is consumed by Raft rather than the metadata journal. Promotion replay must advance their applied index without passing that payload to `JournalBatch::deserialize_compat`.

The old early return also happened before the stale-entry check. Updating the applied index unconditionally at that point would risk moving the high-water mark backwards when replay encounters an already-applied empty entry.

# Changes

- Check the current role-specific applied index before handling an empty entry, so stale no-op entries remain safely ignored.
- Dispatch empty normal entries and committed configuration entries to `AppStorage` even when no client response is required.
- Deserialize journal batches only for non-empty `EntryNormal` entries; advance `fsm_state.applied` for no-op and configuration entries without changing namespace metadata.
- Advance both metadata and UFS applied high-water marks when an empty entry is replayed through the leader path, preserving the existing `set_applied` semantics.
- Add a regression test that commits an empty entry, promotes the node to Leader, and verifies that promotion succeeds with the applied index advanced to the committed index.
- Keep the existing committed-metadata promotion test to verify that non-empty journal replay behavior remains unchanged.

# Deployment Note

This change does not add or modify configuration. Existing journal data remains compatible.

The fix changes only how committed empty Raft entries update in-memory and persisted FSM progress during replay. Empty entries still perform no namespace or UFS mutation. Masters affected by the first-election failure need to be rebuilt and restarted with the fixed version; no journal cleanup or metadata migration is required.

The behavior applies to both fresh elections and later leader promotions where the local committed range contains Raft no-op entries that are newer than the metadata applied high-water mark.

# Verification

- `cargo test -p curvine-server --test journal_test leader_promotion` passes: `3 passed`.
- `leader_promotion_advances_applied_index_over_committed_noop` verifies that an empty committed entry advances `fsm_state.applied` and no longer blocks leader promotion.
- `leader_promotion_advances_over_committed_configuration_entry` verifies that Raft configuration entries advance both promotion high-water marks without journal deserialization.
- `committed_leader_noop_advances_app_storage_applied_index` verifies that a no-op committed through the normal Raft path reaches application storage.
- `leader_promotion_applies_committed_metadata_before_returning` continues to verify ordered replay of non-empty committed metadata before leadership is published.
- `cargo test -p curvine-raft` passes.
- The complete `journal_test` suite passes: `9 passed`.
- `git diff --check` passes.

### DIFF
--- a/curvine-master/src/master/journal/journal_loader.rs
+++ b/curvine-master/src/master/journal/journal_loader.rs
@@ -32,7 +32,7 @@ use curvine_runtime::common::{FileUtils, TimeSpent};
 use curvine_runtime::runtime::{RpcRuntime, Runtime};
 use curvine_runtime::sync::channel::{AsyncChannel, AsyncReceiver, AsyncSender, CallChannel};
 use log::{debug, error, info, warn};
-use raft::eraftpb::Entry;
+use raft::eraftpb::{Entry, EntryType};
 use raft::StateRole;
 use std::path::Path;
 use std::sync::{Arc, Mutex, MutexGuard};
@@ -226,9 +226,9 @@ impl JournalLoader {
             return Ok(());
         }
 
-        // Raft appends an empty no-op entry when a leader is elected. It has no
-        // metadata mutation, but still advances the committed apply high-water mark.
-        if entry.data.is_empty() {
+        // Empty leader no-ops and configuration entries have no metadata mutation,
+        // but still advance the committed apply high-water mark.
+        if entry.get_entry_type() != EntryType::EntryNormal || entry.data.is_empty() {
             let mut applied = if is_leader {
                 cur.ufs_applied
             } else {
--- a/curvine-master/src/master/journal/journal_loader.rs
+++ b/curvine-master/src/master/journal/journal_loader.rs
@@ -229,7 +229,14 @@ impl JournalLoader {
         // Raft appends an empty no-op entry when a leader is elected. It has no
         // metadata mutation, but still advances the committed apply high-water mark.
         if entry.data.is_empty() {
-            self.set_applied(is_leader, Self::build_applied(entry), false)?;
+            let mut applied = if is_leader {
+                cur.ufs_applied
+            } else {
+                cur.applied
+            };
+            applied.term = entry.term;
+            applied.index = entry.index;
+            self.set_applied(is_leader, applied, false)?;
             return Ok(());
         }
 
--- a/curvine-master/src/master/journal/journal_loader.rs
+++ b/curvine-master/src/master/journal/journal_loader.rs
@@ -216,10 +216,6 @@ impl JournalLoader {
         entry: &Entry,
         skip_ufs_error: bool,
     ) -> CommonResult<()> {
-        if entry.data.is_empty() {
-            return Ok(());
-        }
-
         let cur = self.fsm_state_snapshot()?;
         let role_applied = ternary!(is_leader, cur.ufs_applied.index, cur.applied.index);
         if entry.index <= role_applied {
@@ -227,6 +223,13 @@ impl JournalLoader {
                 "skip entry index {}, term {}, fsm_state {:?}",
                 entry.index, entry.term, cur
             );
+            return Ok(());
+        }
+
+        // Raft appends an empty no-op entry when a leader is elected. It has no
+        // metadata mutation, but still advances the committed apply high-water mark.
+        if entry.data.is_empty() {
+            self.set_applied(is_leader, Self::build_applied(entry), false)?;
             return Ok(());
         }
 
--- a/curvine-server/tests/journal_test.rs
+++ b/curvine-server/tests/journal_test.rs
@@ -32,7 +32,8 @@ use curvine_server::master::journal::{
 };
 use curvine_server::master::{Master, MountManager};
 use log::info;
-use raft::eraftpb::Entry;
+use prost::Message;
+use raft::eraftpb::{ConfChange, Entry, EntryType};
 use raft::StateRole;
 use std::collections::HashMap;
 use std::sync::Arc;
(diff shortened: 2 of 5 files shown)


curvine-server/tests/journal_test.rs
```diff
@@ -142,6 +142,30 @@ fn leader_promotion_applies_committed_metadata_before_returning() -> CommonResul
...
+
+    journal_system.append_committed_entry_for_test(Entry {
+        term: 1,
+        index: 1,
+        ..Default::default()
+    })?;
+
+    AsyncRuntime::single().block_on(async { loader.role_change(StateRole::Leader).await })?;
+
+    assert_eq!(loader.get_fsm_state().applied.index, 1);
```
szbr00: The new test exercises the leader-promotion catch-up path (`catch_up_committed_metadata` runs `apply0` with `is_leader=false`, so only `state.applied` advances). The PR description also notes that empty entries advance `ufs_applied` on the leader path via `set_applied(true, ...)`, but that branch is not directly asserted here.

Consider adding an assertion on `ufs_applied` (or a follow-up test that drives the `is_leader=true` scan path) so the dual high-water advancement behavior is locked in and won't silently regress if `set_applied` semantics change.

curvine-server/tests/journal_test.rs
```diff
@@ -142,6 +142,30 @@ fn leader_promotion_applies_committed_metadata_before_returning() -> CommonResul
...
+    Ok(())
```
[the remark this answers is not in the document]
ljy1-lixing: Addressed in 4e16f880. The test now applies a real metadata mutation before the no-op, captures its `op_id`/`rpc_id`, and verifies that both high-water marks retain those values while their Raft index advances.

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -230,6 +226,13 @@ impl JournalLoader {
             return Ok(());
         }
 
+        // Raft appends an empty no-op entry when a leader is elected. It has no
+        // metadata mutation, but still advances the committed apply high-water mark.
+        if entry.data.is_empty() {
+            self.set_applied(is_leader, Self::build_applied(entry), false)?;
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
ljy1-lixing: Fixed in 4e16f880. Empty entries now clone the current role-specific `AppliedIndex` and update only `term` and `index`, so `op_id` and `rpc_id` are preserved instead of being reset by `AppliedIndex::default()`.

curvine-server/tests/journal_test.rs
ljy1-lixing: Addressed in 4e16f880. The regression test now waits for the queued leader UFS replay scan by using a follower role change as an ordering barrier, then asserts that both `applied` and `ufs_applied` advance to the no-op index while preserving their operation progress.

szbr00 (approved):

curvine-master/src/master/journal/journal_loader.rs
```diff
@@ -230,6 +226,20 @@ impl JournalLoader {
             return Ok(());
         }
 
+        // Raft appends an empty no-op entry when a leader is elected. It has no
+        // metadata mutation, but still advances the committed apply high-water mark.
+        if entry.data.is_empty() {
```
szbr00: Checking only `data.is_empty()` does not cover all committed Raft entries without a journal mutation. `EntryConfChange` carries protobuf data, is handled directly by `apply_config_change`, and never advances the applied index in `JournalLoader`. If a later promotion scan reaches that entry, this falls through to `JournalBatch::deserialize_compat` and fails leader promotion.

Please branch on `entry_type`: deserialize only non-empty `EntryNormal` entries as journal batches, and advance the appropriate high-water marks for configuration entries without applying metadata or UFS mutations. A promotion regression test containing a committed configuration entry would cover this path.

szbr00: This handler is currently reached only during scans. The normal committed-entry path in `RaftNode::apply_committed_entries` still skips empty entries before dispatching them to `AppStorage`. In a multi-node election, the leader no-op can commit after the promotion scan captured its commit index, leaving both application high-water marks behind until another entry or replay happens.

Please dispatch empty `EntryNormal` entries to `AppStorage` without producing a client response, or otherwise advance the application state when they commit. A test that commits the no-op after leader promotion starts would cover this ordering.

ljy1-lixing: Addressed in 5bf8068c. Journal replay now deserializes only non-empty `EntryNormal` entries as `JournalBatch`; configuration entries advance the role-specific applied high-water mark without metadata or UFS mutation. Added `leader_promotion_advances_over_committed_configuration_entry`, which verifies both `applied` and `ufs_applied`.

ljy1-lixing: Addressed in 5bf8068c. `RaftNode::apply_committed_entries` no longer drops empty entries: empty normal entries are dispatched to `AppStorage` with no client response, and configuration entries are also dispatched after Raft applies the membership change. Added `committed_leader_noop_advances_app_storage_applied_index` to cover a no-op reaching application storage through the normal committed path.

PR #1571, opened 2026-08-13
feat(version): add unified ComponentVersion type (#1527)

lzjqsdd:
# Summary

Add the unified structured version model to `curvine-sys`: a SemVer-style `ReleaseVersion` type with parsing and precedence, plus `git describe` support captured at build time. The existing `ComponentVersion` (component / release_version / git_commit / git_tag / git_branch / protocol_version / min_protocol_version / capabilities) stays wire- and JSON-compatible, with a `with_capabilities` builder for future capability negotiation.

# Issue Describe / Design

Related to #1527.

This PR implements the second version-management step (T2, depends on T1):

- `ReleaseVersion` parses `X.Y.Z[-pre][+build]` (optional leading `v` for tags), follows SemVer precedence (pre-release sorts before release, numeric pre-release identifiers compare by value, build metadata is ignored for ordering/equality), and serializes to/from JSON.
- `git describe` output is captured at build time (`GIT_DESCRIBE`) via `git describe --tags --always --dirty`. `parse_git_describe` converts the standard forms into SemVer-compatible strings:
  - `v0.4.0` -> `0.4.0`
  - `v0.4.0-alpha` -> `0.4.0-alpha`
  - `v0.4.0-alpha-121-g26b5dc6b` -> `0.4.0-alpha.121+g26b5dc6b`
  - `v0.4.0-alpha-121-g26b5dc6b-dirty` -> `0.4.0-alpha.121+g26b5dc6b.dirty`
  - `dev-g26b5dc6b` -> `0.0.0-dev+g26b5dc6b`
- `ComponentVersion` gains a non-breaking `with_capabilities` builder; the 8-field JSON schema emitted by `--version-json` is unchanged.
- The compatibility checker (later task) can use `ReleaseVersion` ordering directly; no existing caller changes.

# Changes

| Module / File | Change | Impact on existing behavior |
| ------------- | ------ | --------------------------- |
| `crates/core/curvine-sys/build.rs` | Capture `git describe --tags --always --dirty` and emit `GIT_DESCRIBE` build-time constant | None (new constant only) |
| `crates/core/curvine-sys/src/version.rs` | Add `ReleaseVersion`, `PreRelease`, `VersionParseError`, git describe parsing, `git_describe()` / `git_describe_version()`, `with_capabilities` builder, and unit tests | None (additive; existing `ComponentVersion` API and JSON schema unchanged) |

# Test verified

| Test case | Result | Notes |
| --------- | ------ | ----- |
[...]

PR #1580, opened 2026-08-14
feat(proto): add component and compatibility messages (#1527)

### LINKED ISSUE #1527: [FEATURE] Component version, release, and wire-protocol compatibility management

lzjqsdd:
# Summary

Adds the public version/compatibility protocol messages to the common proto: `ComponentInfoProto` (structured component version, git provenance, protocol versions and capabilities), `CompatibilityModeProto` (`diagnose` / `enforce`), and `ServerCompatibilityInfoProto` (server-side compatibility contract with min version bounds and blocklist). These messages are the wire foundation for the version management design (#1527): client-master handshake (T5/T6), worker-master structured version reporting (T7) and client-worker data-plane versioning (T9) will carry them on the reserved 1000+ field range of existing RPC messages.

# Issue Describe / Design

Closes #1527 (sub-task).

Design (from `version-management-design.md` / GitHub issue #1527):

- New messages use low consecutive field numbers starting from 1 (protobuf forbids field number 0). The 1000+ range is reserved for cross-cutting protocol metadata added to existing messages, so these new messages never collide with future business fields.
- `ComponentInfoProto` mirrors the unified `ComponentVersion` type (T2, merged as part of #1571) so every component can report the same structured provenance on the wire.
- `ServerCompatibilityInfoProto` carries the compatibility policy a server advertises during handshake/heartbeat: `compatibility_mode` defaults to `DIAGNOSE` so mixed-version clusters are never rejected by default.
- Purely additive: no existing message, field number or RPC frame is changed; legacy components simply skip the unknown fields.

# Changes

| `crates/common/curvine-proto/proto/common.proto` | Add `ComponentInfoProto`, `CompatibilityModeProto`, `ServerCompatibilityInfoProto` | None (new messages only) |
| `crates/common/curvine-proto/tests/component_compat_test.rs` | New protobuf round-trip and unknown-field compatibility tests | None (test-only) |

# Test verified

| `cargo test -p curvine-proto` | PASS | 14 tests (7 new + 7 existing pipeline tests) |
| `ComponentInfoProto` round-trip (full + defaults) | PASS | |
| `ServerCompatibilityInfoProto` round-trip (full + defaults) | PASS | includes nested `server` message |
| Unknown-field compatibility (varint / length-delimited / fixed32 on 1000+/2000+ ranges, top-level and nested) | PASS | unknown fields skipped, known fields preserved |
| `CompatibilityModeProto` wire values (DIAGNOSE=1, ENFORCE=2) | PASS | |
| `make format` (clippy `--deny warnings` + `cargo fmt`) | PASS | |

# Dependencies

- Related PRs: #1571 (T2 unified `ComponentVersion` type — dependency, merged)
- Related issues: #1527
- Blocks / blocked by: blocked by #1571 (merged); unblocks T5/T6/T7/T9

### DIFF
--- a/crates/common/curvine-proto/proto/common.proto
+++ b/crates/common/curvine-proto/proto/common.proto
@@ -242,17 +242,27 @@ message ComponentInfoProto {
 // Compatibility enforcement mode of a server when a peer version is outside
 // the supported range. diagnose records warnings and allows the request;
 // enforce rejects incompatible peers with an explicit error.
+// UNKNOWN is the zero value so an unset field never maps to an invalid
+// variant on the wire.
 enum CompatibilityModeProto {
+    COMPATIBILITY_MODE_PROTO_UNKNOWN = 0;
     COMPATIBILITY_MODE_PROTO_DIAGNOSE = 1;
     COMPATIBILITY_MODE_PROTO_ENFORCE = 2;
 }
 
 // Server-side compatibility contract exchanged during handshake (e.g.
 // GetFilesystemInfo) and structured version reporting (e.g. worker heartbeat).
+//
+// Only the `server` identity and `compatibility_mode` are required; the
+// version bounds are optional because this message may be produced by
+// different component types that do not all own every bound (e.g. a worker
+// does not dictate min_client_version). Optional bounds keep the contract
+// lenient by default and allow future restructuring without breaking
+// inter-version decoding.
 message ServerCompatibilityInfoProto {
     required ComponentInfoProto server = 1;
-    required string min_worker_version = 2 [default = ""];
-    required string min_client_version = 3 [default = ""];
+    optional string min_worker_version = 2 [default = ""];
+    optional string min_client_version = 3 [default = ""];
     required CompatibilityModeProto compatibility_mode = 4 [default = COMPATIBILITY_MODE_PROTO_DIAGNOSE];
     repeated string blocked_versions = 5;
 }
--- a/crates/common/curvine-proto/proto/common.proto
+++ b/crates/common/curvine-proto/proto/common.proto
@@ -221,3 +221,38 @@ message ListOptionsProto {
     optional int32 limit = 1;
     optional string start_after = 2;
 }
+
+// Structured version/protocol information reported by every Curvine component
+// (master, worker, client, fuse, cli, csi, sdk) on the wire.
+//
+// Internal fields are numbered consecutively from 1. Existing messages that
+// carry this message as cross-cutting protocol metadata use the reserved
+// 1000+ field range instead, so adding it never collides with business fields.
+message ComponentInfoProto {
+    optional string component = 1 [default = ""];
+    optional string release_version = 2 [default = ""];
+    optional string git_commit = 3 [default = ""];
+    optional string git_tag = 4 [default = ""];
+    optional string git_branch = 5 [default = ""];
+    optional uint32 protocol_version = 6 [default = 1];
+    optional uint32 min_protocol_version = 7 [default = 1];
+    repeated string capabilities = 8;
+}
+
+// Compatibility enforcement mode of a server when a peer version is outside
+// the supported range. diagnose records warnings and allows the request;
+// enforce rejects incompatible peers with an explicit error.
+enum CompatibilityModeProto {
+    COMPATIBILITY_MODE_PROTO_DIAGNOSE = 1;
+    COMPATIBILITY_MODE_PROTO_ENFORCE = 2;
+}
+
+// Server-side compatibility contract exchanged during handshake (e.g.
+// GetFilesystemInfo) and structured version reporting (e.g. worker heartbeat).
+message ServerCompatibilityInfoProto {
+    required ComponentInfoProto server = 1;
+    required string min_worker_version = 2 [default = ""];
+    required string min_client_version = 3 [default = ""];
+    required CompatibilityModeProto compatibility_mode = 4 [default = COMPATIBILITY_MODE_PROTO_DIAGNOSE];
+    repeated string blocked_versions = 5;
+}
--- a/crates/common/curvine-proto/proto/common.proto
+++ b/crates/common/curvine-proto/proto/common.proto
@@ -242,8 +242,12 @@ message ComponentInfoProto {
 // Compatibility enforcement mode of a server when a peer version is outside
 // the supported range. diagnose records warnings and allows the request;
 // enforce rejects incompatible peers with an explicit error.
+//
 // UNKNOWN is the zero value so an unset field never maps to an invalid
-// variant on the wire.
+// variant on the wire. Consumers MUST treat UNKNOWN as DIAGNOSE (lenient):
+// the enum's Rust default is UNKNOWN (0) while the compatibility_mode field
+// default is DIAGNOSE (1), so producers must set the field explicitly and
+// consumers must not fail closed when they see UNKNOWN.
 enum CompatibilityModeProto {
     COMPATIBILITY_MODE_PROTO_UNKNOWN = 0;
     COMPATIBILITY_MODE_PROTO_DIAGNOSE = 1;
--- a/crates/common/curvine-proto/tests/component_compat_test.rs
+++ b/crates/common/curvine-proto/tests/component_compat_test.rs
@@ -1,0 +1,235 @@
+use curvine_proto::{CompatibilityModeProto, ComponentInfoProto, ServerCompatibilityInfoProto};
+use prost::Message;
+
+fn sample_component_info() -> ComponentInfoProto {
+    ComponentInfoProto {
+        component: Some("worker".to_string()),
+        release_version: Some("0.4.0-alpha".to_string()),
+        git_commit: Some("24c848719b5b4fea74519d91cbe462bb49761b36".to_string()),
+        git_tag: Some("v0.4.0-alpha".to_string()),
+        git_branch: Some("main".to_string()),
+        protocol_version: Some(1),
+        min_protocol_version: Some(1),
+        capabilities: vec![
+            "short-circuit".to_string(),
+            "batch-write".to_string(),
+            "transfer".to_string(),
+        ],
+    }
+}
+
+fn sample_server_compatibility_info() -> ServerCompatibilityInfoProto {
+    ServerCompatibilityInfoProto {
+        server: sample_component_info(),
+        min_worker_version: "0.2.0".to_string(),
+        min_client_version: "0.2.0".to_string(),
+        compatibility_mode: CompatibilityModeProto::Diagnose as i32,
+        blocked_versions: vec!["0.2.5".to_string(), "0.2.6".to_string()],
+    }
+}
+
+fn sample_server_compatibility_info_enforce() -> ServerCompatibilityInfoProto {
+    ServerCompatibilityInfoProto {
+        compatibility_mode: CompatibilityModeProto::Enforce as i32,
+        ..sample_server_compatibility_info()
+    }
+}
+
+/// Append a raw varint field (field number, wire type 0) to a wire buffer,
+/// emulating a newer peer that sent fields this build does not know about.
+fn push_varint(buf: &mut Vec<u8>, mut value: u64) {
+    loop {
+        let byte = (value & 0x7f) as u8;
+        value >>= 7;
+        if value == 0 {
+            buf.push(byte);
+            break;
+        }
+        buf.push(byte | 0x80);
+    }
+}
+
+/// Append an unknown length-delimited field (wire type 2) to a wire buffer.
+fn push_unknown_len_delimited(buf: &mut Vec<u8>, field: u32, payload: &[u8]) {
+    push_varint(buf, ((field as u64) << 3) | 2);
+    push_varint(buf, payload.len() as u64);
+    buf.extend_from_slice(payload);
+}
+
+/// Append an unknown fixed32 field (wire type 5) to a wire buffer.
+fn push_unknown_fixed32(buf: &mut Vec<u8>, field: u32, value: u32) {
+    push_varint(buf, ((field as u64) << 3) | 5);
+    buf.extend_from_slice(&value.to_le_bytes());
+}
+
+#[test]
+fn test_component_info_round_trip_full() {
+    let info = sample_component_info();
+
+    let encoded = info.encode_to_vec();
+    let decoded = ComponentInfoProto::decode(encoded.as_slice()).unwrap();
+
+    assert_eq!(decoded.component, Some("worker".to_string()));
+    assert_eq!(decoded.release_version, Some("0.4.0-alpha".to_string()));
+    assert_eq!(
+        decoded.git_commit,
+        Some("24c848719b5b4fea74519d91cbe462bb49761b36".to_string())
+    );
+    assert_eq!(decoded.git_tag, Some("v0.4.0-alpha".to_string()));
+    assert_eq!(decoded.git_branch, Some("main".to_string()));
+    assert_eq!(decoded.protocol_version, Some(1));
+    assert_eq!(decoded.min_protocol_version, Some(1));
+    assert_eq!(
+        decoded.capabilities,
+        vec![
+            "short-circuit".to_string(),
+            "batch-write".to_string(),
+            "transfer".to_string()
+        ]
+    );
+}
+
+#[test]
+fn test_component_info_round_trip_defaults() {
+    // An empty component info must survive a round trip with all optional
+    // fields unset (None), which is how a legacy/unknown peer is represented.
+    let info = ComponentInfoProto {
+        component: None,
+        release_version: None,
+        git_commit: None,
+        git_tag: None,
+        git_branch: None,
+        protocol_version: None,
+        min_protocol_version: None,
+        capabilities: vec![],
+    };
+
+    let encoded = info.encode_to_vec();
+    let decoded = ComponentInfoProto::decode(encoded.as_slice()).unwrap();
+
+    assert_eq!(decoded, info);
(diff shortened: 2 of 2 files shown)


crates/common/curvine-proto/proto/common.proto
```diff
@@ -221,3 +221,38 @@ message ListOptionsProto {
...
+enum CompatibilityModeProto {
+    COMPATIBILITY_MODE_PROTO_DIAGNOSE = 1;
+    COMPATIBILITY_MODE_PROTO_ENFORCE = 2;
+}
+
+// Server-side compatibility contract exchanged during handshake (e.g.
+// GetFilesystemInfo) and structured version reporting (e.g. worker heartbeat).
+message ServerCompatibilityInfoProto {
+    required ComponentInfoProto server = 1;
+    required string min_worker_version = 2 [default = ""];
```
szbr00: `ServerCompatibilityInfoProto` marks all four business fields `required`, so any producer must populate every one of them. That is fine for a master that knows its full compatibility contract, but the design doc says this message is also carried on worker heartbeats (T7), where a worker typically does not dictate a `min_client_version`. With `required`, a worker would have to emit a meaningless empty string for fields it does not own.

More importantly, `required` is a wire-level hard constraint: once T5/T6/T7/T9 embed this message, you cannot later split it (e.g. move `blocked_versions` into its own message) or make a field conditionally absent without breaking inter-version decode (prost returns `DecodeError::RequiredFieldNotSet` for a missing required field). The existing proto uses `required` heavily, but those are all single-producer messages; this one is meant to be sent by multiple component types.

Consider making `min_worker_version`, `min_client_version`, and `blocked_versions` `optional` (keeping `server` and `compatibility_mode` required, or even making all of them optional with the `DIAGNOSE` default already declared). That preserves the "lenient by default" policy described in the PR while allowing partial population and future restructuring.

crates/common/curvine-proto/proto/common.proto
```diff
@@ -221,3 +221,38 @@ message ListOptionsProto {
...
+    optional uint32 protocol_version = 6 [default = 1];
+    optional uint32 min_protocol_version = 7 [default = 1];
+    repeated string capabilities = 8;
+}
+
+// Compatibility enforcement mode of a server when a peer version is outside
+// the supported range. diagnose records warnings and allows the request;
+// enforce rejects incompatible peers with an explicit error.
+enum CompatibilityModeProto {
+    COMPATIBILITY_MODE_PROTO_DIAGNOSE = 1;
```
szbr00, line 253: `CompatibilityModeProto` starts at 1 with no zero value, so `ServerCompatibilityInfoProto::default()` produces `compatibility_mode: 0`, which is not a valid variant. Encoding that default and decoding on the peer side preserves the raw 0, and any consumer doing `CompatibilityModeProto::try_from(value)` will error. This is a latent footgun for code paths that build the message via `..Default::default()` (the `server` field already uses that pattern in `test_server_compatibility_info_round_trip_defaults`).

Two options:
- Add `COMPATIBILITY_MODE_PROTO_UNKNOWN = 0` (or `UNSET = 0`) so `Default` maps to a valid variant, consistent with most enums in this file (`StorageTypeProto`, `HeartbeatStatusProto`, `LockTypeProto`, etc.).
- Or keep the current values but add a comment on the enum stating that `Default::default()` is not a valid on-the-wire value and producers must always set `compatibility_mode` explicitly.

The first option is preferable since it removes the footgun entirely.

lzjqsdd: Thanks. Addressed in 6012a071: added `COMPATIBILITY_MODE_PROTO_UNKNOWN = 0` so the zero value maps to a valid variant (consistent with `StorageTypeProto` etc. in this file), and added `test_server_compatibility_info_default_mode_is_valid` asserting `Default::default()` and empty-wire decode both yield `Diagnose` (prost honors the declared `default = DIAGNOSE`, so the default is 1; UNKNOWN=0 removes the footgun for any code path that produces a raw 0).

crates/common/curvine-proto/proto/common.proto
lzjqsdd: Thanks. Addressed in 6012a071: `min_worker_version` and `min_client_version` are now `optional`; `blocked_versions` stays `repeated` (inherently optional). `server` and `compatibility_mode` remain `required` per the design doc. Added a regression test (`test_server_compatibility_info_partial_bounds_round_trip`) covering a producer that leaves `min_client_version` unset.

crates/common/curvine-proto/tests/component_compat_test.rs
```diff
@@ -0,0 +1,279 @@
...
+fn test_server_compatibility_info_unknown_fields_ignored() {
+    // Emulate a future handshake response carrying extra high-range fields
+    // both at the top level and nested inside the embedded component info.
+    let mut compat = sample_server_compatibility_info();
+
+    // Unknown fields inside the nested `server` message (field 1).
+    let mut nested_encoded = compat.server.encode_to_vec();
+    push_varint(&mut nested_encoded, 1000 << 3);
+    push_varint(&mut nested_encoded, 7);
+    compat.server = ComponentInfoProto::decode(nested_encoded.as_slice()).unwrap();
```
szbr00: This does not actually put unknown fields inside the nested `server` bytes of the outer message. prost 0.11 does not retain unknown fields, so `ComponentInfoProto::decode(...)` drops tags 1000/7 before `compat.encode_to_vec()`. The outer decode therefore only sees the top-level unknowns appended below.

To cover the nested case (the one T5/T7 will hit: `ComponentInfo` inside `ServerCompatibilityInfo` inside an RPC), keep the unknown bytes in the nested buffer and splice that buffer in as field 1 yourself, without decoding it first. `test_component_info_unknown_fields_ignored` already covers skip-on-decode for the child type; this test should cover skip-through-parent.

crates/common/curvine-proto/tests/component_compat_test.rs
```diff
@@ -0,0 +1,279 @@
...
+fn test_server_compatibility_info_default_mode_is_valid() {
+    // Default::default() must never produce an invalid wire value for
+    // compatibility_mode (regression for the zero-variant footgun).
+    let default = ServerCompatibilityInfoProto::default();
+    assert_eq!(
+        default.compatibility_mode,
+        CompatibilityModeProto::Diagnose as i32
+    );
+
+    let decoded = ServerCompatibilityInfoProto::decode(&[] as &[u8]).unwrap();
```
szbr00: `decode(&[])` is not a portable wire contract for `ServerCompatibilityInfoProto`.

`server` is `required`, so an empty payload has a missing required nested message. This assertion only passes because prost 0.11 never checks proto2 `required` at decode time; missing fields keep `Default`. Java protobuf-java 3.25 (this repo SDK) still rejects that payload (`isInitialized()` / `parseFrom` fails on missing required fields). The in-memory `ServerCompatibilityInfoProto::default()` path is also not empty on the wire: prost always encodes required fields, so `encode(Default)` writes an empty `server` (tag+len 0) plus `compatibility_mode = DIAGNOSE`.

Please drop the empty-buffer decode (or replace it with a payload that includes `server`), and round-trip `encode(Default)` instead. Otherwise this test documents a Rust-only payload that the Java SDK will fail to parse during T5/T6 handshake.

crates/common/curvine-proto/proto/common.proto
```diff
@@ -221,3 +221,48 @@ message ListOptionsProto {
...
+    repeated string capabilities = 8;
+}
+
+// Compatibility enforcement mode of a server when a peer version is outside
+// the supported range. diagnose records warnings and allows the request;
+// enforce rejects incompatible peers with an explicit error.
+// UNKNOWN is the zero value so an unset field never maps to an invalid
+// variant on the wire.
+enum CompatibilityModeProto {
+    COMPATIBILITY_MODE_PROTO_UNKNOWN = 0;
```
szbr00, line 252: Adding `UNKNOWN = 0` makes 0 a valid variant, but the two `Default` impls now disagree:

- `ServerCompatibilityInfoProto::default().compatibility_mode` is `DIAGNOSE` (1), from the field default
- `CompatibilityModeProto::default()` is `UNKNOWN` (0), because prost's enum `Default` is the zero variant

Producers that write `compatibility_mode: CompatibilityModeProto::default() as i32` (or `i32::default()`) will send `UNKNOWN`, not `DIAGNOSE`. The design says mixed-version clusters must not be rejected by default, but `UNKNOWN` has no stated consumer policy — a handshake that matches only Diagnose/Enforce can error or fail closed.

Please document that consumers must treat `UNKNOWN` as `DIAGNOSE` (lenient), and add a test that `CompatibilityModeProto::default() as i32 == 0` so T5/T6 code does not assume enum default equals field default.

crates/common/curvine-proto/tests/component_compat_test.rs
lzjqsdd: Thanks. Addressed in 41c92aef: dropped the `decode(&[])` case and the test now round-trips `encode(Default)`, asserting the payload is non-empty (prost always writes the required `server` tag+len 0 plus compatibility_mode=DIAGNOSE), decodes back to DIAGNOSE, and the empty server keeps unset fields. This documents a payload the Java SDK (protobuf-java 3.25) can parse.

crates/common/curvine-proto/proto/common.proto
lzjqsdd: Thanks. Addressed in 41c92aef: the proto comment on `CompatibilityModeProto` now states that consumers MUST treat UNKNOWN as DIAGNOSE (lenient) and must not fail closed, and that the enum Rust default (UNKNOWN=0) differs from the compatibility_mode field default (DIAGNOSE=1) so producers must set the field explicitly. Added `assert_eq!(CompatibilityModeProto::default() as i32, 0)` in test_compatibility_mode_enum_values to pin this down for T5/T6.

crates/common/curvine-proto/tests/component_compat_test.rs
lzjqsdd: Thanks. Addressed in 41c92aef: the test now splices the unknown-field bytes into the nested `server` buffer without decoding it first (prost would drop them), and builds the outer wire buffer by hand (field 1 = nested buffer with unknowns, plus known fields 2/3/4 and top-level unknowns). The parent decode now really skips-through-parent, matching the T5/T7 shape (ComponentInfo inside ServerCompatibilityInfo inside an RPC).

szbr00 (approved):

PR #1582, opened 2026-08-14
feat(client): add master handshake with legacy fallback (#1527)

lzjqsdd:
# Summary

Adds the client side of the client-master version handshake (T6 of the version management design #1527): the client reports its own `component_info` on `GetFilesystemInfo` and caches the master's advertised version / protocol / capabilities from the response `compatibility` contract, with a legacy fallback that never rejects old components.

# Issue Describe / Design

Related to #1527 (sub-task T6, depends on T5 #1581).

Design (from `version-management-design.md` / GitHub issue #1527):

- `FsClient::get_filesystem_info` now sends `GetFilesystemInfoRequest.component_info` (built from this client's `ComponentVersion` via the T4/T5 `ProtoUtils::component_version_to_pb` helper) on the reserved 1000+ range. Legacy masters skip the unknown field; legacy clients simply omit it.
- The response's `compatibility` contract (T5) is parsed into a `MasterHandshake` and cached in `FsContext`, shared by all `FsClient` clones: the master's component version, protocol version, min protocol version, capabilities and compatibility mode.
- Legacy fallback: a master that does not advertise a `compatibility` field is treated as a **legacy peer** — no version metadata and never rejected. The cache defaults to legacy before the first successful handshake, an unset/unknown compatibility mode maps to `DIAGNOSE` (lenient), and a failed handshake RPC at FUSE mount time degrades to legacy assumptions with a warning instead of blocking the mount.
- A new `handshake()` API performs the handshake explicitly (used at FUSE mount initialization) and `master_handshake()` exposes the cached contract; both are exposed through `FsClient`, the core `CurvineFileSystem`, and `UnifiedFileSystem`.

# Changes

| `crates/client/curvine-client-core/src/file/handshake.rs` | New `MasterHandshake` type: parses the GetFilesystemInfo response, exposes master version / protocol / capabilities / compatibility mode, and treats absence of `compatibility` (and unset/unknown modes) as legacy — never rejects | None (new type + unit tests) |
| `crates/client/curvine-client-core/src/file/fs_context.rs` | Cache the handshake result in `FsContext` (shared by every `FsClient` clone), defaulting to legacy | None (new field, defaults to legacy) |
| `crates/client/curvine-client-core/src/file/fs_client.rs` | `get_filesystem_info` reports the client's `component_info` and caches the master handshake; new `handshake()` / `master_handshake()` APIs | None (request gains one optional field; response parse unchanged) |
| `crates/client/curvine-client-core/src/file/curvine_filesystem.rs` | Delegate `handshake()` / `master_handshake()` | None |
| `crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs` | Delegate `handshake()` / `master_handshake()` | None |
| `curvine-fuse/src/cli/mount_run.rs` | Perform the handshake at mount initialization; on RPC failure warn and continue with legacy assumptions | None (one extra RPC at startup; failure is non-fatal) |
| `curvine-tests/tests/cluster_test.rs` | New cluster integration tests: new-client handshake against a live master, legacy-master fallback | None (test-only) |

# Test verified

| `cargo test -p curvine-client-core` | PASS | 30 tests (7 new: handshake parsing/legacy/default + request building/backward-compat) |
| `cargo test -p curvine-proto` | PASS | 6 legacy wire-compat tests (new client + legacy master, legacy client + new master, both directions) still pass |
| `cargo test -p curvine-model --lib` | PASS | 23 tests |
| `cargo test -p curvine-tests --test cluster_test` | PASS | 3 tests (2 new handshake integration tests against a live MiniCluster) |
| `cargo clippy --release --all-targets -- --deny warnings` | PASS | via `make format` |
| `make format` (clippy `--deny warnings` + `cargo fmt`) | PASS | |

# Dependencies

- Related PRs: #1581 (T5 handshake fields — dependency, merged); #1580 (T4 proto messages — merged)
- Related issues: #1527
- Blocks / blocked by: blocked by #1581 (merged); unblocks T10 (client-worker protocol negotiation pre-check)

### DIFF
--- a/crates/client/curvine-client-core/src/file/fs_client.rs
+++ b/crates/client/curvine-client-core/src/file/fs_client.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::file::FsContext;
+use crate::file::{FsContext, MasterHandshake};
 use bytes::BytesMut;
 use curvine_config::{ClientConf, ClusterConf, UfsConf, UfsConfBuilder};
 use curvine_core_error::err_box;
@@ -532,10 +532,40 @@ impl FsClient {
     }
 
     pub async fn get_filesystem_info(&self) -> FsResult<FilesystemInfo> {
-        let header = GetFilesystemInfoRequest::default();
+        let header = GetFilesystemInfoRequest {
+            // Client-master handshake: report this client's structured version
+            // on the reserved 1000+ range. Legacy masters skip the unknown
+            // field and legacy clients omit it entirely.
+            component_info: Some(Self::handshake_request_component_info()),
+        };
         let rep: GetFilesystemInfoResponse = self.rpc(RpcCode::GetFilesystemInfo, header).await?;
-        let res = ProtoUtils::filesystem_info_from_pb(rep);
-        Ok(res)
+        // Cache the master's advertised version / protocol / capabilities; a
+        // master without a compatibility contract is recorded as legacy and
+        // never rejected.
+        self.context
+            .set_master_handshake(MasterHandshake::from_response(&rep));
+        Ok(ProtoUtils::filesystem_info_from_pb(rep))
+    }
+
+    /// Client-master version handshake: report this client's `component_info`
+    /// and cache the master's advertised version / protocol / capabilities.
+    /// Returns the cached handshake; a master without a compatibility
+    /// contract is reported as a legacy peer and is never rejected.
+    pub async fn handshake(&self) -> FsResult<MasterHandshake> {
+        let _ = self.get_filesystem_info().await?;
+        Ok(self.master_handshake())
+    }
+
+    /// Cached master handshake (version / protocol / capabilities). Before the
+    /// first handshake and against legacy masters this reports a legacy peer,
+    /// which is never rejected.
+    pub fn master_handshake(&self) -> MasterHandshake {
+        self.context.master_handshake()
+    }
+
+    /// Build the client's own `component_info` payload for the handshake.
+    fn handshake_request_component_info() -> ComponentInfoProto {
+        ProtoUtils::component_version_to_pb(&curvine_sys::version::component_version("client"))
     }
 
     pub async fn get_filesystem_info_bytes(&self) -> FsResult<BytesMut> {
@@ -750,5 +780,45 @@ impl FsClient {
 
     pub fn client_addr(&self) -> &ClientAddress {
         &self.context.client_addr
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use curvine_proto::GetFilesystemInfoRequest;
+
+    #[test]
+    fn handshake_request_carries_client_component_info() {
+        // A new client reports its own structured version during the
+        // handshake so the master can diagnose mixed-version clusters.
+        let info = FsClient::handshake_request_component_info();
+        let req = GetFilesystemInfoRequest {
+            component_info: Some(info),
+        };
+
+        let encoded = req.encode_to_vec();
+        let decoded = GetFilesystemInfoRequest::decode(encoded.as_slice()).unwrap();
+        let decoded_info = decoded.component_info.unwrap();
+        assert_eq!(decoded_info.component.as_deref(), Some("client"));
+        assert_eq!(decoded_info.protocol_version, Some(1));
+        assert_eq!(decoded_info.min_protocol_version, Some(1));
+    }
+
+    #[test]
+    fn handshake_request_is_backward_compatible() {
+        // The request only adds an optional field on the reserved 1000+ range:
+        // a legacy master's view of the message (no component_info) must
+        // decode the payload and ignore the unknown field.
+        #[derive(Clone, PartialEq, ::prost::Message)]
+        struct LegacyGetFilesystemInfoRequest {}
+
+        let req = GetFilesystemInfoRequest {
+            component_info: Some(FsClient::handshake_request_component_info()),
+        };
+        let encoded = req.encode_to_vec();
+
+        let legacy = LegacyGetFilesystemInfoRequest::decode(encoded.as_slice()).unwrap();
+        assert_eq!(legacy, LegacyGetFilesystemInfoRequest {});
     }
 }
--- a/crates/client/curvine-client-core/src/file/fs_client.rs
+++ b/crates/client/curvine-client-core/src/file/fs_client.rs
@@ -25,6 +25,7 @@ use curvine_proto::*;
 use curvine_rpc::client::ClusterConnector;
 use curvine_rpc::message::MessageBuilder;
 use curvine_runtime::runtime::RpcRuntime;
+use log::warn;
 use prost::Message as PMessage;
 use std::collections::LinkedList;
 use std::sync::Arc;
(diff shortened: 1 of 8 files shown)


crates/client/curvine-client-core/src/file/handshake.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Cached outcome of the client-master version handshake.
+#[derive(Clone, Debug)]
+pub struct MasterHandshake {
+    /// Whether the peer master is a legacy component (it did not advertise a
+    /// compatibility contract). Legacy peers are never rejected; the client
+    /// degrades to the pre-handshake behavior for them.
+    pub legacy: bool,
+    /// The compatibility contract advertised by the master, when present.
+    pub compatibility: Option<ServerCompatibilityInfoProto>,
+}
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 9b51b44a: `legacy`/`compatibility` are now private fields with `is_legacy()`/`compatibility()` accessors; instances are only built via `default()`/`legacy()`/`from_response()`, so inconsistent states cannot be constructed.

crates/client/curvine-client-core/src/file/fs_client.rs
```diff
@@ -532,10 +532,40 @@ impl FsClient {
     }
 
     pub async fn get_filesystem_info(&self) -> FsResult<FilesystemInfo> {
-        let header = GetFilesystemInfoRequest::default();
+        let header = GetFilesystemInfoRequest {
+            // Client-master handshake: report this client's structured version
+            // on the reserved 1000+ range. Legacy masters skip the unknown
+            // field and legacy clients omit it entirely.
+            component_info: Some(Self::handshake_request_component_info()),
+        };
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 9b51b44a: `component_info` is now attached to `GetFilesystemInfo` only on the first call per session, gated by an atomic flag in `FsContext` (shared by all `FsClient` clones). Frequent statfs queries omit the payload, and a failed reporting request resets the flag so a retry can still report.

lzjqsdd:
Addressed the suppressed review comment in 77a53a41: `get_filesystem_info_bytes()` now shares the same one-time request builder as the typed `get_filesystem_info()` path, so callers using the bytes API also report the client `component_info` on the first `GetFilesystemInfo` per session (and omit it afterwards to keep statfs queries lean).

crates/client/curvine-client-core/src/file/fs_context.rs
```diff
@@ -49,6 +51,16 @@ pub struct FsContext {
     pub(crate) os_cache: CacheManager,
     pub(crate) failed_workers: Cache<u32, WorkerAddress, BuildHasherDefault<FxHasher>>,
     pub(crate) block_pool: Arc<BlockClientPool>,
+    // Client-master handshake result: the master's advertised version /
+    // protocol / capabilities (or a legacy marker when the master does not
+    // advertise a compatibility contract). Populated by the first
+    // GetFilesystemInfo call and shared by every FsClient clone.
+    master_handshake: Arc<FastRwLock<MasterHandshake>>,
```
szbr00: `FsContext` is already shared through `Arc<FsContext>` (both `CurvineFileSystem` and `FsClient` hold it behind an `Arc`), and `master_handshake` / `handshake_reported` are only ever accessed via `&self`. The extra `Arc` wrapping therefore adds a needless heap allocation and indirection on every access; `FastRwLock<MasterHandshake>` and `AtomicBool` are already interior-mutable and can be stored inline:

    master_handshake: FastRwLock<MasterHandshake>,
    handshake_reported: AtomicBool,

Minor nit — not blocking.

lzjqsdd: Thanks. Fixed in 59948ea7: `master_handshake` and `handshake_reported` are now stored inline in `FsContext` (confirmed it is never cloned directly, only shared behind `Arc`), removing the redundant heap allocation and indirection.

crates/client/curvine-client-core/src/file/fs_client.rs
```diff
@@ -532,15 +532,84 @@ impl FsClient {
...
+
+    /// Build a `GetFilesystemInfo` request, attaching this client's
+    /// `component_info` only on the first call per session (the handshake).
+    /// Both the typed and the raw-bytes RPC paths share this so every
+    /// GetFilesystemInfo caller reports the client version exactly once;
+    /// frequent statfs queries stay lean. Returns the request plus whether
+    /// `component_info` was attached, so the caller can reset the flag when
+    /// the request fails before reaching the master.
+    fn get_filesystem_info_request(&self) -> (GetFilesystemInfoRequest, bool) {
+        let report_component = self.context.claim_handshake_report();
```
szbr00, line 612: The one-shot flag is claimed before awaiting the RPC and reset only by the explicit `Err` branch. If this future is cancelled or dropped, that branch never runs and later calls permanently stop attaching `component_info`. Please make the in-flight claim cancellation-safe (for example, reset from a drop guard unless it is committed after a successful response).

crates/client/curvine-client-core/src/file/fs_client.rs
```diff
@@ -532,15 +532,84 @@ impl FsClient {
...
+        };
+        (header, report_component)
     }
 
     pub async fn get_filesystem_info_bytes(&self) -> FsResult<BytesMut> {
-        let header = GetFilesystemInfoRequest::default();
-        self.rpc_bytes(RpcCode::GetFilesystemInfo, header).await
+        let (header, report_component) = self.get_filesystem_info_request();
+        match self.rpc_bytes(RpcCode::GetFilesystemInfo, header).await {
+            Ok(bytes) => Ok(bytes),
```
szbr00: The bytes path consumes the same one-shot reporting flag as the typed path, but on success it returns the raw response without updating `master_handshake`. A bytes-only caller (including the Java/Python SDK path) therefore keeps the default legacy cache even when the master advertised a compatibility contract. Please decode the response and run the same cache-update logic as the typed path, with a bytes-first regression test.

curvine-fuse/src/cli/mount_run.rs
```diff
@@ -49,6 +49,15 @@ pub fn run_mount(args: FuseRuntimeArgs) -> CommonResult<()> {
         let fs = CurvineFileSystem::new(cluster_conf, fuse_rt.clone())?;
         let conf = fs.conf().clone();
 
+        // Client-master version handshake: report this client's component_info
+        // and cache the master's advertised version / protocol / capabilities.
+        // A master without a compatibility contract is a legacy peer and is
+        // never rejected; a failed handshake degrades to legacy assumptions
+        // instead of blocking the mount.
+        if let Err(e) = fs.fs().handshake().await {
```
szbr00, line 57: This is currently the only production path that performs the handshake proactively. CLI, SDK, data-transfer, and direct `CurvineFileSystem`/`UnifiedFileSystem` users can issue normal master RPCs without ever calling `GetFilesystemInfo`, so they neither report their component metadata nor cache the compatibility contract before doing work. Please move this into a shared initialization path, or add a lazy guard before the first normal master RPC, so handshake behavior is not limited to FUSE mounts.

crates/client/curvine-client-core/src/file/fs_client.rs
lzjqsdd: Thanks. Fixed in cb3a7aef: the one-time `component_info` claim is now held by a `HandshakeReportGuard` that resets the claim on drop unless committed after a successful response, so a cancelled or failed in-flight future never permanently suppresses reporting.

crates/client/curvine-client-core/src/file/fs_client.rs
lzjqsdd: Thanks. Fixed in cb3a7aef: `get_filesystem_info_bytes` now decodes the response and runs the same `master_handshake` cache update as the typed path, so bytes-only callers (e.g. SDK paths) no longer stay at the default legacy cache. Added a bytes-first regression test (`test_bytes_first_get_filesystem_info_caches_handshake`).

curvine-fuse/src/cli/mount_run.rs
lzjqsdd: Thanks. Fixed in cb3a7aef: a lazy one-time handshake now runs before the first ordinary master RPC (mutex + flag in FsContext, best-effort), so CLI/SDK/data-transfer and direct `CurvineFileSystem`/`UnifiedFileSystem` users report component_info and cache the compatibility contract before doing work, not only FUSE mount. The handshake RPC uses raw rpc entry points to avoid recursion. Added `test_lazy_handshake_before_first_ordinary_rpc`.

szbr00 (approved):

PR #1594, opened 2026-08-17
fix(worker): correct sparse buffered block reads

ljy1-lixing:
# Summary

Fix stale data and short reads after a file is truncated down and later grown again.

Curvine Worker reads may need to synthesize a zero-filled sparse tail when the Master-visible logical block length is larger than the bytes currently materialized in the active block file. The affected path trusted the cached `BlockMeta.len` as the physical file length and could continue through `sendfile`, even when the active file was shorter. When the read fell back to the buffered path, `LocalFile::read_region` advanced its internal position twice, causing later chunks to read from the wrong offset or fail with a short physical read.

The combined failure surfaced in the following xfstests:

- `generic/075`: buffered `fsx` coverage.
- `generic/091`: direct I/O with sub-block and buffered I/O coverage.
- `generic/112`: asynchronous I/O variant of the buffered `fsx` coverage.
- `generic/263`: direct I/O mixed with sub-block buffered I/O.

All four tests now pass. 

# Failure Signature

In `generic/091`, operation 4948 read data from a region that had been removed by truncate and later exposed as a hole:

```text
4894 TRUNCATE DOWN 0x79000 -> 0x2f000
4923 TRUNCATE DOWN 0x59000 -> 0xf000
4924 WRITE hole 0x3c000..0x41fff
4937 COPY ... -> 0x3b000..0x3dfff
4940 TRUNCATE UP 0x42000 -> 0x4d000
4948 READ 0x2f000..0x38fff
```

The hole should read as zero, but fsx reported data from an earlier operation:

```text
READ BAD DATA: offset = 0x2f000, size = 0xa000
0x35000 GOOD=0 BAD=operation 63 data
```

The same read path could also fail with a short physical read:

```text
short physical read: expected 131072, got 114688
```

# Minimal Reproducer

Run `generic/091` against a Curvine FUSE mount:

```bash
printf '%s\n' generic/091 > /tmp/curvine-target-091.tests

CURVINE_XFSTESTS_LOCK=/tmp/curvinedevtests-xfstests-fix.lock \
CURVINE_TEST_FS_PATH=/xfstests-fix-test \
CURVINE_SCRATCH_FS_PATH=/xfstests-fix-scratch \
bash /root/ljy/curvinedevtests/xfstests/suites/generic-supported/run.sh \
  --xfstests-root /opt/xfstests \
  --curvine-dist /root/ljy/curvine/build/dist \
  --test-mnt /tmp/curvine-test-fix \
  --scratch-mnt /tmp/curvine-scratch-fix \
  --test-list /tmp/curvine-target-091.tests \
  --timeout 300 \
  --manage-cluster
```

Before the fix, fsx either reported old data in a post-truncate hole or failed with `EIO` after the Worker detected a short physical read.

After the fix:

```text
generic/091      PASS     rc=0   duration=13s
```

# Root Cause

There were two interacting read-path defects.

## Stale metadata was treated as physical file length

`FileLayout::open_reader` used `BlockMeta.len` for both the logical length and the number of bytes physically available from the active block file:

```rust
let physical_len = meta.len;
let logical_len = logical_len.max(physical_len);
```

During truncate, sparse growth, and block-generation publication, the cached Worker metadata can temporarily describe the previous length while the active file already has a different length. A stale-high value caused the read context to request physical bytes beyond the end of the opened file. It could also allow the deferred `sendfile` path to represent bytes that should have been synthesized as zeros.

The opened file descriptor is the authoritative source for physical bytes. Master/client metadata remains the source for the logical block length visible to the caller.

## Buffered reads advanced `LocalFile.pos` twice

`LocalFile::read_full` calls `read_all`, which advances `self.pos`, but `read_region` also advanced `self.pos` after `read_full` returned:

```text
read_region
  -> read_full
     -> read_all
        -> self.pos += bytes_read
  -> self.pos += chunk
```

Each buffered chunk therefore moved the internal device position twice while `BlockReadContext.block_pos` moved once. Subsequent reads used the wrong physical offset and eventually returned stale data, a short chunk, or `EIO`.

# Changes

- Open the active block file before determining physical length and use the opened `LocalFile::len()` as the physical read boundary.
- Keep the client/Master-advertised block length as the logical boundary and synthesize zeros between the physical and logical ends.
- Disable `sendfile` for the entire read session when physical and logical lengths differ, ensuring sparse data is materialized through the buffered zero-fill path.
- Advance `LocalFile.pos` explicitly only for `sendfile`; buffered reads already advance through `read_full` and `read_all`.
- Preserve strict short-read validation and include block position, device position, cached length, and live file length in the error message.
- Stop advertising `FUSE_EXPLICIT_INVAL_DATA` until Curvine issues complete explicit page-cache invalidations for write and resize operations.
- Add regression coverage for stale-high block metadata, sparse zero-filled tails, sendfile eligibility, and single-step buffered position advancement.

# Correctness Notes

The fixed read model separates logical and physical length:

```text
logical length  = length visible to the client
physical length = length obtained from the opened block file

[ physical data ][ zero-filled sparse tail ]
0                                             logical length
```

For a sparse session:

- Reads below the physical boundary come from the opened file.
- Reads crossing the boundary buffer the physical prefix and append zeros.
- Reads entirely above the physical boundary return zeros.
- The block-relative and device positions advance exactly once per physical byte read.
- `sendfile` is used only when the complete logical block is physically materialized.

Removing `FUSE_EXPLICIT_INVAL_DATA` does not disable normal kernel cache coherency. It avoids promising that userspace will explicitly invalidate every affected page while the implementation still relies on the kernel's automatic invalidation behavior.

# Verification

Targeted unit tests:

```text
cargo test -p curvine-io block_device_local_write_read_roundtrip
cargo test -p curvine-storage-local file_reader_uses_open_file_length_when_metadata_is_stale
cargo test -p curvine-storage-local read_context_sparse_logical_tail_returns_zeros
cargo test -p curvine-fuse explicit_data_invalidation_is_not_advertised
```

Release builds:

```text
cargo build --release -p curvine-fuse \
  --no-default-features \
  --features fuse3,curvine-client/opendal-s3,curvine-alloc/jemalloc
cargo build --release -p curvine-server
```

xfstests results:

```text
generic/075      PASS     rc=0   duration=14s
generic/091      PASS     rc=0   duration=13s
generic/112      PASS     rc=0   duration=15s
generic/263      PASS     rc=0   duration=8s
```

Formatting and patch validation:

```text
cargo fmt --all
git diff --check
```

# Scope

This change fixes truncate/sparse-growth data correctness in the Worker and FUSE read path.

### DIFF
--- a/crates/adapters/curvine-storage-local/src/layout/file_layout.rs
+++ b/crates/adapters/curvine-storage-local/src/layout/file_layout.rs
@@ -16,7 +16,7 @@ use crate::layout::{validate_open_offset, BlockLayout};
 use crate::{BlockMeta, BlockState};
 use crate::{BlockReadContext, BlockWriteContext, VfsDir};
 use curvine_core_error::{err_box, try_err, CommonResult};
-use curvine_io::{IOError, IOResult, LocalFile};
+use curvine_io::{IOResult, LocalFile};
 use curvine_model::ExtendedBlock;
 #[cfg(test)]
 use curvine_runtime::common::ByteUnit;
@@ -261,7 +261,12 @@ impl BlockLayout for FileLayout {
         off: i64,
         logical_len: i64,
     ) -> IOResult<BlockReadContext> {
-        let physical_len = meta.len;
+        let file = Self::block_file(dir, meta)?;
+        // Block metadata can briefly retain the pre-resize length while a new
+        // active generation is already published. The opened file is the
+        // source of truth for bytes that can be read physically.
+        let device = LocalFile::with_read(file, 0)?;
+        let physical_len = device.len();
         let logical_len = logical_len.max(physical_len);
         if off < 0 || off > logical_len {
             return err_box!(
@@ -270,12 +275,6 @@ impl BlockLayout for FileLayout {
                 logical_len
             );
         }
-        // Seek within the physical file; sparse logical tail is synthesized.
-        let device_off = off.min(physical_len);
-        let read_off = u64::try_from(device_off)
-            .map_err(|_| IOError::from(format!("Invalid read offset: {}", device_off)))?;
-        let file = Self::block_file(dir, meta)?;
-        let device = LocalFile::with_read(file, read_off)?;
         BlockReadContext::with_physical(device, 0, logical_len, physical_len, off)
     }
 
--- a/crates/adapters/curvine-storage-local/src/layout/file_layout.rs
+++ b/crates/adapters/curvine-storage-local/src/layout/file_layout.rs
@@ -265,9 +265,9 @@ impl BlockLayout for FileLayout {
         // Block metadata can briefly retain the pre-resize length while a new
         // active generation is already published. The opened file is the
         // source of truth for bytes that can be read physically.
-        let device = LocalFile::with_read(file, 0)?;
-        let physical_len = device.len();
-        let logical_len = logical_len.max(physical_len);
+        let inner = OpenOptions::new().read(true).open(&file)?;
+        let device = LocalFile::from_file(&file, inner)?;
+        let physical_len = device.len().min(logical_len);
         if off < 0 || off > logical_len {
             return err_box!(
                 "Invalid block offset: {}, block length: {}",
--- a/crates/adapters/curvine-storage-local/src/block_io_context.rs
+++ b/crates/adapters/curvine-storage-local/src/block_io_context.rs
@@ -284,10 +284,20 @@ impl BlockReadContext {
             .read_region(enable_send_file, physical_chunk as i32)?;
         let got = region.len() as i64;
         if got != physical_chunk {
+            let live_len = self
+                .device
+                .as_local()
+                .and_then(|file| file.metadata().ok())
+                .map(|meta| meta.len() as i64);
             return err_box!(
-                "short physical read: expected {}, got {}",
+                "short physical read: expected {}, got {}, block_pos {}, device_pos {}, physical_len {}, device_len {}, live_len {:?}",
                 physical_chunk,
-                got
+                got,
+                self.block_pos,
+                self.device.pos(),
+                self.physical_len,
+                self.device.len(),
+                live_len
             );
         }
         self.block_pos += got;
@@ -304,7 +314,10 @@ impl BlockReadContext {
     }
 
     pub fn supports_send_file(&self) -> bool {
-        self.device.supports_send_file()
+        // Keep sparse sessions buffered even when the requested region is
+        // currently inside physical bytes. A deferred sendfile can otherwise
+        // race with block-generation replacement and expose pre-truncate data.
+        self.physical_len == self.block_size && self.device.supports_send_file()
     }
 
     pub fn as_local_mut(&mut self) -> Option<&mut LocalFile> {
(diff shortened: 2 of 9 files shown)


crates/adapters/curvine-storage-local/src/layout/file_layout.rs
```diff
@@ -261,7 +261,12 @@ impl BlockLayout for FileLayout {
...
         logical_len: i64,
     ) -> IOResult<BlockReadContext> {
-        let physical_len = meta.len;
+        let file = Self::block_file(dir, meta)?;
+        // Block metadata can briefly retain the pre-resize length while a new
+        // active generation is already published. The opened file is the
+        // source of truth for bytes that can be read physically.
+        let device = LocalFile::with_read(file, 0)?;
+        let physical_len = device.len();
         let logical_len = logical_len.max(physical_len);
```
lzjqsdd: `logical_len.max(physical_len)` expands the client/Master-visible length to the opened file size. That inverts the model in this PR: Master should remain the logical boundary, and extra physical bytes should be ignored.

`BlockReadContext::with_physical` already rejects `physical_len > logical_len`. If the active file is still longer than the advertised block length (truncate-down on Master, or a generation that has not been shrunk yet), this `max` will serve pre-truncate data instead of clamping.

Please clamp instead:

```rust
let physical_len = device.len().min(logical_len);
```

and keep the incoming `logical_len` unchanged. A regression test with `logical_len = 20` against a 50-byte file would lock this in; the current test only covers stale-high metadata with a shorter file.

lzjqsdd (changes requested):
The LocalFile double-advance fix, dropping FUSE_EXPLICIT_INVAL_DATA, and taking physical length from the opened file are the right direction for the reported xfstests failures.

Two stale-metadata gaps remain:

1. FileLayout still expands the client/Master logical length to the opened file size. Please clamp physical length instead so a leftover longer file cannot resurrect truncated data. See the line comment.

2. ReadHandler::open still treats cached BlockMeta.len as physical length when computing BlockReadResponse.len and when allowing short-circuit (`logical_len <= meta.len`). If meta.len is stale-high while the active file is already shorter, a co-located client is still sent onto the short-circuit path, which cannot synthesize a sparse tail. The FileLayout change only helps the remote reader.

I did not find issues in the buffered pos advancement or the init-flag test.

crates/adapters/curvine-storage-local/src/layout/file_layout.rs
ljy1-lixing: Fixed in d8ba6e15. `logical_len` now remains the Master/client boundary and the opened file length is clamped with `device.len().min(logical_len)`. I also added the requested regression test covering a 20-byte logical length backed by a 50-byte file. The reader is now opened without the redundant initial `seek(0)` as well.

ljy1-lixing:
Addressed the remaining stale-metadata gaps in d8ba6e15:

- `ReadHandler::open` now keeps the client/Master length as the response and reader logical boundary instead of expanding it from cached `BlockMeta.len`.
- Short-circuit eligibility now uses the live file length returned with the path; when the active file is shorter than the logical length, the worker falls back to the remote reader so the sparse tail is synthesized correctly.
- `FileLayout` clamps physical bytes to the logical boundary and avoids the redundant initial `seek(0)`.
- Added regression coverage for stale-high metadata/live file length and physical files longer than the logical length.

Validation: all 52 `curvine-storage-local` tests pass, and all PR-related `curvine-worker` tests pass.

lzjqsdd (approved):

PR #1600, opened 2026-08-18
fix(fuse): prevent statfs signed capacity overflow (#1463)

### LINKED ISSUE #1463: [BUG]: FUSE statfs can wrap invalid signed capacity values into huge block counts (by LiAuto)
**Describe the bug**

`CurvineFileSystem::stat_fs()` converts signed capacity values into unsigned FUSE block counts using direct casts:

```rust
let total_blocks = (info.capacity / block_size as i64) as u64;
let free_blocks = (info.available / block_size as i64) as u64;
```

`MasterInfo.capacity` and `MasterInfo.available` are `i64`, while the corresponding `fuse_kstatfs` fields are `u64`.

If a value becomes negative because of malformed data, aggregation overflow, transient state inconsistency, or a compatibility issue, the direct cast wraps it into a very large positive value.

For example, `-1_i64 as u64` becomes `u64::MAX`.

The implementation also does not enforce:

```text
bavail <= bfree <= blocks
```

If `available` exceeds `capacity`, FUSE can report more free blocks than total blocks.

Relevant locations:

- `crates/common/curvine-model/src/master_info.rs`

**To Reproduce**

1. Construct capacity input with `capacity = -4096`.
2. Apply the current byte-to-block conversion.
3. Observe that the resulting block count is close to `u64::MAX`.
4. Construct input with `capacity = 4096` and `available = 8192`.
5. Observe that `bfree` and `bavail` exceed `blocks`.

A focused unit test around an extracted conversion helper is sufficient to reproduce the issue.

**Expected behavior**

Invalid signed values must never wrap into huge unsigned block counts.

The conversion should:

- Convert negative capacity to a defined conservative value or error.
- Convert negative available capacity to a defined conservative value or error.
- Restrict available capacity to total capacity.
- Use checked or saturating arithmetic as appropriate.
- Always satisfy `bavail <= bfree <= blocks`.
- Preserve valid small positive capacity values.
- Define the rounding behavior for values smaller than the 4 KiB fragment size.

- OS: Linux
- Component: `curvine-fuse`

**Additional context**

The 4 KiB fragment size is acceptable, but it should be represented by a constant with explicit statfs semantics rather than an unexplained local expression.

Suggested implementation structure:

```rust
fn capacity_to_statfs(
    capacity: i64,
    available: i64,
    fragment_size: u32,
) -> fuse_kstatfs
```

Alternatively, extract only a helper returning validated block counts.

Acceptance criteria:

- Negative total capacity does not wrap.
- Negative available capacity does not wrap.
- Available capacity is capped at total capacity.
- Returned fields satisfy `bavail <= bfree <= blocks`.
- Zero values are handled.
- Values below one fragment are handled consistently.
- Near-`i64::MAX` values are handled without overflow.
- Unit tests cover negative, zero, inconsistent, normal, and boundary inputs.

gangump82:
# Summary

Prevent FUSE `statfs` from wrapping invalid signed capacity values into huge unsigned block counts. The change also guarantees that reported available/free blocks never exceed total blocks while preserving the existing 4 KiB round-down behavior.

# Issue Describe / Design

Closes #1463

## Root cause

`MasterInfo.capacity` and `MasterInfo.available` are signed `i64` byte counts, but `fuse_kstatfs` stores block counts as `u64`. The previous implementation divided the signed values and then used unchecked `as u64` casts. Negative values therefore wrapped to very large positive block counts, and an inconsistent `available > capacity` response could produce `bavail`/`bfree` greater than `blocks`.

## Reproduction

- `capacity = -4096` produced `(-1_i64) as u64`, a value near `u64::MAX`.
- `capacity = 4096`, `available = 8192` reported more free blocks than total blocks.

## Fix approach

- Define the 4 KiB statfs fragment size as an explicit constant.
- Convert signed values with checked `u64::try_from`, mapping invalid negative inputs to zero.
- Clamp available bytes to validated total capacity before converting to block counts.
- Keep the previous behavior of reporting only complete 4 KiB fragments.
- Emit a conditional `debug!` log with the raw values when invalid master-side filesystem information is normalized.
- Add focused tests for negative, zero, inconsistent, sub-fragment, normal, and `i64::MAX` inputs.

# Changes

| `curvine-fuse/src/fs/curvine_file_system.rs` | Add validated statfs block conversion, boundary tests, and conditional diagnostics for normalized values | Valid non-negative inputs retain existing round-down behavior; invalid inputs are normalized conservatively, logged at debug level, and block-count invariants are enforced |

# Test verified

| `make format` | PASS | Linux container |
| `cargo clippy -p curvine-fuse --lib --tests -- --deny warnings --allow clippy::uninlined-format-args` | PASS | `node4` |
| `cargo fmt --check` + `cargo test -p curvine-fuse statfs_block_counts --lib` | PASS | `node3`; 3 passed |
| `cargo fmt --check` + `cargo test -p curvine-fuse statfs_block_counts --lib` | PASS | `node4`; 3 passed |
| `cargo fmt --check` + `cargo test -p curvine-fuse statfs_block_counts --lib` | PASS | `node5`; 3 passed |
| `cargo test -p curvine-fuse --lib` | BASELINE FAILURE | Both this branch and unmodified `github/main` reproduce the same unrelated parallel metrics assertion in `local_writer_task_body_observes_io_with_local_path_type`; the test passes when run alone |

# Dependencies

- Related PRs: None
- Related issues: #1463
- Blocks / blocked by: None

### DIFF
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -1584,6 +1584,13 @@ impl fs::FileSystem for CurvineFileSystem {
     async fn stat_fs(&self, _: StatFs<'_>) -> FuseResult<fuse_kstatfs> {
         let info = self.fs.get_filesystem_info().await?;
 
+        if info.capacity < 0 || info.available < 0 || info.available > info.capacity {
+            debug!(
+                "Normalizing invalid filesystem info for statfs: capacity={}, available={}",
+                info.capacity, info.available
+            );
+        }
+
         let (total_blocks, free_blocks) = Self::statfs_block_counts(info.capacity, info.available);
 
         let res = fuse_kstatfs {
--- a/curvine-fuse/src/fs/curvine_file_system.rs
+++ b/curvine-fuse/src/fs/curvine_file_system.rs
@@ -49,6 +49,8 @@ use std::ffi::OsStr;
 use std::sync::Arc;
 
 const FUSE_TIME_GRANULARITY_NS: u32 = 1_000_000;
+/// Fragment size used by FUSE statfs block counters.
+const STATFS_FRAGMENT_SIZE: u32 = 4 * ByteUnit::KB as u32;
 
 pub struct CurvineFileSystem {
     fs: UnifiedFileSystem,
@@ -91,6 +93,15 @@ impl CurvineFileSystem {
     fn is_ofd_lock_pid(pid: u32) -> bool {
         // FUSE kernels have used both 0 and -1 for the pid of an OFD lock.
         pid == 0 || pid == u32::MAX
+    }
+
+    fn statfs_block_counts(capacity: i64, available: i64) -> (u64, u64) {
+        let capacity = u64::try_from(capacity).unwrap_or(0);
+        let available = u64::try_from(available).unwrap_or(0).min(capacity);
+        let fragment_size = u64::from(STATFS_FRAGMENT_SIZE);
+
+        // Report only complete fragments, preserving the existing rounding behavior.
+        (capacity / fragment_size, available / fragment_size)
     }
 
     pub fn new(conf: ClusterConf, rt: Arc<Runtime>) -> FuseResult<Self> {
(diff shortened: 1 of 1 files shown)


curvine-fuse/src/fs/curvine_file_system.rs
```diff
@@ -93,6 +95,15 @@ impl CurvineFileSystem {
         pid == 0 || pid == u32::MAX
     }
 
+    fn statfs_block_counts(capacity: i64, available: i64) -> (u64, u64) {
+        let capacity = u64::try_from(capacity).unwrap_or(0);
+        let available = u64::try_from(available).unwrap_or(0).min(capacity);
```
szbr00, line 100: `statfs_block_counts` silently normalizes negative capacity/available and clamps `available > capacity`. Since these cases indicate an inconsistent master-side `FilesystemInfo`, consider emitting a `debug!` (or `warn!` once per occurrence) log when normalization triggers, so the underlying inconsistency is observable instead of hidden behind corrected block counts. Optional — the defensive clamping itself is correct.

gangump82: Thanks. Fixed in 78a13c70: invalid master-side filesystem capacity values are now logged at debug level before statfs normalization, while the conversion helper remains pure.

szbr00 (approved):

PR #1603, opened 2026-08-18
feat(spdk): introduce spdk qpair observability metrics

### LINKED ISSUE #1076: [BUG]: 4k random reads fail with EIO under SPDK (QpairPool exhaustion via read-ahead fan-out)

thuongle2210:
# Summary

- Adds aggregate qpair observability metrics to track acquisition, release, and contention patterns across the SPDK environment.
- Resolves #1076

# Key Changes

- crates/adapters/curvine-storage-spdk/src/spdk_metrics.rs
   + Qpair acquisition counters: `cached`, `allocated`, `timeout`, `shutdown`, `alloc_failed`
   + Qpair release counters: `cached`, `freed_pool_full`
   + Acquisition wait-time histogram (100µs minimum bucket)
   + Aggregate gauges: active, configured, cached qpairs
   + Contention counter and pool shutdown tracking
   + Lazy metric registration with graceful failure handling

- crates/adapters/curvine-storage-spdk/src/spdk_env.rs
   + Record acquisition outcomes on all success/failure paths
   + Accumulate wait duration per acquisition attempt (≥100µs recorded)
   + Track contention on controller capacity blocking
   + Maintain aggregate active, limit, and cached qpair counts
   + Record release outcomes and shutdown drain operations
   + Publish current gauges for metric collection
   + Added qpair metrics validation tests

- curvine-worker/src/worker/worker_metrics.rs
   + Publish SPDK qpair gauges before generating worker metric output
   + Expose metrics via existing Prometheus endpoint

# Impact

- Enables diagnosis of qpair exhaustion and tuning of `io_queues` and `qpair_acquire_timeout`
- Aggregate metrics per worker process across all controllers
- Minimal overhead to qpair acquisition/release paths
- Foundation for future per-controller metrics

# Future Works
- Add per-controller metrics to identify saturated NVMe targets.

### DIFF
--- a/crates/adapters/curvine-storage-spdk/src/spdk_env.rs
+++ b/crates/adapters/curvine-storage-spdk/src/spdk_env.rs
@@ -1,6 +1,7 @@
 #![cfg(feature = "spdk")]
 
 use crate::spdk_ffi;
+use crate::spdk_metrics;
 use crate::spdk_poller::{CtrlHandle, PollerConfig};
 use crate::spdk_poller::{IoRequest, SpdkPoller};
 use curvine_core_error::{err_box, err_msg, CommonResult};
@@ -39,6 +40,12 @@ pub struct QpairPool {
     pub(crate) inner: Mutex<HashMap<usize, Vec<*mut spdk_ffi::spdk_nvme_qpair>>>,
     /// Per-controller active count and max limit, keyed by controller pointer.
     pub(crate) ctrl_state: Mutex<HashMap<usize, CtrlQpairState>>,
+    /// Aggregate active qpair reservations across controllers.
+    pub(crate) total_active: AtomicUsize,
+    /// Aggregate qpair reservation limit across controllers.
+    pub(crate) total_limit: AtomicUsize,
+    /// Aggregate cached idle qpairs across controllers.
+    pub(crate) total_cached: AtomicUsize,
     /// Idle cache limit per controller (soft — excess freed on release).
     pub(crate) max_per_ctrlr: usize,
     /// Wakes blocked acquirers when a qpair is released.
@@ -54,6 +61,9 @@ unsafe impl Sync for QpairPool {}
 impl QpairPool {
     /// Default max idle qpairs per controller (soft cache limit).
     const DEFAULT_MAX_PER_CTRLR: usize = 16;
+    /// Minimum qpair acquire wait recorded in the histogram. Tiny waits are skipped
+    /// to keep observability overhead low on the qpair contention path.
+    const QPAIR_WAIT_METRIC_MIN: Duration = Duration::from_micros(100);
     /// Default qpair acquire timeout, used when SpdkConf does not set `qpair_acquire_timeout`.
     #[cfg(test)]
     const DEFAULT_QPAIR_ACQUIRE_TIMEOUT: Duration = Duration::from_secs(30);
@@ -67,6 +77,9 @@ impl QpairPool {
         Self {
             inner: Mutex::new(HashMap::new()),
             ctrl_state: Mutex::new(HashMap::new()),
+            total_active: AtomicUsize::new(0),
+            total_limit: AtomicUsize::new(0),
+            total_cached: AtomicUsize::new(0),
             max_per_ctrlr: Self::DEFAULT_MAX_PER_CTRLR,
             notify: Condvar::new(),
             shutdown: AtomicBool::new(false),
@@ -85,6 +98,8 @@ impl QpairPool {
             ctrlr_ptr as *const ()
         );
         state.insert(ctrlr_ptr, CtrlQpairState::new(limit));
+        let total_limit = self.total_limit.fetch_add(limit, Ordering::AcqRel) + limit;
+        spdk_metrics::set_qpair_limit(total_limit);
         if limit == 0 {
             warn!(
                 "QpairPool: ctrlr {:p} registered with max_active=0 (0 negotiated IO queues)",
@@ -95,6 +110,12 @@ impl QpairPool {
                 "QpairPool: registered ctrlr {:p} with max_active={}",
                 ctrlr_ptr as *const (), limit
             );
+        }
+    }
+
+    fn record_qpair_wait_if_needed(wait: Duration) {
+        if wait >= Self::QPAIR_WAIT_METRIC_MIN {
+            spdk_metrics::record_qpair_wait(wait);
         }
     }
 
@@ -123,6 +144,7 @@ impl QpairPool {
                     .compare_exchange_weak(cur, cur + 1, Ordering::AcqRel, Ordering::Relaxed)
                     .is_ok()
                 {
+                    self.total_active.fetch_add(1, Ordering::AcqRel);
                     return true;
                 }
             }
@@ -150,6 +172,7 @@ impl QpairPool {
                 return;
             }
             s.active.fetch_sub(1, Ordering::AcqRel);
+            self.total_active.fetch_sub(1, Ordering::AcqRel);
         } else {
             warn!(
                 "QpairPool: release_reservation for unregistered ctrlr {:p} \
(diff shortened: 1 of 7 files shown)


thuongle2210:
Hi @szbr9486 , @lzjqsdd , could you please help me review this PR? I've recently added some observability metrics for qpair spdk.

crates/adapters/curvine-storage-spdk/src/spdk_metrics.rs
```diff
@@ -0,0 +1,138 @@
...
+pub(crate) fn record_qpair_acquire(result: &'static str) {
+    with_metrics(|m| match result {
+        "cached" => m.qpair_acquire_cached.inc(),
+        "allocated" => m.qpair_acquire_allocated.inc(),
+        "timeout" => m.qpair_acquire_timeout.inc(),
+        "shutdown" => m.qpair_acquire_shutdown.inc(),
+        "alloc_failed" => m.qpair_acquire_alloc_failed.inc(),
+        _ => {}
+    });
+}
```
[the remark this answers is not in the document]
thuongle2210: Thanks, I added the warning log.

crates/adapters/curvine-storage-spdk/src/spdk_metrics.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub(crate) fn set_qpair_cached(value: usize) {
+    with_metrics(|m| m.qpair_cached.set(value as i64));
+}
+
+pub(crate) fn record_qpair_release(result: &'static str) {
+    with_metrics(|m| match result {
+        "cached" => m.qpair_release_cached.inc(),
+        "freed_pool_full" => m.qpair_release_freed_pool_full.inc(),
+        _ => {}
```
szbr00: `record_qpair_release` silently drops unknown `result` labels (`_ => {}`), while `record_qpair_acquire` was updated in the follow-up commit to log a warning for unknown labels. The Copilot review flagged exactly this asymmetry and the reply indicated the warning was added, but it only landed in `record_qpair_acquire`. A future release outcome added without updating this match would be invisible. Apply the same `warn!` arm here so a mismatched label is caught early instead of being silently lost.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -292,23 +341,29 @@ impl QpairPool {
...
+        }; // inner lock dropped here
+        spdk_metrics::record_qpair_release(release_result);
         self.release_reservation(key);
         self.notify.notify_one(); // safe to call without holding inner lock
     }
     /// Free all pooled qpairs. Only frees cached (idle) qpairs - active/in-flight
     /// qpairs are tracked by their owners and will be released normally.
     pub(crate) fn drain_all(&self) {
         self.shutdown.store(true, Ordering::Release);
+        spdk_metrics::record_qpair_shutdown();
```
szbr00: `drain_all` is documented as safe to call multiple times, but each invocation increments `spdk_qpair_shutdown_total` and re-runs the drain loop. If a caller ever invokes it twice (e.g. a graceful drain plus a forced shutdown path), the counter over-counts and the second drain does redundant work. Consider early-returning when `self.shutdown` is already true, or noting in the doc comment that the counter reflects `drain_all` invocations rather than distinct shutdown events.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -172,24 +195,31 @@ impl QpairPool {
...
+                spdk_metrics::record_qpair_acquire("cached");
+                log::trace!(
+                    "QpairPool: reusing cached qpair for ctrlr {:p} (pool size now {})",
+                    ctrlr,
+                    stack_len
+                );
+                return Ok(qpair);
             }
             // No cached qpair - fall through to slow path with slot already reserved
         } else {
```
szbr00: `record_qpair_contention` is called whenever `try_reserve` returns false, but `try_reserve` also returns false for an unregistered controller (error path), not only at capacity. That inflates `spdk_qpair_contention_total` on a misconfiguration bug rather than real capacity contention, which is the signal this metric is meant to surface. Consider gating the contention counter on the "at capacity" case only (e.g. have `try_reserve` distinguish the two outcomes, or check `ctrl_state` membership before recording contention).

lzjqsdd (approved):

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
```diff
@@ -1344,6 +1420,31 @@ mod test {
         assert_eq!(active, 1);
```
szbr00, line 1477: This test only asserts that metric *names* appear in the text output, not their
values. For the counters (`spdk_qpair_acquire_total{result="timeout"}`,
`spdk_qpair_contention_total`, `spdk_qpair_shutdown_total`) a series only appears
once it is incremented, so presence implies the recording path ran. The gauges
(`spdk_qpair_active`, `spdk_qpair_limit`, `spdk_qpair_cached`) are different:
they are emitted with value 0 as soon as they are registered, so
`output.contains("spdk_qpair_active")` passes even if `publish_metrics` or the
underlying `total_active`/`total_limit`/`total_cached` accounting is wrong.

Given this PR's core value is correct aggregate accounting, assert concrete
values, e.g. after `register_limit(1)` + `try_reserve` + `acquire(timeout)` +
`drain_all` + `publish_metrics`:

    assert!(output.contains("spdk_qpair_active 1"));
    assert!(output.contains("spdk_qpair_limit 1"));
    assert!(output.contains("spdk_qpair_cached 0"));
    assert!(output.contains("spdk_qpair_contention_total 1"));
    assert!(output.contains("spdk_qpair_acquire_total{result=\"timeout\"} 1"));

That would catch a regression where the atomics drift from the real qpair
counts. Note the existing `try_reserve` call leaves a reservation unreleased
(total_active stays 1), which is fine for the test as long as the expected
`spdk_qpair_active 1` value is asserted.

thuongle2210: Thanks, agreed on validating concrete values. I added exact assertions for the aggregate gauges (active=1, limit=1, cached=0). However, counter values (`spdk_qpair_contention_total`) are checked by presence only because they are process-global value and can be incremented by parallel tests, making exact values flaky.

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
thuongle2210: Thanks, I treated the output of try_reserve as a QpairReserveResult, and track total contention only in the AtCapacity case.

crates/adapters/curvine-storage-spdk/src/spdk_metrics.rs
thuongle2210: I added the warning log. Thanks!

crates/adapters/curvine-storage-spdk/src/spdk_env.rs
thuongle2210: I added the checking that only the first shutdown transition is counted.

szbr00 (approved):

PR #1604, opened 2026-08-19
feat(transfer): add PostgreSQL metadata store

jlon:
# Summary

Add PostgreSQL as a production metadata-store backend for the Transfer service. The backend is selected automatically from `transfer.store_url`, so existing SQLite and MySQL configurations remain unchanged.

# Issue Describe / Design

This PR completes the `TransferStore` contract for PostgreSQL instead of introducing a second Transfer workflow. PostgreSQL uses its native Rust driver with an r2d2 connection pool, explicit transactions, row locks, and a transaction-scoped advisory lock for schema initialization.

The PostgreSQL implementation preserves the MySQL behavior for idempotent submission, target-path conflict detection, execution leases, task planning and reporting, retries, cancellation, recovery, and terminal-job cleanup. PostgreSQL path columns use bytewise `C` collation so path range checks follow Curvine path semantics rather than the database locale.

No new user-facing store-type parameter is required. The backend is inferred from the URL scheme:

```toml
[transfer]
store_url = "postgresql://user:password@postgres-host:5432/curvine_transfer?sslmode=require"
```

TLS always validates certificates and hostnames. Public CAs use the system trust store. For a private PostgreSQL CA, add the CA PEM file through the standard URL parameter:

```toml
[transfer]
store_url = "postgresql://user:password@postgres-host:5432/curvine_transfer?sslmode=require&sslrootcert=/etc/curvine/postgres-ca.pem"
```

# Changes

| `curvine-config` | Recognize `postgres://` and `postgresql://` and infer the PostgreSQL backend from `store_url` | Existing SQLite and MySQL configurations are unchanged |
| `curvine-data-transfer` | Add the complete PostgreSQL `TransferStore` implementation and backend dispatch | Enables PostgreSQL-backed Transfer deployments |
| `curvine-data-transfer` | Add PostgreSQL store metrics and unavailable-state reporting | Adds the `postgres` backend label; existing metrics remain unchanged |
| `curvine-data-transfer` | Accept `sslrootcert` in a PostgreSQL URL and add its PEM CA to the system trust set | Supports private-CA PostgreSQL without weakening certificate or hostname validation |
| `curvine-data-transfer/tests` | Add lifecycle, persistence, path-conflict, concurrent-submission, and private-CA TLS coverage | No production behavior change |
| `Cargo.toml` / `Cargo.lock` | Add PostgreSQL, TLS, r2d2, and URL parsing dependencies | Only the Transfer server feature includes the backend |
| `etc/curvine-cluster.toml` | Add PostgreSQL and private-CA URL configuration examples | Documentation/comment only |

# Test verified

| `cargo test -p curvine-config` | PASS | 73 tests |
| `cargo test -p curvine-data-transfer` | PASS | 10 tests |
| `cargo test -p curvine-data-transfer --features transfer-server` | PASS | 17 tests |/
| `CURVINE_TRANSFER_POSTGRES_URL=... cargo test -p curvine-data-transfer --features transfer-server --test postgres_store_test -- --nocapture` | PASS | Tested against a PostgreSQL 18 container; lifecycle, reopen persistence, path conflicts, and concurrent submissions |
| `CURVINE_TRANSFER_POSTGRES_TLS_URL=... cargo test -p curvine-data-transfer --features transfer-server --test postgres_store_test postgres_store_connects_with_a_private_ca -- --exact` | PASS | `sslmode=require` connection to PostgreSQL with a private-CA certificate |
| `cargo clippy -p curvine-data-transfer --features transfer-server --all-targets -- -D warnings` | PASS | No warnings |
| `cargo check -p curvine-server` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

The repository-wide `--all-features` build still depends on the environment-provided Jindo `jdo_api.h`; that external dependency is unrelated to this PR.


crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1320 @@
...
+
+fn json<T: serde::Serialize>(value: &T) -> FsResult<String> {
+    serde_json::to_string(value).map_err(json_err)
+}
+
+fn json_err(_: serde_json::Error) -> FsError {
+    FsError::common("Transfer metadata store contains invalid data")
+}
+
+fn postgres_err(err: postgres::Error) -> FsError {
```
szbr00, line 722: `postgres_err` maps every `postgres::Error` to `transfer_store_unavailable`. Because `backend.rs::is_store_unavailable_error` treats `TransferStoreUnavailable` as an availability signal, transient failures such as unique-constraint races, serialization/deadlock failures from the submission gate, or statement timeouts will all increment `store_unavailable_total` and light up unavailable alerts even though the database is healthy.

Consider distinguishing connectivity/pool errors (which warrant "unavailable") from per-statement errors like constraint violations or serialization failures (which should surface as a normal `FsError` without flipping the unavailable gauge). At minimum, matching on `postgres::Error` variants (e.g. `Db` with specific SQLSTATE codes vs `Connect`/`Io`) would let only connection-class failures report unavailable.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1320 @@
...
+}
+
+impl PostgresTransferStore {
+    pub fn open(url: &str) -> FsResult<Self> {
+        let config = postgres::Config::from_str(url)
+            .map_err(|_| FsError::common("Invalid PostgreSQL transfer store URL"))?;
+        let tls = TlsConnector::new()
+            .map_err(|_| FsError::common("Unable to initialize PostgreSQL TLS support"))?;
+        let manager = PostgresConnectionManager::new(config, MakeTlsConnector::new(tls));
+        let pool = Pool::builder().build(manager).map_err(postgres_pool_err)?;
```
szbr00, line 53: `Pool::builder().build(manager)` uses r2d2 defaults (max 10 connections, min idle 1). The transfer server issues store calls from multiple scheduler/dispatcher/report lanes, so 10 connections can become the bottleneck under load and requests will block waiting for a pooled connection.

Consider exposing pool size (and optionally min_idle / timeout) via `transfer_conf`, or at least bumping the default `max_size` to something aligned with `max_running_transfers` / `worker_dispatch_concurrency`. This also makes the pool tunable for production deployments without code changes.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1320 @@
...
+                   and stale_deadline_at < $3
+                 order by stale_deadline_at asc
+                 limit $4",
+                &[&job_id, &run_id, &stale_before_ms, &limit],
+            )
+            .map_err(postgres_err)?;
+        rows.iter().map(task_from_row).collect()
+    }
+
+    fn list_tasks_by_state(
```
szbr00, line 1333: `list_tasks_by_state` is defined on `PostgresTransferStore` but is not part of the `TransferStore` trait and has no callers in this PR. Either add it to the trait for parity with the other backends (if callers are planned) or remove it to keep the surface minimal.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1320 @@
...
+use r2d2::{Pool, PooledConnection};
+use r2d2_postgres::PostgresConnectionManager;
+use std::str::FromStr;
+
+use crate::transfer::{
+    apply_task_report_progress, TransferPlannedTasks, TransferRequeueUpdate, TransferStore,
+    TransferTaskStateUpdate,
+};
+
+const TRANSFER_SCHEMA_VERSION: i64 = 1;
```
szbr00, line 38: `TRANSFER_SCHEMA_VERSION` is pinned to 1 and `init_schema` only runs `create table if not exists` / `create index if not exists`. The MySQL backend carries `migrate_mysql_schema_v2_to_v3` / `v3_to_v4` for additive column and index changes; the Postgres backend has no equivalent path.

When the next schema change lands (e.g. a new column on `transfer_jobs`), `create table if not exists` will be a no-op against existing tables and the new column will be missing, silently breaking reads/writes. Consider adding a migration runner keyed off `transfer_schema_version` now (even if it only handles v1 -> v2 stubs) so future changes have a place to go, and so the version check rejects older/newer binaries with a clear message.

crates/server/curvine-data-transfer/tests/postgres_store_test.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+fn test_store_url() -> Option<(String, TestSchema)> {
+    let url = std::env::var("CURVINE_TRANSFER_POSTGRES_URL").ok()?;
+    let name = format!(
+        "curvine_transfer_test_{}_{}",
+        std::process::id(),
+        curvine_runtime::common::LocalTime::mills()
+    );
+    let mut client = Client::connect(&url, NoTls).unwrap();
```
szbr00: The schema setup/drop uses `Client::connect(&url, NoTls)`. If `CURVINE_TRANSFER_POSTGRES_URL` is configured with `sslmode=require` (as the `etc/curvine-cluster.toml` example suggests for production), the admin connection will fail and the test will panic before the store-under-test (which uses `MakeTlsConnector`) gets a chance to run.

Consider reusing the same `MakeTlsConnector` setup (or `postgres::Config::from_str(url)` + `MakeTlsConnector`) for the admin client so the test works against TLS-only servers.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
jlon: Kept the current pool defaults for this PR. MySQL has no exposed pool-size setting today, and there is no production metric demonstrating that PostgreSQL's default pool is the current bottleneck. Adding another Transfer configuration parameter without that evidence would expand the configuration surface. We can tune this from observed pool-wait metrics if production data shows a need.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
jlon: Addressed in da8a9c54. PostgreSQL errors are now classified by connection state and SQLSTATE. Connection, I/O, TLS, shutdown, and closed-connection errors still return `TransferStoreUnavailable`; statement-level errors now return a readable `Common` error and do not mark the backend unavailable.

crates/server/curvine-data-transfer/tests/postgres_store_test.rs
jlon: Addressed in da8a9c54. The integration test now creates and drops its test schema through `postgres::Config` plus the same `MakeTlsConnector` setup used by the production store, so `sslmode=require` is supported by the test harness.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
jlon: Kept the current version check without adding an empty migration framework. PostgreSQL support starts with a new schema at version 1; if the stored version differs, startup already fails with an explicit unsupported-schema error. Any future schema change must add a real migration and bump the version before changing the expected schema. Adding a v1-to-v2 stub now would add surface without a migration to execute.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
jlon: No change needed. `list_tasks_by_state` is already part of the existing `TransferStore` trait in `transfer/store.rs`, and the Memory, SQLite, and MySQL backends implement the same method. PostgreSQL must implement it to satisfy the existing storage contract, even though the current scheduler does not call it directly.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1351 @@
...
+
+pub struct PostgresTransferStore {
+    pool: Pool<PgManager>,
+}
+
+impl PostgresTransferStore {
+    pub fn open(url: &str) -> FsResult<Self> {
+        let config = postgres::Config::from_str(url)
+            .map_err(|_| FsError::common("Invalid PostgreSQL transfer store URL"))?;
+        let tls = TlsConnector::new()
```
szbr00: `PostgresTransferStore::open` always builds `native_tls::TlsConnector::new()`, which verifies the server certificate against the system trust store. rust-postgres `sslmode` only chooses whether TLS is used; it does not implement libpq require vs verify-ca vs verify-full policy.

The documented production URL in `etc/curvine-cluster.toml` uses `sslmode=require`. Under libpq that encrypts without CA verification. Here the same URL fails against a private-CA or self-signed server (common for in-cluster Postgres), and there is no `sslrootcert` / custom-CA path.

Either document that this backend requires a system-trusted cert (so `sslmode=require` is effectively verify-full against the platform CAs), or honor private PKI — for example load a PEM from the URL/`sslrootcert`, or provide an explicit insecure-skip setting for lab use.

jlon: Addressed in f26f5bb0. PostgreSQL store_url now accepts sslrootcert as a PEM CA file. The connector adds it to the system trust set while certificate and hostname verification remain enabled; no insecure-skip option was added. Other query parameters retain their original percent encoding. Added an optional TLS integration test and verified a PostgreSQL container signed by a private CA.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1433 @@
...
+            FsError::common(format!(
+                "PostgreSQL root certificate at '{}' is not valid PEM",
+                path.display()
+            ))
+        })?;
+        builder.add_root_certificate(certificate);
+    }
+    builder
+        .build()
+        .map_err(|_| FsError::common("Unable to initialize PostgreSQL TLS support"))
```
szbr00, line 153: `postgres_tls_connector` always constructs a `TlsConnector` and, when
`sslrootcert` is set, reads and parses the PEM file. When the URL carries
`sslmode=disable`, rust-postgres will not negotiate TLS at all, so this work
is unused on that path. Consider short-circuiting in `postgres_config`/`open`
when `sslmode=disable` is present (or documenting that the cert is loaded
unconditionally and ignored by rust-postgres on the disable path).

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
```diff
@@ -0,0 +1,1433 @@
...
+    }
+}
+
+fn postgres_error_is_unavailable(err: &postgres::Error) -> bool {
+    err.is_closed()
+        || err.code().is_some_and(is_postgres_connectivity_error)
+        || err
+            .source()
+            .is_some_and(|source| source.is::<std::io::Error>() || source.is::<native_tls::Error>())
+}
```
szbr00, line 753: `postgres_error_is_unavailable` classifies any `native_tls::Error` source as
`transfer_store_unavailable`. A persistent TLS failure (expired server cert,
untrusted CA, hostname mismatch) is not a transient connectivity issue — it
will keep failing until the cert/config or process is fixed — but it latches
the `transfer_store_unavailable` gauge on and increments
`transfer_store_unavailable_total` as if it were a recoverable flap. Consider
surfacing TLS/cert verification errors as a distinct, non-transient
classification so operators can distinguish "DB is down" from "TLS is
misconfigured".

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
jlon: Addressed in fd94eee7. After removing sslrootcert from the PostgreSQL URL, the parsed SslMode is checked. With sslmode=disable, the store ignores the CA path and does not read the file. Added an integration test using a missing CA path with TLS disabled.

crates/server/curvine-data-transfer/src/transfer/postgres_store.rs
jlon: Partially addressed in fd94eee7. TLS handshake failures remain TransferStoreUnavailable because the backend cannot operate and the scheduler/readiness checks must pause rather than fail individual jobs. The returned error now identifies a PostgreSQL TLS handshake failure and directs operators to verify the certificate, hostname, and sslrootcert. No new error kind or metric label was added; the existing unavailable metric correctly reflects the backend state.

szbr00 (approved):

PR #1605, opened 2026-08-19
fix(rpc): handle early EOF gracefully in ReadFrame::receive

bigbigxu:
# Summary

`ReadFrame::receive()` propagated tokio's `UnexpectedEof` error as an ERROR log when the remote peer closed the TCP connection before sending a complete RPC frame header. This is normal behaviour in a distributed system (health checks, client disconnects, connection timeouts, etc.) and should not produce error logs. Align `ReadFrame` with `RpcFrame::receive()` which already catches EOF at the head-reading stage and returns `Message::empty()` for a graceful exit.

# Changes

| `crates/core/rpc/src/handler/read_frame.rs` | Catch `read_full` errors in the `FrameSate::Head` branch and return `Ok(Message::empty())` instead of propagating with `?` | Connections that close before sending a frame are now treated as graceful disconnects (same as `RpcFrame`); no more spurious ERROR logs |

# Test verified

| Existing RPC handler tests | Expected PASS | No behavioral change for valid connections |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/core/rpc/src/handler/read_frame.rs
+++ b/crates/core/rpc/src/handler/read_frame.rs
@@ -16,7 +16,6 @@ use std::io;
 use std::mem;
 
 use bytes::BytesMut;
-use log::debug;
 use tokio::io::{AsyncReadExt, ReadHalf};
 use tokio::net::TcpStream;
 
@@ -59,10 +58,7 @@ impl ReadFrame {
                         Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => {
                             return Ok(Message::empty());
                         }
-                        Err(e) => {
-                            debug!("ReadFrame head read error: {}", e);
-                            return Err(e);
-                        }
+                        Err(e) => return Err(e),
                     };
 
                     let (protocol, header_size, data_size) = Message::decode_protocol(&mut buf)?;
@@ -116,11 +112,13 @@ mod tests {
         let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
         let addr = listener.local_addr().unwrap();
 
-        // Connect and immediately drop the connection.
-        let stream = TcpStream::connect(addr).await.unwrap();
-        drop(stream);
-
+        // Accept first, then drop the peer side — matches the production
+        // "peer closed an already-accepted socket" path and avoids the
+        // reset/EOF race on some kernels.
+        let client = TcpStream::connect(addr).await.unwrap();
         let (server_stream, _) = listener.accept().await.unwrap();
+        drop(client);
+
         let (read_half, _) = tokio::io::split(server_stream);
         let mut read_frame = ReadFrame::new(read_half, FrameBuf::new(0));
 
--- a/crates/core/rpc/src/handler/read_frame.rs
+++ b/crates/core/rpc/src/handler/read_frame.rs
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::io;
 use std::mem;
 
 use bytes::BytesMut;
+use log::debug;
 use tokio::io::{AsyncReadExt, ReadHalf};
 use tokio::net::TcpStream;
 
@@ -54,7 +56,13 @@ impl ReadFrame {
                 FrameSate::Head => {
                     let mut buf = match self.read_full(message::PROTOCOL_SIZE).await {
                         Ok(v) => v,
-                        Err(_) => return Ok(Message::empty()),
+                        Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => {
+                            return Ok(Message::empty());
+                        }
+                        Err(e) => {
+                            debug!("ReadFrame head read error: {}", e);
+                            return Err(e);
+                        }
                     };
 
                     let (protocol, header_size, data_size) = Message::decode_protocol(&mut buf)?;
@@ -95,5 +103,28 @@ impl ReadFrame {
                 }
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::net::TcpListener;
+
+    #[tokio::test]
+    async fn receive_returns_empty_on_peer_close_before_sending() {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+
+        // Connect and immediately drop the connection.
+        let stream = TcpStream::connect(addr).await.unwrap();
+        drop(stream);
+
+        let (server_stream, _) = listener.accept().await.unwrap();
+        let (read_half, _) = tokio::io::split(server_stream);
+        let mut read_frame = ReadFrame::new(read_half, FrameBuf::new(0));
+
+        let msg = read_frame.receive().await.unwrap();
+        assert!(msg.is_empty(), "expected empty message on peer EOF");
     }
 }
--- a/crates/core/rpc/src/handler/read_frame.rs
+++ b/crates/core/rpc/src/handler/read_frame.rs
@@ -52,7 +52,10 @@ impl ReadFrame {
         loop {
             match state {
                 FrameSate::Head => {
-                    let mut buf = self.read_full(message::PROTOCOL_SIZE).await?;
+                    let mut buf = match self.read_full(message::PROTOCOL_SIZE).await {
+                        Ok(v) => v,
+                        Err(_) => return Ok(Message::empty()),
+                    };
 
                     let (protocol, header_size, data_size) = Message::decode_protocol(&mut buf)?;
                     let _ = mem::replace(
--- a/crates/core/rpc/src/handler/rpc_frame.rs
+++ b/crates/core/rpc/src/handler/rpc_frame.rs
@@ -21,6 +21,8 @@ use curvine_io::{DataSlice, IOResult};
 use curvine_net::net::ConnState;
 use curvine_sys as sys;
 use curvine_sys::RawIOSlice;
+use log::debug;
+use std::io;
 use std::mem;
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
 use tokio::io::{AsyncReadExt, AsyncWriteExt, Interest};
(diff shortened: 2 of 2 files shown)


crates/core/rpc/src/handler/read_frame.rs
```diff
@@ -52,7 +52,10 @@ impl ReadFrame {
         loop {
             match state {
                 FrameSate::Head => {
-                    let mut buf = self.read_full(message::PROTOCOL_SIZE).await?;
+                    let mut buf = match self.read_full(message::PROTOCOL_SIZE).await {
```
szbr00, line 56: The `Err(_)` arm catches every error from `read_full`, not just `UnexpectedEof`. While this mirrors `RpcFrame::receive()`, `ReadFrame` is used on the client side (`buffer_client.rs::response_future`), where the impact differs: when a non-EOF error occurs at the head stage (e.g. connection reset), `response_future` breaks via `msg.is_empty()` without calling `client_state.set_error`. As a result, `check_calls` reports `take_error() == None` to inflight callers, so they lose the real I/O cause and only see a generic "connection ...: None" message.

Consider narrowing this to `UnexpectedEof` (or at least logging the swallowed error at debug level) so genuine I/O failures remain diagnosable on the client path while still suppressing noise for graceful disconnects.

szbr00, line 56: The PR description lists only "Existing RPC handler tests | Expected PASS", but no test exercises the new behavior (peer closes the connection before sending a full protocol header -> `receive()` returns `Ok(Message::empty())` instead of `Err`). A small regression test that opens a `TcpStream`, drops it immediately, and asserts `receive()` yields an empty message would lock in this behavior and prevent a future refactor from reintroducing the noisy ERROR log.

crates/core/rpc/src/handler/read_frame.rs
```diff
@@ -52,7 +52,10 @@ impl ReadFrame {
...
+                        Ok(v) => v,
+                        Err(_) => return Ok(Message::empty()),
+                    };
```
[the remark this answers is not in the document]
bigbigxu: Good point. Narrowed to `io::ErrorKind::UnexpectedEof` only — real I/O errors (connection reset, permission errors, etc.) are now propagated as before. Also applied the same narrowing to `RpcFrame::receive()` for consistency.

crates/core/rpc/src/handler/read_frame.rs
bigbigxu: Added regression test `receive_returns_empty_on_peer_close_before_sending` — binds a `TcpListener`, connects and immediately drops the stream, then asserts `ReadFrame::receive()` yields `Message::empty()`.

bigbigxu: Fixed. Narrowed to `io::ErrorKind::UnexpectedEof` only. Real I/O errors (e.g. connection reset) are now propagated via `?` so that `response_future` calls `client_state.set_error(e)` and `check_calls` reports the actual cause to inflight callers. Also added `debug!` logging for the non-EOF path. Applied the same narrowing to `RpcFrame::receive()` for consistency.

crates/core/rpc/src/handler/read_frame.rs
```diff
@@ -94,3 +105,26 @@ impl ReadFrame {
...
+mod tests {
+    use super::*;
+    use tokio::net::TcpListener;
+
+    #[tokio::test]
+    async fn receive_returns_empty_on_peer_close_before_sending() {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+
+        // Connect and immediately drop the connection.
```
szbr00: This test drops the client before `accept()`. On some kernels a connection closed while still in the listen queue is surfaced as ConnectionReset on the first read, not UnexpectedEof, so the assertion would fail after the kind filter.

Accept first, then drop the peer, then call `receive()`. That matches the production "peer closed an already-accepted socket" path and avoids the reset/EOF race.

crates/core/rpc/src/handler/read_frame.rs
```diff
@@ -52,7 +54,16 @@ impl ReadFrame {
         loop {
             match state {
                 FrameSate::Head => {
-                    let mut buf = self.read_full(message::PROTOCOL_SIZE).await?;
+                    let mut buf = match self.read_full(message::PROTOCOL_SIZE).await {
+                        Ok(v) => v,
+                        Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => {
```
szbr00, line 58: `read_exact` reports UnexpectedEof both for a 0-byte idle close and for a truncated protocol header (peer wrote 1..PROTOCOL_SIZE-1 bytes, then closed). The second case is a broken frame, not a graceful disconnect.

On the client path, `buffer_client::response_future` treats Message::empty() as a clean break and does not call `client_state.set_error`. Inflight RPCs then fail in `check_calls` with `take_error() == None` (`"connection ...: None"`), while RpcClient::rpc already maps empty to an explicit "Connection is closed" error.

Consider treating only a 0-byte EOF as Message::empty(), and propagating a mid-header UnexpectedEof as Err so inflight callers keep a real cause.

crates/core/rpc/src/handler/rpc_frame.rs
```diff
@@ -245,7 +247,13 @@ impl Frame for RpcFrame {
                 FrameSate::Head => {
                     let mut buf = match self.read_full(message::PROTOCOL_SIZE).await {
                         Ok(v) => v,
-                        Err(_) => return Ok(Message::empty()),
+                        Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => {
```
szbr00, line 249: Narrowing RpcFrame::receive() from Err(_) to UnexpectedEof-only is a server-side behavior change, not just alignment with ReadFrame. ReadFrame is consumed by buffer_client (inflight callbacks need the real I/O cause). RpcFrame is consumed by StreamHandler, and any Err is logged at ERROR in rpc_server.

A peer abort is often ConnectionReset or ConnectionAborted, not UnexpectedEof (process kill, SO_LINGER 0, LB reset). Those used to return Message::empty() and exit quietly; they will now reproduce the noisy ERROR logs this PR is trying to stop.

Please keep RpcFrame's original catch-all at the Head stage, or treat UnexpectedEof | ConnectionReset | ConnectionAborted as graceful disconnect. Do not copy the client-side kind filter onto the server path.

bigbigxu: I understand the concern about `ConnectionReset`/`ConnectionAborted` producing ERROR logs on the server side. However, narrowing to `UnexpectedEof` on both `ReadFrame` and `RpcFrame` is intentional for consistency — both are frame readers and should have the same error semantics. The server-side `StreamHandler`/`BufferHandler` already handles the `Err` propagation path correctly (logging once and cleaning up). If `ConnectionReset`/`ConnectionAborted` becomes noisy in production, we can add those kinds to the graceful set in a follow-up.

crates/core/rpc/src/handler/read_frame.rs
bigbigxu: Fixed. Accept first, then drop the client socket before calling `receive()`. This matches the production 'peer closed an already-accepted socket' path and avoids the reset/EOF race.

crates/core/rpc/src/handler/read_frame.rs
bigbigxu: Thanks for the detailed analysis. However, I prefer to keep the simpler approach for now. The current `UnexpectedEof` filter already addresses the primary noise source (clean disconnects). The mid-header truncation case is extremely rare in practice (requires the peer to send partial bytes then close), and even if it surfaces as `Message::empty()` on the client side, `check_calls` will still notify inflight callers of the connection closure. The two-phase read adds complexity that is not justified by the marginal improvement. Will revisit if this becomes a real diagnostic issue in production.

crates/core/rpc/src/handler/read_frame.rs
```diff
@@ -52,7 +53,13 @@ impl ReadFrame {
...
+                            return Ok(Message::empty());
+                        }
+                        Err(e) => return Err(e),
```
szbr00, line 61: Commit 834902c removed the `debug!` call from this arm (and the identical one at rpc_frame.rs:252), along with the `use log::debug;` imports. The commit message only mentions the regression-test fix, so this looks unintentional.

The debug log was the compensating mitigation for narrowing the filter to UnexpectedEof: it kept the non-EOF cause visible on a path that now returns Err. Without it, a head-stage ConnectionReset carries no context at this layer, which contradicts the earlier reply on this PR stating that debug logging was added for the non-EOF path.

Please restore the `debug!` in both files, or state explicitly in the PR description that the non-EOF path is intentionally silent here.

crates/core/rpc/src/handler/rpc_frame.rs
```diff
@@ -245,7 +246,10 @@ impl Frame for RpcFrame {
...
```
szbr00, line 249: rpc_frame.rs is changed here but has no test coverage in this PR. The only regression test lives in read_frame.rs and exercises just the UnexpectedEof -> Message::empty() branch.

The whole point of this change is discriminating on ErrorKind, so both sides of that discrimination are worth pinning: RpcFrame::receive() returning Message::empty() on a clean peer FIN, and a non-EOF error still propagating as Err. Without the second case, a future refactor could widen the filter back to a catch-all and no test would notice.

crates/core/rpc/src/handler/rpc_frame.rs
szbr00: Following up with one new fact rather than reopening the design decision: commit 834902c also removed the `debug!` logging from RpcFrame's non-EOF arm.

That matters for the trade-off described above. Before this PR, a head-stage ConnectionReset returned Message::empty() and exited quietly. After it, the same event propagates as Err and is logged at ERROR by rpc_server (`Connection[a -> b]: ...`), with no debug-level breadcrumb anywhere in the frame layer. For the server path specifically, that is strictly more log noise than before, which runs against this PR's stated goal.

Also, a small correction on the reasoning: there is no BufferHandler component handling this path. `BufferHandler` appears only inside a panic message in handler_service.rs. The server-side consumer is StreamHandler, which does not log; rpc_server does, at ERROR.

If you still prefer to defer, adding ConnectionReset and ConnectionAborted to the graceful set is a two-line change and avoids shipping a known log-noise regression.

szbr00 (approved):

PR #1606, opened 2026-08-19
feat(libsdk): expose PartialSuccess load job state to clients

xingnailu:
# Summary

Load jobs that finish as transfer `PartialSuccess` were mapped to `JobTaskState::Failed` in libsdk. This PR adds `JobTaskState::PartialSuccess` / `JobTaskStateProto.PARTIAL_SUCCESS` and returns it to callers so they can decide how to handle a partially successful load.

# Issue Describe / Design

No linked issue.

- Transfer already has `TransferState::PartialSuccess`.
- SDK compatibility layer previously collapsed it to `Failed`.
- New job-task enum value is `PARTIAL_SUCCESS = 6` (additive, backward compatible).
- Java `getJobStatus` returns `PARTIAL_SUCCESS`; `waitJobComplete` returns the status for `COMPLETED` and `PARTIAL_SUCCESS` (callers use `isPartialSuccess()`). `FAILED` / `CANCELED` still throw.
- Rust `JobClient::wait_job_complete` still errors on `PartialSuccess` because that API returns `()`.

# Changes

| `crates/common/curvine-proto/proto/job.proto` | Add `PARTIAL_SUCCESS = 6` | Additive enum value |
| `crates/common/curvine-model/src/job.rs` | Add `JobTaskState::PartialSuccess`; include it in `is_finish()` | Terminal-state detection now includes partial success |
| `crates/sdk/curvine-sdk-core/src/core/transfer_compat.rs` | Map `TransferState::PartialSuccess` to `JobTaskState::PartialSuccess` | Status is no longer reported as Failed |
| `curvine-libsdk/java/.../LoadJobStatus.java` | `isFinished` includes partial success; add `isPartialSuccess()` | Callers can distinguish partial success |
| `curvine-libsdk/java/.../CurvineLoadClient.java` | `waitJobComplete` returns status on partial success | Previously would throw as Failed |
| `curvine-cli/src/cmds/load_status.rs` | Treat partial success as terminal for watch | Watch loop stops instead of polling forever |
| `crates/client/curvine-job-client/src/lib.rs` | Wait helper errors on `PartialSuccess` | Avoids hanging after mapping change |
| `curvine-master` / `curvine-worker` | Exhaustive match / cancel / task-state mapping | Compile-safe; task-level proto still maps to Failed |

# Test verified

| `cargo fmt --check` | PASS | Required CI gate |
| `cargo test -p curvine-sdk-core maps_transfer_states_to_job_task_states` | PASS | Mapping now expects `PartialSuccess` |
| `cargo check -p curvine-master -p curvine-worker -p curvine-cli --tests` | PASS | Exhaustive matches compile |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-cli/src/cmds/load_status.rs
+++ b/curvine-cli/src/cmds/load_status.rs
@@ -60,13 +60,7 @@ impl LoadStatusCommand {
         loop {
             let status = handle_rpc_result(client.get_job_status(&self.job_id)).await;
             println!("{}", status);
-            if matches!(
-                status.state,
-                JobTaskState::Completed
-                    | JobTaskState::Failed
-                    | JobTaskState::Canceled
-                    | JobTaskState::PartialSuccess
-            ) {
+            if status.state.is_finish() {
                 break;
             }
             tokio::time::sleep(duration).await;
--- a/curvine-cli/src/cmds/load_status.rs
+++ b/curvine-cli/src/cmds/load_status.rs
@@ -62,7 +62,10 @@ impl LoadStatusCommand {
             println!("{}", status);
             if matches!(
                 status.state,
-                JobTaskState::Completed | JobTaskState::Failed | JobTaskState::Canceled
+                JobTaskState::Completed
+                    | JobTaskState::Failed
+                    | JobTaskState::Canceled
+                    | JobTaskState::PartialSuccess
             ) {
                 break;
             }
--- a/curvine-cli/src/cmds/load_status.rs
+++ b/curvine-cli/src/cmds/load_status.rs
@@ -15,7 +15,7 @@
 use clap::Parser;
 use curvine_core_error::CommonResult;
 use curvine_job_client::{JobMasterClient, TransferClient};
-use curvine_model::{JobTaskState, TransferState};
+use curvine_model::TransferState;
 
 use crate::util::*;
 
--- a/curvine-libsdk/java/src/main/java/io/curvine/CurvineLoadClient.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/CurvineLoadClient.java
@@ -40,6 +40,11 @@ import java.util.concurrent.atomic.AtomicBoolean;
  * {@code fs.cv.transfer.endpoints=transfer-0:9010,transfer-1:9010}. Transfer endpoints
  * are independent from Master addresses and must be reachable from the Java process.
  *
+ * <p><b>Upgrade note:</b> the Java jar and bundled native lib must be upgraded together.
+ * A newer native library may return job states (for example {@code PARTIAL_SUCCESS})
+ * that an older jar cannot classify, which causes {@link #waitJobComplete} to poll
+ * until timeout instead of returning promptly.
+ *
  * <pre>{@code
  * try (CurvineLoadClient client = CurvineLoadClient.from(conf)) {
  *     LoadJobResult result = client.submitLoad(
--- a/curvine-libsdk/java/src/main/java/io/curvine/CurvineLoadClient.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/CurvineLoadClient.java
@@ -128,10 +128,15 @@ public final class CurvineLoadClient implements Closeable {
     /**
      * Poll job status until finished or timeout.
      *
+     * <p>Returns the final status when the job reaches {@code COMPLETED} or
+     * {@code PARTIAL_SUCCESS}. {@code FAILED} and {@code CANCELED} throw
+     * {@link IOException}. Callers should check {@link LoadJobStatus#isPartialSuccess()}
+     * (or {@link LoadJobStatus#getState()}) to decide how to handle partial success.
+     *
      * @param jobId        job id from {@link #submitLoad}
      * @param timeout      max wait duration
      * @param pollInterval sleep between status queries
-     * @return final status when the job reaches a terminal state
+     * @return final status when the job reaches a terminal success / partial-success state
      * @throws TimeoutException if the job is still running after {@code timeout}
      * @throws IOException      if the job fails / is canceled, or RPC fails
      */
@@ -153,7 +158,7 @@ public final class CurvineLoadClient implements Closeable {
         while (true) {
             status = getJobStatus(jobId);
             if (status.isFinished()) {
-                if (status.isSuccessful()) {
+                if (status.isSuccessful() || status.isPartialSuccess()) {
                     return status;
                 }
                 throw new IOException(
--- a/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
@@ -73,11 +73,20 @@ public final class LoadJobStatus {
     public boolean isFinished() {
         return state == JobTaskStateProto.COMPLETED
                 || state == JobTaskStateProto.FAILED
-                || state == JobTaskStateProto.CANCELED;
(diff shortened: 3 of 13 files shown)


curvine-cli/src/cmds/load_status.rs
```diff
@@ -62,7 +62,10 @@ impl LoadStatusCommand {
             println!("{}", status);
             if matches!(
                 status.state,
-                JobTaskState::Completed | JobTaskState::Failed | JobTaskState::Canceled
+                JobTaskState::Completed
+                    | JobTaskState::Failed
+                    | JobTaskState::Canceled
+                    | JobTaskState::PartialSuccess
```
szbr00: The watch-loop termination `matches!` list duplicates `JobTaskState::is_finish()`. Using `status.state.is_finish()` would keep CLI behavior aligned with the master and SDK paths automatically when new terminal states are added, instead of requiring another edit here.

curvine-libsdk/java/src/main/java/io/curvine/CurvineLoadClient.java
```diff
@@ -153,7 +158,7 @@ public LoadJobStatus waitJobComplete(String jobId, Duration timeout, Duration po
         while (true) {
             status = getJobStatus(jobId);
             if (status.isFinished()) {
-                if (status.isSuccessful()) {
+                if (status.isSuccessful() || status.isPartialSuccess()) {
```
szbr00, line 166: Heads up on backward compatibility: the native lib now returns `JobTaskStateProto.PARTIAL_SUCCESS` (6), but older Java client jars whose `LoadJobStatus.isFinished()` only check COMPLETED/FAILED/CANCELED will never treat this as terminal. For those clients, `waitJobComplete` polls until the timeout and then throws `TimeoutException` for a job that has actually finished, which is confusing. Consider documenting that the Java jar and native lib must be upgraded together, or making `isFinished()` default-true for unrecognized terminal enum values.

curvine-master/src/master/job/job_context.rs
```diff
@@ -160,7 +160,11 @@ impl JobContext {
                         detail.task.task_id, detail.progress.message
                     )
                 }
-                _ => (),
+                JobTaskState::UNKNOWN
+                | JobTaskState::Pending
+                | JobTaskState::Loading
+                | JobTaskState::Canceled
+                | JobTaskState::PartialSuccess => (),
```
szbr00, line 169: Note for future work: `update_progress` still collapses any sub-task failure to `JobTaskState::Failed`, so the master path never emits `PartialSuccess` — only the transfer path does (via `map_transfer_state`). The exhaustive match arm added here is therefore defensive only. If we want load jobs served by the master to also surface partial success, `update_progress` would need a "some completed, some failed" branch that sets `JobTaskState::PartialSuccess` instead of `Failed`.

curvine-master/src/master/job/job_manager.rs
```diff
@@ -251,6 +251,7 @@ impl JobManager {
                 if state == JobTaskState::Completed
                     || state == JobTaskState::Failed
                     || state == JobTaskState::Canceled
+                    || state == JobTaskState::PartialSuccess
```
szbr00: This terminal-state guard duplicates `JobTaskState::is_finish()`, which was just extended to include `PartialSuccess`. Inlining the state list here means the next new terminal state will require updating this branch again, or cancellation will silently keep rejecting already-final jobs. Consider `if state.is_finish() { ... }` to stay consistent with the model helper.

curvine-cli/src/cmds/load_status.rs
xingnailu: Thanks. Fixed in aa8fb3ab: watch loop now uses `status.state.is_finish()` instead of an explicit `matches!` list.

curvine-libsdk/java/src/main/java/io/curvine/CurvineLoadClient.java
xingnailu: Thanks. Fixed in aa8fb3ab: `LoadJobStatus.isFinished()` now treats any non-running state (including `UNRECOGNIZED` wire values) as terminal, stores the raw `stateValue` for `isPartialSuccess()`, and documents that the Java jar and native lib must be upgraded together.

curvine-master/src/master/job/job_context.rs
xingnailu: Thanks for the note. Kept current master behavior for this PR (PartialSuccess is transfer-path only). Added an inline comment in `update_progress` documenting that the `PartialSuccess` arm is defensive; master mixed-outcome support can be a follow-up.

curvine-master/src/master/job/job_manager.rs
xingnailu: Thanks. Fixed in aa8fb3ab: cancel guard now uses `state.is_finish()` instead of inlining terminal states.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
```diff
@@ -70,16 +74,34 @@ public JobTaskProgressProto getProgress() {
         return progress;
     }
 
+    /**
+     * True when the job is no longer running (completed, failed, canceled, partial success,
+     * or any future terminal enum value / {@code UNRECOGNIZED} wire value).
+     */
     public boolean isFinished() {
```
szbr00, line 96: `isFinished()` now treats `UNRECOGNIZED` / unknown wire values as terminal, which is what keeps `waitJobComplete` from polling until timeout on a future enum. The new tests cover `PARTIAL_SUCCESS` and `LOADING`, but not an unknown `stateValue`. Consider a `fromProto` case with `setStateValue(99)` asserting `isFinished() == true`, `isSuccessful() == false`, and `isPartialSuccess() == false`, so this forward-compat path cannot regress silently.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
```diff
@@ -24,18 +24,21 @@
 public final class LoadJobStatus {
     private final String jobId;
     private final JobTaskStateProto state;
+    private final int stateValue;
     private final String sourcePath;
     private final String targetPath;
     private final JobTaskProgressProto progress;
 
     public LoadJobStatus(
```
szbr00, line 32: This public constructor now requires `stateValue`, which is a source break for any caller still using the previous 5-arg form. `fromProto` is the only in-repo caller, so this may be fine, but consider keeping the old signature as an overload that derives `stateValue` from `state.getNumber()` (and a private/full constructor for the raw wire value). That preserves compatibility while still supporting `UNRECOGNIZED`.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
xingnailu: Thanks. Fixed in aa1a31b5: added a `fromProto` case with `setStateValue(99)` asserting `isFinished() == true`, `isSuccessful() == false`, and `isPartialSuccess() == false`.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
xingnailu: Thanks. Fixed in aa1a31b5: restored the public 5-arg constructor (derives `stateValue` from `state.getNumber()`) and kept the 6-arg constructor private for the raw wire value used by `fromProto`.

lzjqsdd (approved):

Issue #1272, opened 2026-07-23
[BUG]: FUSE statfs03_64 returns success when permission denied expected


LJF-050:
**Is your feature request related to a problem?**

statfs03_64 succeeds unexpectedly on curvine-fuse.

statfs03 1 TFAIL : statfs03.c:89: call succeeded unexpectedly

Reproduction:

1. Mount curvine-fuse
2. Run LTP statfs03_64
3. Observe TFAIL.

**Impact**

statfs compatibility gap.

**Describe the solution you'd like**

statfs respects permission boundaries.

Fix statfs permission checks on FUSE.

**Additional context**

Root cause stated as hypothesis unless log proves causality.

Root-cause hypothesis: statfs on FUSE allows operations that should fail in statfs03_64 (hypothesis).

Evidence summary: statfs03 1 TFAIL : statfs03.c:89: call succeeded unexpectedly

Related issues: None identified.

Acceptance / test gaps:

- statfs03_64 passes

PR #1610, opened 2026-08-19
fix(master): report allocatable capacity in statfs (#1460)

### LINKED ISSUE #1460: [BUG]: FUSE statfs includes non-writable workers in available capacity (by LiAuto)
**Describe the bug**

`CurvineFileSystem::stat_fs()` reports the `capacity` and `available` values returned by `GetMasterInfo` as FUSE filesystem capacity.

However, `MasterFilesystem::master_info()` adds every worker's capacity statistics before checking its status:

```rust
for (_, worker) in wm.worker_map.workers() {
    info.capacity += worker.capacity;
    info.available += worker.available;
    // ...

    match worker.status {
        WorkerStatus::Live => ...
        WorkerStatus::Blacklist => ...
        WorkerStatus::Decommission => ...
        _ => (),
    }
}
```

As a result, Blacklist and Decommission workers contribute to the capacity reported by `statfs(2)`.

Worker allocation policies only select Live workers. The capacity visible through `df` can therefore be larger than the capacity available for new writes. In the worst case, `df` reports substantial free space while writes fail because no eligible worker exists.

Lost workers are stored separately and are not included by the current loop.

Relevant locations:

- `curvine-master/src/master/fs/master_filesystem.rs:1111`
- `curvine-master/src/master/fs/policy/`

**To Reproduce**

1. Start a Curvine cluster with at least two Live workers.
2. Record the mounted filesystem capacity using `df`.
3. Blacklist or decommission a worker with significant available capacity.
4. Run `df` again on the FUSE mount.
5. Observe that the unavailable worker's capacity remains included.
6. Attempt writes after all workers with sufficient capacity become non-Live.
7. Observe that writes fail despite `df` reporting free space.

**Expected behavior**

The capacity reported by FUSE should represent storage eligible for new Curvine writes.

Blacklist and Decommission workers must not contribute to the reported allocatable capacity.

Administrative cluster statistics may still need to retain total physical capacity across worker states. The fix should not silently change the existing meaning of `MasterInfo.capacity` for unrelated consumers.

- OS: Linux
- Component: `curvine-master`, `curvine-fuse`

**Additional context**

Possible implementations:

1. Add explicit allocatable-capacity fields to `MasterInfo`.
2. Introduce a dedicated capacity RPC that only aggregates eligible workers.
3. Change existing aggregation semantics only after auditing all `MasterInfo` consumers.

`available` at the worker storage layer already excludes reserved bytes, so FUSE should not subtract `reserved_bytes` again.

Acceptance criteria:

- Live workers contribute to allocatable capacity.
- Blacklist workers do not contribute.
- Decommission workers do not contribute.
- Lost workers do not contribute.
- Failed storage directories do not contribute.
- FUSE capacity is consistent with worker-allocation eligibility.
- Tests cover mixed Live, Blacklist, and Decommission states.

gangump82:
# Summary

FUSE `statfs` reported capacity/available summed across all non-lost
workers (Live + Blacklist + Decommission), so `df` could advertise free
space that no new write could use — the allocation policy only selects
Live workers. This PR adds an allocatable capacity view (Live workers
only) and points `statfs` at it, while preserving the existing total
capacity semantics for the CLI report and metrics.

# Issue Describe / Design

Closes #1460

**Root cause**

`MasterFilesystem::filesystem_info()` aggregates `capacity`/`available`
for every worker in `worker_map.workers()` *before* checking
`worker.status`, so Blacklist and Decommission workers inflate the
values that `CurvineFileSystem::stat_fs` feeds into `fuse_kstatfs`.
Worker allocation policies (`random_worker_policy`,
`weighted_worker_policy`, ...) filter via `worker.is_live()`, so only
Live workers are eligible — the aggregation and allocation views
disagree, and in the worst case `df` shows free space while writes
fail with "not enough space".

**Reproduction**

1. Start a cluster with ≥2 Live workers; note `df` on the FUSE mount.
2. Blacklist/decommission a worker with significant available capacity.
3. `df` still reports the unavailable worker's capacity.
4. After all Live capacity is exhausted, writes fail despite `df`
   showing free space.

**Fix approach (issue's option 1 — additive fields)**

- Add `allocatable_capacity` / `allocatable_available` to
  `FilesystemInfo`; aggregate only `WorkerStatus::Live` workers there.
  Failed storage dirs are already excluded by
  `WorkerInfo::add_storage`, so a Live worker's capacity already omits
  them; Lost workers live in a separate map and never enter the loop.
- Add optional proto fields (tags 15/16) to `GetFilesystemInfoResponse`
  and carry them through `ProtoUtils` conversion. Legacy masters
  omitting the fields fall back to the aggregate `capacity`/`available`
  so a new client against an old master never regresses to 0 free
  space.
- `stat_fs` reports the allocatable view; CLI `report` and master
  metrics gain allocatable gauges **without** changing the existing
  `capacity`/`available` meaning.

# Changes

| --- | --- | --- |
| `crates/common/curvine-model/src/filesystem_info.rs` | Add `allocatable_capacity`/`allocatable_available` fields | None (default 0; additive) |
| `crates/common/curvine-proto/proto/master.proto` | Add optional `allocatable_capacity`/`allocatable_available` (tags 15/16) | None (optional; legacy peers unaffected) |
| `crates/common/curvine-model/src/proto_utils.rs` | to/from-pb for new fields; legacy fallback to `capacity`/`available` | Mixed-version clusters keep pre-fix behavior |
| `curvine-master/src/master/fs/master_filesystem.rs` | Aggregate allocatable for Live workers in `filesystem_info()` | Total `capacity`/`available` unchanged |
| `curvine-fuse/src/fs/curvine_file_system.rs` | `stat_fs` uses allocatable fields | **Behavior change (the fix):** `df` no longer counts Blacklist/Decommission workers |
| `curvine-cli/src/cmds/report.rs` | Show `allocatable_capacity`/`allocatable_available` in report + cluster summary | Additive display |
| `curvine-master/src/master/master_metrics.rs` | Add `allocatable_capacity`/`allocatable_available` gauges | Existing `capacity`/`available` metrics unchanged |

# Test verified

| --- | --- | --- |
| `cargo test -p curvine-master --lib master::fs::master_filesystem::tests` | PASS | 8 tests, incl. 3 new: allocatable only-Live, excludes failed dirs, excludes lost workers |
| `cargo test -p curvine-model --test filesystem_info_proto_test` | PASS | 3 tests: round-trip, legacy fallback, zero-preserves-not-confused-with-absent |
| `cargo test -p curvine-proto --test get_filesystem_info_compat_test` | PASS | 6 tests; added allocatable assertions + legacy-absent assertions |
| `cargo check -p curvine-model -p curvine-proto -p curvine-master -p curvine-cli` | PASS | |
| `cargo check -p curvine-fuse` | PASS (for `curvine_file_system.rs`) | |

# Dependencies

- Related issues: #1460, #1272 (statfs permission behavior, not a duplicate)
- Related PRs: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -1944,15 +1944,30 @@ mod tests {
     #[test]
     fn filesystem_info_allocatable_excludes_lost_workers() {
         // Lost workers live in a separate map and must not contribute to either
-        // the total or the allocatable view.
+        // the total or the allocatable view. Insert a lost worker directly into
+        // the lost map (add_test_worker only touches the live map) and assert its
+        // capacity never leaks into aggregation.
         let fs = test_fs("allocatable-excludes-lost");
         fs.add_test_worker(worker_with_status(1, WorkerStatus::Live, 1000, 800));
-        // Unknown/Lost-status workers in the live map are bucketed by the `_ =>`
-        // arm, but real Lost workers are never in worker_map.workers().
+        fs.worker_manager
+            .write()
+            .worker_map
+            .lost_workers
+            .insert(99, worker_with_status(99, WorkerStatus::Lost, 9999, 9999));
+
         let info = fs.filesystem_info().unwrap();
-        assert_eq!(info.capacity, 1000);
-        assert_eq!(info.allocatable_capacity, 1000);
+
+        // The lost worker is reported but its bytes stay out of both views.
+        assert_eq!(info.lost_workers.len(), 1);
+        assert_eq!(
+            info.capacity, 1000,
+            "lost worker capacity must not leak into total"
+        );
+        assert_eq!(info.available, 800);
+        assert_eq!(
+            info.allocatable_capacity, 1000,
+            "lost worker capacity must not leak into allocatable"
+        );
         assert_eq!(info.allocatable_available, 800);
-        assert!(info.lost_workers.is_empty());
     }
 }
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -1131,7 +1131,14 @@ impl MasterFilesystem {
             info.block_num += worker.block_num;
 
             match worker.status {
-                WorkerStatus::Live => info.live_workers.push(worker.clone()),
+                WorkerStatus::Live => {
+                    info.live_workers.push(worker.clone());
+                    // Only Live workers are eligible for new allocations, so the
+                    // allocatable view mirrors the allocation policy. Failed
+                    // storage dirs are already excluded from worker.capacity.
+                    info.allocatable_capacity += worker.capacity;
+                    info.allocatable_available += worker.available;
+                }
                 WorkerStatus::Blacklist => info.blacklist_workers.push(worker.clone()),
                 WorkerStatus::Decommission => info.decommission_workers.push(worker.clone()),
                 _ => (),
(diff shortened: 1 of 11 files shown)


curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1832,4 +1839,120 @@ mod tests {
...
+        let fs = test_fs("allocatable-excludes-lost");
+        fs.add_test_worker(worker_with_status(1, WorkerStatus::Live, 1000, 800));
+        // Unknown/Lost-status workers in the live map are bucketed by the `_ =>`
+        // arm, but real Lost workers are never in worker_map.workers().
+        let info = fs.filesystem_info().unwrap();
+        assert_eq!(info.capacity, 1000);
+        assert_eq!(info.allocatable_capacity, 1000);
+        assert_eq!(info.allocatable_available, 800);
+        assert!(info.lost_workers.is_empty());
+    }
```
szbr00, line 1972: `filesystem_info_allocatable_excludes_lost_workers` never inserts a worker into
`worker_map.lost_workers`, so it does not exercise the exclusion its name
claims. `add_test_worker` only inserts into `workers` (worker_manager.rs:251),
and the test's only assertion about lost workers is `info.lost_workers.is_empty()`,
which is trivially true because nothing was added. The test therefore cannot
catch a regression where a lost worker's capacity leaked into the total or
allocatable aggregation.

To actually verify the exclusion, insert a lost worker directly into the lost
map, e.g. via a test helper that does
`wm.worker_map.lost_workers.insert(id, worker)`, then assert:
- `info.lost_workers.len() == 1`
- `info.capacity` / `info.available` equal the Live worker's values only
- `info.allocatable_capacity` / `info.allocatable_available` equal the Live
  worker's values only

Alternatively, drop the test and keep the exclusion guarantee as a comment on
the `lost_workers()` loop in `filesystem_info()`, since the separation is
already structurally enforced by the distinct map.

gangump82: Fixed in `e10407e0`: the test now inserts a lost worker directly into `worker_map.lost_workers` (id 99, capacity 9999/9999) and asserts `lost_workers.len() == 1` plus that its bytes stay out of both `capacity`/`available` and `allocatable_capacity`/`allocatable_available`. So a future regression that pulls lost workers into the aggregation loop would actually fail here.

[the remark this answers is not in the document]
gangump82: Addressed in `e10407e0`: the test now inserts a lost worker into `worker_map.lost_workers` and asserts `lost_workers.len() == 1` with its capacity (9999) excluded from both total and allocatable aggregation.

szbr00 (commented):
The FUSE statfs path now reports the Live-only allocatable view, which matches
#1460. Two other df-equivalent entry points still advertise
capacity/available (all non-lost workers), so the original "writes fail while
df shows free space" mismatch remains there:

- curvine-cli/src/cmds/fs/df.rs uses filesystem_info.capacity/available
  for `cv df` (commented as HDFS-df-like output).
- CurvineFsStat(GetFilesystemInfoResponse) passes
  info.getCapacity()/getAvailable() into Hadoop FsStatus, which is what
  FileSystem.getStatus() / Spark remaining-space checks see.

This is consistent with the issue's instruction not to silently change
FilesystemInfo.capacity for admin stats, but these two surfaces are
user-facing free-space APIs, not `cv report`. A follow-up can switch them to
allocatable_* with the same legacy fallback already in
ProtoUtils::filesystem_info_from_pb. For Java that fallback must use
hasAllocatableCapacity()/hasAllocatableAvailable(): protobuf returns 0 for an
absent optional int64 with default=0, which would make a new client against
an old master report zero free space.

gangump82:
@szbr00 Both df-equivalent entry points are switched to the allocatable view in `ed484230` (same legacy fallback path as FUSE statfs), so the "writes fail while df shows free space" mismatch is closed everywhere.

**`curvine-cli/src/cmds/fs/df.rs`** — `cv df` now reports `allocatable_capacity` / `allocatable_available` for Size/Available. Used is derived as `allocatable_capacity - allocatable_available` rather than `fs_used + non_fs_used`, so the df arithmetic (`Size = Used + Available`) stays consistent when Blacklist/Decommission workers still hold data (otherwise Used could exceed Size and Use% > 100%). `fs_used` is still surfaced in `cv report` for the physical-usage view. On a legacy master that omits tags 15/16, `FilesystemInfo` falls back to the aggregate totals via `proto_utils::filesystem_info_from_pb`, so `cv df` never regresses to zero.

**`CurvineFsStat.java`** — constructor now uses `hasAllocatableCapacity() ? getAllocatableCapacity() : getCapacity()` (and the available equivalent) exactly as you specified, so the Hadoop `FsStatus` / Spark remaining-space checks only see Live workers and a new client against an old master won't report zero free space. `simple()` also prints the allocatable view (guarded by `hasAllocatableCapacity()`) for parity with `cv report`.

Verified:
- `cargo check -p curvine-cli` clean; `cargo test -p curvine-model --test filesystem_info_proto_test` and `cargo test -p curvine-master -- allocatable` pass; proto compat tests pass.
- Java side: no local `mvn` to build the SDK, so the generated `GetFilesystemInfoResponse` bindings (incl. the `hasAllocatableCapacity()`/`getAllocatableCapacity()` methods) will be exercised by CI. The proto2 optional field → camelCase method names follow the existing convention used elsewhere (`hasMountInfo()` etc.).

szbr00 (approved):

PR #1611, opened 2026-08-19
fix(java-sdk): resolve LoadJobStatus compile error referencing missing proto field

lzjqsdd:
# Summary

Fix the Java SDK compile error introduced by #1606 in the curvine-libsdk/java module. `LoadJobStatus.fromProto` calls `GetJobStatusResponse.getStateValue()` and the unit test calls `.setStateValue(99)`, but `GetJobStatusResponse` has no `state_value` field in `job.proto`, so the generated Java stubs expose no such method and `javac` fails with `cannot find symbol`. The raw state number is now derived from `response.getState().getNumber()`, which restores buildability while keeping the intended behavior for all known job states.

# Issue Describe / Design

- Root cause: #1606 added a `stateValue` field to `LoadJobStatus` for partial-success / forward-compat detection and reads it via `response.getStateValue()` / `.setStateValue(...)`, but `job.proto`'s `GetJobStatusResponse` was never given a `state_value` field, so the generated Java stubs have no such methods and the module cannot compile in any checkout.
- Reproduction: `mvn -f curvine-libsdk/java/pom.xml test` fails with `cannot find symbol` at `LoadJobStatus.java:66` (`getStateValue()`) and `LoadJobClientTest.java:145` (`setStateValue(99)`).
- Fix approach: in `fromProto`, derive the raw state number via `response.getState().getNumber()` (for known enum values this yields the same value the commit intended to store; `isPartialSuccess()` keeps working for `PARTIAL_SUCCESS`). The "unrecognized wire value" test case cannot be constructed through the typed builder without a `state_value` field, so it is replaced with an `UNKNOWN`-state case asserting the actual proto2 decode semantics (not finished, not successful, not partial success). Note: full forward-compat detection of unknown wire enum values would require adding a real `state_value` field to `GetJobStatusResponse`; that is out of scope for this buildability fix and can be a follow-up.

# Changes

| `curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java` | `fromProto` reads the raw state via `response.getState().getNumber()` instead of the non-existent `getStateValue()` | None for known states (fixes compile error) |
| `curvine-libsdk/java/src/test/java/io/curvine/LoadJobClientTest.java` | Replace the `setStateValue(99)` case with an `UNKNOWN`-state case matching proto2 decode semantics | None (test-only) |

# Test verified

| `mvn -f curvine-libsdk/java/pom.xml test -Dtest=LoadJobClientTest` | PASS | 8 tests, 0 failures (previously did not compile) |
| `make format` | PASS | clippy clean, no formatting changes |

# Dependencies

- Related PRs: #1606 (introduced the compile error)
- Related issues: none
- Blocks / blocked by: none

### DIFF
--- a/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
@@ -63,7 +63,7 @@ public final class LoadJobStatus {
         return new LoadJobStatus(
                 response.getJobId(),
                 response.getState(),
-                response.getStateValue(),
+                response.getState().getNumber(),
                 response.getSourcePath(),
                 response.getTargetPath(),
                 response.getProgress());
--- a/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
@@ -14,20 +14,14 @@
 
 package io.curvine;
 
-import com.google.protobuf.CodedInputStream;
-import com.google.protobuf.WireFormat;
 import io.curvine.proto.GetJobStatusResponse;
 import io.curvine.proto.JobTaskProgressProto;
 import io.curvine.proto.JobTaskStateProto;
-
-import java.io.IOException;
 
 /**
  * Load job status snapshot. Mirrors Rust {@code JobStatus}.
  */
 public final class LoadJobStatus {
-    private static final int STATE_FIELD_NUMBER = 2;
-
     private final String jobId;
     private final JobTaskStateProto state;
     private final int stateValue;
@@ -70,40 +64,10 @@ public final class LoadJobStatus {
         return new LoadJobStatus(
                 response.getJobId(),
                 state,
-                rawStateValue(response, state),
+                state.getNumber(),
                 response.getSourcePath(),
                 response.getTargetPath(),
                 response.getProgress());
-    }
-
-    /**
-     * Returns the raw enum wire value for {@code state}. Known enum constants use the mapped
-     * number directly; when proto2 maps an unknown wire value to {@code UNKNOWN}, recover the
-     * original value from the serialized payload so forward-compat helpers stay accurate.
-     */
-    private static int rawStateValue(GetJobStatusResponse response, JobTaskStateProto state) {
-        if (state != JobTaskStateProto.UNKNOWN) {
-            return state.getNumber();
-        }
-        return extractEnumWireValue(response, STATE_FIELD_NUMBER, state.getNumber());
-    }
-
-    private static int extractEnumWireValue(
-            GetJobStatusResponse response, int fieldNumber, int defaultValue) {
-        try {
-            CodedInputStream input = CodedInputStream.newInstance(response.toByteArray());
-            while (!input.isAtEnd()) {
-                int tag = input.readTag();
-                if (WireFormat.getTagFieldNumber(tag) == fieldNumber
-                        && WireFormat.getTagWireType(tag) == WireFormat.WIRETYPE_VARINT) {
-                    return input.readEnum();
-                }
-                input.skipField(tag);
-            }
-        } catch (IOException ignored) {
-            // Keep the mapped enum number when the payload cannot be re-scanned.
-        }
-        return defaultValue;
     }
 
     public String getJobId() {
@@ -127,8 +91,10 @@ public final class LoadJobStatus {
     }
 
     /**
-     * True when the job is no longer running (completed, failed, canceled, partial success,
-     * or any future terminal enum value / {@code UNRECOGNIZED} wire value).
+     * True when the job is no longer running (completed, failed, canceled, or partial success).
+     * Under proto2, unknown wire enum values decode to {@code UNKNOWN} and are treated as
+     * non-terminal; full forward-compat for future terminal states requires a preserved raw
+     * state field on the wire (for example {@code state_value} in {@code job.proto}).
      */
     public boolean isFinished() {
         switch (state) {
--- a/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
@@ -14,14 +14,20 @@
 
 package io.curvine;
 
+import com.google.protobuf.CodedInputStream;
+import com.google.protobuf.WireFormat;
 import io.curvine.proto.GetJobStatusResponse;
 import io.curvine.proto.JobTaskProgressProto;
 import io.curvine.proto.JobTaskStateProto;
+
+import java.io.IOException;
 
 /**
  * Load job status snapshot. Mirrors Rust {@code JobStatus}.
  */
 public final class LoadJobStatus {
+    private static final int STATE_FIELD_NUMBER = 2;
+
     private final String jobId;
     private final JobTaskStateProto state;
     private final int stateValue;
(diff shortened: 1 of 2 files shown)


szbr00 (commented):
Suggestion (follow-up — not in this diff): the `isFinished()` javadoc states the result is true for "any future terminal enum value / {@code UNRECOGNIZED} wire value", but `job.proto` is proto2 and the generated `GetJobStatusResponse.getState()` maps an unknown wire value to `UNKNOWN` (not `UNRECOGNIZED`), for which `isFinished()` returns false. After this change `stateValue` is derived from `getState().getNumber()`, so an unknown wire value decodes to `UNKNOWN`(0) — not finished / not successful / not partial — which is exactly what the updated test now asserts.

Consider updating the javadoc to drop the `UNRECOGNIZED` reference and describe the actual proto2 behavior (unknown wire values decode to `UNKNOWN` and are treated as non-terminal), so callers don't expect future terminal states to be reported as finished.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
```diff
@@ -60,15 +66,46 @@ private LoadJobStatus(
...
+            return state.getNumber();
+        }
+        return extractEnumWireValue(response, STATE_FIELD_NUMBER, state.getNumber());
+    }
+
+    private static int extractEnumWireValue(
+            GetJobStatusResponse response, int fieldNumber, int defaultValue) {
+        try {
+            CodedInputStream input = CodedInputStream.newInstance(response.toByteArray());
+            while (!input.isAtEnd()) {
```
szbr00: This re-scan does not implement the forward-compat it claims, on either the production entry path or the public helpers.

1) Production never feeds an unknown enum into fromProto. CurvineLoadClient.getJobStatus uses GetJobStatusResponse.parseFrom (not parsePartialFrom). Generated proto2 mergeFrom stores an unknown enumerant via mergeUnknownVarintField(2, tmpRaw) and does not set hasState(); isInitialized() then fails, so parseFrom throws UninitializedMessageException. Separately, JNI already collapses unknowns: job_client maps `JobTaskState::from(status.state as i8)` (num_enum default = UNKNOWN) and re-encodes `state: i32::from(status.state as i8)` before Java sees the bytes (java_filesystem.rs get_job_status). The scanner therefore only runs for genuine UNKNOWN=0, where toByteArray() + scan returns 0 — the same as state.getNumber().

2) Even if a raw 99 were recovered, no helper uses it. isFinished() switches on `state` and treats UNKNOWN as not finished; isPartialSuccess() is true only for wire value 6, which is a known enumerant so getState() would already be PARTIAL_SUCCESS. There is no getStateValue() accessor. waitJobComplete only inspects isPartialSuccess() after isFinished() is true, so a recovered PARTIAL_SUCCESS number on an older generated enum would still poll until timeout — the failure mode already documented on CurvineLoadClient.

Suggestion: drop extractEnumWireValue and keep `state.getNumber()` for the compile fix. Real forward-compat needs a preserved int on the wire (state_value field, or proto3 getStateValue) plus isFinished() consulting that int; a client-side re-parse of toByteArray() cannot substitute for that.

curvine-libsdk/java/src/test/java/io/curvine/LoadJobClientTest.java
```diff
@@ -140,22 +143,54 @@ public void loadJobStatusFinishedHelpers() {
...
+
+    @Test
+    public void loadJobStatusHandlesUnknownWireEnumValue() throws Exception {
+        JobTaskProgressProto progress = JobTaskProgressProto.newBuilder()
+                .setLoadedSize(1)
+                .setTotalSize(10)
+                .setUpdateTime(1)
+                .setState(99)
+                .setMessage("future-state")
+                .build();
```
szbr00: This test does not fail on the defect the scanner is supposed to fix.

It builds status via PARSER.parsePartialFrom, which skips the required-field check that CurvineLoadClient.getJobStatus actually uses (parseFrom). On proto2, an unknown value of required enum `state` leaves hasState() false, so the production parse throws and never reaches fromProto.

The assertions (getState()==UNKNOWN, isFinished/isSuccessful/isPartialSuccess all false) are identical to fromProto(GetJobStatusResponse with setState(UNKNOWN)). stateValue is private and never asserted, so deleting extractEnumWireValue would still leave this test green.

If the scanner stays, parse with parseFrom (or drive CurvineLoadClient.getJobStatus) and assert the recovered raw number through a public helper whose result actually changes (for example isFinished() true for a non-zero unknown wire value, if that becomes the contract). Otherwise delete this case; the UNKNOWN builder test above already covers mapped proto2 semantics.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
```diff
@@ -63,7 +63,7 @@ public static LoadJobStatus fromProto(GetJobStatusResponse response) {
         return new LoadJobStatus(
                 response.getJobId(),
                 response.getState(),
-                response.getStateValue(),
+                response.getState().getNumber(),
                 response.getSourcePath(),
                 response.getTargetPath(),
                 response.getProgress());
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Addressed in 4f15e31b: kept `state.getNumber()` for the compile fix. `GetJobStatusResponse` has no generated `getStateValue()` (proto2), and the payload re-scan added in ce59dcf4 was removed after maintainer review because the production `parseFrom` path never delivers unknown enum wire values into `fromProto`.

lzjqsdd:
Thanks for the follow-up review. Addressed in 4f15e31b: dropped the payload re-scan helper, kept `state.getNumber()` for the compile fix, removed the misleading `parsePartialFrom` test, and updated `isFinished()` javadoc to document proto2 `UNKNOWN` (non-terminal) semantics instead of `UNRECOGNIZED`.

curvine-libsdk/java/src/test/java/io/curvine/LoadJobClientTest.java
```diff
@@ -140,22 +140,22 @@ public void loadJobStatusFinishedHelpers() {
...
+                        .setState(JobTaskStateProto.UNKNOWN.getNumber())
+                        .setMessage("unknown-state")
                         .build())
                 .build());
-        Assert.assertTrue(unrecognized.isFinished());
-        Assert.assertFalse(unrecognized.isSuccessful());
-        Assert.assertFalse(unrecognized.isPartialSuccess());
+        Assert.assertFalse(unknownState.isFinished());
+        Assert.assertFalse(unknownState.isSuccessful());
+        Assert.assertFalse(unknownState.isPartialSuccess());
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. The `parsePartialFrom` unknown-wire test from ce59dcf4 was removed in 4f15e31b; it did not exercise the `CurvineLoadClient.getJobStatus` (`parseFrom`) path. The `UNKNOWN` builder case in `loadJobStatusFinishedHelpers` covers the actual proto2 decode semantics.

curvine-libsdk/java/src/main/java/io/curvine/LoadJobStatus.java
lzjqsdd: Thanks. Fixed in 4f15e31b: removed `extractEnumWireValue` / `rawStateValue` and derive `stateValue` directly from `state.getNumber()`. Updated `isFinished()` javadoc to describe proto2 `UNKNOWN` behavior instead of `UNRECOGNIZED`.

curvine-libsdk/java/src/test/java/io/curvine/LoadJobClientTest.java
lzjqsdd: Thanks. Fixed in 4f15e31b: deleted `loadJobStatusHandlesUnknownWireEnumValue` and the manual encoder helper. The `UNKNOWN` builder assertions in `loadJobStatusFinishedHelpers` remain as the proto2 contract test.

szbr00 (approved):

PR #1613, opened 2026-08-20
fix(fuse): reject unsupported mount options (#1399)

### LINKED ISSUE #1399: [BUG]: Reject silently ignored FUSE mount options [LiAuto]
**Describe the bug**

FUSE mount options are stored as strings but only a limited set of exact values is interpreted. Comma-separated forms and security options such as `nosuid`, `nodev`, or `noexec` can be silently ignored, causing actual mount behavior to differ from operator intent.

**To Reproduce**

1. Start FUSE with `-o allow_other,nodev` or an equivalent `--options` value.
2. Inspect the final mount flags.
3. Observe that unsupported or combined options can be ignored without an error.

**Expected behavior**

Options must be normalized, mapped through an explicit support table, and rejected with the offending name when unknown or unimplemented.

- Linux
- Component: FUSE CLI and mount option mapping

**Additional context**

Tests must assert final mount flags and FUSE parameters, not only the parsed CLI structure.

gangump82:
# Summary

Reject FUSE mount options that Curvine's direct mount backend cannot implement instead of silently ignoring them. Normalize options after CLI/config merging, separate Linux VFS flags from FUSE mount data, and reject conflicting positive/negative options both within `fuse_opts` and across the `readonly` configuration field.

# Issue Describe / Design

Closes #1399

## Root cause

Curvine previously treated `fuse_opts` as opaque strings. Comma-separated values were not normalized, several Linux VFS options were appended to FUSE mount data rather than converted into `mount(2)` flags, and unknown or unimplemented options could be accepted without taking effect. In addition, accepting both `sync` and `async` left `MS_SYNCHRONOUS` enabled whenever `sync` was present, silently ignoring `async`.

The explicit `readonly` field is stored separately from `fuse_opts`. After `rw` became a supported positive VFS option, `readonly = true` could still append `ro` during raw mount construction and silently override an explicit `rw` request.

## Reproduction

1. Pass comma-separated options such as `--options nodev,nosuid,noexec` or a known but unimplemented option such as `auto_unmount`.
2. The old path could accept the input while not applying the requested semantics.
3. Pass `--options sync,async`; both options were accepted even though the resulting mount remained synchronous.
4. Pass `--readonly --options rw`, or configure `readonly = true` with `fuse_opts = ["rw"]`; the mount silently became read-only instead of reporting the conflict.

## Fix approach

- Split and trim comma-separated options and deduplicate supported options in order.
- Validate against an explicit support table after CLI options have replaced config-file options, and include the offending option in errors.
- Reject known but unimplemented direct-mount options.
- Convert VFS options to `mount(2)` flags while keeping FUSE parameters in mount data.
- Accept standard positive VFS forms (`rw`, `dev`, `suid`, `exec`, and `atime`) as explicit defaults.
- Reject opposite pairs such as `ro`/`rw`, `nodev`/`dev`, and `sync`/`async`, preserving both option names in the conflict error.
- Reject explicit `rw` when `fuse.readonly=true`, both during final configuration normalization and during the defensive mount-time validation.
- Remove unsupported defaults and benchmark usages.

# Changes

| `crates/common/curvine-config/src/fuse_conf.rs` | Normalize and validate mount options at the FUSE boundary; support positive VFS forms; reject opposite pairs and `rw` with `readonly=true` | Non-FUSE services can load shared legacy config; invalid final mount options and cross-field conflicts fail explicitly |
| `curvine-fuse/src/cli/mount_args.rs` | Merge config/CLI/default options before final validation and remove unsupported defaults | CLI `--options` can replace legacy config values as documented; `--readonly --options rw` fails before mounting |
| `curvine-fuse/src/raw/fuse_pure.rs` | Separate VFS mount flags from FUSE mount data and retain defensive conflict validation | Security/read-only options use the correct kernel interface; direct callers cannot silently override `rw` with `readonly` |
| `curvine-tests/benchmark/master-stress/*.yaml` | Remove unsupported `auto_unmount` usage | Benchmark deployments use only implemented mount options |

# Test verified

| `cargo fmt --all -- --check` | PASS | Local and Ubuntu checkout |
| `git diff --check github/main...HEAD` | PASS | No whitespace errors |
| `cargo test -p curvine-config --lib` | PASS | Ubuntu, 61 passed |
| `cargo test -p curvine-fuse --no-default-features --features fuse2 cli::mount_args --lib` | PASS | Ubuntu, 11 passed |
| `cargo test -p curvine-fuse --no-default-features --features fuse2 raw::fuse_pure::tests --lib` | PASS | Ubuntu, 17 passed |
| `cargo test -p curvine-fuse --no-default-features --features fuse3 cli::mount_args --lib` | PASS | Ubuntu, 11 passed |
| `cargo clippy -p curvine-config --lib --no-deps -- -D warnings` | PASS | Local and Ubuntu checkout |

Linux FUSE verification is listed above. The macOS FUSE test target retains the existing platform-gating compile failure because `fuse_mount_pure` is Linux-only.

# Dependencies

- Related PRs: None
- Related issues: #1399
- Blocks / blocked by: None

### DIFF
--- a/crates/common/curvine-config/src/fuse_conf.rs
+++ b/crates/common/curvine-config/src/fuse_conf.rs
@@ -105,7 +105,13 @@ pub struct FuseConf {
     // Read and write file request queue size, default is 0
     pub stream_channel_size: usize,
 
-    // Mount the configuration, needs to be passed to the linux kernel.
+    // Mount options for Curvine's direct Linux FUSE backend. Supported VFS
+    // pairs are `ro`/`rw`, `nodev`/`dev`, `nosuid`/`suid`, `noexec`/`exec`,
+    // `noatime`/`atime`, and `sync`/`async`; `dirsync` is also supported.
+    // Supported FUSE-side options are `allow_other`, `default_permissions`, and
+    // `big_write` (negotiated through FUSE_INIT). Opposite options conflict.
+    // Legacy libfuse options `auto_unmount`, `allow_root`, and `max_write` are
+    // not supported in this field by the direct mount backend.
     pub fuse_opts: Vec<String>,
 
     // Mount the whole FUSE filesystem read-only at the kernel level.
@@ -287,8 +293,6 @@ impl FuseConf {
                 self.max_background
             );
         }
-
-        self.fuse_opts = Self::normalize_fuse_opts(&self.fuse_opts)?;
 
         self.attr_ttl = Duration::from_millis(self.attr_timeout_ms);
         self.entry_ttl = Duration::from_millis(self.entry_timeout_ms);
@@ -480,7 +484,7 @@ impl FuseConf {
         args
     }
 
-    fn normalize_fuse_opts(opts: &[String]) -> CommonResult<Vec<String>> {
+    fn normalized_fuse_opts(opts: &[String]) -> CommonResult<Vec<String>> {
         let mut normalized = Vec::new();
 
         for entry in opts {
@@ -500,10 +504,15 @@ impl FuseConf {
 
                 let option = match name {
                     "ro"
+                    | "rw"
                     | "nodev"
+                    | "dev"
                     | "nosuid"
+                    | "suid"
                     | "noexec"
+                    | "exec"
                     | "noatime"
+                    | "atime"
                     | "dirsync"
                     | "sync"
                     | "allow_other"
@@ -530,6 +539,16 @@ impl FuseConf {
                 };
 
                 let conflicting_option = match option.as_str() {
+                    "ro" => Some("rw"),
+                    "rw" => Some("ro"),
+                    "nodev" => Some("dev"),
+                    "dev" => Some("nodev"),
+                    "nosuid" => Some("suid"),
+                    "suid" => Some("nosuid"),
+                    "noexec" => Some("exec"),
+                    "exec" => Some("noexec"),
+                    "noatime" => Some("atime"),
+                    "atime" => Some("noatime"),
                     "sync" => Some("async"),
                     "async" => Some("sync"),
                     _ => None,
@@ -553,11 +572,21 @@ impl FuseConf {
         Ok(normalized)
     }
 
+    /// Normalize and validate mount options after config-file values, CLI
+    /// overrides, and defaults have been merged for the FUSE process. Generic
+    /// cluster loading deliberately does not call this method because master and
+    /// worker processes do not consume FUSE mount options.
+    pub fn normalize_fuse_opts(&mut self) -> CommonResult<()> {
+        self.fuse_opts = Self::normalized_fuse_opts(&self.fuse_opts)?;
+        Ok(())
+    }
+
     /// Appends options that belong in the FUSE mount data string. Linux VFS
-    /// options (`ro`, `nodev`, `nosuid`, `noexec`, `noatime`, `dirsync`, and
-    /// `sync`) are deliberately handled as `mount(2)` flags by the raw backend.
+    /// options (`ro`/`rw`, `nodev`/`dev`, `nosuid`/`suid`, `noexec`/`exec`,
+    /// `noatime`/`atime`, `dirsync`, and `sync`/`async`) are deliberately handled
+    /// as `mount(2)` flags, or as the absence of a flag, by the raw backend.
     pub fn set_fuse_opts(&self, mount_options: &mut String) -> CommonResult<()> {
-        let opts = Self::normalize_fuse_opts(&self.fuse_opts)?;
+        let opts = Self::normalized_fuse_opts(&self.fuse_opts)?;
         let mut default_permissions_added = false;
 
         // The kernel can distinguish an executable load from a normal read while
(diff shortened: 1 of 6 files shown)


crates/common/curvine-config/src/fuse_conf.rs
```diff
@@ -288,6 +288,8 @@ impl FuseConf {
             );
         }
 
+        self.fuse_opts = Self::normalize_fuse_opts(&self.fuse_opts)?;
```
szbr00: normalize_fuse_opts is invoked from FuseConf::init, and ClusterConf::from always
calls fuse.init() (cluster_conf.rs:99) before FuseMountArgs::get_conf applies
`--options` (mount_args.rs:182, then 316–322).

Two effects:

1. A shared curvine-cluster.toml that still lists previously accepted values
   (`auto_unmount`, `allow_root`, `max_write=...`) now fails to load for
   master/worker, not only curvine-fuse. Those names were fuse2/fuse3 CLI
   defaults; this PR is teaching operators to stop using them.
2. The documented CLI contract ("when provided, replace options from the
   config file") never runs, so `--options allow_other` cannot override a
   legacy TOML list. init() already documents a two-pass sequence so CLI
   overrides are not frozen (tasks_per_mnt); fuse_opts does not follow that.

Keep the allowlist, but apply it after CLI/config merge. set_fuse_opts already
re-validates before mount. Also document the supported names and the
auto_unmount / allow_root / max_write break on the fuse_opts field (it still
says the values are passed through to the kernel).

crates/common/curvine-config/src/fuse_conf.rs
```diff
@@ -478,14 +480,86 @@ impl FuseConf {
...
+                    }
+                    // These names are known FUSE/libfuse options, but the direct
+                    // mount backend cannot implement their required semantics.
+                    "allow_root" | "max_write" | "auto_unmount" => {
+                        return err_box!(
+                            "FUSE mount option '{}' is recognized but not supported by the direct mount backend",
+                            raw_opt
+                        );
+                    }
+                    _ => return err_box!("Unknown FUSE mount option '{}'", raw_opt),
```
szbr00, line 539: The allowlist accepts `async` as the explicit inverse of `sync` (and rejects
the pair), but rejects `rw` as unknown. `rw` is the matching inverse of `ro`
and is a common `-o rw,nosuid,nodev` form; those mounts will now fail.

Accept `rw` as a no-op VFS option and reject `rw`+`ro` the same way as
`async`+`sync`. Same consideration for other standard inverses (`dev`/`suid`/
`exec`/`atime`) if you want mount(8)-compatible input rather than only the
negative flags.

curvine-fuse/src/raw/fuse_pure.rs
```diff
@@ -387,16 +401,84 @@ pub fn fuse_umount_pure(mnt: &Path) -> IOResult<()> {
...
+    use curvine_config::FuseConf;
     use curvine_io::IOError;
     use std::ffi::OsStr;
     use std::fs::{self, File};
     use std::os::unix::ffi::OsStrExt;
     use std::os::unix::io::{AsRawFd, OwnedFd};
     use std::path::{Path, PathBuf};
 
+    #[test]
+    fn mount_builder_maps_security_options_to_vfs_flags() {
```
szbr00: mount_builder_maps_security_options_to_vfs_flags asserts MS_NODEV/NOSUID/NOEXEC.
`sync`, `noatime`, `dirsync`, and `ro` (as a fuse_opt, not the readonly
boolean) are classified as VFS flags in normalize_fuse_opts and passed through
the same builder, but nothing asserts the corresponding MS_* bits.

#1399 asked for final mount flags, and `sync` was a named bug (accepted while
MS_SYNCHRONOUS stayed on). A regression that drops those names from
options_to_flag would still pass. Please assert the flags for that full VFS
set.

crates/common/curvine-config/src/fuse_conf.rs
gangump82: Thanks. Fixed in 1dc4b3dd: FUSE option normalization/validation now runs after config, CLI overrides, and defaults are merged. Generic ClusterConf loading no longer rejects legacy mount-only values, while the final FUSE path and set_fuse_opts still validate them. The fuse_opts field now documents the supported and unsupported names.

crates/common/curvine-config/src/fuse_conf.rs
gangump82: Thanks. Fixed in 1dc4b3dd: rw, dev, suid, exec, and atime are accepted as explicit default/no-op VFS forms. Each positive/negative pair is rejected in either order with an error containing both original option names.

curvine-fuse/src/raw/fuse_pure.rs
gangump82: Thanks. Fixed in 1dc4b3dd: the mount-builder regression test now asserts MS_RDONLY, MS_NODEV, MS_NOSUID, MS_NOEXEC, MS_NOATIME, MS_DIRSYNC, and MS_SYNCHRONOUS, and also verifies that positive/default forms produce no VFS flags or FUSE mount-data entries.

curvine-fuse/src/raw/fuse_pure.rs
```diff
@@ -97,10 +97,40 @@ pub fn options_to_flag(mount_option: &str) -> libc::c_ulong {
...
+
+    // Set the FUSE subtype so /proc/mounts reports `type fuse.curvinefs` instead of
+    // the generic `type fuse` (same mechanism sshfs uses for `fuse.sshfs`). `c_type`
+    // stays "fuse" (the only registered kernel FUSE fs type); the Linux kernel FUSE
+    // module parses `subtype=` from this mount data and sets sb->s_subtype.
+    mount_options.push_str(",subtype=curvinefs");
+
+    // VFS options belong in the mount(2) flag argument, not in FUSE mount data.
+    let mut vfs_options = conf.fuse_opts.join(",");
+    if conf.readonly {
```
szbr00, line 122: `rw` is now an accepted explicit option, but it can still be silently overridden
from another field: when `conf.readonly` is true this block appends `ro`
unconditionally, so `--options rw` (or `fuse_opts = ["rw"]`) combined with
`readonly = true` mounts read-only even though the operator explicitly asked for
rw. This is the same class of silent conflict this PR rejects for ro+rw and
sync+async inside fuse_opts, and it is the only remaining opposite pair because
`readonly` lives outside fuse_opts.

`normalize_fuse_opts(&mut self)` can see `self.readonly`; consider rejecting `rw`
there when readonly is set (or rejecting the pair at the same boundary) so this
combination fails loudly instead of resolving silently toward ro.

gangump82: Thanks. Fixed in 1eb2ae16: explicit `rw` now conflicts with `fuse.readonly=true` after final CLI/config merging, and the same cross-field validation is retained at the mount boundary for defensive coverage. Added config, CLI, and raw mount-builder regression tests.

szbr00 (approved):

PR #1614, opened 2026-08-20
fix(config): trim whitespace in host:port address lists

lzjqsdd:
# Summary

Trim whitespace when parsing comma-separated `host:port` lists and related hostname config. Mixed spacing after commas (common in Hadoop XML / CLI) previously left a leading space on some hostnames, so DNS failed for those nodes only. If that node was the Raft leader, clients treated the whole master set as unavailable even though the cluster was healthy.

# Issue Describe / Design

No tracking issue. Found in production.

- **Root cause:** `master_addrs` (and similar `host:port` lists) were split on `,` / `:` without `trim`. A value such as `h1:8995, h2:8995,h3:8995` produced hostname `" h2"`, which fails DNS. Followers still answered "Not a leader"; the padded leader never connected.
- **Reproduction:** Configure Java/CLI `master_addrs` with a space after only some commas; elect the padded node as leader; client reports all masters unavailable.
- **Fix approach:** Add `InetAddr::parse_list()` and trim at construction, serde, ClusterConf load, Java Hadoop config, Fluid endpoints, and DCM worker lists. Empty tokens are skipped.

# Changes

| `crates/core/net/src/net/inet_addr.rs` | Trim in `new`/`from_str`/serde; add `parse_list()` | Hostnames with surrounding whitespace are normalized |
| `crates/sdk/curvine-sdk-core/src/filesystem_conf.rs` | Parse `master_addrs` via `parse_list` | Java SDK HA list with mixed comma spacing works |
| `curvine-cli/src/main.rs`, `curvine-fuse/src/cli/mount_args.rs` | Use `parse_list` for `--master_addrs` | CLI/FUSE override no longer keeps leading spaces |
| `crates/common/curvine-config/src/cluster_conf.rs` | `normalize_whitespace()` on load; trim hostname env vars | TOML hostnames / `data_dir` / endpoints are trimmed |
| `client_conf.rs`, `raft_peer.rs`, `transfer_conf.rs`, `worker_conf.rs` | Trim hostnames, endpoints, data-dir paths | Same values with padding become valid |
| `curvine-libsdk/java/.../FilesystemConf.java` | Trim Hadoop values; normalize CSV `master_addrs` / transfer endpoints | XML whitespace no longer poisons native TOML |
| `curvine-docker/fluid/config-parse.py` | Strip hostname/port after split | ThinRuntime endpoints tolerate padding |
| `curvine-master/src/master/router_handler.rs` | Trim DCM worker CSV tokens | Query lists ignore empty/padded names |

# Test verified

| `cargo test -p curvine-net --lib` | PASS | `parse_list` mixed spacing + empty tokens |
| `cargo test -p curvine-config --lib` | PASS | TOML hostname / data_dir / endpoints trim |
| `cargo test -p curvine-sdk-core --lib` | PASS | Production-style `master_addrs` string |
| `python3 -m unittest curvine-docker/fluid/tests/test_config_parse.py` | PASS | Fluid mixed comma spacing |
| `make format` | PASS | Pre-commit format + check |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/common/curvine-config/src/raft_peer.rs
+++ b/crates/common/curvine-config/src/raft_peer.rs
@@ -14,16 +14,25 @@
 
 use curvine_net::net::InetAddr;
 use curvine_runtime::common::Utils;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use std::fmt::{Display, Formatter};
 
 pub type NodeId = u64;
+
+fn deserialize_trimmed_string<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value = String::deserialize(deserializer)?;
+    Ok(value.trim().to_string())
+}
 
 // Represents a raft address
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(default)]
 pub struct RaftPeer {
     pub id: NodeId,
+    #[serde(default, deserialize_with = "deserialize_trimmed_string")]
     pub hostname: String,
     pub port: u16,
 }
@@ -65,5 +74,27 @@ impl Default for RaftPeer {
             hostname: "".to_string(),
             port: 0,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::RaftPeer;
+    use curvine_net::net::InetAddr;
+
+    #[test]
+    fn serde_trims_hostname() {
+        let peer: RaftPeer = toml::from_str(
+            r#"
+                id = 1
+                hostname = " master1 "
+                port = 8996
+            "#,
+        )
+        .unwrap();
+
+        assert_eq!(peer.hostname, "master1");
+        assert_eq!(peer.to_addr().hostname, "master1");
+        assert_eq!(peer.to_addr(), InetAddr::new(" master1 ", 8996));
     }
 }
--- a/crates/common/curvine-config/src/raft_peer.rs
+++ b/crates/common/curvine-config/src/raft_peer.rs
@@ -32,13 +32,14 @@ impl RaftPeer {
     pub fn new<T: AsRef<str>>(id: NodeId, hostname: T, port: u16) -> Self {
         Self {
             id,
-            hostname: hostname.as_ref().to_string(),
+            hostname: hostname.as_ref().trim().to_string(),
             port,
         }
     }
 
     pub fn from_addr<T: AsRef<str>>(hostname: T, port: u16) -> Self {
-        let id = Self::create_id(format!("{}{}", hostname.as_ref(), port));
+        let hostname = hostname.as_ref().trim();
+        let id = Self::create_id(format!("{}{}", hostname, port));
         Self::new(id, hostname, port)
     }
 
--- a/crates/common/curvine-config/src/transfer_conf.rs
+++ b/crates/common/curvine-config/src/transfer_conf.rs
@@ -121,6 +121,13 @@ impl TransferConf {
     pub const DEFAULT_TASK_MAX_RETRIES: usize = 3;
 
     pub fn init(&mut self) -> FsResult<()> {
+        self.hostname = self.hostname.trim().to_string();
+        self.endpoints = self
+            .endpoints
+            .iter()
+            .map(|endpoint| endpoint.trim().to_string())
+            .filter(|endpoint| !endpoint.is_empty())
+            .collect();
         self.infer_store_url();
         if self.endpoints.is_empty() {
             self.endpoints
--- a/crates/common/curvine-config/src/transfer_conf.rs
+++ b/crates/common/curvine-config/src/transfer_conf.rs
@@ -122,6 +122,7 @@ impl TransferConf {
 
     pub fn init(&mut self) -> FsResult<()> {
         self.hostname = self.hostname.trim().to_string();
+        let had_configured_endpoints = !self.endpoints.is_empty();
         self.endpoints = self
             .endpoints
             .iter()
@@ -130,6 +131,11 @@ impl TransferConf {
             .collect();
         self.infer_store_url();
         if self.endpoints.is_empty() {
+            if had_configured_endpoints {
+                return Err(FsError::common(
+                    "transfer.endpoints must contain host:port values; whitespace-only entries are invalid",
+                ));
+            }
             self.endpoints
                 .push(format!("{}:{}", self.hostname, self.rpc_port));
         }
@@ -424,6 +430,20 @@ mod tests {
         assert_eq!(conf.rpc_port, TransferConf::DEFAULT_RPC_PORT);
         assert_eq!(conf.web_port, TransferConf::DEFAULT_WEB_PORT);
         assert_eq!(conf.log.file_name, "transfer.log");
+    }
+
+    #[test]
+    fn rejects_whitespace_only_endpoints() {
+        let mut conf = TransferConf {
+            endpoints: vec![" ".to_string()],
+            ..Default::default()
+        };
+
+        let err = conf.init().unwrap_err().to_string();
+        assert!(
+            err.contains("transfer.endpoints must contain host:port values"),
+            "unexpected error: {err}"
+        );
     }
 
     #[test]
--- a/crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
+++ b/crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
@@ -197,7 +197,13 @@ impl FilesystemConf {
 
         let master_addrs = match InetAddr::parse_list(&self.master_addrs) {
             Ok(addrs) => addrs,
-            Err(_) => return err_box!("wrong format fs.cv.master_addrs {}", self.master_addrs),
+            Err(e) => {
+                return err_box!(
+                    "wrong format fs.cv.master_addrs {}: {}",
+                    self.master_addrs,
+                    e
+                );
+            }
         };
 
         let client = ClientConf {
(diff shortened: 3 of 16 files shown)


crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
```diff
@@ -191,24 +191,18 @@ impl FilesystemConf {
...
-            if vec.len() != 2 {
-                return err_box!("wrong format fs.cv.master_addrs {}", self.master_addrs);
-            }
-            let hostname = vec[0].to_string();
-            let port: u16 = vec[1].parse()?;
-            master_addrs.push(InetAddr::new(hostname, port));
-        }
+        let master_addrs = match InetAddr::parse_list(&self.master_addrs) {
+            Ok(addrs) => addrs,
+            Err(_) => return err_box!("wrong format fs.cv.master_addrs {}", self.master_addrs),
```
szbr00: `InetAddr::parse_list` error is discarded here (`Err(_)`), so a misconfigured
`fs.cv.master_addrs` surfaces only as a generic "wrong format" message with no
indication of which token or port failed. The CLI path in
`curvine-cli/src/main.rs` includes the parse error in its message
(`...: {}", master_addrs, e`). For consistency and easier diagnosis, consider
propagating the error here too, e.g.
`Err(e) => return err_box!("wrong format fs.cv.master_addrs {}: {}", self.master_addrs, e)`.

curvine-fuse/src/cli/mount_args.rs
```diff
@@ -298,16 +298,10 @@ impl FuseMountArgs {
...
-                if tmp.len() != 2 {
-                    return err_box!("wrong format master_addrs {}", master_addrs);
-                }
-                let hostname = tmp[0].to_string();
-                let port: u16 = tmp[1].parse()?;
-                vec.push(InetAddr::new(hostname, port));
-            }
+            let vec = match InetAddr::parse_list(master_addrs) {
+                Ok(vec) => vec,
+                Err(_) => return err_box!("wrong format master_addrs {}", master_addrs),
```
szbr00: Same as the SDK path: `InetAddr::parse_list` error is dropped (`Err(_)`), so
FUSE mount failures from a bad `--master_addrs` give only "wrong format
master_addrs" without the underlying reason. The CLI path already includes
the parse error in its message. Align this site by propagating it, e.g.
`Err(e) => return err_box!("wrong format master_addrs {}: {}", master_addrs, e)`.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
```diff
@@ -191,24 +191,18 @@ impl FilesystemConf {
...
+        };
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in 0cd75529: `InetAddr::parse_list` errors are now included in the `fs.cv.master_addrs` message, matching the CLI path. Added a unit test that a bad port surfaces the underlying parse error.

curvine-fuse/src/cli/mount_args.rs
```diff
@@ -298,16 +298,10 @@ impl FuseMountArgs {
...
+            };
```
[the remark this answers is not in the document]
lzjqsdd: Fixed in 0cd75529: `parse_list` errors are now included in the FUSE `master_addrs` message. Added a unit test that a bad `--master-addrs` value surfaces the underlying parse error.

crates/sdk/curvine-sdk-core/src/filesystem_conf.rs
lzjqsdd: Fixed in 0cd75529: the SDK path now propagates the parse error as `wrong format fs.cv.master_addrs {}: {}`, matching `curvine-cli`. Covered by `into_cluster_conf_includes_parse_error_for_invalid_master_addrs`.

curvine-fuse/src/cli/mount_args.rs
lzjqsdd: Fixed in 0cd75529: the FUSE path now propagates the parse error as `wrong format master_addrs {}: {}`. Covered by `get_conf_includes_parse_error_for_invalid_master_addrs`.

curvine-libsdk/java/src/main/java/io/curvine/FilesystemConf.java
```diff
@@ -147,10 +147,15 @@ public FilesystemConf(Configuration conf) throws IllegalAccessException {
             }
 
             field.setAccessible(true);
+            value = value.trim();
             String fieldType = field.getType().getName();
             switch (fieldType) {
                 case "java.lang.String":
-                    field.set(this, value);
+                    if ("master_addrs".equals(field.getName())) {
```
szbr00, line 153: The constructor now `normalizeCsv`s `fs.cv.master_addrs`, but that is not the only Hadoop entry path.

`CurvineFileSystem.initialize` overwrites the field afterwards:

```java
filesystemConf = new FilesystemConf(conf);
if (StringUtils.isNotEmpty(authority)) {
    filesystemConf.master_addrs = getMasterAddrs(authority); // raw fs.cv.{ns}.master_addrs
}
```

`getMasterAddrs` returns `conf.get(...)` with no trim/`normalizeCsv`. That is the `cv://cluster/` namespaced HA path. Native `parse_list` will still connect, but:

- `toToml()` re-injects padded CSV into the native config string this PR claims to stop poisoning
- `MOUNT_CACHE` is keyed on the raw string, so `"h1:8995, h2:8995"` and `"h1:8995,h2:8995"` create two native runtimes

The same constructor also sets `client_hostname = getClientHostname()` from `POD_IP` / `CURVINE_CLIENT_HOSTNAME` **before** `value.trim()`. Rust now trims those env vars; Java does not. A trailing newline in `POD_IP` is written by `toToml()` as a raw newline inside quotes and can fail TOML parse.

Apply `normalizeCsv` in `getMasterAddrs` (and trim in `getClientHostname`) so every Java path matches the constructor.

crates/common/curvine-config/src/raft_peer.rs
```diff
@@ -32,13 +32,14 @@ impl RaftPeer {
     pub fn new<T: AsRef<str>>(id: NodeId, hostname: T, port: u16) -> Self {
         Self {
             id,
-            hostname: hostname.as_ref().to_string(),
+            hostname: hostname.as_ref().trim().to_string(),
```
szbr00, line 44: `RaftPeer::new` / `from_addr` now trim, but serde still assigns `hostname` directly. TOML `journal_addrs` therefore keep padding until `ClusterConf::normalize_whitespace`.

That is load-path dependent, and it is now unsafe: `JournalConf::local_addr()` uses `InetAddr::new`, which always trims, while `RaftGroup::get_node_id` compares `peer.hostname == addr.hostname` against the serde value.

If `journal_addrs` are ever deserialized without `from()` / `from_transfer()` (or `normalize_whitespace` is skipped), local addr is `"master1"` and the peer is `" master1 "` → `"Not a master role"` at startup.

`InetAddr` already uses `deserialize_with` trim. Do the same on `RaftPeer.hostname` so identity does not depend on one ClusterConf helper.

crates/common/curvine-config/src/transfer_conf.rs
```diff
@@ -121,6 +121,13 @@ impl TransferConf {
     pub const DEFAULT_TASK_MAX_RETRIES: usize = 3;
 
     pub fn init(&mut self) -> FsResult<()> {
+        self.hostname = self.hostname.trim().to_string();
+        self.endpoints = self
+            .endpoints
+            .iter()
+            .map(|endpoint| endpoint.trim().to_string())
+            .filter(|endpoint| !endpoint.is_empty())
```
szbr00, line 130: Whitespace-only endpoints are dropped, then empty lists are replaced with `{hostname}:{rpc_port}`.

`endpoints = [" "]` used to fail later in `NodeAddr::from_str`. After this change it silently talks to localhost. Treat whitespace-only the same as an invalid endpoint (error), or only default when the list was omitted, not when it was present and empty after trim.

curvine-libsdk/java/src/main/java/io/curvine/FilesystemConf.java
lzjqsdd: Fixed in a91b9687: namespaced `fs.cv.{ns}.master_addrs` now goes through `normalizeCsv`, and `POD_IP` / `CURVINE_CLIENT_HOSTNAME` are trimmed before `toToml()`. Covered by `namespacedMasterAddrsAreNormalized` and `clientHostnameEnvValuesAreTrimmed`.

crates/common/curvine-config/src/raft_peer.rs
lzjqsdd: Fixed in a91b9687: `RaftPeer.hostname` now uses a trim `deserialize_with`, matching `InetAddr`. Identity no longer depends on `ClusterConf::normalize_whitespace`. Covered by `serde_trims_hostname`.

crates/common/curvine-config/src/transfer_conf.rs
lzjqsdd: Fixed in a91b9687: whitespace-only `transfer.endpoints` is now an error. The localhost default still applies only when the list was omitted or empty. Covered by `rejects_whitespace_only_endpoints` and `rejects_whitespace_only_transfer_endpoints`.

curvine-libsdk/java/src/test/java/io/curvine/LibFsTest.java
```diff
@@ -38,6 +38,33 @@ public void conf() throws Exception {
...
+        conf.set("fs.cv.io_threads", " 16 ");
+
+        FilesystemConf filesystemConf = new FilesystemConf(conf);
+
+        assert filesystemConf.master_addrs.equals("localhost:9001,localhost:9002,localhost:9003");
+        assert filesystemConf.io_threads == 16;
+    }
+
+    @Test
+    public void namespacedMasterAddrsAreNormalized() {
```
szbr00: This test would still pass if the actual fix were reverted: it only asserts on
`FilesystemConf.normalizeCsv` in isolation, while the regression lived in
`CurvineFileSystem.getMasterAddrs` returning the raw `fs.cv.{ns}.master_addrs`
value. If someone removes the `normalizeCsv(addrs)` call there, nothing fails.

Consider making `getMasterAddrs` package-private and asserting through it with
a `Configuration` carrying a padded `fs.cv.ns1.master_addrs` — that exercises
the real seam without needing a live cluster (`initialize` itself requires a
native mount, so testing the helper directly is reasonable, but the assertion
should cover the method that was broken).

lzjqsdd: Fixed in fce7f436: `getMasterAddrs` is package-private. `namespacedMasterAddrsAreNormalized` now sets `fs.cv.ns1.master_addrs` with mixed comma spacing and asserts through `getMasterAddrs(\"ns1\")`.

szbr00 (approved):

PR #1615, opened 2026-08-21
fix(sdk): derive CurvineFsStat used from allocatable view (#1460)

### LINKED ISSUE #1460: [BUG]: FUSE statfs includes non-writable workers in available capacity (by LiAuto)

gangump82:
# Summary

Follow-up to #1610 (which closed #1460). The merged `CurvineFsStat` constructor passed `info.getFsUsed()` as the Hadoop `FsStatus` used, while `Capacity` and `Remaining` already use the Live-only allocatable view. `fs_used` sums every non-lost worker (including Blacklist/Decommission), so when non-writable workers still hold data, `getUsed()` could exceed `getCapacity() - getRemaining()`, yielding a misleading usage ratio. This derives `used = max(0, capacity - remaining)` so the `FsStatus` triple is self-consistent.

# Issue Describe / Design

- Root cause: mixed scopes inside one FsStatus — used was all-workers, capacity/remaining were Live-only.
- Fix: derive used from the same allocatable view, mirroring the Rust cv df logic from #1610.
- Trade-off: FsStatus.getUsed() no longer equals the master-reported total fs_used; that value is still surfaced in simple() for the physical-usage view. Spark write decision uses getRemaining() (unchanged, already correct).

# Changes

| curvine-libsdk/java/.../CurvineFsStat.java | Constructor derives used = max(0, capacity - remaining) instead of info.getFsUsed() | FsStatus.getUsed() now Live-only-consistent; may differ from master total fs_used (still in simple()) |

# Test verified

| Java build (mvn) | pending CI | No local mvn; CI regenerates proto bindings and compiles CurvineFsStat |

# Dependencies

- Follow-up to: #1610 (merged)
- Related issues: #1460 (closed)

### DIFF
--- a/curvine-libsdk/java/src/main/java/io/curvine/CurvineFsStat.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/CurvineFsStat.java
@@ -24,23 +24,10 @@ public class CurvineFsStat extends FsStatus {
     private final GetFilesystemInfoResponse info;
 
     public CurvineFsStat(GetFilesystemInfoResponse info) {
-        // Report the allocatable (writable) view so Hadoop/Spark remaining-space
-        // checks only see Live workers. Legacy masters omit tags 15/16; protobuf
-        // returns 0 for an absent optional int64 with default=0, so we must guard
-        // with hasAllocatableCapacity()/hasAllocatableAvailable() and fall back to
-        // the aggregate totals — otherwise a new client against an old master
-        // would report zero free space.
-        //
-        // Used is derived as Capacity - Remaining rather than info.getFsUsed()
-        // (which sums every non-lost worker, including Blacklist/Decommission).
-        // Capacity/Remaining are Live-only, so deriving Used keeps the FsStatus
-        // triple self-consistent (Used == Capacity - Remaining) even when
-        // non-writable workers still hold data — otherwise getUsed() could
-        // exceed getCapacity() - getRemaining() and report a misleading ratio.
-        // The master-reported total fs_used is still surfaced in simple().
-        long capacity = info.hasAllocatableCapacity() ? info.getAllocatableCapacity() : info.getCapacity();
-        long remaining = info.hasAllocatableAvailable() ? info.getAllocatableAvailable() : info.getAvailable();
-        super(capacity, Math.max(0, capacity - remaining), remaining);
+        // super(...) must be the first statement, so the allocatable fallback
+        // and used derivation live in static helpers below. See the comment on
+        // allocatableCapacity/allocatableAvailable for the rationale.
+        super(allocatableCapacity(info), allocatableUsed(info), allocatableRemaining(info));
         this.info = info;
     }
 
@@ -204,5 +191,34 @@ public class CurvineFsStat extends FsStatus {
         }
 
         return builder.toString();
+    }
+
+    // The allocatable (writable) view: capacity/available eligible for new
+    // writes (Live workers only). Legacy masters omit tags 15/16; protobuf
+    // returns 0 for an absent optional int64 with default=0, so we must guard
+    // with hasAllocatableCapacity()/hasAllocatableAvailable() and fall back to
+    // the aggregate totals — otherwise a new client against an old master
+    // would report zero free space.
+    //
+    // Used is derived as Capacity - Remaining rather than info.getFsUsed()
+    // (which sums every non-lost worker, including Blacklist/Decommission).
+    // Capacity/Remaining are Live-only, so deriving Used keeps the FsStatus
+    // triple self-consistent (Used == Capacity - Remaining) even when
+    // non-writable workers still hold data — otherwise getUsed() could
+    // exceed getCapacity() - getRemaining() and report a misleading ratio.
+    // The master-reported total fs_used is still surfaced in simple().
+    // These are static so they can run before super(...) completes.
+    private static long allocatableCapacity(GetFilesystemInfoResponse info) {
+        return info.hasAllocatableCapacity() ? info.getAllocatableCapacity() : info.getCapacity();
+    }
+
+    private static long allocatableRemaining(GetFilesystemInfoResponse info) {
+        return info.hasAllocatableAvailable() ? info.getAllocatableAvailable() : info.getAvailable();
+    }
+
+    private static long allocatableUsed(GetFilesystemInfoResponse info) {
+        long capacity = allocatableCapacity(info);
+        long remaining = allocatableRemaining(info);
+        return Math.max(0, capacity - remaining);
     }
 }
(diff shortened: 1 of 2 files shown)


curvine-libsdk/java/src/main/java/io/curvine/CurvineFsStat.java
```diff
@@ -30,11 +30,17 @@ public CurvineFsStat(GetFilesystemInfoResponse info) {
...
+        // Used is derived as Capacity - Remaining rather than info.getFsUsed()
+        // (which sums every non-lost worker, including Blacklist/Decommission).
+        // Capacity/Remaining are Live-only, so deriving Used keeps the FsStatus
+        // triple self-consistent (Used == Capacity - Remaining) even when
+        // non-writable workers still hold data — otherwise getUsed() could
+        // exceed getCapacity() - getRemaining() and report a misleading ratio.
+        // The master-reported total fs_used is still surfaced in simple().
+        long capacity = info.hasAllocatableCapacity() ? info.getAllocatableCapacity() : info.getCapacity();
+        long remaining = info.hasAllocatableAvailable() ? info.getAllocatableAvailable() : info.getAvailable();
+        super(capacity, Math.max(0, capacity - remaining), remaining);
```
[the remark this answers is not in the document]
gangump82: fixed in `2cb0e3a3`. The `super(...)` call is now the first statement; the allocatable fallback and `used = max(0, capacity - remaining)` derivation moved into static helpers (`allocatableCapacity` / `allocatableRemaining` / `allocatableUsed`) so they can run before `super(...)` completes, with no ternary repetition.

curvine-libsdk/java/src/main/java/io/curvine/CurvineFsStat.java
```diff
@@ -24,17 +24,10 @@ public class CurvineFsStat extends FsStatus {
...
-        // with hasAllocatableCapacity()/hasAllocatableAvailable() and fall back to
-        // the aggregate totals — otherwise a new client against an old master
-        // would report zero free space.
-        super(
-                info.hasAllocatableCapacity() ? info.getAllocatableCapacity() : info.getCapacity(),
-                info.getFsUsed(),
-                info.hasAllocatableAvailable() ? info.getAllocatableAvailable() : info.getAvailable()
-        );
+        // super(...) must be the first statement, so the allocatable fallback
+        // and used derivation live in static helpers below. See the comment on
```
szbr00, line 28: This regression fix (mixed-scope `getUsed()` from #1610) ships without a test,
so the same defect could re-introduce silently. The Java SDK already has
proto-level unit tests (e.g. `SetAttrOptsTest`); a `CurvineFsStatTest` would
fit the same pattern and lock in the invariant.

Suggested cases:
- Allocatable present: build a `GetFilesystemInfoResponse` with
  capacity=1000, available=400, allocatable_capacity=600,
  allocatable_available=200, fs_used=300. Assert getCapacity()==600,
  getRemaining()==200, getUsed()==400 (== max(0, 600-200)), and that
  getUsed() <= getCapacity() - getRemaining() no longer overflows when
  Blacklist/Decommission workers hold the other 100.
- Legacy master (no allocatable fields): assert it falls back to
  capacity/available and still derives used = max(0, capacity - remaining).

This mirrors the assertion already enforced on the Rust side via the
`df` path and `get_filesystem_info_compat_test.rs`.

gangump82: Added `CurvineFsStatTest` in `75d8c61f` covering both cases:

- `allocatableViewIsSelfConsistentAndIgnoresFsUsed`: capacity=1000/available=400 (all workers), allocatable_capacity=600/allocatable_available=200, fs_used=300 (distractor). Asserts getCapacity()==600, getRemaining()==200, getUsed()==400 (== max(0, 600-200)) and explicitly that getUsed() != fs_used, so the derived value can't silently regress to the master's all-workers fs_used.
- `legacyMasterFallsBackToTotalAndStillDerivesUsed`: allocatable fields omitted → asserts hasAllocatableCapacity()==false, falls back to capacity/available, and still derives used = max(0, capacity - remaining).

A shared `assertSelfConsistent` helper pins the invariant `used == max(0, capacity - remaining)` for both. Mirrors `get_filesystem_info_compat_test.rs` / the `cv df` path. Java run depends on CI (no local mvn).

szbr00: The CurvineFsStatTest in 75d8c61f addresses this. Both cases hit the constructor entry path and the assertions fail if getUsed() falls back to fs_used or if the legacy allocatable fallback is dropped. No further comment from me.

lzjqsdd (approved):

PR #1618, opened 2026-08-24
feat(discovery): add etcd service discovery

yangcx000:
# Summary

Adds a reusable service discovery foundation for Curvine backed by etcd. The new crate provides common endpoint schema, registry/resolver traits, etcd list/watch/lease registration support, and test-only fake discovery utilities for future dynamically scalable services such as MDS.

# Issue Describe / Design

Related to the service discovery design in `etcd-service-discovery-design.md`.

- Keep service discovery generic and not tied to MDS-specific config or routing logic.
- Gate the etcd implementation behind the `etcd` feature so default builds do not pull etcd/tonic dependencies.
- Rely on etcd lease/keepalive for liveness; endpoint picking, consistent hashing, RPC failure blacklist, and business routing stay in callers.
- Do not include username/password/TLS options in the first phase because etcd is expected to run inside the cluster network.

# Changes

| `crates/adapters/curvine-service-discovery` | Adds service discovery traits, endpoint key/value schema, etcd provider, registration guard, watch cache, and tests | None by default; etcd provider is feature-gated |
| `crates/common/curvine-config` | Adds `[discovery]` config with validation and ClusterConf integration | None when discovery is disabled, which is the default |
| `Cargo.toml` / `Cargo.lock` | Adds workspace crate and optional `etcd-client` dependency | Default feature set remains lightweight |
| `etc/*.toml` and docker deployment configs | Documents disabled-by-default discovery settings | Existing deployments remain unchanged |
| `etcd-service-discovery-design.md` | Adds the service discovery design and implementation status | Documentation only |

# Test verified

| `make format` | PASS | Runs project formatting and Clippy checks |
| `cargo test -p curvine-service-discovery` | PASS | Unit tests without etcd feature |
| `cargo test -p curvine-service-discovery --features etcd` | PASS | Includes etcd integration tests when `CURVINE_ETCD_ADDR` is available |
| `cargo test -p curvine-config discovery` | PASS | Discovery config tests |
| `git diff --check` | PASS | Whitespace check |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None


crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+                tokio::select! {
+                    changed = shutdown_rx.changed() => {
+                        if changed.is_err() || *shutdown_rx.borrow() {
+                            return;
+                        }
+                    }
+                    result = keeper.keep_alive() => {
+                        if let Err(error) = result {
+                            mark_keepalive_lost(&status_tx, &kind, &service_id, error.to_string());
+                            return;
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: keepalive now tolerates transient failures with a consecutive-failure threshold before reporting KeepAliveLost.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+                    .with_prefix()
+                    .with_start_revision(start_revision)
+                    .with_progress_notify(),
+            ),
+        )
+        .await?;
+    let mut snapshot = reader.cached_snapshot().await;
+    let mut watch_confirmed = false;
+
+    while let Some(response) = stream.message().await? {
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: watch reconnect sleeps and stream reads now observe receiver closure so dropping the resolver handle can stop the background watch task.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+            .await?;
+        let revision = response
+            .header()
+            .map(|header| header.revision())
+            .unwrap_or(0);
+        let mut endpoints = Vec::new();
+        for kv in response.kvs() {
+            let key = kv
+                .key_str()
+                .map_err(|error| DiscoveryError::InvalidEndpointValue(error.to_string()))?;
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: list now logs and skips malformed non-UTF-8 keys instead of failing the whole discovery query.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+        }
+        for event in events {
+            match event.event_type() {
+                EventType::Put => {
+                    let Some(kv) = event.kv() else {
+                        continue;
+                    };
+                    let key = kv
+                        .key_str()
+                        .map_err(|error| DiscoveryError::InvalidEndpointValue(error.to_string()))?;
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: watch PUT and DELETE paths now drop malformed non-UTF-8 keys and continue processing later events.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+        .saturating_mul(multiplier)
+        .min(config.watch_reconnect_max_ms);
+    let jitter_bound = (base as f64 * config.watch_reconnect_jitter_ratio).round() as u64;
+    if jitter_bound == 0 {
+        base
+    } else {
+        let jitter = now_millis()
+            .wrapping_add(jitter_seed)
+            .wrapping_add(u64::from(attempt))
+            % (jitter_bound + 1);
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: reconnect jitter now uses saturating arithmetic so max-range configs cannot overflow or divide by zero.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+                "watch canceled: {}",
+                response.cancel_reason()
+            )));
+        }
+
+        let is_create_response = response.created();
+        let events = response.events();
+        if is_create_response && events.is_empty() {
+            continue;
+        }
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: watch creation/progress responses now confirm the cache and clear stale state instead of being skipped.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,872 @@
...
+        .await?;
+        let lease_id = lease.id();
+
+        if let Err(error) = with_registration_timeout(
+            timeout_duration,
+            client.put(
+                key.clone(),
+                value,
+                Some(PutOptions::new().with_lease(lease_id)),
+            ),
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: registration is now an atomic create-if-absent transaction and updates compare the current key lease before writing.

crates/adapters/curvine-service-discovery/src/resolver.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub(crate) async fn cached_snapshot(&self) -> ServiceSnapshot {
+        self.cache.read().await.clone()
+    }
+
+    #[allow(dead_code)]
+    pub(crate) async fn mark_stale(&self, last_update_ms: u64) {
+        let mut snapshot = self.cache.write().await;
+        snapshot.stale = true;
+        snapshot.last_update_ms = last_update_ms;
+    }
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: marking a cache stale no longer overwrites last_update_ms, preserving the last successful update timestamp.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -56,6 +56,8 @@ pub struct ClusterConf {
 
     pub worker: WorkerConf,
 
+    pub discovery: DiscoveryConf,
```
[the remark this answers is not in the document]
yangcx000: Thanks. Fixed in b5977a13: moved the new ClusterConf discovery field to append-only position after existing fields.

yangcx000:
Addressed the review feedback in b5977a13. In addition to the inline threads, I also handled the related suppressed notes: removed FsError coupling from the discovery crate, added InvalidDiscoveryConfig/RegistrationAlreadyExists errors, kept runtime ownership with guards/handles, and documented duplicate-registration and malformed-key behavior. Validation passed: make format; cargo test -p curvine-service-discovery; cargo test -p curvine-service-discovery --features etcd; cargo test -p curvine-config discovery; git diff --check.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,994 @@
...
+    service_id: &str,
+    message: String,
+) {
+    log::warn!(
+        "etcd discovery registration keepalive lost: kind={}, service_id={}, error={}",
+        kind,
+        service_id,
+        message
+    );
+    let _ = status_tx.send(RegistrationStatus::KeepAliveLost { message });
```
szbr00: On terminal keepalive loss this path only publishes KeepAliveLost and
returns. It does not revoke the lease, so the etcd key stays bound to
the dying lease until TTL expiry.

register() then uses Compare::version(key, Equal, 0) (line 71). A
still-running process that follows the documented recovery
(RegistrationGuard docs / design: after KeepAliveLost, call
register() again for a new lease) therefore gets
RegistrationAlreadyExists for the remaining TTL. shutdown() can
revoke, but that is not part of the advertised recovery path, and
callers that drop the guard and re-register will still collide.

On KeepAliveLost, best-effort lease_revoke so the key disappears
before the caller retries register(). If revoke fails because etcd is
down, AlreadyExists until TTL is unavoidable — document that case
explicitly.

szbr00 (commented):
The etcd provider is the core of this PR (~1000 lines) but it is
cfg(feature = "etcd") and that feature is never enabled in CI.
`.github/workflows/build.yml` runs `cargo clippy --all-targets` and
`cargo build` without `--features etcd` / `--all-features`, so
etcd_provider.rs (including its unit tests for reconnect delay and
config validation) is not compiled or linted on the PR gate.

The advertised recovery behavior is also untested: watch cancel /
compaction -> Reset, transient disconnect -> stale + resume from
revision+1, and allow_stale_cache true/false. Integration tests skip
when CURVINE_ETCD_ADDR is unset, so those paths can regress without a
failing default test.

Minimum: add a CI step
`cargo clippy -p curvine-service-discovery --features etcd --all-targets`
and `cargo test -p curvine-service-discovery --features etcd` (existing
tests already skip without etcd). Please also add a deterministic
assertion for compact/reconnect/stale — a unit test of watch_loop
error handling is enough if a live etcd sidecar is out of scope.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,994 @@
...
+                }
+            }
+            _ = interval.tick() => {
+                tokio::select! {
+                    changed = shutdown_rx.changed() => {
+                        if changed.is_err() || *shutdown_rx.borrow() {
+                            return;
+                        }
+                    }
+                    result = keeper.keep_alive() => {
```
szbr00: keepalive_loop retries transient keep_alive()/stream errors up to
KEEPALIVE_FAILURE_THRESHOLD, but it never recreates LeaseKeeper /
LeaseKeepAliveStream. After the first connection drop the stream is
dead, so the next two attempts fail on the same broken objects and
the task exits with KeepAliveLost.

Call chain: EtcdServiceResolver::register -> keepalive_loop ->
keeper.keep_alive() / stream.message(). With the default 3s interval
and threshold 3, one etcd blip permanently stops renewals even though
the lease (TTL 10s) may still be valid. The design retry-then-threshold
contract requires recreating the keepalive session (e.g.
client.lease_keep_alive(lease_id)) while the lease is still alive, and
only treating ttl<=0 or a failed recreate as terminal loss.

yangcx000:
I did another systematic self-review and pushed 813d9b96 with additional hardening beyond the previous review fixes: re-exported RegistrationControl so the public guard constructor is usable downstream, aligned the test fake with real duplicate-registration semantics, ignored unknown DELETE events instead of emitting false Removed events, made watch creation confirmation avoid advancing the resume revision, sorted endpoint snapshots by id for deterministic caller behavior, tightened direct EtcdDiscoveryConfig validation, and ensured shutdown publishes Revoked status after revoke attempts. Validation passed locally: make format; cargo test -p curvine-service-discovery; cargo test -p curvine-service-discovery --features etcd; cargo test -p curvine-config discovery; git diff --check.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,1047 @@
...
+                response.cancel_reason()
+            );
+            return Err(DiscoveryError::EtcdUnavailable(format!(
+                "watch canceled: {}",
+                response.cancel_reason()
+            )));
+        }
+
+        let events = response.events();
+        if response.created() && events.is_empty() {
```
szbr00, line 978: The new created() && events.is_empty() branch confirms the cache and
continues before reading response.header(). That is a regression vs
b5977a13, which applied the create/progress header revision and then
wrote it on confirm.

After this, a quiet watch keeps cache.revision at the original list
revision. Later progress notifies update only the local snapshot
(watch_confirmed is already true, so replace_snapshot is skipped).
watch_loop then reconnects from cached_snapshot().revision + 1, which
is hours/days stale. etcd compaction turns every quiet disconnect into
WatchCompacted + full Reset, instead of resuming from the last observed
revision as the design requires.

Apply header.revision() before the created() continue, and persist
progress-only header revisions to the cache so reconnect can start from
a live revision.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,1047 @@
...
+
+    async fn shutdown(&self) -> DiscoveryResult<()> {
+        let _ = self.status_tx.send(RegistrationStatus::Revoking);
+        let _ = self.shutdown_tx.send(true);
+        let mut client = self.client.lock().await;
+        let result = client
+            .lease_revoke(self.lease_id)
+            .await
+            .map_err(DiscoveryError::from);
+        let _ = self.status_tx.send(RegistrationStatus::Revoked);
```
szbr00: shutdown() now always sends RegistrationStatus::Revoked, then returns
the lease_revoke error. Subscribers therefore observe a successful
teardown while the etcd key may still be bound to the lease.

Callers that treat Revoked as "key is gone" will immediately
re-register and hit RegistrationAlreadyExists (Compare::version == 0
still fails). Keep Revoking until revoke succeeds; only publish
Revoked after a successful lease_revoke. On revoke failure, leave the
status as Revoking (or a failed-revoke signal) and return the error so
the caller can retry.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
```diff
@@ -0,0 +1,1047 @@
...
+        .and_then(vec![TxnOp::put(
+            key,
+            value,
+            Some(PutOptions::new().with_lease(lease_id)),
+        )]);
+    let response = client.txn(txn).await.map_err(DiscoveryError::from)?;
+    if response.succeeded() {
+        Ok(())
+    } else {
+        Err(DiscoveryError::RegistrationLost(format!(
```
szbr00, line 264: put_endpoint_if_owned returns RegistrationLost when the key is no
longer on this lease, but EtcdRegistrationControl::update_endpoint /
update_status do not publish KeepAliveLost. status_rx stays
Registered until keepalive_loop happens to notice.

subscribe_status() is the advertised liveness signal. A process that
only watches status will keep serving after update_* already proved
the registration is gone. On this txn failure, send KeepAliveLost
(and stop keepalive) so status and the sync API agree.

crates/adapters/curvine-service-discovery/src/resolver.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    pub fn kind(&self) -> &ServiceKind {
+        &self.kind
+    }
+
+    pub fn reader(&self) -> SnapshotReader {
+        self.reader.clone()
+    }
+
+    pub fn into_parts(self) -> (SnapshotReader, ServiceWatch) {
```
szbr00, line 83: into_parts(self) returns (reader, events) and drops
ServiceResolverHandle._lifecycle_owner. SnapshotReader has its own
owner, so keeping only the event receiver lets the last Runtime Arc
drop while watch_loop is still supposed to run; the task is then
shut down and next_event() goes silent.

Transfer the handle's lifecycle owner onto the returned reader (or
document that both parts must be kept, and that dropping the reader
cancels the watch).

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
yangcx000: Thanks. Fixed in 896012c5: terminal KeepAliveLost now stops keepalive and best-effort revokes the lease; if revoke cannot reach etcd, retry may still see RegistrationAlreadyExists until TTL expiry.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
yangcx000: Thanks. Fixed in 896012c5: keepalive failures now recreate the lease_keep_alive session with the shared client instead of retrying the same dead keeper/stream.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
yangcx000: Thanks. Fixed in 896012c5: watch responses now apply header revision before the created branch and persist progress/header-only revisions to the cached snapshot.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
yangcx000: Thanks. Fixed in 896012c5: shutdown now publishes Revoked only after lease_revoke succeeds; revoke errors are returned while status remains Revoking.

crates/adapters/curvine-service-discovery/src/etcd_provider.rs
yangcx000: Thanks. Fixed in 896012c5: update_endpoint/update_status now publish KeepAliveLost, stop keepalive, and best-effort revoke when the ownership txn proves registration is lost.

crates/adapters/curvine-service-discovery/src/resolver.rs
yangcx000: Thanks. Fixed in 896012c5: ServiceWatch is now a wrapper that carries the lifecycle owner, so into_parts keeps the watch runtime alive when callers keep only the event stream.

.github/workflows/build.yml
```diff
@@ -86,6 +86,9 @@ jobs:
           echo "Running clippy (static code analysis)"
           # Run clippy in debug mode to match build
           cargo clippy --all-targets --jobs ${{ env.CARGO_BUILD_JOBS }} -- --${CLIPPY_LEVEL}=warnings --allow clippy::uninlined-format-args
+
+          echo "Running etcd service discovery clippy"
+          cargo clippy -p curvine-service-discovery --features etcd --tests --jobs ${{ env.CARGO_BUILD_JOBS }} -- --${CLIPPY_LEVEL}=warnings --allow clippy::uninlined-format-args
```
szbr00, line 91: CI now clippy/builds `curvine-service-discovery --features etcd --tests`,
which closes the compile gap. It still never runs
`cargo test -p curvine-service-discovery --features etcd`.

The unit tests inside etcd_provider.rs (reconnect delay, upsert/sort,
config validation) do not need a live etcd; they are simply cfg'd
behind the feature, so the default workspace test job never executes
them. The integration tests are already #[ignore] and will stay skipped.

Please add that cargo test step (without --ignored) so a regression in
the feature-gated unit tests fails the gate.

yangcx000: Thanks. Fixed in 9ef8029d: CI now runs `cargo test -p curvine-service-discovery --features etcd` without `--ignored`, so the feature-gated unit tests run in the gate while live-etcd integration tests remain explicitly ignored.

szbr00 (approved):

Issue #1619, opened 2026-08-25
[FEATURE] Unify component configuration loading, validation, and hot-reload extension points


lzjqsdd:
# Is your feature request related to a problem?

Configuration handling is currently scattered and inconsistent:

- Loading logic lives in one large `ClusterConf::from()` function; env vars are applied ad hoc (only 5 hand-coded variables).
- CLI overrides use three different styles: ~30 hand-written `if let Some(...)` branches in FUSE mount args, a generated `ClientConfCliOverrides` for the client section, and manual string parsing in the CLI.
- There is no unified validation: `#[serde(default)]` silently drops misspelled keys (only FUSE has a local unknown-key audit), so typos never take effect and are never noticed.
- Multiple entrypoints (CLI, server, lancedb, SDK) each reimplement config-file path discovery.
- No hot-reload capability exists at all; tuning any parameter requires a restart.

A previous attempt (#613) was closed after the crate layout was restructured; this issue tracks the redesign on the current architecture.

# Describe the solution you'd like

Introduce a layered loading pipeline in `crates/common/curvine-config`:

```
serde defaults -> file (TOML) -> environment (explicit allowlist) -> CLI (typed overrides)
      -> single serde deserialization -> normalize -> validate -> freeze
```

Design decisions:

- Hand-written layering instead of the `config` crate: its `Environment::separator("_")` conflicts with Curvine's underscore-heavy key names (`format_master`, `master_addrs`, ...), which would force an explicit mapping table anyway while adding a heavy dependency linked into every component.
- A single deserialization pass preserves native serde semantics (defaults, aliases, arrays of tables, error messages).
- Environment variables are applied through an explicit allowlist table (`CURVINE_*_HOSTNAME` -> dotted paths), keeping backward compatibility.
- A uniform `ConfValidate` trait wraps the existing per-section `init()` checks (single source of truth), implemented by every section and by `ClusterConf` as a whole.
- Unknown-key audit generalizes the FUSE-only check to the entire document by deriving the canonical key schema from `ClusterConf::default()`.
- Typed CLI overrides generalize the existing `ClientCliArgs` derive macro (with opt-in mode) to other sections, eliminating hand-written override branches.
- Hot reload is out of scope for implementation, but extension points are added: a `Reloadable` marker, an ephemeral `RuntimeOverlay` (admin-push, lost on restart by design), and an `Arc`-snapshot `ConfigHandle`. Two future mechanisms are supported: file-watch reload and admin-API pushdown.

[...]

PR #1620, opened 2026-08-25
feat(config): add unified layered config loading pipeline (#1619)

lzjqsdd:
# Summary

First PR of a stacked series (1/7) that unifies component configuration handling. This PR introduces the layered loading pipeline core: file (TOML) -> environment (explicit allowlist) -> single serde deserialization, plus shared config-file discovery. No component is rewired yet; behavior-neutral foundation for the follow-up commits.

# Issue Describe / Design

Related to #1619 (partial — pipeline core only).

- **Hand-written layering instead of the `config` crate**: `Environment::separator("_")` conflicts with Curvine's underscore-heavy key names (`format_master`, `master_addrs`, ...), which would force an explicit mapping table anyway while adding a heavy dependency linked into every component.
- **Environment allowlist table**: each recognized env var maps to explicit dotted paths (`CURVINE_MASTER_HOSTNAME` -> `master.hostname` + `journal.hostname`), preserving current semantics including `net_interface` precedence and its ignore-warnings.
- **Single serde deserialization**: layers merge as raw `toml::Value` documents, so native serde semantics (defaults, aliases, arrays of tables, error messages) are preserved verbatim.
- **Injectable env lookup** (`build_document_with`) for testability without mutating process-global environment in parallel tests.
- **`ConfigLoader::discover()`**: unified discovery order (explicit path > `CURVINE_CONF_FILE` > well-known locations) plus a `discover_configured()` variant for embedders that must not probe default locations.

# Changes

| `crates/common/curvine-config/src/pipeline.rs` (new) | Layer model, deep merge, env allowlist, `ConfigLoader` discovery | None (new module, no callers yet) |
| `crates/common/curvine-config/src/lib.rs` | Export `pipeline` module and loader types | None |
| same (review fixes, ebcdf5a7) | net_interface trim persisted into document; collision error names the actual segment; env value restored in precedence warning; injectable-lookup build variants | None |

# Test verified

| `cargo test -p curvine-config` | PASS (88 lib tests) | Pipeline: merge priority, dotted paths + collision segment, env trimming, transfer scoping, discovery order |
| `net_interface_wins_over_env_on_typed_config` | PASS | NIC precedence asserted on deserialized ClusterConf (all five hostnames), env values ignored |
| `whitespace_only_net_interface_falls_back_to_env` | PASS | Trim decision persisted: typed `net_interface == ""` without relying on normalize |
| `default_candidates_follow_documented_order` | PASS | Discovery order asserted without mutating process env |
| Per-commit build check | PASS | Commit compiles standalone (verified via detached worktree) |

# Dependencies

- Related PRs: stacked series to follow on top (validation framework, ClusterConf routing, generated CLI overrides, discovery convergence, hot-reload scaffolding, docs)
- Related issues: #1619
- Blocks / blocked by: None (first of series)


crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,536 @@
...
+            }
+        }
+    }
+    Ok(())
+}
+
+fn net_interface_set(doc: &toml::Value) -> bool {
+    doc_string(doc, &[NET_INTERFACE_KEY])
+        .map(|v| !v.trim().is_empty())
+        .unwrap_or(false)
```
szbr00: `net_interface_set` treats a whitespace-only value as unset (`v.trim().is_empty()`), but the raw `net_interface` string is left in the document. After the planned single deserialize, `ClusterConf.net_interface` is still `" "`, and every existing caller uses `!self.net_interface.is_empty()` (`apply_hostname_overrides`, `apply_transfer_hostname_overrides`, `check_master_hostname` in `cluster_conf.rs`). Those will take the NIC branch even though this layer already decided the NIC was unset and applied env hostnames instead.

That is not 1:1 with today's loader (no trim on the set-check), and it will split-brain when the routing PR lands.

Pick one definition and persist it: either drop the trim-as-unset check, or write the trimmed value back (`""` / the trimmed NIC name) so later typed checks agree. A `toml` → `ClusterConf` assertion on `whitespace_only_net_interface_falls_back_to_env` would have caught this.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,536 @@
...
+            // entries would be surprising for a data-plane-only service.
+            LoadProfile::Transfer => {
+                o.var == ClusterConf::ENV_CLIENT_HOSTNAME
+                    || o.var == ClusterConf::ENV_TRANSFER_HOSTNAME
+            }
+        })
+        .collect();
+
+    if net_interface_set(doc) {
+        let interface = doc_string(doc, &[NET_INTERFACE_KEY])
```
szbr00: This `net_interface_set` branch is the production path called out in #1619's test plan (NIC precedence), but nothing exercises it. `whitespace_only_net_interface_falls_back_to_env` only covers the unset path.

Add a loopback case (same `lo` / `lo0` pattern as `cluster_conf.rs`): set `net_interface`, export a `CURVINE_*_HOSTNAME`, and assert that every Full-profile hostname becomes `127.0.0.1` and the env value is ignored. Deserialize the merged document to `ClusterConf` so the assertion is on the typed config, not only `doc_string`. Without that, a regression that still applies env on top of a set NIC — or that forgets `journal.hostname` — stays green.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,536 @@
...
+        // as-is regardless of existence (load errors surface later with context).
+        // We cannot safely mutate global env in parallel tests, so we exercise
+        // only the explicit branch here; env/default branches share the same
+        // code path verified by construction.
+        let found = ConfigLoader::discover(Some("/tmp/explicit.toml")).unwrap();
+        assert_eq!(found.source, "--conf");
+        assert_eq!(found.as_str(), "/tmp/explicit.toml");
+    }
+
+    #[test]
```
szbr00, line 743: This only checks that every candidate ends in `curvine-cluster.toml`. #1619's test plan asks for discovery order; that can be asserted without mutating process env: cwd file, `conf/`, `$HOME/.curvine/`, `$CURVINE_HOME/conf/` in that sequence (HOME / CURVINE_HOME entries only when those vars are already present).

`build_document_parses_file_and_applies_env` also never calls `build_document` — it re-runs the stages. Rename it, or thread the lookup into `build_document` so the public entry point is actually covered.

szbr00 (commented):
The raw-TOML merge + explicit hostname allowlist + injectable lookup is the right foundation for #1619, and discovery order matches today's CLI. No wire/compat or hot-path performance issue in this PR.

Two must-fixes below before this document becomes the source of truth for ClusterConf::from / from_transfer: net_interface "set" semantics are not persisted into the document, and the issue test plan (NIC precedence) is not actually asserted.

Not repeating the Copilot notes (warning omits the env value; set_dotted collision names the next segment) — those are optional diagnostics.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,536 @@
...
+        // an operator who exported a hostname override but sees it have no
+        // effect can tell why. Runs before Logger::init — use eprintln!.
+        let mut touched = Vec::new();
+        for entry in &scoped {
+            if lookup(entry.var).is_some() {
+                eprintln!(
+                    "[WARN] net_interface '{}' is set (resolved to {}); ignoring {}.",
+                    interface, ip, entry.var
+                );
+            }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in ebcdf5a7: the warning now includes the env value (`ignoring CURVINE_X_HOSTNAME='value'`), matching the previous loader.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,536 @@
...
+                "config path '{}' collides with a non-table value at '{}'",
+                path,
+                seg
+            );
+        }
+        let table = cur.as_table_mut().expect("checked is_table");
+        cur = table
+            .entry(seg.to_string())
+            .or_insert_with(|| toml::Value::Table(toml::value::Table::new()));
+    }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in ebcdf5a7: `set_dotted` now tracks the segment it last descended into and names that one in the collision error.

crates/common/curvine-config/src/pipeline.rs
lzjqsdd: Thanks, good catch on the implicit invariant. Fixed in ebcdf5a7: the env layer now persists the trimmed value back into the document (whitespace-only becomes `""`), so the deserialized `ClusterConf.net_interface` agrees with the layer decision without depending on the later normalize step. Covered by `whitespace_only_net_interface_falls_back_to_env`, which now deserializes to `ClusterConf` and asserts `net_interface == ""`.

crates/common/curvine-config/src/pipeline.rs
lzjqsdd: Thanks. Added in ebcdf5a7: `net_interface_wins_over_env_on_typed_config` sets `net_interface` in the file layer (lo/lo0 pattern), injects all four `CURVINE_*_HOSTNAME` values, runs the real `build_document_with` entry point, deserializes to `ClusterConf`, and asserts all five hostnames resolve to `127.0.0.1` with none of the env values applied.

crates/common/curvine-config/src/pipeline.rs
lzjqsdd: Thanks. Both addressed in ebcdf5a7: (1) `default_candidates_follow_documented_order` asserts the cwd > conf/ > $HOME/.curvine/ > $CURVINE_HOME/conf/ sequence by reading HOME/CURVINE_HOME without mutating process env; (2) added `build_document_with`/`build_transfer_document_with` injectable-lookup variants so tests exercise the actual public entry point — the old stage-reimplementation test was replaced with `build_document_with_parses_file_and_applies_env`.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,660 @@
...
+        .collect();
+
+    let trimmed_interface = doc_string(doc, &[NET_INTERFACE_KEY])
+        .unwrap_or_default()
+        .trim()
+        .to_string();
+    // Persist the trim decision into the document: after deserialization,
+    // `ClusterConf.net_interface` must agree with what this layer decided,
+    // instead of relying on a later normalize step to reconcile " " vs "".
+    if doc.get(NET_INTERFACE_KEY).is_some() {
```
szbr00: `doc.get(NET_INTERFACE_KEY).is_some()` is true for any type, not just a string. `doc_string` then yields None, `trimmed_interface` is `""`, and this block rewrites the key to `toml::Value::String("")`.

A file with `net_interface = 1` (or `true`, or a table) would fail today's `toml::from_str::<ClusterConf>` with a type error. After this layer it deserializes as `net_interface == ""` and silently falls through to env hostnames. That breaks the module's own contract: "native serde semantics / error messages are preserved".

Only persist when the existing value is a string. Leave any other type untouched so the later deserialize still fails.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,660 @@
...
+        }
+        for path in touched {
+            set_dotted(doc, path, toml::Value::String(ip.clone()))?;
+        }
+        return Ok(());
+    }
+
+    for entry in scoped {
+        if let Some(value) = lookup(entry.var) {
+            let value = value.trim().to_string();
```
szbr00, line 231: After trim, a whitespace-only `CURVINE_*_HOSTNAME` becomes `""` and is written to every mapped path, wiping a valid file hostname. Today's loader assigns the raw env value (`apply_hostname_overrides` does not trim); this is worse, not 1:1.

It is also inconsistent with the net_interface handling a few lines above, which treats whitespace as unset. Skip empty-after-trim (do not call `set_dotted`) so `CURVINE_MASTER_HOSTNAME="   "` cannot blank `master.hostname` / `journal.hostname`. A unit test with a file hostname plus a whitespace-only env var should keep the file value — the current trim test only covers `" master-01 "`.

crates/common/curvine-config/src/pipeline.rs
```diff
@@ -0,0 +1,660 @@
...
+pub fn build_transfer_document(
+    file_text: &str,
+    overlays: &[toml::Value],
+) -> CommonResult<toml::Value> {
+    build_transfer_document_with(file_text, overlays, &real_env_lookup)
+}
+
+/// [`build_transfer_document`] with an injected environment lookup.
+pub fn build_transfer_document_with(
+    file_text: &str,
```
szbr00, line 134: `build_transfer_document_with` has no test. `transfer_profile_ignores_master_worker_env_vars` only covers the env branch via `apply_env_layer_with`.

`cluster_conf.rs::transfer_net_interface_keeps_master_address` is an existing invariant: Transfer + `net_interface` must rewrite client/transfer only and leave `master.hostname` (and therefore `resolve_master_addrs`) on the file value. Please add a loopback case through `build_transfer_document_with`, deserialize to `ClusterConf`, and assert master/journal stay put while client/transfer become `127.0.0.1`. A later change that applies the Full allowlist on the NIC branch would not be caught today.

szbr00 (commented):
The ebcdf5a7 follow-up addresses the previous review: net_interface trim is persisted, NIC precedence is asserted on deserialized ClusterConf, discovery order is checked, and tests now go through build_document_with.

Two remaining contract issues before this document is safe to wire into ClusterConf::from / from_transfer: the persist block can hide a serde type error by rewriting a non-string net_interface to "", and a whitespace-only CURVINE_*_HOSTNAME now blanks the file hostname instead of being treated as unset.

crates/common/curvine-config/src/pipeline.rs
lzjqsdd: Thanks. Already fixed in 92d2e1a4 before this comment landed: the trim-persist step now only rewrites string values, leaving any other type untouched so deserialization fails with the native type error. Regression test `nonstring_net_interface_is_not_silently_coerced` asserts exactly this.

crates/common/curvine-config/src/pipeline.rs
lzjqsdd: Fixed in 9107a668: empty-after-trim env values are now skipped entirely so the file hostname survives. One factual note — the legacy loader did trim (`v.trim().to_string()` assigned unconditionally), so it also blanked hostnames on whitespace-only vars; skipping is therefore a deliberate hardening over strict 1:1. Covered by `whitespace_only_env_value_keeps_file_hostname`.

crates/common/curvine-config/src/pipeline.rs
lzjqsdd: Thanks. Added in 9107a668: `transfer_nic_rewrites_only_client_and_transfer` runs the real `build_transfer_document_with` entry point with a loopback NIC plus client/transfer env vars set, deserializes to `ClusterConf`, and asserts master stays on its file value while client/transfer become 127.0.0.1 (and derived master_addrs still point at master).

szbr00 (approved):

PR #1621, opened 2026-08-25
feat(mds): add standalone service scaffold

yangyang233333:
# Summary

Add the initial standalone `curvine-mds` service scaffold. The service is opt-in and disabled by default, so existing Master, Worker, and client behavior remains unchanged.

# Issue Describe / Design

This is the first implementation step of the stateless MDS design. It establishes an independent crate and process lifecycle before metadata storage, RPC handlers, service discovery, or client routing are introduced.

Design decisions:
- Keep stateless MDS separate from `curvine-master`.
- Use runtime configuration instead of a Cargo feature.
- Preserve backward compatibility by defaulting `mds.enabled` to `false` when the section is absent.
- Expose `curvine-server --service mds` without adding placeholder metadata RPC handlers.

# Changes

| `curvine-mds` | Add standalone MDS crate with lifecycle and graceful shutdown scaffold | None unless explicitly enabled and started |
| `curvine-config` | Add `MdsConf`, validation, server configuration helpers, and compatibility tests | Existing configurations default to MDS disabled |
| `curvine-server` | Add the `mds` service option and version component name | Existing service options are unchanged |
| `Cargo.toml` / `Cargo.lock` | Register the new workspace crate and dependency | Build graph only |
| Configuration examples | Add disabled `[mds]` sections | No runtime behavior change |

# Test verified

| `CARGO_TARGET_DIR=/tmp/curvine-mds-target cargo test -p curvine-config mds_conf::tests` | PASS | 2 tests |
| `CARGO_TARGET_DIR=/tmp/curvine-mds-target cargo test -p curvine-config legacy_config_without_mds_uses_disabled_defaults` | PASS | Backward compatibility |
| `CARGO_TARGET_DIR=/tmp/curvine-mds-target cargo test -p curvine-config loads_workspace_cluster_config` | PASS | Example config parsing |
| `CARGO_TARGET_DIR=/tmp/curvine-mds-target cargo test -p curvine-mds` | PASS | 2 tests |
| `CARGO_TARGET_DIR=/tmp/curvine-mds-target cargo test -p curvine-server --bin curvine-server` | PASS | 3 CLI tests |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/common/curvine-config/src/cluster_conf.rs
+++ b/crates/common/curvine-config/src/cluster_conf.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::{
-    CliConf, ClientConf, DBConf, FuseConf, JobConf, JournalConf, MasterConf, TransferConf,
+    CliConf, ClientConf, DBConf, FuseConf, JobConf, JournalConf, MasterConf, MdsConf, TransferConf,
     WorkerConf,
 };
 use curvine_core_error::{err_box, try_err, CommonResult};
@@ -50,6 +50,8 @@ pub struct ClusterConf {
     pub net_interface: String,
 
     pub master: MasterConf,
+
+    pub mds: MdsConf,
 
     // Log synchronization configuration.
     pub journal: JournalConf,
@@ -96,6 +98,7 @@ impl ClusterConf {
         conf.normalize_whitespace();
         conf.apply_hostname_overrides()?;
         conf.master.init()?;
+        conf.mds.init()?;
         conf.client.init()?;
         conf.fuse.init()?;
         conf.job.init()?;
@@ -328,6 +331,22 @@ impl ClusterConf {
         web_conf
     }
 
+    pub fn mds_server_conf(&self) -> ServerConf {
+        let mut conf = ServerConf::with_hostname(&self.mds.hostname, self.mds.rpc_port);
+        conf.name = format!("{}-mds", self.cluster_id);
+        conf.io_threads = self.mds.io_threads;
+        conf.worker_threads = self.mds.worker_threads;
+        conf
+    }
+
+    pub fn mds_web_conf(&self) -> ServerConf {
+        let mut conf = ServerConf::with_hostname(&self.mds.hostname, self.mds.web_port);
+        conf.name = format!("{}-mds", self.cluster_id);
+        conf.io_threads = self.mds.io_threads;
+        conf.worker_threads = self.mds.worker_threads;
+        conf
+    }
+
     pub fn worker_addr(&self) -> InetAddr {
         InetAddr::new(self.worker.hostname.clone(), self.worker.rpc_port)
     }
@@ -494,6 +513,7 @@ impl Default for ClusterConf {
             cluster_id: "curvine".to_string(),
             net_interface: String::new(),
             master: Default::default(),
+            mds: Default::default(),
             journal: Default::default(),
             worker: Default::default(),
             fault_injection: Default::default(),
@@ -679,7 +699,24 @@ mod tests {
         let conf = ClusterConf::from(path).unwrap();
 
         assert_eq!(conf.master.rpc_port, ClusterConf::DEFAULT_MASTER_PORT);
+        assert!(!conf.mds.enabled);
         assert!(!conf.client.master_addrs.is_empty());
+    }
+
+    #[test]
+    fn legacy_config_without_mds_uses_disabled_defaults() {
+        let conf: ClusterConf = toml::from_str(
+            r#"
+                cluster_id = "legacy"
+
+                [master]
+            "#,
+        )
+        .unwrap();
+
+        assert!(!conf.mds.enabled);
+        assert_eq!(conf.mds.rpc_port, crate::MdsConf::DEFAULT_RPC_PORT);
+        assert_eq!(conf.mds.web_port, crate::MdsConf::DEFAULT_WEB_PORT);
     }
 
     #[test]
--- a/crates/common/curvine-config/src/cluster_conf.rs
+++ b/crates/common/curvine-config/src/cluster_conf.rs
@@ -88,6 +88,7 @@ impl ClusterConf {
     pub const DEFAULT_FUSE_WEB_PORT: u16 = 9003;
 
     pub const ENV_MASTER_HOSTNAME: &'static str = "CURVINE_MASTER_HOSTNAME";
+    pub const ENV_MDS_HOSTNAME: &'static str = "CURVINE_MDS_HOSTNAME";
     pub const ENV_WORKER_HOSTNAME: &'static str = "CURVINE_WORKER_HOSTNAME";
     pub const ENV_CLIENT_HOSTNAME: &'static str = "CURVINE_CLIENT_HOSTNAME";
     pub const ENV_TRANSFER_HOSTNAME: &'static str = "CURVINE_TRANSFER_HOSTNAME";
@@ -98,7 +99,9 @@ impl ClusterConf {
         conf.normalize_whitespace();
         conf.apply_hostname_overrides()?;
         conf.master.init()?;
-        conf.mds.init()?;
+        if conf.mds.enabled {
+            conf.mds.init()?;
+        }
         conf.client.init()?;
         conf.fuse.init()?;
         conf.job.init()?;
@@ -126,6 +129,7 @@ impl ClusterConf {
     fn normalize_whitespace(&mut self) {
         self.net_interface = self.net_interface.trim().to_string();
         self.master.hostname = self.master.hostname.trim().to_string();
+        self.mds.hostname = self.mds.hostname.trim().to_string();
         self.journal.hostname = self.journal.hostname.trim().to_string();
         self.worker.hostname = self.worker.hostname.trim().to_string();
         self.client.hostname = self.client.hostname.trim().to_string();
@@ -170,6 +174,7 @@ impl ClusterConf {
             // eprintln! to make the warning visible on stderr.
             for env_key in [
                 Self::ENV_MASTER_HOSTNAME,
+                Self::ENV_MDS_HOSTNAME,
                 Self::ENV_WORKER_HOSTNAME,
                 Self::ENV_CLIENT_HOSTNAME,
                 Self::ENV_TRANSFER_HOSTNAME,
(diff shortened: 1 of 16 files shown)


curvine-mds/src/server.rs
```diff
@@ -0,0 +1,118 @@
...
+            shutdown_rx,
+        })
+    }
+
+    pub fn conf(&self) -> &ClusterConf {
+        &self.conf
+    }
+
+    pub async fn start(&mut self) -> CommonResult<()> {
+        info!(
```
szbr00: `start()` logs `rpc_addr`/`web_addr` as if the process is serving, but this scaffold never binds those ports. Master, Worker, and Transfer all call `Logger::init` in `with_conf` before any `info!`; MDS does not, so these messages are also dropped at runtime.

Initialize logging here (add `MdsConf.log` like the other roles, then `Logger::init` in `with_conf`). Until RPC/web exist, log that the process is up as a scaffold without advertising listen addresses, or bind a health endpoint so the addresses are real.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -679,9 +699,26 @@ mod tests {
         let conf = ClusterConf::from(path).unwrap();
 
         assert_eq!(conf.master.rpc_port, ClusterConf::DEFAULT_MASTER_PORT);
+        assert!(!conf.mds.enabled);
         assert!(!conf.client.master_addrs.is_empty());
     }
 
+    #[test]
+    fn legacy_config_without_mds_uses_disabled_defaults() {
```
szbr00, line 701: This compatibility test only `toml::from_str`s. It never runs `ClusterConf::from` / `mds.init()`, which is the Master/Worker load path. `loads_workspace_cluster_config` also no longer covers the legacy case because `etc/curvine-cluster.toml` now contains `[mds]`.

Please load a temp TOML that omits `[mds]` through `ClusterConf::from` (or deserialize then `init`) and assert `enabled == false` after validation. Otherwise a required/non-defaulted MDS field can break existing cluster configs without failing this test.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -96,6 +98,7 @@ impl ClusterConf {
         conf.normalize_whitespace();
         conf.apply_hostname_overrides()?;
         conf.master.init()?;
+        conf.mds.init()?;
```
szbr00: `ClusterConf::from` now validates MDS on every Master/Worker start, even when `mds.enabled` is false. That diverges from `from_transfer`, which skips unused-role `init` so a placeholder `conn_limit = 0` on master does not block Transfer.

An invalid disabled `[mds]` section will now fail Master/Worker startup. Prefer `if conf.mds.enabled { conf.mds.init()?; }` and validate again in `Mds::with_conf`.

The same load path still omits MDS from `normalize_whitespace` and `apply_hostname_overrides` (`net_interface` / `CURVINE_*_HOSTNAME`). `mds.hostname` will stay `localhost` in interface-based deployments. Include MDS in those helpers (and add `ENV_MDS_HOSTNAME`) before this service binds.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -328,6 +331,22 @@ impl ClusterConf {
         web_conf
     }
 
+    pub fn mds_server_conf(&self) -> ServerConf {
```
szbr00: `mds_server_conf` / `mds_web_conf` have no callers in this PR; `Mds::start` does not use them, and `conn_limit` / `global_limit` on `MdsConf` are similarly unused. Expanding the public `ClusterConf` surface ahead of an RPC server makes the next change harder to review.

Please drop these helpers (and the unused limit knobs) until the service actually binds, or wire `start()` to them in this PR so they are covered by the lifecycle tests.

szbr00 (commented):
Reviewed the MDS scaffold against Master/Worker/Transfer lifecycle and ClusterConf load paths. Inline comments cover logging/bind claims, the legacy-config test, load-path validation, and unused public helpers.

New Rust sources are missing the standard Apache 2.0 header used by every other crate (`curvine-mds/src/lib.rs`, `curvine-mds/src/server.rs`, `crates/common/curvine-config/src/mds_conf.rs`). Please add the same copyright header as `curvine-master/src/lib.rs`.

curvine-mds/src/server.rs
yangyang233333: Thanks. Fixed in dbc41df2: MDS now initializes logging in `with_conf`, validates its config there, and no longer advertises unbound RPC/Web addresses.

crates/common/curvine-config/src/cluster_conf.rs
yangyang233333: Thanks. Fixed in dbc41df2: the legacy compatibility test now writes a config without `[mds]` and loads it through `ClusterConf::from`.

crates/common/curvine-config/src/cluster_conf.rs
yangyang233333: Thanks. Fixed in dbc41df2: disabled MDS config skips validation, enabled config is validated again by `Mds::with_conf`, and MDS hostname normalization/overrides are included.

crates/common/curvine-config/src/cluster_conf.rs
yangyang233333: Thanks. Fixed in dbc41df2: removed the unused MDS server/web helpers and connection limit fields until the RPC server is introduced.

yangyang233333:
Thanks for the review. The four inline issues are addressed in dbc41df2. I kept the new files without the OPPO copyright header intentionally, following the contributor requirement for this change.

szbr00 (commented):
The CLI flag is wired, but the packaged process lifecycle is not. `build/bin/launch-process.sh` only starts `master|worker|transfer`; `mds` hits `Unknown service`. There is no `curvine-mds.sh` next to `curvine-transfer.sh`. The Docker entrypoint still documents `Service type: master, worker, transfer`.

Transfer was added with the wrapper script, launcher whitelist, and entrypoint comment in the same change. As written, `curvine-server --service mds` works only if someone invokes the binary by hand; `bin/` and the documented container start path cannot run this scaffold. Please add `mds` to `launch-process.sh`, add `build/bin/curvine-mds.sh`, and update the Docker entrypoint comment.

The `net_interface` comments in `etc/curvine-cluster.toml` and the Docker example still say the resolved IP is applied to master/journal/worker/client hostnames, but this PR also overwrites `mds.hostname`.

crates/common/curvine-config/src/mds_conf.rs
```diff
@@ -0,0 +1,87 @@
...
+impl Default for MdsConf {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            hostname: "localhost".to_string(),
+            rpc_port: Self::DEFAULT_RPC_PORT,
+            web_port: Self::DEFAULT_WEB_PORT,
+            io_threads: 4,
+            worker_threads: 8,
+            log: Default::default(),
```
szbr00: `MdsConf` log uses `LogConf::default()`, so `file_name` is empty. `Logger::create_writer` then falls back to `"curvine"` whenever `log_dir` is a directory. Transfer sets `file_name: "transfer.log"`; the example `[master]` / `[worker]` stanzas also set an explicit file name. `[mds]` in `etc/curvine-cluster.toml` and the Docker example omit `log` entirely.

If an operator enables MDS and points `log_dir` at `./logs` (the usual Master/Worker pattern), MDS will write `curvine.log` and can collide with other components. Default `file_name` to `"mds.log"` and add a `log = { ... file_name = "mds.log" }` line to both example configs.

curvine-docker/deploy/example/conf/curvine-cluster.toml
```diff
@@ -23,6 +23,14 @@ meta_dir = "./data/meta"
 audit_logging_enabled = true
 log = { level = "info", log_dir = "./logs", file_name = "master.log" }
 
+# Optional standalone stateless metadata service. It is disabled by default;
+# start it with `curvine-server --service mds` only after enabling this flag.
+[mds]
+enabled = false
+hostname = "0.0.0.0"
```
szbr00: Docker MDS sets `hostname = "0.0.0.0"` while `[master]` / `[worker]` leave hostname unset (localhost / later `CURVINE_*_HOSTNAME` or `net_interface`). That mixes a listen-all address into the same field the rest of the cluster uses as an advertise name.

`net_interface` and `CURVINE_MDS_HOSTNAME` can override this, but a typical k8s deploy that only sets the master/worker hostname env vars will leave MDS on `0.0.0.0`. Please use the same hostname convention as master/worker in this example, or document listen vs advertise before the service binds.

crates/common/curvine-config/src/mds_conf.rs
yangyang233333: Thanks. Fixed in 2ccf30e1: MDS now defaults to `mds.log`, the default is covered by the config test, and both example configurations set an explicit MDS log file.

curvine-docker/deploy/example/conf/curvine-cluster.toml
yangyang233333: Thanks. Fixed in 2ccf30e1: the Docker example no longer uses `0.0.0.0` as the MDS hostname, so it follows the same default, environment override, and `net_interface` conventions as the other roles.

yangyang233333:
Thanks for the follow-up review. Fixed in 2ccf30e1: added `build/bin/curvine-mds.sh`, allowed `mds` in `launch-process.sh`, updated the Docker entrypoint service documentation, and corrected both `net_interface` comments to include MDS. Shell syntax checks and all targeted Rust tests pass.

curvine-docker/deploy/entrypoint.sh
```diff
@@ -46,7 +46,7 @@ if is_fluid_mode "${1:-}"; then
 fi
 
 # Start the curvine service process
-# Service type: master, worker, transfer
+# Service type: master, worker, mds, transfer
```
szbr00, line 49: The comment now says `Service type: master, worker, mds, transfer`, but the dispatch below is still `(master|worker|transfer)`. `mds` falls through to `(*) exec "$@"`, so `docker ... mds start` tries to exec `mds` instead of calling `start_service`.

Please add `mds` to that case pattern so the container start path matches the comment, `launch-process.sh`, and `curvine-mds.sh`.

yangyang233333: Thanks. Fixed in the latest push: the Docker entrypoint dispatch now includes `mds`, so `docker ... mds start|stop|restart` uses the same service lifecycle as Master, Worker, and Transfer. `bash -n curvine-docker/deploy/entrypoint.sh` passes.

szbr00 (approved):

PR #1622, opened 2026-08-25
feat(config): add validation framework and unknown-key audit (#1619)

lzjqsdd:
# Summary

Second PR of the stacked series for #1619 (2/7, base: #1620). Adds the unified validation framework: a `ConfValidate` trait wrapping each section's existing `init()` checks, and a workspace-wide unknown-key audit that generalizes the FUSE-only typo detection to the entire config document.

# Issue Describe / Design

Related to #1619 (partial — validation framework only). Depends on #1620 (pipeline core).

- **`ConfValidate`**: uniform validation hook; implementations delegate to each section's own `init()` so rules keep a single source of truth. `ClusterConf` implements it too, running every section in the historical init order.
- **Unknown-key audit**: serializing `ClusterConf::default()` derives the canonical key schema; any raw TOML key not consumed is reported with its dotted path. Covers nested sections (`master.compatibility`) and `#[serde(skip)]` fields (values that silently never take effect).
- Warnings never abort startup (forward/backward compatibility across versions).

# Changes

| `crates/common/curvine-config/src/validation.rs` (new) | `ConfValidate` trait + section impls + unknown-key audit | None |
| `crates/common/curvine-config/src/fuse_conf.rs` | Legacy fuse audit delegates to shared audit; hand-rolled schema helpers removed | Warning text unchanged |
| `crates/common/curvine-config/src/lib.rs` | Export `validation` module | None |

# Test verified

| `cargo test -p curvine-config` | PASS | Audit tests: typos, unknown sections, skip fields, sample-config-clean guard |

# Dependencies

- Related PRs: base #1620
- Related issues: #1619

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,10 @@ curvine-tests/regression_result
 .agents/skills/*
 !.agents/skills/cv-*/
 .codegraph/
+
+# Local agent tooling & planning artifacts
+.kiro/
+.plans/
+plans/
+.mcp.json
+opencode.jsonc
--- a/crates/common/curvine-config/src/fuse_conf.rs
+++ b/crates/common/curvine-config/src/fuse_conf.rs
@@ -382,48 +382,16 @@ impl FuseConf {
         }
     }
 
-    /// Canonical TOML key names for `FuseConf` fields, derived by serializing the
-    /// default so it cannot drift from the struct definition. `#[serde(skip_*)]`
-    /// runtime-only fields (the `*_ttl` durations) are excluded automatically.
-    fn known_field_names() -> Vec<String> {
-        match toml::Value::try_from(FuseConf::default()) {
-            Ok(toml::Value::Table(t)) => t.keys().cloned().collect(),
-            _ => vec![],
-        }
-    }
-
-    /// Returns keys in a raw `[fuse]` TOML table that do not match any current
-    /// `FuseConf` field. These are surfaced as warnings by `validate-config`:
-    /// they are silently dropped on load (FuseConf is `#[serde(default)]` with
-    /// no `deny_unknown_fields`), so a typo would otherwise never take effect
-    /// and never be noticed. Result is sorted for stable output.
-    ///
-    /// Note: a renamed field kept alive by `#[serde(alias)]` (e.g. the old
-    /// `mnt_per_task`) is reported here too — its canonical name is what
-    /// `FuseConf` exposes — so the value DOES still take effect via the alias.
-    /// The warning is intentionally phrased to cover both cases (typo vs.
-    /// deprecated/renamed) rather than claiming the setting was ignored.
-    pub fn unrecognized_fuse_keys(fuse_table: &toml::value::Table) -> Vec<String> {
-        let known: std::collections::HashSet<String> =
-            Self::known_field_names().into_iter().collect();
-
-        let mut unknown: Vec<String> = fuse_table
-            .keys()
-            .filter(|k| !known.contains(k.as_str()))
-            .cloned()
-            .collect();
-        unknown.sort();
-        unknown
-    }
-
     /// Parses raw cluster TOML text and returns the unrecognized keys found
     /// under the `[fuse]` table. Missing `[fuse]` table yields an empty list.
+    /// Parses raw cluster TOML text and returns the unrecognized keys found
+    /// under the `[fuse]` table. Delegates to the workspace-wide
+    /// [`crate::validation`] audit and filters to this section.
     pub fn unrecognized_fuse_keys_from_toml(raw: &str) -> CommonResult<Vec<String>> {
-        let value: toml::Value = try_err!(toml::from_str(raw));
-        match value.get("fuse").and_then(|v| v.as_table()) {
-            Some(t) => Ok(Self::unrecognized_fuse_keys(t)),
-            None => Ok(vec![]),
-        }
+        Ok(crate::validation::audit_unknown_keys(raw)?
+            .into_iter()
+            .filter_map(|key| key.strip_prefix("fuse.").map(str::to_string))
+            .collect())
     }
 
     pub fn parse_fuse_opts(&self) -> Vec<CString> {
@@ -1304,25 +1272,24 @@ mnt_per_task = 7
 
     #[test]
     fn unrecognized_fuse_keys_flags_typo_and_legacy() {
-        // Every key not matching a current FuseConf field is flagged so the user
-        // can investigate: the typo (direct_iox), the dropped legacy param
-        // (node_cache_size), and the renamed key still alive via #[serde(alias)]
-        // (mnt_per_task). A current field (io_threads) is NOT flagged.
+        // Keys no FuseConf field consumes are flagged so the user can
+        // investigate: the typo (direct_iox) and the dropped legacy param
+        // (node_cache_size). The renamed-but-aliased key (mnt_per_task) is
+        // still consumed via #[serde(alias)], and a current field (io_threads)
+        // likewise — neither is flagged. max_readahead_kb (Option<u32>, None
+        // default) is also consumed and must not be flagged.
         let raw = r#"
 [fuse]
 io_threads = 16
 node_cache_size = 200000
 mnt_per_task = 7
+max_readahead_kb = 1024
 direct_iox = true
 "#;
         let unknown = FuseConf::unrecognized_fuse_keys_from_toml(raw).unwrap();
         assert_eq!(
             unknown,
-            vec![
-                "direct_iox".to_string(),
-                "mnt_per_task".to_string(),
-                "node_cache_size".to_string(),
-            ]
+            vec!["direct_iox".to_string(), "node_cache_size".to_string()]
         );
     }
 
--- a/crates/common/curvine-config/src/fuse_conf.rs
+++ b/crates/common/curvine-config/src/fuse_conf.rs
@@ -387,6 +387,9 @@ impl FuseConf {
     /// Parses raw cluster TOML text and returns the unrecognized keys found
     /// under the `[fuse]` table. Delegates to the workspace-wide
     /// [`crate::validation`] audit and filters to this section.
+    /// Parses raw cluster TOML text and returns the unrecognized keys found
+    /// under the `[fuse]` table. Delegates to the workspace-wide
+    /// [`crate::validation`] audit and filters to this section.
     pub fn unrecognized_fuse_keys_from_toml(raw: &str) -> CommonResult<Vec<String>> {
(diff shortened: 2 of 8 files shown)


crates/common/curvine-config/src/validation.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/// [`audit_unknown_keys`] over an already-parsed document.
+pub fn audit_unknown_keys_in(doc: &toml::Value) -> Vec<String> {
+    // Serializing the default derives the schema from the struct definitions,
+    // so it cannot drift from the code as fields come and go. If a field type
+    // ever becomes non-TOML-serializable, degrade to "no schema" instead of
+    // failing config load — the audit is advisory.
+    let known = match toml::Value::try_from(ClusterConf::default()) {
+        Ok(v) => v,
```
szbr00: The schema is derived by serializing `ClusterConf::default()`, but that is not
the set of keys serde will consume on load.

toml 0.8 omits `Option::None` and `#[serde(skip_serializing)]` from the
serialized Value (verified: `try_from` succeeds and simply drops those keys).
Those fields still deserialize:

- `fuse.max_readahead_kb` (`Option<u32>`, default `None`) — documented,
  consumed, and already on the `validate-config` path via
  `FuseConf::unrecognized_fuse_keys_from_toml`.
- `transfer.store_url`, `store_type`, `sqlite_path`, `mysql_url`
  (`skip_serializing` only; `init()` / `infer_store_url()` read them).

A config that sets any of these is reported as unknown even though the value
takes effect. #1623 then prints that warning on every `ClusterConf::from`.
The module docs ("skipped fields … never take effect") are wrong for
`skip_serializing` without `skip_deserializing`.

`skipped_fields_are_reported_as_unknown` does not catch this: it uses
`fuse.attr_ttl_ms`, which is not a serde field at all (the skip field is
`attr_ttl`). Please add cases that must stay clean:

    [fuse]
    max_readahead_kb = 1024
    [transfer]
    store_url = "sqlite:///tmp/t.db"

Fix direction: derive the schema from deserialize-visible names (struct field
+ `alias`, minus `skip_deserializing` / `skip`), not from a serialized
default. A `JsonSchema`-style walk, a serde `Deserializer` that records
accepted keys, or an explicit allowlist built from the same attributes would
all work; serializing `Default` cannot see aliases or skip_serializing-only
fields.

crates/common/curvine-config/src/validation.rs
```diff
@@ -0,0 +1,260 @@
...
+        (Some(r), Some(k)) => (r, k),
+        // Non-table values carry no keys to audit.
+        _ => return,
+    };
+
+    for (key, value) in raw_table {
+        let path = if prefix.is_empty() {
+            key.clone()
+        } else {
+            format!("{prefix}.{key}")
```
szbr00: `walk_unknown` returns as soon as either side is not a table, so keys inside
array-of-tables are never audited. `etc/curvine-cluster.toml` already ships
`[[worker.spdk_disk.targets]]`; a typo such as `trtypo` / `traddrr` is
silently dropped by serde and will not appear in the warning list.

If the audit is meant to cover the whole document, walk array elements when
both values are arrays (compare each raw table element against
`known[0]` / a default element schema). Fine to defer if this PR is
table-sections only — worth a comment on `walk_unknown` so the gap is
explicit.

crates/common/curvine-config/src/validation.rs
lzjqsdd: Thanks — confirmed, the serialized-default schema cannot see aliases or skip_serializing-only fields, and toml 0.8 also omits Option::None. Replaced with an empirical probe in 6ca0a66a: each scalar leaf path is individually probed by swapping in an opposite-typed value and attempting a real ClusterConf deserialization — a consumed key fails the type check, an ignored key passes and is reported. This derives acceptance from actual serde behavior, so max_readahead_kb, transfer.store_url/store_type/sqlite_path/mysql_url, and aliased keys are all handled with no special-casing. Added your two clean cases plus a corrected #[serde(skip)] case (attr_ttl, not the nonexistent attr_ttl_ms), and fixed the module-doc claim.

crates/common/curvine-config/src/validation.rs
lzjqsdd: Thanks. Deferred per your option, now explicitly documented: `collect_leaf_paths` carries a comment stating array-of-tables internals are outside audit scope, mirrored in the module docs, with a test (`array_of_tables_internals_are_not_audited`) pinning the current behavior so widening it later is a conscious change.

crates/common/curvine-config/src/validation.rs
```diff
@@ -0,0 +1,350 @@
...
+            for (key, child) in table {
+                let path = if prefix.is_empty() {
+                    key.clone()
+                } else {
+                    format!("{prefix}.{key}")
+                };
+                collect_leaf_paths(child, path, out);
+            }
+        }
+        toml::Value::Array(_) => {
```
szbr00, line 172: Required: The Array arm drops the path instead of treating the array itself as
a probeable leaf. `collect_leaf_paths` only pushes scalar leaves, so an unknown
key whose value is an array is never a candidate.

    [worker]
    data_dirz = []

is silently dropped by serde and will not appear in the warning list. The
previous `typoed_keys` case (`worker.data_dirz`) was removed rather than kept
as a regression. The documented gap is element keys *inside* arrays of tables
(`[[worker.spdk_disk.targets]]`); skipping the array-valued key itself is
broader than that.

Push `prefix` here (same as the scalar arm) so the probe replaces the whole
array with a string. A consumed field such as `worker.data_dir` fails
deserialization; a typo succeeds and is reported. Restore a test that
`worker.data_dirz = []` is flagged.

.gitignore
```diff
@@ -58,3 +58,10 @@ curvine-tests/regression_result
 .agents/skills/*
 !.agents/skills/cv-*/
 .codegraph/
+
+# Local agent tooling & planning artifacts
+.kiro/
```
szbr00, line 63: Optional: These ignores (`.kiro/`, `.plans/`, `opencode.jsonc`, …) are
unrelated to the validation framework. Please drop them from this PR (or
land them separately) so the diff stays on the #1619 audit.

crates/common/curvine-config/src/validation.rs
```diff
@@ -0,0 +1,350 @@
...
+        warn_unknown_keys("this is not = = valid toml");
+    }
+
+    #[test]
+    fn fuse_section_audit_matches_legacy_helper() {
+        let raw = r#"
+            [fuse]
+            mnt_pathh = "/mnt/x"
+        "#;
+        let legacy = FuseConf::unrecognized_fuse_keys_from_toml(raw).unwrap();
```
szbr00, line 324: Required: `curvine-fuse validate-config` still calls
`FuseConf::unrecognized_fuse_keys_from_toml`, which serializes
`FuseConf::default()` — the schema that false-positives `max_readahead_kb`
and cannot see aliases the way this probe does. This PR no longer delegates
that helper (it did in the earlier stacked head), so the only production
unknown-key path keeps the old bug while the new audit is test-only until
#1623.

`fuse_section_audit_matches_legacy_helper` only compares a typo, so it stays
green even though the two algorithms disagree on `Option` / alias keys.

Point `unrecognized_fuse_keys_from_toml` at `audit_unknown_keys` and filter
`fuse.*`, then update `unrecognized_fuse_keys_flags_typo_and_legacy`:
`mnt_per_task` is consumed by the alias under the probe (do not require it
in the unknown list) while `direct_iox` / `node_cache_size` still must be
flagged. Expand this lockstep test to include `max_readahead_kb = 1024` so
the two paths cannot drift again.

crates/common/curvine-config/src/validation.rs
lzjqsdd: Thanks. Fixed in 2c7657f8: collect_leaf_paths now pushes array-valued keys (same as scalars), so the probe replaces the whole array with a string — a consumed field like worker.data_dir fails, a typo like worker.data_dirz succeeds and is flagged. The typoed test case now includes worker.data_dirz = []. The documented gap was narrowed to element keys inside arrays of tables only.

.gitignore
lzjqsdd: Thanks. Addressed in 2c7657f8: the non-project entries (.kiro/, .plans/, opencode.jsonc, .mcp.json) have been removed from .gitignore in this PR. The remaining plans/ entry is kept as it is the standardized working directory for project plans.

crates/common/curvine-config/src/validation.rs
lzjqsdd: Thanks. Fixed in 2c7657f8: FuseConf::unrecognized_fuse_keys_from_toml now delegates to audit_unknown_keys with a fuse.* filter. The old serialized-default helpers (known_field_names, unrecognized_fuse_keys) are removed. The lockstep test is expanded to include max_readahead_kb = 1024, and the alias case (mnt_per_task) is no longer expected in the unknown list — matching the empirical probe contract.

crates/common/curvine-config/src/validation.rs
```diff
@@ -0,0 +1,350 @@
...
+        }
+        toml::Value::Array(_) => {
+            // Documented gap: element keys inside arrays are not audited.
+        }
+        _ => {
+            if !prefix.is_empty() {
+                out.push(prefix);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
lzjqsdd: Already fixed in 2c7657f8 (previous round): collect_leaf_paths now pushes array-valued keys as probe candidates, so misspelled array keys like worker.data_dirz = [] are flagged. The module docs and test coverage have been updated accordingly.

crates/common/curvine-config/src/fuse_conf.rs
```diff
@@ -382,48 +382,16 @@ impl FuseConf {
...
-            .collect();
-        unknown.sort();
-        unknown
-    }
-
     /// Parses raw cluster TOML text and returns the unrecognized keys found
     /// under the `[fuse]` table. Missing `[fuse]` table yields an empty list.
+    /// Parses raw cluster TOML text and returns the unrecognized keys found
+    /// under the `[fuse]` table. Delegates to the workspace-wide
+    /// [`crate::validation`] audit and filters to this section.
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 92e2c369 (p2) and propagated to all stacked branches: the doc comment on unrecognized_fuse_keys_from_toml has been restored, explaining the delegation to the workspace-wide audit. The duplicate was caused by the earlier conflict resolution dropping the comment entirely.

szbr00 (approved):

PR #1623, opened 2026-08-25
refactor(config): load ClusterConf through the unified pipeline (#1619)

lzjqsdd:
# Summary

Third PR of the stacked series for #1619 (3/7). Routes `ClusterConf::from()` / `from_transfer()` through the layered pipeline: file -> env allowlist -> single deserialization -> normalize -> validate -> resolve. Public signatures are unchanged; behavior verified equivalent.

# Issue Describe / Design

Related to #1619 (partial). Depends on #1620 + #1622.

- Hostname/net_interface handling moves into the pipeline's env layer, replacing the hand-written override methods.
- `from_transfer()` keeps its reduced scope via the transfer profile (only client/transfer hostnames resolve) and reduced validation set.
- Unknown-key warnings now print at load time for every component.
- Error messages propagate verbatim; failures abort the load exactly as before.

# Changes

| `crates/common/curvine-config/src/cluster_conf.rs` | `from()/from_transfer()` routed through pipeline; hostname overrides moved to env layer | None observable: all pre-existing assertion tests pass unchanged |

# Test verified

| `cargo test -p curvine-config` | PASS | Includes new file<env priority test and end-to-end transfer NIC test on the real load path |

# Dependencies

- Related PRs: base #1622 (stack: #1620 -> #1622 -> this)
- Related issues: #1619

### DIFF
--- a/crates/common/curvine-config/src/cluster_conf.rs
+++ b/crates/common/curvine-config/src/cluster_conf.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::validation::ConfValidate;
+use crate::{pipeline, validation};
 use crate::{
     CliConf, ClientConf, DBConf, DiscoveryConf, FuseConf, JobConf, JournalConf, MasterConf,
     MdsConf, TransferConf, WorkerConf,
@@ -26,9 +28,7 @@ use curvine_runtime::common::{LogConf, Utils};
 use log::info;
 use nix::ifaddrs::getifaddrs;
 use serde::{Deserialize, Serialize};
-use std::env;
 use std::fmt::{Display, Formatter};
-use std::fs::read_to_string;
 use std::time::Duration;
 
 // Cluster configuration files.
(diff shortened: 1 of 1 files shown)


crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -707,24 +642,73 @@ mod tests {
         assert!(!conf.client.master_addrs.is_empty());
     }
 
+    // End-to-end layer-priority check at the struct level, mirroring from()'s
+    // stages but with an injected env lookup so parallel tests never mutate
+    // process-global environment variables.
+    #[test]
+    fn file_layer_loses_to_env_layer() {
```
szbr00: file_layer_loses_to_env_layer never calls ClusterConf::from(). It re-runs
apply_env_layer_with + try_into + run_section_validation. A from() that skipped
build_document, or that used LoadProfile::Transfer by mistake, would still pass
this test and the other existing from() cases: trims_whitespace_in_hostname_config
uses a whitespace-only NIC (env layer is a no-op), and
loads_workspace_cluster_config does not set net_interface.

from_transfer() already has a real-path NIC test. Add the Full-profile counterpart:
write a temp file with net_interface = lo/lo0, call ClusterConf::from(), and assert
master/journal/worker/client/transfer hostnames are 127.0.0.1. That needs no
process-env mutation and is the assertion that fails if from() is not actually
wired through the Full pipeline.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -91,33 +91,52 @@ impl ClusterConf {
...
+    /// Single serde deserialization pass — native semantics (defaults, aliases,
+    /// arrays of tables, error messages) preserved verbatim.
+    fn try_from_doc(doc: toml::Value) -> CommonResult<Self> {
+        Ok(try_err!(doc.try_into()))
+    }
+
+    /// Section hard-validation + derived-field init, in the historical order.
+    /// Each impl delegates to the section's own init(), keeping one source of
+    /// truth for the rules; failures abort the load exactly as before.
+    fn run_section_validation(&mut self) -> CommonResult<()> {
```
szbr00: run_section_validation is a second copy of ConfValidate for ClusterConf
(validation.rs:54): same five sections, same order. That impl is documented as
"what the pipeline invokes as its validate stage", but from() never calls it.
A later section that grows a validate() body will be easy to add in only one list.

from() should call self.validate() and drop this helper. from_transfer() can keep
its reduced client+transfer set inline.

crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -91,33 +91,52 @@ impl ClusterConf {
...
     pub const ENV_CONF_FILE: &'static str = "CURVINE_CONF_FILE";
 
+    /// Loads through the unified pipeline:
+    /// `file(toml) → env(allowlist) → deserialize once → normalize →
+    /// validate(section init) → resolve_master_addrs`.
+    /// See [`crate::pipeline`] for layer semantics.
     pub fn from<T: AsRef<str>>(path: T) -> CommonResult<Self> {
-        let mut conf = Self::read(path)?;
+        let text = pipeline::read_file_text(path.as_ref())?;
+        validation::warn_unknown_keys(&text);
```
szbr00, line 105: from() now calls warn_unknown_keys on every load. curvine-fuse validate-config
(validate_run.rs) calls get_conf() → from(), then re-audits [fuse] keys and prints
a second warning. A typo under [fuse] will appear twice, with different prefixes
("[WARN] unrecognized config key 'fuse.X'" vs "Warning: unrecognized [fuse]
config key 'X'").

Either drop the second audit in validate-config in a follow-up, or keep load-time
warnings here and have validate-config only add the state_dir check.

crates/common/curvine-config/src/cluster_conf.rs
lzjqsdd: Thanks. Added in 7ed1429d: `from_resolves_all_hostnames_via_net_interface` writes a temp file with net_interface (lo/lo0) plus file hostnames and loads it through the real `ClusterConf::from()`, asserting all five role hostnames resolve to 127.0.0.1. No process-env mutation needed.

crates/common/curvine-config/src/cluster_conf.rs
lzjqsdd: Already addressed: `run_section_validation` was removed and `ClusterConf` now implements `ConfValidate`, so `from()` calls `self.validate()` — the trait impl is the single copy of the section list. See commit 01b54058 / current cluster_conf.rs.

crates/common/curvine-config/src/cluster_conf.rs
lzjqsdd: Thanks. Fixed in a7a87f2a (carried on the stack): validate-config dropped its [fuse]-scoped re-audit since the loader itself warns once via the shared validation audit; the subcommand keeps only the state_dir filesystem check.

szbr00 (approved):

PR #1624, opened 2026-08-25
refactor(fuse): generate mount CLI overrides from FuseConf (#1619)

lzjqsdd:
# Summary

Fourth PR of the stacked series for #1619 (4/7). Generates the `curvine-fuse mount` CLI overrides from `FuseConf` via the existing `ClientCliArgs` derive (opt-in mode), replacing ~30 hand-written `if let Some(...)` branches so the CLI surface cannot drift from the config struct.

# Issue Describe / Design

Related to #1619 (partial). Depends on the stack #1620 -> #1622 -> #1623.

- Derive macro extended with an `alias` option.
- 27 historically tunable `[fuse]` fields annotated once; `FuseConfCliOverrides` is generated.
- Hand-written args remain only where no 1:1 scalar mapping exists: `--debug`/`--readonly` (SetTrue), `--options` list, `--master-addrs`.
- Deprecated `--mnt-per-task` spelling preserved as a clap alias; `--meta-cache-ttl` maps to `meta_cache_timeout`.

# Changes

| `crates/common/curvine-macros/src/client_cli.rs` | Add `alias` support to generated `#[arg]`s | None for existing users |
| `crates/common/curvine-config/src/fuse_conf.rs` | Derive + opt-in annotations; export `FuseConfCliOverrides` | None |
| `curvine-fuse/src/cli/mount_args.rs` | Replace manual override branches with one `apply_to` call | Known trade-off: generated flags lose per-flag help strings (restoration tracked as follow-up) |
| `curvine-fuse/src/cli/fuse_cli.rs` | Test accessors updated (`args.mount.fuse.*`) | Test-only |

# Test verified

| `cargo test -p curvine-fuse --lib cli` | PASS (35) | Includes legacy-alias and unset-flag-keeps-config-value cases |
| `cargo test -p curvine-macros` | PASS | |

# Dependencies

- Related PRs: base #1623
- Related issues: #1619


crates/common/curvine-macros/src/client_cli.rs
```diff
@@ -318,6 +345,12 @@ fn client_cli_attr(field: &Field) -> syn::Result<Option<ClientCliAttr>> {
     Ok(found)
 }
 
+/// Extra clap `alias` declared via `#[client_cli(alias = "...")]`, keeping
```
szbr00, line 348: The new `#[client_cli(alias = "...")]` option is the public contract this PR
adds to `ClientCliArgs`, but `crates/common/curvine-macros/src/lib.rs` rustdoc
still lists only `skip` and `long`. Callers of the derive (and the next stack
PR) will not see `alias` in the attribute list. Please add it there — e.g.
`#[client_cli(alias = "...")]` — extra clap alias for a deprecated flag name —
in the same field-level list as `long`.

curvine-fuse/src/cli/mount_args.rs
```diff
@@ -381,152 +194,51 @@ mod tests {
...
 
     #[test]
-    fn get_conf_cli_fuse_opts_replace_unsupported_config_fuse_opts() {
-        let conf = get_conf(
-            "[fuse]\nio_threads = 1\nfuse_opts = [\"auto_unmount\"]\n",
-            &["--options", "allow_other"],
-        );
-
-        assert_eq!(conf.fuse.fuse_opts, vec!["allow_other"]);
+    fn generated_overrides_apply_io_threads() {
```
szbr00, line 197: These new happy-path cases cover apply/alias/mapping, but this diff also
deletes the `FuseMountArgs::get_conf` tests that were the only CLI-boundary
checks for still-live wiring:

- `--io-threads 0` rejected after apply_to + `FuseConf::init()`
- default `fuse_opts` when neither CLI nor config supplies any
- invalid `--master-addrs` parse error

`FuseConf` unit tests still cover `init()` / `normalize_fuse_opts()` as
helpers; they will not fail if `get_conf` stops calling them after the
generated `apply_to`. Please restore those three cases (the VFS-conflict
suite can stay on `FuseConf` if you prefer).

crates/common/curvine-macros/src/client_cli.rs
lzjqsdd: Thanks. Fixed in a7a87f2a: the derive rustdoc now lists `#[client_cli(alias = "...")]` alongside `long` in the field-level attribute list.

curvine-fuse/src/cli/mount_args.rs
lzjqsdd: Thanks. Restored in a7a87f2a: `get_conf_rejects_zero_io_threads_cli_override`, `get_conf_applies_default_mnt_opts_when_neither_source_supplies_any`, and `get_conf_includes_parse_error_for_invalid_master_addrs` — all asserting through the real `get_conf` wiring rather than the FuseConf helpers.

crates/common/curvine-config/src/fuse_conf.rs
```diff
@@ -384,9 +418,10 @@ impl FuseConf {
 
     /// Parses raw cluster TOML text and returns the unrecognized keys found
     /// under the `[fuse]` table. Missing `[fuse]` table yields an empty list.
-    /// Parses raw cluster TOML text and returns the unrecognized keys found
-    /// under the `[fuse]` table. Delegates to the workspace-wide
-    /// [`crate::validation`] audit and filters to this section.
+    ///
+    /// Delegates to the workspace-wide [`crate::validation`] audit and filters
+    /// to this section, so semantics stay in lockstep with full-document
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in f0b8ad38: the duplicated paragraph on `unrecognized_fuse_keys_from_toml` is removed, leaving a single complete doc comment.

curvine-fuse/src/cli/mount_args.rs
```diff
@@ -381,112 +194,54 @@ mod tests {
     }
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in f0b8ad38: added `get_conf_splits_comma_separated_cli_options`, which asserts `--options allow_other,nodev` splits to `["allow_other", "nodev"]` through the real `get_conf` wiring (CLI parse → replace config fuse_opts → `FuseConf::normalize_fuse_opts`).

szbr00 (approved):

PR #1625, opened 2026-08-25
refactor(config): converge config discovery on ConfigLoader (#1619)

lzjqsdd:
# Summary

Fifth PR of the stacked series for #1619 (5/7). Converges config-file discovery on the shared `ConfigLoader`, removing the per-crate duplicates in the CLI, the server entrypoint, and curvine-lancedb.

# Issue Describe / Design

Related to #1619 (partial). Depends on the preceding stacked PRs.

- CLI: hand-rolled candidate list replaced by `ConfigLoader::discover` (identical order).
- lancedb: env fallback switched to `discover_configured` (explicit sources only), preserving its storage-option > master-addrs > env priority.
- server entrypoint: gains well-known-location fallback when launched without `--conf` or `CURVINE_CONF_FILE`.

# Changes

| `curvine-cli/src/main.rs` | Use `ConfigLoader::discover` | None (same candidates/order) |
| `curvine-server/src/bin/curvine-server.rs` | Discovery + clearer missing-config error | Improvement: falls back to well-known locations instead of failing on empty `--conf` |
| `curvine-lancedb/src/object_store.rs` | Env branch via `ConfigLoader::discover_configured` | None |

# Test verified

| `cargo test -p curvine-cli` / `-p curvine-server` / `-p curvine-lancedb` | PASS | |

# Dependencies

- Related PRs: base #1624
- Related issues: #1619

### DIFF
--- a/curvine-server/src/bin/curvine-server.rs
+++ b/curvine-server/src/bin/curvine-server.rs
@@ -14,7 +14,7 @@
 
 use clap::Parser;
 use curvine_alloc as _;
-use curvine_config::{ConfigLoader, ClusterConf};
+use curvine_config::{ClusterConf, ConfigLoader};
 use curvine_core_error::{err_box, CommonResult};
 use curvine_data_transfer::transfer::TransferServer;
 use curvine_master::master::Master;
@@ -113,15 +113,17 @@ impl ServerArgs {
         // Unified discovery: `--conf` > CURVINE_CONF_FILE > well-known
         // locations (launch scripts export the env var; bare local runs fall
         // back to ./conf/curvine-cluster.toml).
-        let found = ConfigLoader::discover(Some(&self.conf))
-            .ok_or_else(|| -> curvine_core_error::CommonError {
+        let found = ConfigLoader::discover(Some(&self.conf)).ok_or_else(
+            || -> curvine_core_error::CommonError {
                 format!(
                     "no configuration file found: pass --conf or set {}, or place \
-                     curvine-cluster.toml in a well-known location",
-                    ClusterConf::ENV_CONF_FILE
+                     {} in {{./, ./conf/, ~/.curvine/, $CURVINE_HOME/conf/}}",
+                    ClusterConf::ENV_CONF_FILE,
+                    ConfigLoader::DEFAULT_FILE_NAME
                 )
                 .into()
-            })?;
+            },
+        )?;
         println!("Loading config from {} ({})", found.as_str(), found.source);
 
         let load = match service {
@@ -160,6 +162,27 @@ mod tests {
 
         assert!(args.version_json);
         assert_eq!(args.component_name(), "server");
+    }
+
+    #[test]
+    fn get_conf_loads_from_discovered_config() {
+        let tmpdir = std::env::temp_dir().join(format!("cv-test-srv-conf-{}", std::process::id()));
+        let _ = std::fs::remove_dir_all(&tmpdir);
+        std::fs::create_dir_all(tmpdir.join("conf")).unwrap();
+        let conf_path = tmpdir.join("conf").join(ConfigLoader::DEFAULT_FILE_NAME);
+        std::fs::write(&conf_path, "[master]\nhostname = \"test-master\"\n").unwrap();
+
+        let args = ServerArgs::try_parse_from([
+            "curvine-server",
+            "--conf",
+            conf_path.to_str().unwrap(),
+            "--service",
+            "master",
+        ])
+        .expect("parse server args");
+        let conf = args.get_conf(&ServiceType::Master).unwrap();
+        assert_eq!(conf.master.hostname, "test-master");
+        let _ = std::fs::remove_dir_all(&tmpdir);
     }
 
     #[test]
--- a/curvine-server/src/bin/curvine-server.rs
+++ b/curvine-server/src/bin/curvine-server.rs
@@ -14,7 +14,7 @@
 
 use clap::Parser;
 use curvine_alloc as _;
-use curvine_config::ClusterConf;
+use curvine_config::{ConfigLoader, ClusterConf};
 use curvine_core_error::{err_box, CommonResult};
 use curvine_data_transfer::transfer::TransferServer;
 use curvine_master::master::Master;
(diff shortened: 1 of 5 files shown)


curvine-server/src/bin/curvine-server.rs
```diff
@@ -102,10 +102,26 @@ impl ServerArgs {
...
 
     pub fn get_conf(&self, service: &ServiceType) -> CommonResult<ClusterConf> {
-        match service {
-            ServiceType::Transfer => ClusterConf::from_transfer(&self.conf),
-            ServiceType::Master | ServiceType::Worker => ClusterConf::from(&self.conf),
-        }
+        // Unified discovery: `--conf` > CURVINE_CONF_FILE > well-known
+        // locations (launch scripts export the env var; bare local runs fall
+        // back to ./conf/curvine-cluster.toml).
+        let found = ConfigLoader::discover(Some(&self.conf)).ok_or_else(
```
szbr00: `ServerArgs.conf` defaults to `""`, so this `discover(Some(&self.conf))` call is the production contract for "no --conf": `discover_configured` must treat empty/whitespace as unset and fall through to `CURVINE_CONF_FILE` and the well-known locations. That empty-explicit path is not covered — `discovery_prefers_explicit_over_env_var` only exercises a non-empty path. A unit test on `ConfigLoader::discover(Some(""))` / `discover_configured(Some(""))` (and `Some("  ")`) that asserts the empty value does not win would lock the fallback this PR advertises; dropping the trim/empty filter would silently restore `ClusterConf::from("")` and break bare local starts.

lzjqsdd: Thanks. Locked in 3657aa7c: `empty_or_blank_explicit_path_does_not_win` asserts that `discover(Some(\"\"))` / `discover_configured(Some(\"\"))` / `Some(\"  \")` never produce a `--conf` source, so dropping the trim/empty filter would fail the test.

curvine-server/src/bin/curvine-server.rs
```diff
@@ -110,12 +110,25 @@ impl ServerArgs {
...
-        match service {
-            ServiceType::Transfer => ClusterConf::from_transfer(&self.conf),
-            ServiceType::Master | ServiceType::Worker | ServiceType::Mds => {
-                ClusterConf::from(&self.conf)
-            }
-        }
+        // Unified discovery: `--conf` > CURVINE_CONF_FILE > well-known
+        // locations (launch scripts export the env var; bare local runs fall
+        // back to ./conf/curvine-cluster.toml).
+        let found = ConfigLoader::discover(Some(&self.conf))
```
szbr00: Required: `ServerArgs.conf` still defaults to `""`, so `discover(Some(&self.conf))` is the production "no --conf" path. The trim/empty filter in `ConfigLoader::discover_configured` is what makes fallback to `CURVINE_CONF_FILE` and well-known locations work. The follow-up test `empty_or_blank_explicit_path_does_not_win` landed in 3657aa7c after the previous review, but it is not on this head (cc8cb1dc, rebased onto main). `discovery_prefers_explicit_over_env_var` only covers a non-empty path, and `cargo test -p curvine-server` does not call `get_conf`. Restore that unit test on `discover(Some(""))` / `discover_configured(Some(""))` / `Some("  ")` so dropping the filter cannot silently restore `ClusterConf::from("")` and break bare local starts.

szbr00 (changes requested):
Requesting changes on the empty-`--conf` discovery contract.

`get_conf` now calls `ConfigLoader::discover(Some(&self.conf))` with the clap default `""`. That fallback is correct only while `discover_configured` treats empty/whitespace as unset. The unit test that locked this (`empty_or_blank_explicit_path_does_not_win` in 3657aa7c) is not on this head after the rebase onto main.

Inspected: `curvine-cli` resolve path, `curvine-server` `get_conf`, `curvine-lancedb` `resolve_curvine_conf` / `curvine_store_identity`, and `ConfigLoader::{discover,discover_configured}` plus its tests on cc8cb1dc. CLI/lancedb wiring matches the prior behavior; no other blockers.

curvine-server/src/bin/curvine-server.rs
lzjqsdd: Addressed. Fixed in the commit that restored empty_or_blank_explicit_path_does_not_win (also included in the rebased branch).

curvine-server/src/bin/curvine-server.rs
```diff
@@ -110,12 +110,25 @@ impl ServerArgs {
...
+            .ok_or_else(|| -> curvine_core_error::CommonError {
+                format!(
+                    "no configuration file found: pass --conf or set {}, or place \
+                     curvine-cluster.toml in a well-known location",
+                    ClusterConf::ENV_CONF_FILE
+                )
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 5cfb300f: the error now enumerates the actual probe directories (./, ./conf/, ~/.curvine/, $CURVINE_HOME/conf/) and uses ConfigLoader::DEFAULT_FILE_NAME so the message stays in sync with discovery.

curvine-server/src/bin/curvine-server.rs
```diff
@@ -110,12 +110,25 @@ impl ServerArgs {
...
-            ServiceType::Transfer => ClusterConf::from_transfer(&self.conf),
-            ServiceType::Master | ServiceType::Worker | ServiceType::Mds => {
-                ClusterConf::from(&self.conf)
-            }
-        }
+        // Unified discovery: `--conf` > CURVINE_CONF_FILE > well-known
+        // locations (launch scripts export the env var; bare local runs fall
+        // back to ./conf/curvine-cluster.toml).
+        let found = ConfigLoader::discover(Some(&self.conf))
+            .ok_or_else(|| -> curvine_core_error::CommonError {
```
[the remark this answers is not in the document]
lzjqsdd: Thanks. Fixed in 5cfb300f: added get_conf_loads_from_discovered_config which creates a temp dir with conf/curvine-cluster.toml, passes --conf, and asserts the config loads correctly.

curvine-server/src/bin/curvine-server.rs
lzjqsdd: Thanks. Fixed: the empty_or_blank_explicit_path_does_not_win test was restored by cherry-picking the lost commit after rebase onto main. The discover(Some("")) / discover(Some("  ")) scenario is now covered again.

szbr00 (approved):

PR #1626, opened 2026-08-25
feat(config): add hot-reload extension points (#1619)

lzjqsdd:
# Summary

Sixth PR of the stacked series for #1619 (6/7). Adds hot-reload extension points only — no watcher or admin RPC is wired yet.

# Issue Describe / Design

Related to #1619 (partial). Depends on the preceding stacked PRs.

- **`Reloadable`** marker: sections safe to swap at runtime (first implementation: `LogConf`).
- **`RuntimeOverlay`**: ephemeral admin-pushed dotted-path overrides, deliberately lost on restart.
- **`ConfigHandle`**: `Arc`-snapshot holder; readers clone the Arc, so a swap never tears an in-flight consumer. No new dependencies.
- Module docs describe the two future mechanisms: file-watch reload and admin-API pushdown — both are direct follow-ups since the pipeline is a pure function from layers to a validated `ClusterConf`.

# Changes

| `crates/common/curvine-config/src/reload.rs` (new) | Extension-point types + docs | None (no production callers yet) |
| `crates/common/curvine-config/src/lib.rs` | Export `reload` module | None |

# Test verified

| `cargo test -p curvine-config` | PASS | Overlay set/replace/clear; handle swap leaves old snapshots untouched |

# Dependencies

- Related PRs: base #1625
- Related issues: #1619

### DIFF
--- a/crates/common/curvine-config/src/reload.rs
+++ b/crates/common/curvine-config/src/reload.rs
@@ -1,0 +1,187 @@
+// Copyright 2025 OPPO.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Hot-reload extension points.
+//!
+//! This module wires up the *shapes* only; no file watcher or admin RPC is
+//! connected yet. The pipeline in [`crate::pipeline`] is a pure function from
+//! layers to a validated [`ClusterConf`], which makes both planned mechanisms
+//! straightforward follow-ups:
+//!
+//! 1. **File-change reload** (persistent semantics): a watcher re-runs
+//!    `build_document` + deserialization + `ConfValidate` on the new file text,
+//!    then swaps the result into a [`ConfigHandle`]. Failed validation keeps
+//!    the previous snapshot — a bad edit must never take a process down.
+//!
+//! 2. **Admin-API push** (ephemeral semantics): handlers write dotted-path
+//!    overrides into a [`RuntimeOverlay`] and trigger the same
+//!    rebuild-and-swap flow with the overlay merged above the environment layer.
+//!    Overrides are never persisted: restarting the process deliberately
+//!    reverts to on-disk configuration.
+//!
+//! Effective layer order with an overlay active:
+//! `file -> environment -> RuntimeOverlay -> deserialize`, so an operator's
+//! pushed value cannot be clobbered by `CURVINE_*` env vars.
+//!
+//! Readers go through [`ConfigHandle::get`], cloning an `Arc` snapshot — no
+//! locks held while the config is in use, so a swap never tears a request
+//! that started under the previous values.
+
+use crate::ClusterConf;
+use curvine_core_error::CommonResult;
+use std::sync::{Arc, RwLock};
+
+/// Marker for a configuration value that is safe to swap at runtime.
+///
+/// Sections whose state is captured by long-lived components (open pools,
+/// bound sockets, data directories) must NOT implement this trait: swapping
+/// their value would silently diverge from actual component behavior.
+///
+/// Known future candidates, deliberately NOT yet implemented:
+/// - `LogConf` — `Logger::init` captures it once behind a `OnceCell` and
+///   installs a global subscriber; hot-swapping the value would not change
+///   the running logger until Logger gains a real reload hook.
+pub trait Reloadable {
+    /// Section name used in reload audit logs (e.g. `"log"`).
+    fn reload_name(&self) -> &'static str;
+}
+
+/// Ephemeral override layer for admin-pushed values.
+///
+/// Entries are dotted paths into the cluster TOML document (same addressing as
+/// [`crate::pipeline`]) and are merged above the environment layer on every rebuild.
+/// Intentionally memory-only: restart discards them by design.
+#[derive(Default)]
+pub struct RuntimeOverlay {
+    entries: RwLock<Vec<(String, toml::Value)>>,
+}
+
+impl RuntimeOverlay {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Pushes (or replaces) one override at a dotted path.
+    pub fn set(&self, path: impl Into<String>, value: toml::Value) {
+        let path = path.into();
+        let mut entries = self.entries.write().expect("runtime overlay poisoned");
+        match entries.iter_mut().find(|(p, _)| *p == path) {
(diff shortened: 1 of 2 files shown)


crates/common/curvine-config/src/reload.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Merges the overrides into a document, above the file layer. Public for
+    /// the future file-watcher/admin rebuild flow (and tests) until that flow
+    /// is wired; unused in production code paths today.
+    pub fn apply_to(&self, doc: &mut toml::Value) {
+        let entries = self.entries.read().expect("runtime overlay poisoned");
+        for (path, value) in entries.iter() {
+            // set_dotted is infallible here in practice: overlay paths target
+            // existing keys that already deserialized successfully once.
+            if let Err(e) = crate::pipeline::set_dotted(doc, path, value.clone()) {
+                eprintln!("[WARN] dropping invalid runtime override '{}': {}", path, e);
```
szbr00: apply_to drops set_dotted errors and continues, so an admin-pushed override can
appear in list() while never landing in the document. set_dotted is not
infallible: a path that collides with a non-table parent (e.g.
client.rpc_timeout_ms.extra when rpc_timeout_ms is an integer) returns Err. The
comment that overlays only retarget keys that already deserialized is also
wrong — set_dotted creates missing tables, which is how the unit test writes
log.level into a document that had only [client].

Return CommonResult from apply_to (fail the rebuild, keep the previous
ConfigHandle snapshot) instead of eprintln! + drop. Add a collision-path unit
test so a swallowed override cannot stay green.

crates/common/curvine-config/src/reload.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Sections whose state is captured by long-lived components (open pools,
+/// bound sockets, data directories) must NOT implement this trait: swapping
+/// their value would silently diverge from actual component behavior.
+/// First candidate: [`LogConf`] (log level/dir are consulted per call).
+pub trait Reloadable {
+    /// Section name used in reload audit logs (e.g. `"log"`).
+    fn reload_name(&self) -> &'static str;
+}
+
+impl Reloadable for LogConf {
```
szbr00: LogConf is not safe to mark Reloadable under this trait's own rule. Logger::init captures the value once via OnceCell and installs a global tracing subscriber (level via with_max_level / LevelFilter; writers/appenders from log_dir). A later ConfigHandle::store of a new ClusterConf does not change the running logger, so swapping log.level / log.log_dir would silently diverge from actual process behavior — the exact failure mode the trait docs forbid.

Either drop this impl until Logger has a real reload hook, or keep LogConf as a documented future candidate and do not implement Reloadable for it yet. The "consulted per call" claim on the trait docs is not true of the current Logger.

crates/common/curvine-config/src/reload.rs
```diff
@@ -0,0 +1,185 @@
...
+    fn reload_name(&self) -> &'static str {
+        "log"
+    }
+}
+
+/// Ephemeral override layer for admin-pushed values.
+///
+/// Entries are dotted paths into the cluster TOML document (same addressing as
+/// [`crate::pipeline`]) and are merged above the file layer on every rebuild.
```
szbr00, line 63: These two comments in the same PR disagree on overlay rank. Here RuntimeOverlay is "merged above the file layer"; pipeline.rs:102-104 says the admin/runtime overlay ranks ABOVE the environment layer and is merged between apply_env_layer_with and deserialization. "Above the file layer" is also how build_document's overlays parameter is described, and those sit BELOW env.

Pick one contract and write it in both places. The safe direction for admin push is file → env → RuntimeOverlay → deserialize, so an operator override cannot be clobbered by CURVINE_*_HOSTNAME. apply_to should then be documented as running on the post-env document, not as a file-layer peer.

crates/common/curvine-config/src/reload.rs
lzjqsdd: Thanks, both points confirmed. Fixed in 4f6ef43d: `apply_to` now returns `CommonResult<()>` and propagates the first `set_dotted` error so the caller can abort the rebuild and keep the previous `ConfigHandle` snapshot; the incorrect "already deserialized" comment is gone (set_dotted does create missing tables); a scalar-parent collision test (`apply_to_fails_on_colliding_override_path`) guards against swallowing.

crates/common/curvine-config/src/reload.rs
lzjqsdd: Thanks — correct, the "consulted per call" claim was wrong for the current Logger. Fixed in 4f6ef43d: the `Reloadable` impl for LogConf is removed and LogConf is documented as a known future candidate that stays unimplemented until Logger gains a real reload hook.

crates/common/curvine-config/src/reload.rs
lzjqsdd: Fixed in 4f6ef43d: both docs now state the same contract — file -> environment -> RuntimeOverlay -> deserialize — and `apply_to` documents that it runs on the post-environment document. Agreed this is the safe direction: an operator push cannot be clobbered by CURVINE_* env vars.

crates/common/curvine-config/src/reload.rs
```diff
@@ -0,0 +1,187 @@
...
+///   the running logger until Logger gains a real reload hook.
+pub trait Reloadable {
+    /// Section name used in reload audit logs (e.g. `"log"`).
+    fn reload_name(&self) -> &'static str;
+}
+
+/// Ephemeral override layer for admin-pushed values.
+///
+/// Entries are dotted paths into the cluster TOML document (same addressing as
+/// [`crate::pipeline`]) and are merged above the file layer on every rebuild.
```
szbr00, line 63: Optional: These two remaining "merged above the file layer" sentences (this struct doc, and the Admin-API bullet in the module docs) still describe the same rank as `build_document`'s `overlays` parameter, which sits below env. `apply_to` and the "Effective layer order" line now correctly say file → env → RuntimeOverlay → deserialize. Please replace "above the file layer" with "above the environment layer" (or "on the post-environment document") so the leftover phrase cannot send the wiring PR into `build_document(..., overlays)`.

szbr00 (commented):
Re-reviewed head c96ee5b3 against base unified-conf-p5. Dirty crate: curvine-config (reload.rs + lib.rs export).

Previous findings on apply_to error swallowing, LogConf Reloadable, and overlay rank are addressed: apply_to returns CommonResult and the collision test fails closed; LogConf is a documented future candidate only; apply_to and the module Effective layer order match pipeline.rs (file → env → overlay → deserialize).

Inspected overlay set/replace/clear and snapshot isolation tests, the set_dotted collision path, ConfigHandle Arc swap (lock not held after get), no production callers (intentional per #1619), no proto/SDK/lockfile changes, and no data-path impact.

One leftover: RuntimeOverlay / Admin-API docs still say "above the file layer" (see inline). Not blocking.

crates/common/curvine-config/src/reload.rs
lzjqsdd: Thanks. Fixed: both remaining occurrences of "above the file layer" in reload.rs have been replaced with "above the environment layer" to match the correct file → env → RuntimeOverlay → deserialize rank.

szbr00 (approved):

PR #1627, opened 2026-08-25
docs(config): document env vars and load priority in sample config (#1619)

lzjqsdd:
# Summary

Seventh and final PR of the stacked series for #1619 (7/7). Documents the unified priority chain, the recognized environment variables, and the unknown-key warning semantics in the shipped sample config.

# Changes

| `etc/curvine-cluster.toml` | Header comment block: priority chain (CLI > env > file > defaults), env var list, warning semantics | Comment-only |

# Test verified

| `workspace_sample_config_audits_clean` (in #1622) | PASS | Guards the sample against unrecognized keys |

# Dependencies

- Related PRs: base #1626; closes out the series started by #1620
- Related issues: #1619

### DIFF
--- a/etc/curvine-cluster.toml
+++ b/etc/curvine-cluster.toml
@@ -1,4 +1,25 @@
 
+# ---------------------------------------------------------------------------
+# Configuration loading priority (unified pipeline, all components):
+#
+#   CLI flags  >  CURVINE_* environment variables  >  this file  >  built-in defaults
+#
+# Exception: a non-empty `net_interface` below outranks the four hostname
+# env vars — when it is set, the NIC-resolved IPv4 replaces every role
+# hostname and CURVINE_*_HOSTNAME is ignored (with a warning).
+#
+# Recognized environment variables:
+#   CURVINE_CONF_FILE          config file path used when --conf is omitted
+#   CURVINE_MASTER_HOSTNAME    overrides [master].hostname (and seeds journal.hostname)
+#   CURVINE_WORKER_HOSTNAME    overrides [worker].hostname
+#   CURVINE_CLIENT_HOSTNAME    overrides [client].hostname
+#   CURVINE_TRANSFER_HOSTNAME  overrides [transfer].hostname
+#
+# Keys in this file that no component recognizes are reported as warnings at
+# load time (possible typos or settings removed by upgrades); they never abort
+# startup.
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


etc/curvine-cluster.toml
```diff
@@ -1,4 +1,21 @@
 
+# ---------------------------------------------------------------------------
+# Configuration loading priority (unified pipeline, all components):
+#
+#   CLI flags  >  CURVINE_* environment variables  >  this file  >  built-in defaults
```
szbr00, line 5: The layer order here is right (CLI > env allowlist > file > defaults), but it is not the rule operators will actually hit for the four hostname variables listed below.

`apply_env_layer_with` treats a non-empty `net_interface` in this file as a hard winner: it writes the NIC IPv4 onto every role hostname and ignores `CURVINE_*_HOSTNAME` (see `pipeline.rs` 200–226, and the comment at lines 25–29 of this same file). An operator who sets both will follow this header and expect the env var to win.

Please add a one-line exception next to the chain, e.g. that a non-empty `net_interface` outranks the hostname env vars (point at the existing comment below).

etc/curvine-cluster.toml
```diff
@@ -1,4 +1,21 @@
...
+#
+# Recognized environment variables:
+#   CURVINE_CONF_FILE          path to this file (highest-priority source discovery)
```
szbr00: `CURVINE_CONF_FILE` is not the highest-priority discovery source. `ConfigLoader::discover` takes an explicit `--conf` path first, then this env var, then the well-known fallbacks (`pipeline.rs` 333–338).

The parenthetical also blurs discovery with value overlay. Suggest: "config file path when `--conf` is omitted" (or similar) so it cannot be read as beating CLI.

etc/curvine-cluster.toml
lzjqsdd: Thanks. Added in 4c7de622: a one-line exception right under the priority chain noting that a non-empty net_interface outranks the four hostname env vars, pointing at the detailed comment below in the file.

etc/curvine-cluster.toml
lzjqsdd: Thanks. Rephrased in 4c7de622 to: \"config file path used when --conf is omitted\" — no longer reads as beating the CLI flag and keeps discovery separate from value overlay.

szbr00 (approved):

PR #1631, opened 2026-08-26
feat(master): local policy assigns remaining replicas by weight

bigbigxu:
# Summary

LocalWorkerPolicy keeps the local worker as the first replica, but now delegates the remaining replicas to WeightedWorkerPolicy instead of RobinWorkerPolicy, so extra replicas are distributed proportionally to `worker.weight` and nodes with weight 0 receive no new data.

# Issue Describe / Design

No related issue.

- Local-first selection is unchanged except for drain semantics: a local worker with weight 0 is now skipped as well, so a drained node receives no new replicas even from co-located clients.
- Remaining replicas previously rotated round-robin regardless of node capacity; weighted random selection lets operators steer replicas to higher-capacity nodes via `worker.weight` and drain a node by setting its weight to 0.
- Under-replication is preserved: when a local replica is selected but no eligible remote worker remains (single-node clusters, drained or dead peers), selection returns the local replica alone instead of failing, matching the previous RobinWorkerPolicy behavior and the WorkerManager contract. The weighted error still propagates when nothing was selected at all.
- `choose_workers` (no block context) keeps pure weighted behavior, since the local worker cannot be determined without `client_host`.

# Changes

| `curvine-master/src/master/fs/policy/local_worker_policy.rs` | Replace RobinWorkerPolicy fallback with WeightedWorkerPolicy; skip weight-0 local workers; under-replicate instead of failing when no remote worker is eligible | Behavior change: remaining replicas are weighted-random instead of round-robin; drained (weight 0) local nodes no longer receive the first replica |
| `etc/curvine-cluster.toml` | Document local policy semantics in the worker_policy comment | None |

# Test verified

| `local_first_then_weighted_remaining` | Added | Local worker first; weight-0 remotes never chosen by the fallback |
| `under_replicates_when_no_eligible_remote_worker` | Added | Local selected + no eligible remote returns Ok([local]), not an error |
| `no_local_worker_falls_back_to_weighted` | Added | No local worker -> full weighted selection |
| `errors_when_no_worker_is_eligible` | Added | No local worker and all weight 0 -> error propagates |
| `skips_drained_local_worker` | Added | Weight-0 local worker receives no replica |
| `cargo test -p curvine-master` | Not run locally | Windows dev box cannot build platform dependencies; rely on CI |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/fs/policy/local_worker_policy.rs
+++ b/curvine-master/src/master/fs/policy/local_worker_policy.rs
@@ -17,7 +17,10 @@ use curvine_core_error::{err_box, CommonResult};
 use curvine_model::{WorkerAddress, WorkerInfo};
 use indexmap::IndexMap;
 
-/// Local workers are preferred, and weighted policies are used if there are no local workers
+/// A local worker is preferred for the first replica, and the remaining replicas are assigned
+/// by the weighted policy. If no eligible local worker exists, all replicas are assigned by
+/// the weighted policy. Local workers with weight 0 are skipped so that drained nodes
+/// receive no new data.
 pub struct LocalWorkerPolicy {
     inner: WeightedWorkerPolicy,
 }
@@ -56,6 +59,7 @@ impl WorkerPolicy for LocalWorkerPolicy {
             if !ctx.exclude_workers.contains(id)
                 && worker.address.is_local(&ctx.client_host)
                 && worker.available > ctx.block_size
+                && worker.weight > 0
                 && worker.is_live()
             {
                 res.push(worker.address.clone());
@@ -66,11 +70,22 @@ impl WorkerPolicy for LocalWorkerPolicy {
 
         let replicas = ctx.replicas;
         if res.len() < replicas as usize {
-            let remote_res = self.inner.choose(workers, ctx)?;
-            for item in remote_res {
-                res.push(item);
-                if res.len() == replicas as usize {
-                    break;
+            match self.inner.choose(workers, ctx) {
+                Ok(remote_res) => {
+                    for item in remote_res {
+                        res.push(item);
+                        if res.len() == replicas as usize {
+                            break;
+                        }
+                    }
+                }
+                // The weighted policy errors when no remote worker is eligible. A local
+                // replica is still valid, so under-replicate instead of failing; keep
+                // the error when nothing was selected at all.
+                Err(e) => {
+                    if res.is_empty() {
+                        return Err(e);
+                    }
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


curvine-master/src/master/fs/policy/local_worker_policy.rs
```diff
@@ -12,14 +12,14 @@
...
+use crate::master::fs::policy::{ChooseContext, WeightedWorkerPolicy, WorkerPolicy};
 use curvine_core_error::{err_box, CommonResult};
 use curvine_model::{WorkerAddress, WorkerInfo};
 use indexmap::IndexMap;
 
-/// Local workers are preferred, and polling policies are used if there are no local workers
+/// Local workers are preferred, and weighted policies are used if there are no local workers
 pub struct LocalWorkerPolicy {
-    inner: RobinWorkerPolicy,
+    inner: WeightedWorkerPolicy,
```
szbr00, line 25: Critical: Replacing RobinWorkerPolicy with WeightedWorkerPolicy changes the
failure mode of the existing `self.inner.choose(workers, ctx)?` call (line 69).

RobinWorkerPolicy::choose returns Ok([]) when nothing remaining is eligible.
WeightedWorkerPolicy::choose returns Err("No eligible workers for weighted
selection") when total_weight == 0. The `?` then fails the whole Local
selection even though a local replica is already in `res`.

WorkerManager::choose_worker already accepts under-replication (non-empty
result with len <= replicas). This composition breaks that contract on the
default policy (`master.worker_policy = "local"`). Concrete cases that used
to succeed and now fail addBlock/create:

- 1 live worker, replicas > 1 (local is excluded, then weighted sees no one)
- local worker found, every remaining worker has weight 0 / no space / is dead

Fix at the composition site, not by changing Robin: if `res` is already
non-empty, treat "no eligible weighted workers" as an empty remote set and
return the local replica. Keep the error when `res` is empty (no local
worker). Add a unit test that fails today: local worker selected, remaining
ineligible, choose must return Ok([local]) rather than Err.

szbr00, line 25: Consider: The PR summary says nodes with weight 0 receive no new data, but
local-first eligibility (not excluded, same host, available > block_size,
live) still does not check `worker.weight`. A co-located client will keep
writing replica 1 to a weight-0 node, so drain is only applied to extra
replicas.

If drain-by-weight-0 is a goal for the default local policy, skip weight == 0
in the local-first loop as well. If local-first is meant to override drain,
say so in the struct / cluster.toml comment so operators do not assume
weight 0 fully silences a node under `worker_policy = "local"`.

szbr00 (changes requested):
Required: This changes default master placement (`worker_policy = "local"`)
and has no tests. LocalWorkerPolicy has no `#[cfg(test)]` module; weighted
and robin each test their own policy, which does not cover this composition.
CI compile does not catch the under-replication regression above.

Please add tests in `local_worker_policy.rs` whose assertions fail on the
intended defects, at least:

1. Local worker is first; remaining replicas come from other workers and
   respect weight (weight-0 remotes are not chosen).
2. Local worker found and no remaining eligible workers → Ok with the local
   worker only, not an error.
3. No local worker → full selection is weighted (including error when every
   worker has weight 0).

etc/curvine-cluster.toml
```diff
@@ -34,6 +34,7 @@ auth_token_env = "CURVINE_FAULT_TOKEN"
 [master]
 meta_dir = "testing/meta"
 # Worker selection policy: local, robin, random, load_based, or weighted.
+# local prefers a local worker for the first replica and assigns the rest by weight.
```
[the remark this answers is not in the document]
bigbigxu: Fixed in e1ae4409a: the comment now references the policy name directly - 'assigns the rest via the weighted policy'.

curvine-master/src/master/fs/policy/local_worker_policy.rs
```diff
@@ -12,14 +12,14 @@
...
 // limitations under the License.
 
-use crate::master::fs::policy::{ChooseContext, RobinWorkerPolicy, WorkerPolicy};
+use crate::master::fs::policy::{ChooseContext, WeightedWorkerPolicy, WorkerPolicy};
 use curvine_core_error::{err_box, CommonResult};
 use curvine_model::{WorkerAddress, WorkerInfo};
 use indexmap::IndexMap;
 
-/// Local workers are preferred, and polling policies are used if there are no local workers
+/// Local workers are preferred, and weighted policies are used if there are no local workers
```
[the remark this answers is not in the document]
bigbigxu: Fixed in e1ae4409a: rewrote the struct doc comment to describe the actual behavior - a local worker is preferred for the first replica, the remaining replicas are always assigned by the weighted policy, and drained local workers (weight 0) are skipped.

curvine-master/src/master/fs/policy/local_worker_policy.rs
```diff
@@ -84,7 +84,8 @@ impl WorkerPolicy for LocalWorkerPolicy {
         count: Option<usize>,
         exclude_workers: Vec<u32>,
     ) -> CommonResult<Vec<WorkerAddress>> {
-        // Since you do not rely on block information, you cannot judge the local worker, and use the polling strategy directly
+        // Since you do not rely on block information, you cannot judge the local worker,
+        // and use the weighted strategy directly
```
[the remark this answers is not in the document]
bigbigxu: Fixed in e1ae4409a: rephrased to 'Without block or client context the local worker cannot be determined, so delegate directly to the weighted policy.'

curvine-master/src/master/fs/policy/local_worker_policy.rs
bigbigxu: Fixed in e1ae4409a. Chose the first option: the local-first loop now skips weight == 0 as well, so a drained node receives no new replicas even from co-located clients, keeping the weight-0 drain semantics consistent for the default local policy. Documented in the struct doc comment and covered by test skips_drained_local_worker.

bigbigxu: Fixed in e1ae4409a. Fixed at the composition site as suggested: the weighted error is now treated as an empty remote set when a local replica is already selected (under-replicate instead of fail), and still propagates when nothing was selected at all. Added unit test under_replicates_when_no_eligible_remote_worker covering exactly this case: local worker selected, remaining worker weight 0, choose returns Ok([local]) instead of Err.

szbr00 (approved):

PR #1635, opened 2026-08-27
fix(libsdk): fix mount lifecycle to prevent use-after-free and double close

bigbigxu:
# Summary

Fixes CurvineFileSystem mount lifecycle so the shared native mount is released
exactly once, only when the last FileSystem instance using it is closed, and is
never raced with in-flight JNI calls. Removes the JVM shutdown-hook cleanup that
crashed Spark/YARN containers at exit, and makes reads on closed streams throw.

# Issue Describe / Design

**Root cause**: the native handle is freed unconditionally by
closeWriter/closeReader/closeFilesystem once Java decides to release it. The
previous code never closed the cached mount (delegated cleanup to a JVM
shutdown hook), and that hook raced with application threads still issuing JNI
calls at JVM exit -> use-after-free (SIGSEGV in tokio::runtime::handle::Handle::
enter, observed on Spark ApplicationMaster containers).

**Fix approach**:
- CachedMount refcount: close() releases the mount only when the last
  CurvineFileSystem instance is closed (refcount reaches 0); the cache entry is
  removed under the cache lock;
- closeOnce() is synchronized and idempotent -> the native handle is released
  exactly once (no double free);
- getOrCreateMount performs lookup + refcount under the cache lock -> callers
  can never observe a closed mount (no use-after-free);
- Removed the JVM shutdown-hook cleanup: leaked mounts are left to OS
  reclamation at process exit instead of racing with in-flight calls;
- CurvineInputStream.read() now checks closed first so reads on closed streams
  throw IOException instead of silently returning EOF.

# Changes

| `curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java` | mount refcount lifecycle, idempotent closeOnce, lock-protected cache, remove shutdown hook | Behavior change: native mount is released when the last instance closes; no exit-time fallback cleanup |
| `curvine-libsdk/java/src/main/java/io/curvine/CurvineInputStream.java` | check closed before read | Behavior change: reads on closed streams throw IOException |
| `crates/client/curvine-client-core/src/file/curvine_filesystem.rs` | log git commit in filesystem creation | None (logging only) |

# Test verified

| `mvn compile` (curvine-libsdk/java) | PASS | JDK 8 / Maven 3.9 |
| Spark TPC-H on YARN (Spark 3.5, Hadoop 3.3) | PASS | Crash at JVM exit disappeared after removing shutdown-hook cleanup |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
+++ b/curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
@@ -20,7 +20,6 @@ import java.net.URI;
 import java.nio.file.DirectoryNotEmptyException;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.classification.InterfaceAudience;
@@ -60,30 +59,34 @@ public class CurvineFileSystem extends FileSystem {
      * Value: CachedMount containing the shared CurvineFsMount and reference count
      */
     private static final ConcurrentHashMap<String, CachedMount> MOUNT_CACHE = new ConcurrentHashMap<>();
-    
+
     private static class CachedMount {
         final CurvineFsMount mount;
-        final AtomicInteger refCount;
-        
+        int refCount;
+        private boolean closed;
+
         CachedMount(CurvineFsMount mount) {
             this.mount = mount;
-            this.refCount = new AtomicInteger(1);
+            this.refCount = 1;
         }
-    }
-    
-    static {
-        // Register shutdown hook to clean up all cached mounts
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            LOGGER.info("Shutting down CurvineFileSystem, closing {} cached mounts", MOUNT_CACHE.size());
-            for (CachedMount cached : MOUNT_CACHE.values()) {
-                try {
-                    cached.mount.close();
-                } catch (IOException e) {
-                    LOGGER.warn("Error closing cached mount", e);
-                }
+
+        /**
+         * Close the underlying mount exactly once; later calls are no-ops.
+         * The lock makes it idempotent against concurrent closes from multiple
+         * CurvineFileSystem instances sharing the mount, so the native handle
+         * is never released twice (double free).
+         */
+        synchronized void closeOnce() {
+            if (closed) {
+                return;
             }
-            MOUNT_CACHE.clear();
-        }));
+            closed = true;
+            try {
+                mount.close();
+            } catch (IOException e) {
+                LOGGER.warn("Error closing cached mount", e);
+            }
+        }
     }
 
     private CurvineFsMount libFs;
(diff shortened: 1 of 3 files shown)


curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
```diff
@@ -321,15 +319,18 @@ public boolean mkdirs(Path f, FsPermission permission) throws IOException {
...
+            synchronized (MOUNT_CACHE) {
+                cached = MOUNT_CACHE.get(cacheKey);
+                last = cached != null && --cached.refCount <= 0;
+                if (last) {
+                    MOUNT_CACHE.remove(cacheKey);
+                }
+            }
+            if (last) {
+                cached.closeOnce();
             }
```
szbr00, line 337: Critical: cached.closeOnce() calls CurvineFsMount.close() → CurvineNative.closeFilesystem, which drops the native JavaFilesystem and its tokio Runtime (java_abi.rs closeFilesystem → FFIUtils::free_raw_ptr). The refcount only tracks CurvineFileSystem instances, not open streams or in-flight JNI.

CurvineInputStream/CurvineOutputStream keep the shared CurvineFsMount and will still call JNI after this FileSystem.close(); Hadoop FileSystem.close() does not close those streams. Hadoop FileSystem.CACHE also invokes this close() from its JVM shutdown hook (fs.automatic.close defaults to true on hadoop-common 3.3.4). That is the Spark AM exit path this PR is fixing — removing Curvine's unordered Runtime.addShutdownHook is correct, but freeing native memory here reintroduces SIGSEGV in tokio::runtime::handle::Handle::enter.

Do not free the native mount from this path. Drop the cache entry only and leave the runtime to process-exit reclamation, or tombstone CurvineFsMount so every JNI entry throws IOException, hold an inflight/stream count, skip native close when ShutdownHookManager.isShutdownInProgress(), and run super.close() before any native teardown.

szbr00 (changes requested):
Required: There is no test that fails on the lifecycle regressions this PR claims to fix. mvn compile does not exercise close; one Spark TPC-H pass does not lock last-close against in-flight JNI.

Please add assertions on the real Java entry path: (1) two FileSystem instances sharing master_addrs — an intermediate close must not call closeFilesystem; (2) read/close on an open stream after last FileSystem.close() must throw IOException, not crash; (3) CurvineInputStream.read after stream close throws IOException (the checkClosed move).

curvine-libsdk/java/src/main/java/io/curvine/CurvineFileSystem.java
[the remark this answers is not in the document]
bigbigxu: Fixed in af3b45e79: closeOnce() now propagates the IOException from mount.close() instead of swallowing it, and CurvineFileSystem.close() clears its per-instance state (libFs = null) in a finally block so a failed native close cannot break close() idempotency. Callers can now detect abnormal native cleanup on the explicit close path.

bigbigxu: Disagree. We intentionally leave the FileSystem lifecycle under Hadoop's management: with our own shutdown hook removed, close() is only invoked through the normal Hadoop lifecycle (explicit user close, or Hadoop's FileSystem.CACHE closeAll() at JVM exit when fs.automatic.close is on). In that flow refcount reaching 0 means every CurvineFileSystem instance has been closed, so releasing the native mount at that point is the expected contract, not a race. The race we fixed was caused by our own unordered shutdown hook closing mounts concurrently with Hadoop's closeAll() hook - registering any extra hook of our own (including a shutdown-progress check) conflicts with Hadoop's lifecycle management and is exactly the premature-release path this PR removes. Using a stream after its owning FileSystem has been closed violates the Hadoop close-before-last-use contract and is a caller bug, not something the SDK can or should defend against; per earlier discussion, streams are intentionally not thread-safe and not guarded. Skipping native close on ShutdownHookManager.isShutdownInProgress() would also silently leak the mount in normal explicit-close flows that happen to run during shutdown, weakening resource release without adding safety.

bigbigxu:
Tests added in 26443fd67: **CurvineFileSystemLifecycleTest** (pure-Java, no cluster/native library required - the native mount is replaced by a FakeMount allocated via Unsafe, and the private MOUNT_CACHE/CachedMount state is driven via reflection).

Coverage against the requested assertions:
1. **Intermediate close must not release the mount** - \intermediateCloseMustNotReleaseSharedMount\: two instances share one mount, closing the first leaves the cache entry in place, refcount drops by one only, and closeFilesystem is never called.
2. **Last close releases exactly once** - \lastCloseReleasesSharedMountExactlyOnce\: after the second close the entry is removed and closeFilesystem is called exactly once.
3. **Closed stream read throws IOException** - \
eadOnClosedStreamThrowsIOException\ + \
eadOnClosedStreamDoesNotDependOnFileSize\: the checkClosed move is pinned so a closed stream throws instead of returning EOF/0 regardless of file size.
4. Bonus: \closeFailureIsPropagatedToCaller\ verifies a failed native close is propagated to the caller (per the Copilot comment) and does not break per-instance close idempotency.

\mvn test -Dtest=CurvineFileSystemLifecycleTest\: 5 tests, all PASS.

szbr00 (approved):

PR #1639, opened 2026-08-27
feat(mds): add generic KV abstraction and memory backend

yangyang233333:
# Summary

Add a business-agnostic byte-level KV abstraction to `curvine-mds`, together with an in-memory backend, retry handling, fault injection, metrics, and reusable contract tests. This establishes the storage boundary needed by subsequent stateless MDS repositories.

# Issue Describe / Design

This is PR 2 of the stateless MDS implementation plan.

Key design decisions:
- Keep keys and values opaque as byte arrays, without introducing mount, path, or metadata domain types.
- Define transaction boundaries and optimistic conflict behavior through `KvBackend` and `KvTransaction`.
- Keep retry ownership in the shared `run_txn` helper and expose stable retryable error classification.
- Provide a deterministic memory backend and fault injector for tests without external services.

The next PR will implement the FoundationDB backend on top of this abstraction.

# Changes

| `curvine-mds/src/kv/backend.rs` | Add KV backend and transaction traits, convenience operations, CAS, and transaction retry helper | None; new API |
| `curvine-mds/src/kv/error.rs` | Add common KV errors and retry classification | None; new API |
| `curvine-mds/src/kv/memory.rs` | Add in-memory transactional backend and fault injection | None; used for tests and local execution |
| `curvine-mds/src/kv/metrics.rs` | Add operation, latency, error, conflict, retry, and in-flight metrics | Adds MDS KV observability |
| `curvine-mds/src/kv/contract_tests.rs` | Add reusable backend contract tests | Test-only |
| `curvine-mds/src/lib.rs` | Export the KV interfaces and memory backend | Adds public MDS APIs |
| `curvine-mds/Cargo.toml`, `Cargo.lock` | Add required dependencies | None |

# Test verified

| `CARGO_TARGET_DIR=/tmp/curvine-target cargo test -p curvine-mds` | PASS | 8 tests passed |
| Repeated parallel `cargo test -p curvine-mds --quiet` | PASS | Passed 5 consecutive runs after isolating the metrics test label |

# Dependencies

- Related PRs: #1621
- Related issues: None
- Blocks / blocked by: The next PR will add the FoundationDB backend using these interfaces

### DIFF
--- a/curvine-mds/src/kv/backend.rs
+++ b/curvine-mds/src/kv/backend.rs
@@ -1,0 +1,275 @@
+//! Byte-level, business-agnostic KV abstraction.
+//!
+//! This module defines the stable boundary the stateless MDS repositories and
+//! the future FoundationDB backend both build on. It intentionally knows
+//! NOTHING about MDS domain types (`MetaKey`, `MetaValue`, mount tables, paths,
+//! ...). Keys and values are opaque byte strings; ordering is unsigned
+//! lexicographic over the raw bytes, which every intended backend (memory,
+//! FDB) preserves.
+//!
+//! ## Transaction model
+//!
+//! Transactions are the first-class primitive, not batches. A
+//! [`KvTransaction`] buffers writes (`put`/`delete`) and tracks the keys it
+//! reads (`get`/`multi_get`, plus explicit
+//! [`KvTransaction::add_read_conflict`]) in a read-conflict set. `commit`
+//! applies the buffered writes atomically only if no key in the read-conflict
+//! set was changed by a concurrent committed transaction; otherwise it returns
+//! [`KvError::Conflict`]. This optimistic-concurrency contract is what lets
+//! multiple stateless MDS processes write safely without a single owning node,
+//! and it mirrors FoundationDB's serializable-snapshot-isolation semantics so
+//! the FDB backend (PR 3) can implement the same trait with native behavior.
+//!
+//! Use [`run_txn`] to execute a transaction closure with automatic retry on
+//! retryable errors ([`KvError::is_retryable`]); it re-runs the closure from a
+//! fresh snapshot so read-modify-write and compare-and-set become correct
+//! under contention.
+
+use crate::kv::error::KvResult;
+use crate::kv::metrics;
+use async_trait::async_trait;
+use std::future::Future;
+use std::pin::Pin;
+use std::time::Duration;
+
+/// A single transaction against a [`KvBackend`].
+///
+/// Reads observe a consistent snapshot taken when the transaction began and add
+/// the accessed keys to the read-conflict set. Writes are buffered and applied
+/// atomically by [`KvTransaction::commit`]. After `commit` or
+/// [`KvTransaction::rollback`] the transaction must not be used again.
+#[async_trait]
+pub trait KvTransaction: Send {
+    /// Reads a single key, adding it to the read-conflict set.
+    async fn get(&mut self, key: &[u8]) -> KvResult<Option<Vec<u8>>>;
+
+    /// Reads many keys in one call, adding each to the read-conflict set.
+    /// The returned vector has one entry per input key, in order.
+    async fn multi_get(&mut self, keys: &[Vec<u8>]) -> KvResult<Vec<Option<Vec<u8>>>>;
+
+    /// Reads a single key from the transaction's snapshot WITHOUT adding it to
+    /// the read-conflict set (maps to FDB `snapshotGet`).
+    ///
+    /// Same snapshot as [`get`], so not a dirty read; the only difference is a
+    /// concurrent change to this key won't make the transaction conflict at
+    /// commit. Use only when the value read doesn't affect what the transaction
+    /// writes (advisory reads, wide scans); otherwise use [`get`].
+    ///
+    /// [`get`]: KvTransaction::get
+    async fn snapshot_get(&mut self, key: &[u8]) -> KvResult<Option<Vec<u8>>>;
+
+    /// Buffers a write of `value` at `key`.
+    fn put(&mut self, key: &[u8], value: &[u8]);
+
+    /// Buffers a delete of `key`.
+    fn delete(&mut self, key: &[u8]);
+
+    /// Explicitly adds `key` to the read-conflict set without reading it. Use
+    /// when a decision depends on a key's absence or when the value was read
+    /// through another channel.
+    fn add_read_conflict(&mut self, key: &[u8]);
+
+    /// Applies the buffered writes atomically. Returns [`KvError::Conflict`]
+    /// when a key in the read-conflict set was changed concurrently; the
+    /// transaction is then discarded and the caller should retry from a fresh
+    /// transaction (see [`run_txn`]).
+    async fn commit(&mut self) -> KvResult<()>;
+
+    /// Discards all buffered writes and releases the transaction.
+    fn rollback(&mut self);
+}
+
+/// A pluggable byte-level KV store.
+///
+/// Implementations provide transactions via [`KvBackend::begin`]; the
+/// non-transactional convenience methods have default implementations that run
+/// a single auto-committed transaction, so a backend only has to implement
+/// `begin` and `name` to be fully usable.
+#[async_trait]
+pub trait KvBackend: Send + Sync {
+    /// Short, stable backend identifier used as a metrics label
+    /// (e.g. `"memory"`, `"fdb"`). Must be low-cardinality.
+    fn name(&self) -> &'static str;
+
+    /// Starts a new transaction.
+    async fn begin(&self) -> KvResult<Box<dyn KvTransaction>>;
+
+    /// Point read of a single key.
+    async fn get(&self, key: &[u8]) -> KvResult<Option<Vec<u8>>> {
+        let mut txn = self.begin().await?;
+        let value = txn.get(key).await?;
+        txn.commit().await?;
+        Ok(value)
+    }
+
+    /// Point snapshot read of a single key (no read-conflict tracking).
+    ///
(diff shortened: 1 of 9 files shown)


curvine-mds/src/kv/backend.rs
```diff
@@ -0,0 +1,275 @@
...
+            Err(error) => {
+                txn.rollback();
+                Err(error)
+            }
+        };
+
+        match outcome {
+            Ok(value) => return Ok(value),
+            Err(error) => {
+                if error.is_retryable() && attempt < max_retries {
```
szbr00, line 276: Critical: run_txn retries every KvError::is_retryable, including MaybeCommitted. That is unsafe for any non-idempotent closure.

error.rs says MaybeCommitted is only safe to retry when the body is idempotent, then run_txn retries it anyway because "the closure is re-run from a fresh snapshot". A fresh snapshot re-applies the logic: the txn_read_modify_write increment would bump twice if the first commit applied but returned MaybeCommitted. compare_and_set (backend.rs:184) has the same bug: applied write + retry sees current == new != expected and returns Ok(false), so a successful create-if-absent is reported as failed.

Do not retry MaybeCommitted in run_txn. Surface it to the caller (or take an explicit idempotent opt-in). Retry Conflict / Timeout / Unavailable only. Also retry retryable begin() failures (line 239 currently propagates with ?).

FaultInjector returns MaybeCommitted before applying writes, so it cannot reproduce FDB commit_unknown_result. Add an apply-then-unknown fault and a test that CAS / increment do not double-apply or flip a successful CAS to false.

curvine-mds/src/kv/memory.rs
```diff
@@ -0,0 +1,368 @@
...
+                metrics::op::COMMIT,
+                start,
+                &Err(error.clone()),
+            );
+            return Err(error);
+        }
+
+        let mut store = self.store.lock().unwrap();
+
+        // Conflict check: any read key advanced past our snapshot?
```
szbr00: Critical: MemTxn::commit / read_key do not implement the SSI/OCC contract this module documents, so the memory backend is a false oracle for the FDB backend.

1) Deletes drop the key (memory.rs:340) and the conflict loop only inspects keys that still exist (313-327). T1 get(k) -> T2 delete(k)+commit -> T1 commit succeeds. A concurrent delete of a read key must Conflict; keep a tombstone (Entry { value: Option<Vec<u8>>, version }) and never remove the version.

2) The write set is not validated. Two begin+put(k)+commit transactions both succeed (last writer wins). FDB SSI treats overlapping write ranges as conflicts. Check write-set keys the same way as the read set.

3) read_key (209-218) reads the live map and ignores read_version. Trait docs say reads observe the begin snapshot; snapshot_get of a key written after begin returns the new value and then commits. Capture an Arc snapshot of the map at begin for reads; use the live store only at commit.

Contract tests cover only "get then concurrent put". Add cases for delete-of-read-key, blind write-write, and snapshot_get of a key changed after begin. Without those, PR 3 can pass the same suite and still diverge from memory.

curvine-mds/src/kv/backend.rs
```diff
@@ -0,0 +1,275 @@
...
+pub trait KvBackend: Send + Sync {
+    /// Short, stable backend identifier used as a metrics label
+    /// (e.g. `"memory"`, `"fdb"`). Must be low-cardinality.
+    fn name(&self) -> &'static str;
+
+    /// Starts a new transaction.
+    async fn begin(&self) -> KvResult<Box<dyn KvTransaction>>;
+
+    /// Point read of a single key.
+    async fn get(&self, key: &[u8]) -> KvResult<Option<Vec<u8>>> {
```
szbr00, line 102: Required: KvBackend::get / multi_get start a transaction, add the keys to the read-conflict set, then commit. Under concurrent writers this returns Conflict for a point read. put / delete / batch_delete do not go through run_txn, so after write-write detection is fixed they will also surface Conflict to every convenience caller.

FDB's Database get/set retry internally; a raw get+commit can not_committed. Future MDS repositories will call backend.get() and treat Conflict as a hard failure.

Wrap the convenience methods in run_txn (same as compare_and_set), or implement standalone get/multi_get via snapshot_get + rollback so a point read cannot fail with Conflict and does not pay a commit RTT on FDB.

curvine-mds/src/kv/metrics.rs
```diff
@@ -0,0 +1,153 @@
...
+                "KV transaction retries performed by run_txn",
+                &["backend"],
+            )?,
+            txn_in_flight: Metrics::new_gauge(
+                "mds_kv_txn_in_flight",
+                "Currently in-flight KV transactions",
+            )?,
+        })
+    }
+
```
szbr00, line 95: Optional: Instrumentation is split. Trait defaults observe put / delete / batch_delete / CAS; MemoryBackend observes begin / get / snapshot_get / multi_get / commit. KvBackend::get records nothing at the convenience layer. The FDB backend will either duplicate MemoryBackend's observe calls or silently lose read/begin/commit metrics.

Pull observe into a single decorator (or into the trait defaults and run_txn only) so backends implement begin/commit and do not own Prometheus labels.

szbr00 (changes requested):
Request changes after reviewing curvine-mds kv traits, the memory OCC backend, run_txn/CAS, contract tests, and the lockfile.

The abstraction is the right shape for a later FDB backend, but two contract bugs would make that backend diverge from what this suite certifies:

- MemoryBackend does not implement the documented SSI/OCC rules (no delete tombstones, no write-write conflicts, reads are live rather than snapshotted).
- run_txn retries MaybeCommitted, which double-applies non-idempotent closures and can flip a successful CAS to false.

Convenience get/multi_get can also return Conflict under concurrent writers; wrap those paths or use snapshot_get + rollback. Metrics placement is optional.

Inspected: curvine-mds (kv/), Cargo.toml, Cargo.lock. Author verification (cargo test -p curvine-mds) is present; the missing cases are the ones above.

curvine-mds/src/kv/backend.rs
yangyang233333: @szbr00 Fixed in 4c60674f. `is_retryable` no longer includes `MaybeCommitted`, so `run_txn` surfaces it to the caller instead of re-running a non-idempotent body. `begin()` failures are now folded into the retry budget too. Added a memory `apply_then_unknown_next` fault that applies the write then returns `MaybeCommitted`, with tests asserting an increment does not double-apply and a create-if-absent CAS is not flipped to `false`.

curvine-mds/src/kv/memory.rs
yangyang233333: @szbr00 Fixed the two real divergences in 4c60674f: (1) deletes now keep a version-bumped tombstone (never remove the key), so a concurrent delete of a read key conflicts; (2) reads observe an `Arc` snapshot captured at `begin`, not the live store. Added contract tests for delete-of-read-key and begin-snapshot visibility.

On write-set validation: I kept last-writer-wins for blind writes. FDB only conflicts when a write range intersects another txn's *read* range; two pure blind writes (no reads) never conflict. Validating the write set would make the memory backend stricter than FDB — a false oracle in the other direction. Added `blind_write_write_does_not_conflict` to pin this to FDB semantics. Happy to discuss if you're seeing a case where FDB would actually conflict here.

curvine-mds/src/kv/backend.rs
yangyang233333: @szbr00 Fixed in 4c60674f. `get`/`multi_get` now run through `run_txn`, so a transient conflict is retried rather than surfaced (a point read only returns `Conflict` if it persists past `DEFAULT_MAX_RETRIES`). `snapshot_get` reads the begin snapshot without conflict tracking and rolls back — conflict-free by design. Writes go through `run_txn` as well.

curvine-mds/src/kv/metrics.rs
yangyang233333: @szbr00 Agreed the instrumentation is split. Kept this PR focused on the correctness fixes and filed #1641 to centralize the observe calls into a single decorator (backends implement only begin/commit, no duplicated Prometheus wiring). Will address it in a follow-up PR.

yangyang233333:
For visibility: the automated review points are all addressed in 4c60674f.

- `read_key` reads a begin-time snapshot, not the live store (snapshot isolation).
- Deletes keep a version-bumped tombstone, so a concurrent delete of a read key conflicts.
- Convenience reads/writes go through `run_txn`, which rolls back on the error path; `snapshot_get` rolls back explicitly.
- `run_txn` now retries retryable `begin()` failures.
- Removed the redundant `add_read_conflict` in `compare_and_set` (the prior `get` already tracks the key).
- The assertion-less terminal-error test now proves exactly one error is consumed and the second remains queued.

New contract tests cover delete-of-read-key, blind write-write, begin-snapshot visibility, and apply-then-unknown (no double-apply, no CAS flip). `cargo fmt`, `cargo test -p curvine-mds`, and `cargo clippy --all-targets` are green.

yangyang233333:
@Copilot thanks — on the two suppressed comments:

- **backend.rs backoff cap**: fixed in 32316140. Bumped the shift limit to 7 (`1 << attempt.min(7)`) so the `.min(100)` cap actually applies; previously it maxed at 64ms and the "~100ms" doc comment was wrong.
- **metrics.rs `txn_conflicts_total` overcounting**: good catch — now that `get`/`put`/`delete`/`multi_get` go through `run_txn`, they can also surface `Conflict` after exhausting retries, so counting it on every op overcounts commit conflicts. Tracked in #1641 to gate the counter on `op == COMMIT`, landing with the instrumentation-centralization refactor rather than expanding this PR.

szbr00 (approved):

Issue #1634, opened 2026-08-26
[BUG]: Block replication remote write fails on zero-length staging block


LJF-050:
**Is your feature request related to a problem?**

When the master schedules block replication to a target worker, the target worker's remote-block write handler opens a staging block with a declared block size but zero physical bytes on disk. The replication pipeline then fails immediately with offset exceeds block length, length=0, offset=0 from BlockReadContext::read_region, so replica count never increases.

Replication jobs error on the target worker during Write remote-block; GetBlockLocations continues to show only two replicas and integration tests time out or report that replication did not increase replica count.

Reproduction:

1. Build Curvine at upstream/main d7b6e382 and start a multi-worker mini-cluster with block replication enabled.
2. Create a file spanning multiple 64KB blocks with replication factor 3 across three workers.
3. Simulate under-replication for one block (remove one replica location) and wait for master replication.
4. Observe target worker logs: Write remote-block followed by offset exceeds block length, length=0, offset=0; replica count does not increase.

**Impact**

Under-replicated blocks cannot be repaired automatically; data durability goals for replication=3 clusters are not met when a worker needs a new replica.

**Describe the solution you'd like**

Target workers must accept remote-block replication writes for the full source block size, persist the replica, and report the additional location to the master so replica count increases to the policy target.

Ensure remote-block replication initializes staging block physical length (or uses a write-only path without read_region on an empty file) before streaming source bytes. Review interaction between master replica placement (#1631) and worker remote-block write initialization.

**Additional context**

test_replication_honors_source_block_capacity and test_block_replication_e2e passed at 29b066a1 and regressed before d7b6e382. Candidate related changes include master local weighted replica policy (#1631) and allocatable capacity reporting (#1610).

Root-cause hypothesis: Block replication to a newly selected target worker fails because the remote-block write path opens a staging block with block_size set but physical length zero, so the first read_region call errors with offset exceeds block length. Master still reports two replicas and retries never complete, breaking replication honor and e2e tests.

Evidence summary: replication_test_test_replication_honors_source_block_capacity.log and replication_test_test_block_replication_e2e.log: worker_replication_manager.rs:78 ERROR offset exceeds block length, length=0, offset=0; master_replication_manager.rs:261 replication error to target worker. Prior run 20260813T181439Z at 29b066a1 passed both replication_test cases.

Related issues: None identified.

Acceptance / test gaps:

- test_replication_honors_source_block_capacity passes.
- test_block_replication_e2e passes.
[...]

PR #1642, opened 2026-08-28
fix(worker): cancel remote writer on replication failure (#1636)

### LINKED ISSUE #1636: [BUG]: Block replication failures do not cancel remote block writers
**Describe the bug**

The Worker block replication path does not explicitly cancel a remote block
writer when replication fails after `BlockWriterRemote::new()` succeeds.

`WorkerReplicationManager::replicate_block` creates the remote writer and then
uses early-returning `?` operations for:

- reading the next source block region;
- writing a region to the target Worker;
- flushing the remote writer;
- completing the remote writer.

If source reading, remote writing, or flushing fails, the function returns
without calling `BlockWriterRemote::cancel()`.

`BlockWriterRemote::cancel()` sends a cancel commit to the target Worker.
The target `WriteHandler` handles that request through
`complete(msg, false)`, which calls `abort_block`. Without an explicit cancel,
the target may retain an unfinished staging block, writing metadata, or reserved
storage until some separate cleanup mechanism removes it.

This is not the direct cause of #1634, but it was exposed while investigating
that issue.

**To Reproduce**

1. Start a multi-worker cluster with block replication enabled.
2. Trigger replication of a finalized block to another Worker.
3. Allow `BlockWriterRemote::new()` to open the target staging block.
4. Inject one of the following failures:
   - source `read_region` failure;
   - remote `writer.write()` failure;
   - remote `writer.flush()` failure.
5. Observe that `replicate_block` returns immediately without invoking
   `writer.cancel()`.
6. Inspect the target Worker for unfinished staging state or writing metadata.

**Expected behavior**

After a remote writer has been created, every replication failure before a
successful `complete()` should make a best-effort call to `writer.cancel()`.

The original replication error must remain the primary error. A cancellation
failure should be logged but must not replace or hide the original failure
reported to the Master.

The successful replication path must continue to call `complete()` and must
not call `cancel()`.

- OS: Platform-independent
- Affected branch: `main`
- Analyzed revision: `69d975caf670`
- Affected module:
  `curvine-worker/src/worker/replication/worker_replication_manager.rs`

**Additional context**

`BlockWriterRemote` already exposes the required cancellation operation:

```rust
pub async fn cancel(&mut self) -> FsResult<()>
```

The target Worker maps a cancel request to:

```rust
WriteHandler::complete(msg, false)
```

which aborts the block through `BlockStore::abort_block`.

The missing behavior is therefore in the replication caller's error handling,
not in the remote writer protocol itself.

Related:

- #1634 — discovered while fixing the replication source-reader length bug.
- #1233 — implemented general propagation of writer cancellation, but does not
  cover the replication-specific caller.

Suggested acceptance tests:

1. A source read failure after remote writer creation aborts the target staging
   block.
2. A remote write failure triggers a best-effort `cancel()`.
3. A remote flush failure triggers a best-effort `cancel()`.
4. A cancellation failure is logged while the original replication error is
   still reported to the Master.
5. Successful replication calls `complete()` and does not call `cancel()`.
6. Failed replication leaves no target writing metadata, staging block, or
   reserved-capacity leak.

gangump82:
# Summary

Cancel the target remote block writer when block replication fails after the writer has been created.

Cancellation is best-effort: the original replication error remains the primary error, while cancellation failures are logged separately.

# Issue Describe / Design

Closes #1636

Root cause:

`WorkerReplicationManager::replicate_block` used early-returning `?` operations after creating `BlockWriterRemote`. If source reading, remote writing, flushing, or completion failed, the function returned without sending a cancellation request to the target Worker.

This could leave an unfinished staging block, writing metadata, or reserved storage capacity on the target Worker.

Fix approach:

- Collect the read, write, flush, and complete operations into one replication result.
- If the replication result fails, call `BlockWriterRemote::cancel()` once.
- Preserve and return the original replication error.
- Log cancellation failures without replacing the original error.
- Do not cancel after successful completion.

# Changes

| `curvine-worker/src/worker/replication/worker_replication_manager.rs` | Add best-effort remote writer cancellation after replication failures | Failed replication now cleans up target staging state when possible |
| `curvine-worker/src/worker/replication/worker_replication_manager.rs` | Add tests for successful completion, cancellation, and cancellation failure | No production behavior change outside the replication failure path |

# Test verified

| `cargo test -p curvine-worker worker::replication::worker_replication_manager::tests -- --nocapture` | PASS | 3 passed after rebase |
| `cargo test -p curvine-worker` | PASS | 34 passed, 1 ignored |
| `cargo test -p curvine-storage-local abort_spdk` | PASS | Target abort path |
| `cargo test -p curvine-storage-local abort_and_failed_prepare_write_preserve_capacity` | PASS | Capacity cleanup |
| `cargo test -p curvine-storage-local failed_file_deallocate_keeps_block_and_capacity_reserved` | PASS | Failed deallocation behavior |
| `cargo test -p curvine-tests --test replication_test test_replication_honors_source_block_capacity -- --nocapture` | PASS | Replication capacity regression |
| `cargo test -p curvine-tests --test replication_test test_block_replication_e2e -- --nocapture` | PASS | Replica count and data integrity |
| Three-node post-open fault injection on `tikv-nxxy03`, `tikv-nxxy04-1`, and `tikv-nxxy05` | PASS | A temporary test-only fault point failed replication after the target writer opened. The target logged `is commit: false`, retained no staging/active replica, and reported `reserved_bytes = 0`; the Master did not add the failed target location. |
| Fault-free recovery read after clearing the injected rule | PASS | Restored the second replica, downloaded 2 MiB, and matched SHA-256 `399902e71e4ac567b3d6efc9f46eb903bfa93a9faf3f4c7e7a2b227f8250c337` |
| Rebase and final diff verification | PASS | `github/main` is an ancestor; one #1636 commit and one changed file remain; the #1634 patch is already present in `main` |
| `git diff --check` | PASS | Rechecked after rebase |
| `cargo clippy -p curvine-worker --lib --tests -- --deny warnings --allow unused_macros --allow unused_imports --allow unused_variables --allow clippy::uninlined-format-args` | PASS | Platform-specific existing warnings allowed |
| `make format` | BLOCKED | Existing macOS issues: unavailable `fuse_mount_pure` and unused `sys_call` macro under `-D warnings` |

# Dependencies

- Related PRs: #1637 (merged)
- Related issues: #1634, #1233, #1645
- Blocks / blocked by: None
- Stronger repeatable caller-level or worker fault-injection regression coverage is tracked in #1645.

### DIFF
--- a/curvine-worker/src/worker/replication/worker_replication_manager.rs
+++ b/curvine-worker/src/worker/replication/worker_replication_manager.rs
@@ -18,15 +18,41 @@ use curvine_client_core::block::BlockWriterRemote;
 use curvine_client_core::file::FsContext;
 use curvine_config::ClusterConf;
 use curvine_core_error::{err_box, CommonResult};
+use curvine_error::FsResult;
 use curvine_fs_api::RpcCode;
 use curvine_model::{ExtendedBlock, FileType};
 use curvine_proto::{ReportBlockReplicationRequest, ReportBlockReplicationResponse};
 use curvine_runtime::runtime::{AsyncRuntime, RpcRuntime};
-use log::{error, info};
+use log::{error, info, warn};
 use once_cell::sync::OnceCell;
+use std::future::Future;
 use std::sync::Arc;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio::sync::Semaphore;
+
+async fn finish_replication_with_cleanup<C, F>(
+    result: CommonResult<()>,
+    cancel: C,
+    block_id: i64,
+    target_worker_id: u32,
+) -> CommonResult<()>
+where
+    C: FnOnce() -> F,
+    F: Future<Output = FsResult<()>>,
+{
+    match result {
+        Ok(()) => Ok(()),
+        Err(replication_error) => {
+            if let Err(cancel_error) = cancel().await {
+                warn!(
+                    "Failed to cancel remote writer for block {} on worker {} after replication error '{}': {}",
+                    block_id, target_worker_id, replication_error, cancel_error
+                );
+            }
+            Err(replication_error)
+        }
+    }
+}
 
 #[derive(Clone)]
 pub struct WorkerReplicationManager {
@@ -147,17 +173,28 @@ impl WorkerReplicationManager {
             target_capacity,
         )
         .await?;
-        let mut remaining = block_meta.len;
-        while remaining > 0 {
-            let size = remaining.min(self.replicate_chunk_size as i64);
-            let slice = reader.read_region(true, size as i32)?;
-            let read_len = slice.len() as i64;
-            writer.write(slice).await?;
-            remaining -= read_len;
+        let replication_result: CommonResult<()> = async {
+            let mut remaining = block_meta.len;
+            while remaining > 0 {
+                let size = remaining.min(self.replicate_chunk_size as i64);
+                let slice = reader.read_region(true, size as i32)?;
+                let read_len = slice.len() as i64;
+                writer.write(slice).await?;
+                remaining -= read_len;
+            }
+            writer.flush().await?;
+            writer.complete().await?;
+            Ok(())
         }
-        writer.flush().await?;
-        writer.complete().await?;
-        Ok(())
+        .await;
+
+        finish_replication_with_cleanup(
+            replication_result,
+            || writer.cancel(),
+            job.block_id,
+            job.target_worker_addr.worker_id,
+        )
+        .await
     }
 
     pub fn accept_job(&self, job: ReplicationJob) -> CommonResult<()> {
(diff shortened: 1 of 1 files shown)


curvine-worker/src/worker/replication/worker_replication_manager.rs
```diff
@@ -169,3 +208,77 @@ impl WorkerReplicationManager {
         let _ = self.master_client.set(master_client);
     }
 }
+
+#[cfg(test)]
+mod tests {
```
szbr00, line 213: Optional: These tests pin the helper policy but would still pass if
`replicate_block` stopped calling `finish_replication_with_cleanup` — the
actual #1636 defect. A TrackingWriter-style test on `replicate_block` (see
`FuseWriter::cancel_failure_does_not_mask_worker_error`), or an integration
assertion that a post-`new()` read/write/flush failure leaves no target
writing metadata or reserved capacity, would fail on the caller bug. The
helper coverage can stay.

szbr00 (approved):
Inspected curvine-worker replication, BlockWriterRemote::cancel, WriteHandler::complete(commit=false), abort_block idempotency, and BlockClient close-on-network-error.

Failure-path cancel is wired only after new(), success skips cancel, and the original replication error is preserved. Cancel after a failed complete() is safe: is_commit is a no-op, and a network error closes the same client so cancel cannot land on a new handler and abort a finalized replica.

Approve. The Optional comment on the helper tests is not a merge blocker.

curvine-worker/src/worker/replication/worker_replication_manager.rs
gangump82: Thanks, agreed that the current helper tests do not protect the
`replicate_block` caller wiring itself. The production cleanup path is correct,
but a meaningful caller-level regression test requires either a small writer
injection seam or worker fault-injection coverage.

I verified the production path with a three-node fault-injection test. After the target remote
writer opened, an injected replication failure caused the source to send
cancel. The target logged `is commit: false`, retained no staging/active block
for the failed replica, and reported `reserved_bytes = 0`; the Master also did
not add the failed target as a block location. After clearing the injected
rule and restoring the second replica, a 2 MiB recovery read matched the
original SHA-256 (`399902e71e4ac567b3d6efc9f46eb903bfa93a9faf3f4c7e7a2b227f8250c337`).

I’ve tracked that stronger coverage in #1645. I’ll keep this PR focused on the
#1636 production fix. It has now been rebased onto `main` after #1637 merged,
and the final PR diff contains only the #1636 change.

PR #1644, opened 2026-08-31
feat(worker): add disk free-space ratio protection (free_ratio)

### LINKED ISSUE #1643: [FEATURE] disk free-space ratio protection (free_ratio) for worker data dirs
## Summary

Add a per-disk **free-space ratio floor** (`free_ratio`) so a worker stops writing to a data directory once its free-space ratio drops below a threshold — preventing disk exhaustion on shared / differently-sized NVMe disks where a fixed-byte `dir_reserved` cannot adapt, and giving an early-warning signal before the disk is full. Inspired by juicefs `--free-space-ratio`.

## Problem

curvine workers currently only support a fixed-byte per-dir reservation (`dir_reserved`). On a disk shared with other consumers (kubelet, logs, other stores) or on mixed disk sizes, a fixed byte limit:

- cannot adapt as other processes consume space, so curvine keeps writing until the OS `ENOSPC`s — which can take down co-located critical processes;
- provides no "disk is getting full" signal before the disk is actually full;
- doesn't translate across disks of different sizes.

## Proposed solution

A new `WorkerConf.free_ratio` (`f64`, default `0.0` = disabled, backward compatible). When a data dir's free ratio drops below the threshold, the dir is treated as non-writable: new writes + block allocation are rejected (no eviction — that's an orthogonal metadata-entry quota).

- **FS mode** (`[SSD]/data/data1`): ratio = `statvfs.available_space / total_space`.
- **SPDK raw-device mode**: ratio = `(capacity - used) / capacity` (no statvfs available).
- **Single lever**: `VfsDir::available()` returns 0 when `raw_free_ratio() < free_ratio`. All write-admission paths flow through it (worker `can_allocate`, rewrite precheck, heartbeat → master `allocatable_available` → chooser), so **worker-side rejection and master-side stop-allocation both apply with zero master code change**. The worker is the last line of defense in the heartbeat gap.
- **Metrics**: `disk_free_ratio` (per-disk GaugeVec, labeled `dir_id`/`dir_path`/`storage_type`, in basis points) and `disk_full_rejected_writes` (Counter, instrumented at the write-handler layer via stable error-message matching so `curvine-storage-local` stays decoupled from `curvine-metrics`).

## Cluster verification (internal 3-node dev cluster)

Built `curvine-server` (linux x86_64, release) from this branch and ran the worker on one node only (master/journal/fuse untouched):

| State | `available` | `disk_free_ratio` | `disk_full_rejected_writes` |
|---|---|---|---|
| `free_ratio=0.0` (disabled) | ~3.8 TB | 9905 (99.05%) | 0 |
| `free_ratio=0.995` (gate, above actual 99.05%) | **0** | 9905 (raw, pre-guard) | 0 |
| `free_ratio=0.0` (recovered) | ~3.8 TB | 9905 | 0 |

- Gate fires correctly: with `free_ratio=0.995` and actual free ratio 99.05%, `available` drops to 0; `capacity` unchanged; the gauge keeps reporting the real pre-guard ratio (the "why").
- Master-side stop-allocation confirmed: a FUSE write was routed to a *healthy* worker (the gated worker excluded via heartbeat), so the write succeeded cluster-wide and `disk_full_rejected_writes` stayed 0 — i.e. the master correctly stopped allocating to the zero-available worker. (The counter path itself is covered by unit tests; in-cluster it only ticks in the heartbeat-gap race, which the working feature prevents.)
- Fully reversible: `free_ratio=0.0` restores `available` immediately.
- Production worker restored to the original binary after the test.

## Backward compatibility

Default `0.0` disables the guard, so `available()` behaves identically to today; old configs without `free_ratio` get `0.0` via existing `#[serde(default)]`.

## Linked PR

Implementation: #1644

gangump82:
## What & why

Adds a per-disk **free-space ratio floor** (`free_ratio`) so a worker stops writing to a storage directory once its free-space ratio drops below a threshold. Closes #1643.

curvine currently only supports a fixed-byte per-dir reservation (`dir_reserved`), which cannot adapt to disks shared with other consumers or to mixed disk sizes, and provides no early-warning signal before the disk fills. `free_ratio` (default `0.0`, disabled, backward compatible) gates `VfsDir::available()` to 0 when `raw_free_ratio() < free_ratio`, covering both FS mode (statvfs available/total) and SPDK raw-device mode ((capacity − used)/capacity).

## Design

- **One lever**: `VfsDir::available()` returns 0 below the threshold. All write-admission flows pass through it — worker `can_allocate`, rewrite precheck, and heartbeat → master `allocatable_available` → chooser — so worker-side rejection **and** master-side stop-allocation apply with **no master code change**. Eviction is untouched (orthogonal metadata-entry quota).
- **Config**: `WorkerConf.free_ratio: f64` (default `0.0`; range `[0.0, 1.0)`; out-of-range clamped with a warning).
- **Metrics**: `disk_free_ratio` per-disk GaugeVec (labeled `dir_id`/`dir_path`/`storage_type`; basis points, since the crate's gauges are i64-only) + `disk_full_rejected_writes` Counter, instrumented at the write-handler layer via stable error-message matching so `curvine-storage-local` stays decoupled from `curvine-metrics`.
- Example config entry + unit tests (FS/SPDK gate, `raw_free_ratio`, disabled-by-default, config parse/clamp).

## Cluster verification

Tested on an internal 3-node dev cluster with a release build of this branch running on one worker node only (master/journal/fuse untouched). See #1643 for the full result matrix. Summary: with `free_ratio=0.995` above the disk's actual 99.05% free, `available` correctly dropped to 0 (gauge kept reporting the real pre-guard ratio), a FUSE write was routed to a healthy worker (master excluded the gated one), and `free_ratio=0.0` restored `available` immediately. The production worker was restored to the original binary afterward.

## Backward compatibility

Default `0.0` disables the guard → `available()` behaves identically to today; configs without `free_ratio` get `0.0` via existing `#[serde(default)]`.


crates/adapters/curvine-storage-local/src/vfs_dir.rs
```diff
@@ -39,6 +39,9 @@ pub struct VfsDir {
     pub(crate) storage_type: StorageType,
     pub(crate) conf_capacity: i64,
     pub(crate) reserved_bytes: i64,
+    /// Free-space ratio floor for write admission. `0.0` disables. See
+    /// `available()` guard and `raw_free_ratio()`.
+    pub(crate) free_ratio: f64,
```
szbr00, line 44: Required: Adding `free_ratio` breaks the same-file SPDK test helpers
`spdk_dir` / `local_dir` (vfs_dir.rs ~424 and ~446). Those literals omit
the new field, so `cargo test --features spdk -p curvine-storage-local`
fails to compile. The vfs_dataset.rs helpers were updated; these were
not. Add `free_ratio: 0.0` to both literals (same default as
`VfsDir::from_dir`).

curvine-worker/src/worker/worker_metrics.rs
```diff
@@ -95,6 +100,16 @@ impl WorkerMetrics {
...
             )?,
 
+            disk_free_ratio: m::new_gauge_vec(
+                "disk_free_ratio",
+                "Per-disk free-space ratio in basis points (10000 = 100%, 1000 = 10%); pre-reserved, pre-guard",
+                &["dir_id", "dir_path", "storage_type"],
+            )?,
+            disk_full_rejected_writes: m::new_counter(
+                "disk_full_rejected_writes",
+                "Writes rejected because a data dir dropped below free_ratio",
```
szbr00: Required: `disk_full_rejected_writes` help text says rejections are
caused by `free_ratio`, but `WriteHandler::open` increments on any
storage-layer "Not enough space" / "storage capacity" admission error
(dir_reserved, capacity, and free_ratio). Operators will misread the
counter. Either narrow the increment to the free_ratio path, or change
the help/doc to "writes rejected because the storage layer reported
insufficient capacity".

crates/adapters/curvine-storage-local/src/dataset.rs
```diff
@@ -41,6 +53,9 @@ pub trait Dataset {
         ByteUnit::byte_to_string(self.available() as u64)
     }
 
+    /// Per-directory raw free-space ratios for the `disk_free_ratio` metric.
+    fn dir_free_ratios(&self) -> Vec<DirFreeRatio>;
```
szbr00: Optional: `dir_free_ratios` is a new required `Dataset` method whose
only production caller is `WorkerMetrics::text_output`. The only
implementor is `VfsDataset`. Put this on `VfsDataset` / `DirList` (or
give the trait method a default `vec![]`) instead of widening the
public trait for one metrics scrape.

crates/adapters/curvine-storage-local/src/vfs_dir.rs
```diff
@@ -182,6 +187,14 @@ impl VfsDir {
     }
 
     pub fn available(&self) -> i64 {
+        // free_ratio guard: when the directory's raw free ratio drops below the
+        // configured floor, treat it as unavailable so writes are rejected and
+        // the master stops allocating to this worker (via heartbeat available).
+        // `0.0` disables the guard. Applies to both FS and SPDK paths.
+        if self.free_ratio > 0.0 && self.raw_free_ratio() < self.free_ratio {
```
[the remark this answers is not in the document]
gangump82: Thanks. Fixed in ab7f5613: the filesystem path now samples total and available space once per available() call and reuses those values for both the free-ratio guard and capacity calculation.

crates/adapters/curvine-storage-local/src/vfs_dataset.rs
```diff
@@ -125,6 +125,21 @@ impl VfsDataset {
...
+                warn!(
+                    "worker.free_ratio={} is negative; clamping to 0.0 (disabled)",
+                    raw
+                );
+                0.0
+            } else if raw >= 1.0 {
+                warn!("worker.free_ratio={} >= 1.0; clamping to 0.99", raw);
+                0.99
+            } else {
+                raw
```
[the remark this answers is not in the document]
gangump82: Thanks. Fixed in ab7f5613: the normalization was extracted into a helper with unit coverage for valid, negative, >=1.0, NaN, and +/-infinity inputs.

curvine-worker/src/worker/worker_metrics.rs
```diff
@@ -45,6 +45,11 @@ pub struct WorkerMetrics {
     pub(crate) store_total_disks: Gauge,
     pub(crate) num_blocks_to_delete: Gauge,
 
+    /// Per-directory free-space ratio (0.1 = 10%). Pre-reserved, pre-guard.
+    pub(crate) disk_free_ratio: GaugeVec,
+    /// Writes rejected because a data dir dropped below `free_ratio`.
+    pub(crate) disk_full_rejected_writes: Counter,
```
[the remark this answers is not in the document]
gangump82: Thanks. Fixed in ab7f5613: both the Rust doc comment and Prometheus help now describe generic storage admission rejections caused by insufficient capacity.

crates/adapters/curvine-storage-local/src/vfs_dataset.rs
```diff
@@ -125,6 +125,21 @@ impl VfsDataset {
...
+            }
+        };
```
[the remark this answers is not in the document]
gangump82: Thanks. Fixed in ab7f5613: non-finite values now warn and disable the guard, while values >= 1.0 are clamped to the largest representable f64 below 1.0.

crates/adapters/curvine-storage-local/src/dataset.rs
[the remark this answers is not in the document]
gangump82: Thanks. Fixed in ab7f5613 by removing dir_free_ratios() from the public Dataset trait and keeping it as a VfsDataset inherent method, avoiding both a breaking required method and a silent empty default.

gangump82: Thanks. Fixed in ab7f5613 by moving dir_free_ratios() off the public Dataset trait and onto VfsDataset, which is the sole production type used by WorkerMetrics.

curvine-worker/src/worker/worker_metrics.rs
gangump82: Thanks. Fixed in ab7f5613: the counter help and doc text now state that the storage layer reported insufficient capacity, covering free_ratio, reserved space, and general capacity exhaustion.

crates/adapters/curvine-storage-local/src/vfs_dir.rs
gangump82: Thanks. Fixed in ab7f5613 by adding free_ratio: 0.0 to both same-file SPDK test helpers. The SPDK test targets now pass cargo check with --features spdk --tests on all three Linux test hosts.

crates/adapters/curvine-storage-local/src/vfs_dir.rs
```diff
@@ -162,44 +167,88 @@ impl VfsDir {
...
+        // once so the filesystem path performs only one total-space and one
+        // available-space query per call.
         if self.storage_type == StorageType::SpdkDisk {
             let capacity = self.capacity();
             let fs_used = self.fs_used();
-            let reserved_bytes = self.reserved_bytes;
-            return 0.max(capacity - fs_used - reserved_bytes);
+            let free = capacity.saturating_sub(fs_used).max(0);
+            if self.free_ratio > 0.0 && Self::calculate_free_ratio(free, capacity) < self.free_ratio
+            {
```
szbr00, line 199: Required: Gating `available()` to 0 also fails in-place rewrite of
already-placed FileLayout blocks. `BlockStore::open_block` ->
`reserve_file_open` / `Dataset::open_block` both reject when
`required_bytes > dir.available()`. After the master stops new
allocation, a client overwrite of a file whose block lives on the
gated dir still has to come back to this worker; it now gets a space
error even when the disk still has gigabytes free (e.g. 9% free with
`free_ratio=0.1`) and the rewrite is not attracting new data.
FileLayout rewrite always charges a full staging copy
(`preserves_committed_on_write`, then `fs::copy`), so `available()==0`
is a hard write-freeze, not just "stop allocating". Cluster
verification only covered a new FUSE write routed to another worker.

Split the lever: keep the guard on `can_allocate` and on the
heartbeat `StorageInfo.available` used by the master chooser, and let
rewrite prechecks use ungated remaining bytes (so an existing block
can still be updated while the 10% headroom is reserved for other
apps and for the staging copy). Add a test: finalize a FileLayout
block, set `free_ratio` above `raw_free_ratio`, rewrite the same
size, and assert the open succeeds. If write-freeze of existing
files is the intended product behavior, document that on
`WorkerConf.free_ratio` and assert the rejection instead — do not
leave it implicit.

gangump82: Thanks. Addressed in 374ff4fe by choosing and documenting the write-freeze behavior. `worker.free_ratio` is now a strict physical free-space admission floor for both new writes and FileLayout rewrite staging copies.

A rejected rewrite leaves the committed block intact and readable, but the existing file can remain temporarily non-writable until sufficient headroom is restored. FileLayout keeps the committed and staging copies concurrently, so a same-size rewrite needs another full block copy, while an expanding rewrite reserves the full target size (for example, rewriting 10 GB to 20 GB may require about 30 GB of peak physical space).

We intentionally did not use the proposed bypass behavior because allowing rewrite staging below the floor could consume space reserved for other applications sharing the device, especially during expanding rewrites.

Storage-capacity rejection is now returned as `FsError::DiskOutOfSpace`, which maps to FUSE `ENOSPC`. The rejection test verifies that committed data remains readable and that no staging file or reservation state leaks.

Automatic relocation before rewrite is tracked in #1647.

Verified on Linux:
- `cargo fmt --check`
- `cargo test -p curvine-config free_ratio -- --nocapture` (3 passed)
- `cargo test -p curvine-storage-local --lib` (62 passed)
- `cargo test -p curvine-worker --lib` (36 passed, 1 ignored)
- `cargo test -p curvine-fuse disk_out_of_space_maps_to_enospc -- --nocapture` (1 passed)

szbr00 (approved):

PR #1650, opened 2026-09-02
test(config): isolate environment-dependent failure paths

uhmh123:
# Summary

Make configuration and block-store tests deterministic in root-based automation environments. The change isolates hostname environment variables and replaces permission-based deallocation failure simulation with an injected failure gate.

# Issue Describe / Design

The daily and integration profiles failed for two independent test-environment reasons:

- The runner-provided `CURVINE_*_HOSTNAME` variables overrode hostname values from temporary TOML files.
- The block-store tests used `chmod 0500` to simulate deallocation failures, but the tests run as root and deletion still succeeded.

The hostname tests now save, clear, and automatically restore the relevant environment variables. The block-store tests inject a deterministic failure before real deallocation, while production paths continue to execute the real deallocator unchanged.

# Changes

| `crates/common/curvine-config/src/cluster_conf.rs` | Save, clear, and restore hostname environment variables in the whitespace normalization test | Test-only isolation |
| `curvine-server/src/bin/curvine-server.rs` | Isolate `CURVINE_MASTER_HOSTNAME` while loading the temporary server configuration | Test-only isolation |
| `curvine-worker/src/worker/block/block_store.rs` | Add an internal deallocation gate and replace chmod-based failure simulation with a deterministic stub | Production deletion behavior remains unchanged |

# Test verified

| `cargo test -p curvine-config` | PASS | Includes 107 unit tests and integration tests with `CURVINE_MASTER_HOSTNAME=localhost` |
| `cargo test -p curvine-server get_conf_loads_from_discovered_config` | PASS | Executed with `CURVINE_MASTER_HOSTNAME=localhost` |
| `cargo test -p curvine-worker --lib` | PASS | 36 passed, 1 manual benchmark ignored |
| `cargo fmt --all` | PASS | Rust 1.92.0 |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/common/curvine-config/src/cluster_conf.rs
+++ b/crates/common/curvine-config/src/cluster_conf.rs
@@ -548,11 +548,10 @@ mod tests {
     #[test]
     fn trims_whitespace_in_hostname_config() {
         const HOSTNAME_ENV_KEYS: &[&str] = &[
-            "CURVINE_MASTER_HOSTNAME",
-            "CURVINE_JOURNAL_HOSTNAME",
-            "CURVINE_WORKER_HOSTNAME",
-            "CURVINE_CLIENT_HOSTNAME",
-            "CURVINE_TRANSFER_HOSTNAME",
+            ClusterConf::ENV_MASTER_HOSTNAME,
+            ClusterConf::ENV_WORKER_HOSTNAME,
+            ClusterConf::ENV_CLIENT_HOSTNAME,
+            ClusterConf::ENV_TRANSFER_HOSTNAME,
         ];
         let _hostname_env = EnvVarsGuard::unset(HOSTNAME_ENV_KEYS);
         let path = std::env::temp_dir().join(format!(
--- a/crates/common/curvine-config/src/cluster_conf.rs
+++ b/crates/common/curvine-config/src/cluster_conf.rs
@@ -447,6 +447,33 @@ mod tests {
     use crate::RaftPeer;
     use curvine_runtime::common::Utils;
 
+    struct EnvVarsGuard(Vec<(&'static str, Option<std::ffi::OsString>)>);
+
+    impl EnvVarsGuard {
+        fn unset(keys: &'static [&'static str]) -> Self {
+            let saved = keys
+                .iter()
+                .map(|key| {
+                    let value = std::env::var_os(key);
+                    std::env::remove_var(key);
+                    (*key, value)
+                })
+                .collect();
+            Self(saved)
+        }
+    }
+
+    impl Drop for EnvVarsGuard {
+        fn drop(&mut self) {
+            for (key, value) in self.0.drain(..) {
+                match value {
+                    Some(value) => std::env::set_var(key, value),
+                    None => std::env::remove_var(key),
+                }
+            }
+        }
+    }
+
     #[test]
     fn normalize_whitespace_trims_cluster_id() {
         let mut conf = ClusterConf {
@@ -520,6 +547,14 @@ mod tests {
 
     #[test]
     fn trims_whitespace_in_hostname_config() {
+        const HOSTNAME_ENV_KEYS: &[&str] = &[
+            "CURVINE_MASTER_HOSTNAME",
+            "CURVINE_JOURNAL_HOSTNAME",
+            "CURVINE_WORKER_HOSTNAME",
+            "CURVINE_CLIENT_HOSTNAME",
+            "CURVINE_TRANSFER_HOSTNAME",
+        ];
+        let _hostname_env = EnvVarsGuard::unset(HOSTNAME_ENV_KEYS);
         let path = std::env::temp_dir().join(format!(
             "curvine-trim-conf-{}-{}.toml",
             std::process::id(),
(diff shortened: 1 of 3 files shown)


crates/common/curvine-config/src/cluster_conf.rs
```diff
@@ -520,6 +547,14 @@ mod tests {
 
     #[test]
     fn trims_whitespace_in_hostname_config() {
+        const HOSTNAME_ENV_KEYS: &[&str] = &[
+            "CURVINE_MASTER_HOSTNAME",
+            "CURVINE_JOURNAL_HOSTNAME",
+            "CURVINE_WORKER_HOSTNAME",
+            "CURVINE_CLIENT_HOSTNAME",
+            "CURVINE_TRANSFER_HOSTNAME",
```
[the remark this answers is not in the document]
uhmh123: Thanks. Fixed in 8b7c34f4: replaced the hostname string literals with canonical ClusterConf constants and removed the unsupported journal hostname key.

curvine-server/src/bin/curvine-server.rs
```diff
@@ -166,6 +188,7 @@ mod tests {
 
     #[test]
     fn get_conf_loads_from_discovered_config() {
+        let _master_hostname = EnvVarGuard::unset("CURVINE_MASTER_HOSTNAME");
```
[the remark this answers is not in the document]
uhmh123: Thanks. Fixed in 42ec9e0e: the server test now uses ClusterConf::ENV_MASTER_HOSTNAME.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -315,13 +315,39 @@ impl BlockStore {
     }
 
     pub fn remove_block(&self, id: i64) -> CommonResult<()> {
+        self.remove_block_with(id, || Ok(()))
+    }
+
+    fn remove_block_with(
```
szbr00: Required: `remove_block` / `async_remove_block` now wrap new `*_with(..., deallocation_gate)` helpers, and sync `remove_block` inlines `get_block` + `remove_block_state_by_id` + `deallocate` + `restore_removed_block` + `release`. That is a second copy of `Dataset::remove_block` / `VfsDataset::remove_block_by_id`. With the default `|| Ok(())` the I/O result matches today, but the worker delete path no longer goes through the dataset method that already owns restore-on-deallocate-failure.

Keep production `remove_block` as `state.remove_block(&block)`, and keep `async_remove_block` calling `removed.deallocate()` directly. Do not add a test gate to the production delete path.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -464,10 +472,10 @@ mod tests {
     #[test]
     fn async_remove_deallocate_error_keeps_block_for_retry() -> CommonResult<()> {
         let store = create_store("deallocate-error")?;
-        let _delete_denied = deny_block_delete(&store)?;
+        finalized_block(&store)?;
         store.read()?.increment_blocks_to_delete();
 
-        let result = store.async_remove_block(1);
+        let result = store.async_remove_block_with(1, injected_deallocate_error);
```
szbr00: Required: `async_remove_deallocate_error_keeps_block_for_retry` and `remove_deallocate_error_keeps_block_for_retry` call `*_with` and fail `deallocation_gate()` before `removed.deallocate()`. They would still pass if `deallocate()` errors were not restored, or if the public wrappers stopped forwarding the error. That is the retry invariant these tests were written to protect; heartbeat delete uses `async_remove_block`, not the helper.

`chmod 0500` is a no-op as root, but this crate already has a root-safe way to make `FileLayout::deallocate` fail: replace the block file with a non-empty directory (`VfsDataset::failed_file_deallocate_keeps_block_and_capacity_reserved`). Use that setup and call `async_remove_block` / `remove_block` so the assertion fails on a real deallocate error through the public entry path.

szbr00 (changes requested):
Required: Hostname env isolation in `curvine-config` and `curvine-server` looks correct and is verified with `CURVINE_MASTER_HOSTNAME=localhost`.

The worker change should not land as written. Production `remove_block` / `async_remove_block` now carry a test-only deallocation gate, and the updated tests call the helpers instead of failing real `deallocate()` on the public entry path. Reuse `state.remove_block` / `removed.deallocate()`, and make deallocate fail the same root-safe way `VfsDataset::failed_file_deallocate_keeps_block_and_capacity_reserved` already does.

Inspected: `curvine-config`, `curvine-server` bin tests, `curvine-worker` `BlockStore` plus `VfsDataset` remove/restore callers.

curvine-worker/src/worker/block/block_store.rs
uhmh123: Fixed in 07c6b93f: restored remove_block/async_remove_block to their production implementations and removed the test-only deallocation gates.

curvine-worker/src/worker/block/block_store.rs
uhmh123: Fixed in 07c6b93f: the tests now replace the real block file with a non-empty directory, call the public removal APIs, verify state is retained on deallocation failure, then clean up and verify retry succeeds.

szbr00 (approved):

PR #1651, opened 2026-09-02
feat(mds): add FoundationDB KV backend

yangyang233333:
# Summary

Implement the FoundationDB (FDB) KV backend on top of the generic byte-level KV abstraction added in #1639, and wire it into `curvine-mds` via configuration. This is the persistent storage backend the stateless MDS repositories will build on. The same backend-agnostic contract test suite that pins the memory backend runs unchanged against a real FDB cluster.

# Issue Describe / Design

This is PR 3 of the stateless MDS implementation plan (parent: #951). It builds directly on the KV abstraction from #1639 (PR 2) and introduces no MDS domain types.

Key design decisions:

- **Network lifecycle.** The FDB C client requires a process-global network event loop started exactly once; it is booted lazily via a `OnceCell` and the `NetworkAutoStop` guard is intentionally leaked. Dropping it would call `fdb_stop_network`, which blocks on the network thread and could wedge process shutdown if the cluster is unreachable. Leaking it lets the OS reclaim the thread on exit, satisfying "FDB unavailable or shut down must not block process exit".
- **Fail-fast, not hang.** Every transaction is created with a per-transaction `Timeout` so an unreachable cluster surfaces a bounded error instead of blocking forever.
- **Error classification.** Native `FdbError` is mapped to the abstraction's `KvError` without leaking SDK types. `commit_unknown_result` -> `MaybeCommitted` (NOT retryable — the write may already have applied); `not_committed` (1020) is the only code mapped to `Conflict`, so the conflict metric counts real read/write-set clashes; `transaction_timed_out` (1031) -> `Timeout`; other retryable codes (connection loss, coordinator change, stale read version, …) -> `Unavailable`, kept distinct so availability problems do not pollute conflict metrics.
- **Begin-time snapshot.** `begin()` eagerly resolves the read version so reads observe the snapshot taken when the transaction began, matching the trait contract and the memory backend (FDB otherwise picks the read version lazily on the first read).
- **Metrics.** Transaction latency, conflicts, retries, errors, in-flight gauge, and a KV size distribution histogram.
- **Config.** `kv_backend = "memory" | "fdb"` selects the backend; `fdb_cluster_file` points at a standard FDB cluster file; `fdb_txn_timeout_ms` bounds each transaction.

Out of scope (deferred): range/scan operations are not implemented yet — they will be added in PR 4 when the MountTable repository first needs prefix/range scans. A separate follow-up (tracked in design notes) will add fail-fast / circuit-breaking for prolonged cluster unavailability at the `run_txn` layer.

# Changes

| `curvine-mds/src/kv/fdb.rs` | New FDB backend: network bootstrap, transaction lifecycle, error mapping, point ops, begin-time read version | None; new backend |
| `curvine-mds/src/kv/mod.rs` | Wire up the `fdb` module and re-export `FdbBackend` | None; new API |
| `curvine-mds/src/kv/metrics.rs` | Add KV key/value size histograms and `observe_kv_size` | Adds MDS metrics |
| `curvine-mds/src/kv/contract_tests.rs` | Run the shared contract suite against a real cluster (gated by `CURVINE_MDS_FDB_CLUSTER`); per-run key prefix isolates the keyspace; add an unreachable-cluster fail-fast test | Test-only |
| `curvine-mds/src/lib.rs` | Export `FdbBackend` | None; new API |
| `curvine-mds/src/server.rs` | Select memory/FDB backend from config | None when `kv_backend = memory` |
| `crates/common/curvine-config/src/mds_conf.rs` | Add `kv_backend`, `fdb_cluster_file`, `fdb_txn_timeout_ms` with validation | New config fields; defaults preserve behavior |
| `crates/common/curvine-config/src/lib.rs` | Export `KvBackendType` | None; new API |
| `curvine-mds/Cargo.toml`, `Cargo.toml`, `Cargo.lock` | Add `foundationdb` dependency | New dependency |
| `etc/curvine-cluster.toml`, `curvine-docker/deploy/example/conf/curvine-cluster.toml` | Document the new MDS KV backend options | Docs/config examples |

# Test verified

- `cargo test -p curvine-mds -p curvine-config` — memory backend contract suite, config validation, and server backend selection all pass.
- Against a real FoundationDB 7.3 cluster (`CURVINE_MDS_FDB_CLUSTER=<cluster file>`): the identical contract suite (`fdb_backend_satisfies_contract`) passes, covering point ops, batch delete, arbitrary/UTF-8 keys, transactional read-modify-write, CAS, concurrent write/delete conflicts, snapshot reads, begin-time snapshot, and unknown-commit behavior.
- `fdb_unreachable_cluster_does_not_hang` — points the backend at an unreachable address and asserts `begin()` fails fast within the per-transaction timeout (bounded, does not block process exit).
- `cargo fmt --check` and `cargo clippy --all-targets -- --deny=warnings` clean for both crates, with and without a reachable cluster.

# Dependencies

Builds on #1639 (PR 2: generic KV abstraction and memory backend). No dependency on other in-flight PRs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,27 @@ jobs:
           rustup default stable
           rustc --version
           cargo --version
+
+      - name: Install FoundationDB client library
+        run: |
+          # curvine-mds links libfdb_c unconditionally (foundationdb crate,
+          # fdb-7_3). The client rpm ships libfdb_c.so needed at link time;
+          # embedded-fdb-include already provides the headers at build time, so
+          # only the client library is required here.
+          set -euo pipefail
+          FDB_VERSION=7.3.77
+          curl -fsSL -o /tmp/foundationdb-clients.rpm \
+            "https://github.com/apple/foundationdb/releases/download/${FDB_VERSION}/foundationdb-clients-${FDB_VERSION}-1.el7.x86_64.rpm"
+          rpm -i --nodeps /tmp/foundationdb-clients.rpm
+          # The rpm installs libfdb_c.so under /usr/lib; also expose it under
+          # /usr/lib64 (Rocky's default 64-bit lib dir) so the linker finds it
+          # regardless of its default search paths.
+          if [ -f /usr/lib/libfdb_c.so ] && [ ! -e /usr/lib64/libfdb_c.so ]; then
+            ln -s /usr/lib/libfdb_c.so /usr/lib64/libfdb_c.so
+          fi
+          echo "/usr/lib" > /etc/ld.so.conf.d/foundationdb.conf
+          ldconfig
+          ldconfig -p | grep libfdb_c || (echo "libfdb_c not registered" && exit 1)
       
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,8 +61,12 @@ jobs:
           # only the client library is required here.
           set -euo pipefail
           FDB_VERSION=7.3.77
+          # Pinned SHA256 of the official release asset (supply-chain safety):
+          # verify the download before installing.
+          FDB_CLIENTS_SHA256=9255a4730655ac1a8461d8da5148a9d87c8978944188477f5b745318bb6d40ba
           curl -fsSL -o /tmp/foundationdb-clients.rpm \
             "https://github.com/apple/foundationdb/releases/download/${FDB_VERSION}/foundationdb-clients-${FDB_VERSION}-1.el7.x86_64.rpm"
+          echo "${FDB_CLIENTS_SHA256}  /tmp/foundationdb-clients.rpm" | sha256sum -c -
           rpm -i --nodeps /tmp/foundationdb-clients.rpm
           # The rpm installs libfdb_c.so under /usr/lib; also expose it under
           # /usr/lib64 (Rocky's default 64-bit lib dir) so the linker finds it
--- a/curvine-mds/src/kv/contract_tests.rs
+++ b/curvine-mds/src/kv/contract_tests.rs
@@ -325,7 +325,10 @@ async fn reads_observe_begin_snapshot(be: Arc<dyn KvBackend>, p: &[u8]) {
 
 fn error_classification() {
     assert!(KvError::Conflict.is_retryable());
-    assert!(KvError::Timeout.is_retryable());
+    assert!(KvError::TransientUnavailable("down".into()).is_retryable());
+    // DeadlineExceeded is NOT retryable: the per-txn deadline (the caller's wait
+    // budget) is exhausted, so it must surface rather than re-run.
+    assert!(!KvError::DeadlineExceeded.is_retryable());
     // MaybeCommitted is NOT retryable: the write may already have applied.
     assert!(!KvError::MaybeCommitted.is_retryable());
     assert!(!KvError::Aborted("a".into()).is_retryable());
@@ -576,12 +579,15 @@ async fn fdb_unreachable_cluster_does_not_hang() {
         Err(_elapsed) => panic!("begin() hung on an unreachable cluster (blocked >30s)"),
         Ok(Ok(_txn)) => panic!("begin() unexpectedly succeeded against an unreachable cluster"),
         Ok(Err(err)) => {
-            // Fail-fast surfaced as a retryable error (Timeout, or another
-            // transient class). It must NOT be a silent hang, which is all this
-            // criterion requires.
+            // Fail-fast surfaced as a bounded error instead of a silent hang —
+            // which is all this criterion requires. The exact class depends on
+            // how FDB reports the unreachable cluster: `DeadlineExceeded` if the
+            // per-txn deadline is hit, or `TransientUnavailable` for a
+            // connection-level failure. Both are acceptable; a hang (handled
+            // above) is not.
             assert!(
-                err.is_retryable(),
-                "expected a fail-fast retryable error, got {err:?}"
+                matches!(err, KvError::DeadlineExceeded | KvError::TransientUnavailable(_)),
+                "expected a bounded fail-fast error (DeadlineExceeded/TransientUnavailable), got {err:?}"
             );
         }
     }
--- a/curvine-mds/src/kv/contract_tests.rs
+++ b/curvine-mds/src/kv/contract_tests.rs
@@ -82,9 +82,20 @@ async fn arbitrary_bytes_round_trip(be: Arc<dyn KvBackend>, p: &[u8]) {
 }
 
 async fn distinct_invalid_utf8_keys_remain_distinct(be: Arc<dyn KvBackend>, p: &[u8]) {
-    let key_1 = k(p, &[b'f', b'o', 0x80]);
-    let key_2 = k(p, &[b'f', b'o', 0x81]);
+    // Two distinct invalid-UTF8 byte suffixes that collapse to the SAME lossy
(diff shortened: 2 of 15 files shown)


curvine-mds/src/kv/fdb.rs
```diff
@@ -0,0 +1,354 @@
...
+            // signature matches the trait (no Result).
+            let _ = trx.add_conflict_range(key, &end, ConflictRangeType::Read);
+        }
+    }
+
+    async fn commit(&mut self) -> KvResult<()> {
+        let start = Instant::now();
+        self.finished = true;
+        metrics::metrics().txn_in_flight.dec();
+        let trx = match self.trx.take() {
```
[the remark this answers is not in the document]
yangyang233333: @Copilot Good catch — fixed in 3d16e801. `txn_in_flight.dec()` is now called only after `self.trx.take()` returns `Some`, so a repeat commit on a finished transaction returns the error without decrementing the gauge.

curvine-mds/src/kv/contract_tests.rs
```diff
@@ -59,21 +73,19 @@ async fn arbitrary_bytes_round_trip(be: Arc<dyn KvBackend>) {
...
-    let key_2 = vec![b'f', b'o', 0x81];
+async fn distinct_invalid_utf8_keys_remain_distinct(be: Arc<dyn KvBackend>, p: &[u8]) {
+    let key_1 = k(p, &[b'f', b'o', 0x80]);
+    let key_2 = k(p, &[b'f', b'o', 0x81]);
 
-    assert_eq!(
-        String::from_utf8_lossy(&key_1),
-        String::from_utf8_lossy(&key_2)
-    );
+    assert_ne!(key_1, key_2);
```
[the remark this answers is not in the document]
yangyang233333: @Copilot Fixed in 3d16e801. Restored the precondition assertion that both byte keys share the same `String::from_utf8_lossy` representation (0x80/0x81 → U+FFFD), so the test again proves lossy-colliding keys stay distinct as raw bytes.

.github/workflows/build.yml
```diff
@@ -52,6 +52,27 @@ jobs:
...
+        run: |
+          # curvine-mds links libfdb_c unconditionally (foundationdb crate,
+          # fdb-7_3). The client rpm ships libfdb_c.so needed at link time;
+          # embedded-fdb-include already provides the headers at build time, so
+          # only the client library is required here.
+          set -euo pipefail
+          FDB_VERSION=7.3.77
+          curl -fsSL -o /tmp/foundationdb-clients.rpm \
+            "https://github.com/apple/foundationdb/releases/download/${FDB_VERSION}/foundationdb-clients-${FDB_VERSION}-1.el7.x86_64.rpm"
+          rpm -i --nodeps /tmp/foundationdb-clients.rpm
```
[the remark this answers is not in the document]
yangyang233333: @Copilot Addressed in 3d16e801. The client rpm is now verified against a pinned SHA256 (from the official release `.sha256`) via `sha256sum -c` before install, so a tampered or corrupted download fails the CI step.

curvine-mds/src/kv/fdb.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// `fdbcli -C <path>` accepts). Boots the process-global network on first
+    /// use.
+    pub fn open(cluster_file: &str, txn_timeout_ms: i32) -> KvResult<Self> {
+        ensure_network_started();
+        let cluster_file = cluster_file.trim();
+        if cluster_file.is_empty() {
+            return Err(KvError::Backend("fdb cluster file path is empty".into()));
+        }
+        let db = Database::from_path(cluster_file).map_err(map_fdb_error)?;
+        Ok(Self {
```
szbr00, line 219: Required: `FdbBackend::open` stores `txn_timeout_ms` with no check. FoundationDB `TransactionOption::Timeout(0)` means no timeout (wait forever), which breaks the fail-fast contract this PR states. `MdsConf` rejects `<= 0`, but `open` is public and is the call that applies the option (`begin` -> `set_option(Timeout(...))`). Reject `txn_timeout_ms <= 0` here so a direct caller cannot construct a backend that hangs on `begin`/`commit`.

curvine-mds/src/kv/fdb.rs
```diff
@@ -0,0 +1,357 @@
...
+    //   1020 not_committed          — optimistic-concurrency conflict
+    //   1031 transaction_timed_out  — deadline exceeded
+    const NOT_COMMITTED: i32 = 1020;
+    const TRANSACTION_TIMED_OUT: i32 = 1031;
+
+    // Timeout first, for semantic precision. 1031 is itself retryable, so this
+    // branch is not needed for correctness (it would otherwise fall through to
+    // the retryable arm), but surfacing Timeout distinguishes "deadline hit"
+    // from conflict/unavailability in errors and metrics.
+    if err.code() == TRANSACTION_TIMED_OUT {
```
szbr00: Critical: `map_fdb_error` turns FDB `transaction_timed_out` (1031) into `KvError::Timeout`, which `is_retryable()` treats as safe to replay. The call chain is `FdbTxn::commit` -> `trx.commit()` -> `map_fdb_error` -> `run_txn`. A commit that hits the per-txn timeout may already have applied or still be in-flight (FDB `on_error` does not retry 1031; `commit_unknown_result` is the only "unknown" FDB itself will retry). `run_txn` will re-run the closure and can double-apply a counter/CAS. The comment that "1031 is itself retryable" is not FDB's commit semantics.

Split the mapper by operation (explicit dispatcher): on `commit`, map 1031 and `is_maybe_committed()` to `MaybeCommitted`; on `begin`/`get`, keep 1031 as `Timeout` so an unreachable cluster remains retryable. Add a table-driven test with `FdbError::from_code` for 1020 / 1021 / 1031 (commit vs read) so a swapped arm fails.

szbr00 (changes requested):
Requesting changes after reviewing the FDB backend (`curvine-mds` kv/fdb + contract tests, `curvine-config` MdsConf, CI/lockfile).

The implementation matches the stated intent (FDB `KvBackend` on the #1639 contract, fail-fast timeout, same suite as memory) and the author's verification story is solid. One correctness issue must land first: commit-path `transaction_timed_out` (1031) is mapped to retryable `KvError::Timeout`, so `run_txn` can double-apply a write. Details are on the inline comments.

curvine-mds/src/kv/fdb.rs
yangyang233333: @szbr00 Good catch — fixed in a80def9f. Split the mapper by operation (`map_fdb_read_error` / `map_fdb_commit_error`). On a commit, 1031 and every code that is not provably-uncommitted now map to `MaybeCommitted` (not retryable), so `run_txn` never re-runs a possibly-applied write. On a read, 1031 maps to `DeadlineExceeded` (also not retryable — the per-txn timeout is the caller's wait budget). Verified against FDB source that 1031 is an outer deadline that can fire while a commit RPC is in flight, so its commit status is genuinely unknown. Added a table-driven test (1020/1021/1031/terminal × read vs commit) so a swapped arm fails.

curvine-mds/src/kv/fdb.rs
yangyang233333: @szbr00 Agreed — fixed in a80def9f. `FdbBackend::open` now rejects `txn_timeout_ms <= 0` and returns a `Backend` error, so the guard sits at the layer that actually applies `TransactionOption::Timeout`; a direct caller can no longer build a backend that hangs on begin/commit even if it bypasses `MdsConf` validation.

szbr00 (approved):

PR #1653, opened 2026-09-03
perf: demote expected IO errors to debug logging

bigbigxu:
# Summary

Demote expected/anticipated IO error logs from `error`/`info` to `debug` in the FUSE and client buffer paths, so that normal races (e.g. a file being deleted while readers/writers are still active, or callers being cancelled) no longer flood the error log.

# Issue Describe / Design

No linked issue.

Design notes:
- `FileNotFound` errors are expected during teardown/race scenarios; they are now logged at `debug` level in the FUSE reader/writer task loops and the fs read/write buffer channel tasks, while other errors remain at `error` level.
- In `BufferClient`, callback-send failures (which simply mean the caller has already been cancelled/gone) are now logged at `debug` with clearer messages instead of `info`.
- `ErrorMonitor` now exposes a public `error()` accessor, makes `has_error`/`error` fields public, and uses the relaxed `AtomicBool` helper instead of raw `SeqCst` load/store.

# Changes

| `curvine-fuse/src/fs/fuse_reader.rs` | Log `FileNotFound` as debug, others as error | None (log level only) |
| `curvine-fuse/src/fs/fuse_writer.rs` | Log `FileNotFound` as debug, others as error | None (log level only) |
| `crates/client/curvine-client-core/src/file/fs_reader_buffer.rs` | Log `FileNotFound` as debug | None (log level only) |
| `crates/client/curvine-client-core/src/file/fs_writer_buffer.rs` | Log `FileNotFound` as debug | None (log level only) |
| `crates/core/rpc/src/client/buffer_client.rs` | Downgrade callback-failure logs to `debug` with clearer text | None (log level only) |
| `crates/core/runtime/src/sync/error_monitor.rs` | Add `error()` accessor; public fields; relaxed `AtomicBool` | None (additive API) |
| `curvine-fuse/src/fuse_error.rs` | Add `is_enoent()` helper | None (additive API) |

# Test verified

| Not run locally | N/A | To be verified by CI |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/core/runtime/src/sync/error_monitor.rs
+++ b/crates/core/runtime/src/sync/error_monitor.rs
@@ -12,14 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::sync::AtomicBool;
 use std::error::Error;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 
 // Used to set errors in asynchronous environments.
 pub struct ErrorMonitor<E: Error> {
-    has_error: AtomicBool,
-    error: Mutex<Option<E>>,
+    pub has_error: AtomicBool,
+    pub error: Mutex<Option<E>>,
 }
 
 impl<E: Error> ErrorMonitor<E> {
@@ -31,7 +31,11 @@ impl<E: Error> ErrorMonitor<E> {
     }
 
     pub fn has_error(&self) -> bool {
-        self.has_error.load(Ordering::SeqCst)
+        self.has_error.get()
+    }
+
+    pub fn error(&self) -> &Mutex<Option<E>> {
+        &self.error
     }
 
     // If this error has been set, it will be returned directly.
@@ -40,7 +44,7 @@ impl<E: Error> ErrorMonitor<E> {
             return;
         }
         let mut e = self.error.lock().unwrap();
-        self.has_error.store(true, Ordering::SeqCst);
+        self.has_error.set(true);
         let _ = (*e).replace(error);
     }
 
--- a/crates/core/runtime/src/sync/error_monitor.rs
+++ b/crates/core/runtime/src/sync/error_monitor.rs
@@ -38,14 +38,17 @@ impl<E: Error> ErrorMonitor<E> {
         &self.error
     }
 
-    // If this error has been set, it will be returned directly.
+    // Keep the first error; later callers return without overwriting.
     pub fn set_error(&self, error: E) {
         if self.has_error() {
             return;
         }
         let mut e = self.error.lock().unwrap();
+        if e.is_some() {
+            return;
+        }
         self.has_error.set(true);
-        let _ = (*e).replace(error);
+        *e = Some(error);
     }
 
     pub fn take_error(&self) -> Option<E> {
--- a/crates/core/runtime/src/sync/error_monitor.rs
+++ b/crates/core/runtime/src/sync/error_monitor.rs
@@ -75,21 +75,32 @@ impl<E: Error> Default for ErrorMonitor<E> {
 
 #[cfg(test)]
 mod tests {
-    use std::sync::Arc;
+    use std::fmt;
+    use std::sync::{Arc, Barrier};
     use std::thread;
-    use std::thread::Barrier;
 
     use crate::sync::ErrorMonitor;
+
+    #[derive(Debug, PartialEq)]
+    struct TestError(&'static str);
+
+    impl fmt::Display for TestError {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(f, "{}", self.0)
+        }
+    }
+
+    impl std::error::Error for TestError {}
 
     /// A later set_error must not overwrite the first stored error.
     #[test]
     fn first_error_wins() {
-        let monitor = ErrorMonitor::<String>::new();
-        monitor.set_error("first".to_string());
-        monitor.set_error("second".to_string());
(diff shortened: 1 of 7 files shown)


crates/core/runtime/src/sync/error_monitor.rs
```diff
@@ -40,7 +44,7 @@ impl<E: Error> ErrorMonitor<E> {
             return;
         }
         let mut e = self.error.lock().unwrap();
-        self.has_error.store(true, Ordering::SeqCst);
+        self.has_error.set(true);
         let _ = (*e).replace(error);
     }
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 8356032f: `set_error` now re-checks under the mutex and keeps the first stored error so concurrent callers cannot overwrite it.

crates/core/runtime/src/sync/error_monitor.rs
```diff
@@ -31,17 +31,24 @@ impl<E: Error> ErrorMonitor<E> {
...
             return;
         }
         let mut e = self.error.lock().unwrap();
-        self.has_error.store(true, Ordering::SeqCst);
-        let _ = (*e).replace(error);
+        if e.is_some() {
+            return;
+        }
+        self.has_error.set(true);
+        *e = Some(error);
```
szbr00, line 51: Required: `set_error` now keeps the first stored error under the mutex, but there is no assertion that fails if a later caller overwrites it. `curvine-runtime` already unit-tests other sync primitives (`state_monitor`, `async_map`, `count_down_latch`). Add a `#[cfg(test)]` on this type that (1) sets two errors sequentially and asserts `take_error()` returns the first, and (2) races two threads/tasks calling `set_error` and asserts exactly one error is stored and it is one of the two inputs. Compile CI will not catch a regression to `replace()`.

crates/core/runtime/src/sync/error_monitor.rs
```diff
@@ -12,14 +12,14 @@
...
+use crate::sync::AtomicBool;
 use std::error::Error;
-use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 
 // Used to set errors in asynchronous environments.
 pub struct ErrorMonitor<E: Error> {
-    has_error: AtomicBool,
-    error: Mutex<Option<E>>,
+    pub has_error: AtomicBool,
```
szbr00, line 21: Optional: `has_error` and `error` were made public and `error()` was added, but nothing in this PR or the fuse/client-core callers reads either field or the accessor. Public fields let callers mutate the mutex and the flag independently and bypass the first-error invariant just added in `set_error`. Keep the fields private and drop `error()` unless there is a current consumer.

curvine-fuse/src/fuse_error.rs
```diff
@@ -52,6 +52,10 @@ impl FuseError {
     pub(crate) fn errno(&self) -> i32 {
         self.errno
     }
+
+    pub fn is_enoent(&self) -> bool {
```
szbr00, line 56: Optional: `FuseError::is_enoent()` has no callers. The FUSE reader/writer paths classify `ErrorKind::FileNotFound` on `FsError`, so this helper is unused. Drop it from this PR, or use it if a `FuseError`-level check is intended.

szbr00 (dismissed):
Required: The FileNotFound / cancelled-callback log demotion is sound (errors are still stored and returned; BufferClient cb.send failures are caller teardown). The ErrorMonitor first-wins change needs a unit test before merge — compile CI will not catch a regression to replace(). See inline comments.

Inspected: curvine-fuse reader/writer/error, curvine-client-core buffer paths, curvine-rpc BufferClient, curvine-runtime ErrorMonitor/AtomicBool (still SeqCst). No proto/wire change. Verification story is CI-only.

crates/core/runtime/src/sync/error_monitor.rs
bigbigxu: Thanks. Added the requested unit tests in 0eff7820: (1) sequential set_error keeps the first error, take_error() returns it and then None; (2) two threads race set_error behind a Barrier, asserting exactly one error is stored and it is one of the two inputs.

crates/core/runtime/src/sync/error_monitor.rs
bigbigxu: Thanks for the suggestion. Keeping the public fields and the error() accessor for now: an upcoming PR has callers that need direct access to the error mutex, so the fields will gain consumers shortly. We will revisit narrowing visibility once those callers land.

curvine-fuse/src/fuse_error.rs
bigbigxu: Thanks. Keeping is_enoent() — an upcoming PR will use it at the FuseError level, so it will gain callers shortly. If that changes, we will drop it.

crates/core/runtime/src/sync/error_monitor.rs
bigbigxu: Follow-up: CI caught two issues in the initial tests, fixed in 8d416ed2 — Barrier is in std::sync (not std::thread), and String does not implement std::error::Error, so the tests now use a small local TestError type.

bigbigxu: One more compile fix in e6be78fb: the second race thread now uses a cloned Arc so the original monitor stays owned by the test thread for the final assertions.

lzjqsdd (approved):

PR #1659, opened 2026-09-04
fix(java): harden SDK runtime packaging

jlon:
# Summary

Harden the Java SDK single-jar runtime packaging path for mixed Hadoop/Flink/StarRocks environments. This PR removes avoidable runtime conflicts from the client artifact, improves native library fallback/loading, and keeps JNI compatibility with older Linux baselines.

# Issue Describe / Design

Related issues: None

Root cause:
- The Java SDK client artifact could leak Hadoop/Scala/bench/javax/jakarta classes into runtime classpaths, causing conflicts with host frameworks such as Flink and StarRocks.
- Native loading relied on one exact OS-specific library name, so Ubuntu/CentOS variants could fail even when a compatible generic Linux or CentOS 7 JNI library was packaged.
- Packaged JNI libraries need Jindo runtime extraction into the same directory so `$ORIGIN` can resolve `libjindosdk_c.so.6`.
- CentOS 7 era glibc environments require static-link compatibility shims and ARM linker wrapping as well as x86.

Fix approach:
- Add Java-only helpers and move shell/stat/runtime packaging code away from Scala bench utilities.
- Tighten Maven jar/shade/client packaging rules and relocate Curvine-owned dependencies under `curvine_shaded.*`.
- Load native libraries through an ordered candidate list: exact OS first, then generic Linux/CentOS fallback; extract Jindo dependency beside the selected JNI library.
- Keep native library name resolution testable without triggering JNI loading.
- Add glibc compatibility shims for old Linux runtimes and extend the static link wrapper to ARM.
- Allow OpenDAL HDFS to select JVM or native backend through configuration while keeping existing default behavior.

# Changes

| `curvine-libsdk/java/pom.xml` | Harden main/shade/client packaging, mark Hadoop as provided, relocate bundled dependencies, remove bench/logging/runtime-conflict entries | Behavior change: client jar becomes safer for host-managed Hadoop/Flink/StarRocks classpaths |
| `curvine-libsdk/java/src/main/java/io/curvine/CurvineNative.java` | Add native candidate fallback and colocated Jindo extraction/loading | Behavior change: wider Linux distro compatibility and better diagnostics |
| `curvine-libsdk/java/src/main/java/io/curvine/CurvineNativeLibraryResolver.java` | Split native OS/arch library name resolution from JNI loading | Test-only path no longer requires packaged native `.so` |
| `curvine-libsdk/java/src/main/java/io/curvine/CurvineJavaUtils.java` | Add Java-only size parsing/formatting and Curvine Hadoop conf helper | Removes runtime dependency on Scala bench utility code |
| `curvine-libsdk/java/src/main/java/io/curvine/*Stream.java` | Remove `javax.annotation.Nonnull` from stream overrides | Avoids unnecessary javax dependency in client artifact |
| `crates/adapters/curvine-ufs-opendal/src/lib.rs` | Route HDFS provider selection through `hdfs.provider` / `hdfs.backend` / `hdfs.native` | Existing default remains JVM HDFS when available; native backend can be selected explicitly |
| `.cargo/config.toml`, `build/static-link-wrapper.sh`, `curvine-libsdk/src/cxx_compat.rs` | Extend static-link wrapper and add CentOS 7 compatibility symbols | Improves JNI loadability on old glibc and ARM environments |
| `Cargo.toml`, `Cargo.lock`, `curvine-libsdk/Cargo.toml` | Bump workspace/package version and add required libc dependency | Version and dependency alignment for the SDK build |
| `curvine-libsdk/java/src/test/java/io/curvine/*` | Add Java utility and native fallback coverage | Test-only impact |

# Test verified

| `make format` | PASS | Runs project pre-commit formatting/check path, including Rust release workspace checks |
| `cargo check -p curvine-libsdk --no-default-features --features java-sdk-minimal` | PASS | Validates minimal Java SDK native feature graph |
| `mvn -B -ntp -f curvine-libsdk/java/pom.xml -Dtest=io.curvine.CurvineJavaUtilsTest,io.curvine.LibFsTest#linuxLibraryNamesFallBackToGenericLinux,io.curvine.LibFsTest#osVersion test` | PASS | 4 tests passed |
| `mvn -B -ntp -f curvine-libsdk/java/pom.xml -Prelease -DskipTests package` | PASS | Built main/client/shade jars |
| `jar tf curvine-libsdk/java/target/curvine-hadoop-0.2.2-client.jar | rg '(^scala/|^javax/|^jakarta/|^org/apache/hadoop/|^io/curvine/bench/|^META-INF/versions/|StorageSize.class)'` | PASS | No forbidden entries found |
| `jar tf curvine-libsdk/java/target/curvine-hadoop-0.2.2.jar | rg '^io/curvine/bench/'` | PASS | No forbidden bench entries found |
| `jar tf curvine-libsdk/java/target/curvine-hadoop-0.2.2-shade.jar | rg '(^scala/|^io/curvine/bench/)'` | PASS | No forbidden Scala/bench entries found |
| `javap -verbose -classpath curvine-libsdk/java/target/curvine-hadoop-0.2.2-client.jar io.curvine.CurvineNative | rg 'major version'` | PASS | Java bytecode major version 52 |
| `mvn -B -ntp -f curvine-libsdk/java/pom.xml -Dtest=io.curvine.CurvineJavaUtilsTest,io.curvine.LibFsTest test` | FAIL | `LibFsTest.jni1` requires local Curvine master on `localhost:9001/9002`; current machine has no local master, so it fails with connection refused |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None

### DIFF
--- a/crates/adapters/curvine-ufs-opendal/src/lib.rs
+++ b/crates/adapters/curvine-ufs-opendal/src/lib.rs
@@ -354,64 +354,8 @@ impl OpendalFileSystem {
                 Self::add_stability_layers(base_op, &conf)?
             }
 
-            #[cfg(feature = "opendal-hdfs")]
-            "hdfs" => {
-                use curvine_hdfs_jni::jni::{register_jvm, JVM};
-
-                register_jvm();
-
-                let _ = JVM.get_or_init().map_err(|e| {
-                    FsError::common(format!("Failed to initialize JVM for HDFS: {}", e))
-                })?;
-
-                let mut builder = Hdfs::default();
-
-                let namenode = if let Some(namenode_config) = conf.get("hdfs.namenode") {
-                    namenode_config.clone()
-                } else {
-                    format!("hdfs://{}", bucket_or_container)
-                };
-
-                builder = builder.name_node(&namenode);
-
-                let root_path = conf.get("hdfs.root").map(|s| s.as_str()).unwrap_or("/");
-                builder = builder.root(root_path);
-
-                let hdfs_user = conf
-                    .get("hdfs.user")
-                    .cloned()
-                    .or_else(|| std::env::var("HADOOP_USER_NAME").ok())
-                    .or_else(|| std::env::var("USER").ok());
-
-                if let Some(user) = hdfs_user {
-                    builder = builder.user(&user);
-                }
-
-                if let Some(ccache) = conf.get("hdfs.kerberos.ccache") {
-                    builder = builder.kerberos_ticket_cache_path(ccache);
-                } else if let Ok(ccache) = std::env::var("KRB5CCNAME") {
-                    builder = builder.kerberos_ticket_cache_path(&ccache);
-                }
-
-                if let Some(krb5_conf) = conf.get("hdfs.kerberos.krb5_conf") {
-                    std::env::set_var("KRB5_CONFIG", krb5_conf);
-                }
-
-                if conf
-                    .get("hdfs.atomic_write_dir")
-                    .map(|s| s == "true")
-                    .unwrap_or(false)
-                {
-                    let atomic_dir = format!("{}/atomic_write_dir", root_path);
-                    builder = builder.atomic_write_dir(&atomic_dir);
-                }
-
-                let base_op = Operator::new(builder)
-                    .map_err(|e| FsError::common(format!("Failed to create HDFS operator: {}", e)))?
-                    .finish();
-
-                Self::add_stability_layers(base_op, &conf)?
-            }
+            #[cfg(any(feature = "opendal-hdfs", feature = "opendal-hdfs-native"))]
+            "hdfs" => Self::create_hdfs_operator(&bucket_or_container, &conf)?,
 
             #[cfg(feature = "opendal-webhdfs")]
             "webhdfs" => {
@@ -553,9 +497,6 @@ impl OpendalFileSystem {
                 Self::add_stability_layers(base_op, &conf)?
             }
 
-            #[cfg(all(feature = "opendal-hdfs-native", not(feature = "opendal-hdfs")))]
-            "hdfs" => Self::create_hdfs_native_operator(&bucket_or_container, &conf)?,
-
             _ => {
                 return Err(FsError::unsupported(format!(
                     "Unsupported scheme: {}",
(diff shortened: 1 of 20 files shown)


curvine-libsdk/java/src/main/java/io/curvine/CurvineJavaUtils.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public static Configuration getCurvineConf() {
+        String confDir = System.getProperty("curvine.conf.dir");
+        Configuration conf = new Configuration(false);
+        File file = new File(Paths.get(confDir, "curvine-site.xml").toString());
+        conf.addResource(new Path(file.getPath()));
+
+        if (conf.get("fs.cv.impl") == null) {
+            conf.set("fs.cv.impl", "io.curvine.CurvineFileSystem");
```
szbr00, line 155: Required: `getCurvineConf()` now registers `fs.cv.impl` and `fs.AbstractFileSystem.cv.impl` (replacing the incorrect `fs.curvine.impl` in `Utils.scala`), but `CurvineJavaUtilsTest` only covers `byteFromString`. `CurvineShell` is the production caller; a revert to `fs.curvine.impl` would compile and the current tests would still pass. Assert the two `cv` scheme keys (and that `fs.curvine.impl` is not the one set) so the CLI path cannot silently lose FileSystem registration.

curvine-libsdk/java/src/main/java/io/curvine/CurvineNativeLibraryResolver.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CurvineNativeLibraryResolver.class);
+
+    static final String OS_RELEASE_FILE = "/etc/os-release";
+    static final String LINUX_ID_PREFIX = "ID=";
+    static final String LINUX_VERSION_PREFIX = "VERSION_ID=";
+
+    private CurvineNativeLibraryResolver() {
+    }
+
+    static String[] getLibraryNames(String sysOs, String osVersion, String arch) {
```
szbr00, line 39: Required: `curvine-libsdk/README.md` still says the JNI library name must match `CurvineNative.getLibraryName()`. After this change the loader walks an ordered candidate list (exact OS → `linux` → `centos7` → x86 `rocky9` / `libcurvine_libsdk.so`). Update that Java SDK section so packagers and `java.library.path` / `LD_PRELOAD` users know which names are tried.

Consider: official `build-java-sdk.yml` ships `ubuntu22`/`ubuntu24`/`rocky9`/`amzn2`/`alinux3`, not `linux_*` or `centos7_*`. On an unmatched distro the only packaged fallback that is actually tried is x86 `rocky9`, which is often too new; `amzn2`/`alinux3` (older glibc) sit in the jar unused. Either emit the generic/centos7 names in the release workflow, or add those shipped names to the candidate list (including aarch).

curvine-libsdk/java/pom.xml
```diff
@@ -73,13 +73,48 @@
...
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-hdfs-client</artifactId>
             <version>${hadoop.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-cli</groupId>
```
szbr00, line 87: Optional: `commons-cli` is only imported by `io.curvine.bench.BenchParams`, and bench classes are now excluded from the main/client/shade jars. A compile-scoped CLI dependency will still be pulled into the client artifact (and relocated under `curvine_shaded.org.apache.commons`). Drop it from production dependencies, or confine it to the bench/test classpath, so the client jar does not grow an unused Commons CLI tree.

crates/adapters/curvine-ufs-opendal/src/lib.rs
```diff
@@ -576,6 +517,119 @@ impl OpendalFileSystem {
...
+        }
+
+        #[cfg(all(not(feature = "opendal-hdfs"), feature = "opendal-hdfs-native"))]
+        {
+            Self::create_hdfs_native_operator(bucket_or_container, conf)
+        }
+    }
+
+    #[cfg(any(feature = "opendal-hdfs", feature = "opendal-hdfs-native"))]
+    fn use_hdfs_native(conf: &HashMap<String, String>) -> bool {
```
szbr00, line 599: Required: `use_hdfs_native` introduces four user-visible config keys (`hdfs.provider`, `hdfs.backend`, `opendal.hdfs.driver`, `hdfs.native`) with no test and no owning-doc update. Default JVM behavior is preserved, but nothing would fail if the dispatcher inverted that default, treated an unknown provider as native, or stopped honoring `hdfs.native=true`. Author verification also did not run `curvine-ufs-opendal` tests. Add a unit test of this helper (empty map -> false; `native`/`hdfs-native` -> true; unknown provider -> false; `hdfs.native=true` -> true) and document one canonical key on `create_hdfs_native_operator`'s config list. Prefer a single key over four aliases unless a current consumer needs each name.

szbr00 (changes requested):
Required: The Java packaging and JNI fallback direction is sound, but three gaps should land before merge: unit-test `use_hdfs_native` and document one HDFS backend key; assert `getCurvineConf()` registers `fs.cv.impl` / `fs.AbstractFileSystem.cv.impl`; and update `curvine-libsdk/README.md` so the Java SDK section describes the ordered native candidate list instead of a single `getLibraryName()`.

Optional: `commons-cli` is bench-only after the jar excludes and should not ship on the client classpath.

curvine-libsdk/java/src/main/java/io/curvine/CurvineNativeLibraryResolver.java
jlon: Fixed. The README now documents the ordered native candidate list. The resolver also tries the workflow-produced amzn2, alinux3, and rocky9 names for both x86_64 and aarch64, after linux and centos7 fallbacks. The existing build-java-sdk workflow already emits the distribution-specific names.

crates/adapters/curvine-ufs-opendal/src/lib.rs
jlon: Fixed. use_hdfs_native now uses hdfs.provider as the canonical selector and retains hdfs.native only as a legacy fallback. Tests cover empty config, native/hdfs-native, JVM, unknown provider, boolean fallback, and provider precedence; the canonical key is documented.

curvine-libsdk/src/lib.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+mod cxx_compat;
```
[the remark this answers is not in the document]
jlon: Fixed. cxx_compat is now compiled only for target_os=linux and target_env=gnu at the module declaration in curvine-libsdk/src/lib.rs.

curvine-libsdk/java/src/main/java/io/curvine/CurvineJavaUtils.java
jlon: Fixed. CurvineJavaUtilsTest now verifies fs.cv.impl and fs.AbstractFileSystem.cv.impl, and verifies that the old fs.curvine.impl key is not used. It also covers missing configuration directory and missing curvine-site.xml failures.

curvine-libsdk/java/pom.xml
jlon: Addressed. commons-cli is now provided-scoped: benchmark sources still compile and test, but the dependency is not part of the runtime dependency tree or any packaged jar.

curvine-libsdk/java/src/main/java/io/curvine/CurvineJavaUtils.java
```diff
@@ -0,0 +1,184 @@
...
+            default:
+                throw new IllegalArgumentException("Unsupported size unit: " + unit);
+        }
+    }
+
+    public static Configuration getCurvineConf() {
+        String confDir = System.getProperty("curvine.conf.dir");
+        Configuration conf = new Configuration(false);
+        File file = new File(Paths.get(confDir, "curvine-site.xml").toString());
+        conf.addResource(new Path(file.getPath()));
```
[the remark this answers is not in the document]
jlon: Fixed. getCurvineConf() now validates curvine.conf.dir before constructing the Hadoop Configuration and verifies that curvine-site.xml is a regular file. Both failures use a clear IllegalArgumentException.

szbr00 (approved):

PR #1663, opened 2026-09-07
feat(master): add inode id to resize and assign worker RPCs (#1662)

### LINKED ISSUE #1662: [FEATURE] Support resolving files by inode id in resize and assign worker RPCs
When a client performs file resize or block assignment, the master resolves the file by path lookup (`resolve_path`). The client already knows the inode id, and path-based resolution can mismatch when the file is renamed/moved concurrently, and adds unnecessary namespace traversal on hot write paths.

Add an optional `inode_id` field to `FileResizeRequest` and `AssignWorkerRequest`. The master resolves the file inode by id when provided (with path fallback), refactored into `FsDir::resize_inode`/`assign_worker_inode` and `MasterFilesystem::resize_by_id`/`assign_worker_by_id`. The client writer passes the known inode id via `resize_by_id`/`assign_worker_by_id`.

Keep path-only resolution and re-validate that the resolved inode id matches the client-side id after the operation. This still pays the full namespace traversal cost and can fail on concurrent renames instead of resolving directly.

**Additional context**
Implemented in branch `feat/resize-inode-id`.

# Goal

Design scope: list sub-requirements and the problem each solves.

| Optional `inode_id` in resize/assign-worker RPCs | Master can locate the file directly by inode |
| Inode-based resolution in master with path fallback | Avoid rename/move races and redundant traversal |
| Client writer passes inode id | End-to-end inode-based flow |

# Not Goal

Modules, features, or boundaries this issue must **not** touch.

- Changing journal format or audit log semantics (path is still used for logging)
- Extending inode-id based resolution to other RPCs

# Test Plan

| Resize by inode id | Resize resolves file via inode id | curvine-tests/fs_test.rs |
| Existing resize/assign-worker flows | Backward compatible when inode_id absent | master, client |

bigbigxu:
# Summary

Pass the client-known inode id in resize and assign-worker requests so the master can resolve the file by inode instead of path lookup, avoiding rename/move races and redundant namespace traversal on hot write paths.

# Issue Describe / Design

Closes #1662

- `FileResizeRequest` and `AssignWorkerRequest` gain an optional `inode_id` field (proto2 optional, backward compatible).
- The master resolves the file by inode id when provided, with path lookup as fallback (`resolve_file_inode`), so older clients keep working.
- `FsDir::resize_inode` / `FsDir::assign_worker_inode` take an already-resolved `InodePtr`; journal and audit logging still use the path string, so journal format and audit semantics are unchanged.
- The client writer already holds the file inode id in `file_blocks.status.id`, so it calls `resize_by_id` / `assign_worker_by_id` directly.

# Changes

| `crates/common/curvine-proto/proto/master.proto` | Add optional `inode_id` to `FileResizeRequest` (=3) and `AssignWorkerRequest` (=5) | None, optional field |
| `curvine-master/src/master/meta/fs_dir.rs` | Extract `resize_inode` / `assign_worker_inode` operating on a resolved `InodePtr` | None, original path-based methods delegate |
| `curvine-master/src/master/fs/master_filesystem.rs` | Add `resize_by_id` / `assign_worker_by_id` with inode-based resolution and path fallback; return blocks from the same resolved inode | None when `inode_id` is absent |
| `curvine-master/src/master/master_handler.rs` | Forward `inode_id` from RPC requests | None |
| `crates/client/curvine-client-core/src/file/fs_client.rs` | Add `resize_by_id` / `assign_worker_by_id`; requests carry `inode_id` | None |
| `crates/client/curvine-client-core/src/file/fs_writer_base.rs` | Writer passes `file_blocks.status.id` on resize / assign-worker; flush instead of full complete before resize | Assign/resize now resolve by inode id |
| `curvine-tests/tests/fs_test.rs` | Add resize-by-inode-id coverage | None |

Review updates:

- `resize_by_id` now holds the `fs_dir` write guard across the whole flow (resolve, resize, `remove_blocks`, response build), so concurrent `complete_file` / `add_block` / second resize can no longer be omitted from the returned `FileBlocks`; the legacy (`inode_id == None`) path is also protected against delete/recreate races, since resolution and response build are atomic under one write lock.
- Corrected the regression test comment to describe the actual resize trigger (write at an offset past EOF).

# Test verified

| `cargo test -p curvine-tests --test fs_test` | Passed | Includes the new `open_file_rename_sparse_write` regression |
| Existing resize / assign-worker regression | Passed | Backward-compatible path (no `inode_id`) covered by regression suite |

# Dependencies

- Related PRs: None
- Related issues: #1662
- Blocks / blocked by: None

### DIFF
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -1679,8 +1679,9 @@ impl MasterFilesystem {
         } else {
             self.worker_manager.read().available_bytes()
         };
+
+        let mut fs_dir = self.fs_dir.write();
         let (del_res, inode) = {
-            let mut fs_dir = self.fs_dir.write();
             let mut inode = Self::resolve_file_inode(&fs_dir, path, inode_id)?;
             let file = inode.as_file_ref()?;
             Self::validate_alloc_capacity(file.len, file.replicas, &opts, available)?;
@@ -1693,7 +1694,6 @@ impl MasterFilesystem {
         }
 
         let blocks = {
-            let fs_dir = self.fs_dir.read();
             let file = inode.as_file_ref()?;
             let locs = self.get_block_locs(path, &fs_dir, file)?;
             let status = inode.to_file_status(path)?;
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -1659,6 +1659,15 @@ impl MasterFilesystem {
     }
 
     pub fn resize<T: AsRef<str>>(&self, path: T, opts: FileAllocOpts) -> FsResult<FileBlocks> {
+        self.resize_by_id(path, None, opts)
+    }
+
+    pub fn resize_by_id<T: AsRef<str>>(
+        &self,
+        path: T,
+        inode_id: Option<i64>,
+        opts: FileAllocOpts,
+    ) -> FsResult<FileBlocks> {
         opts.validate()?;
 
         let path = path.as_ref();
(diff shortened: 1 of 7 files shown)


curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1670,30 +1679,26 @@ impl MasterFilesystem {
...
-        let blocks = self.get_block_locations(path)?;
-        if blocks.status.id != inode_id {
-            return err_box!(
-                "Path {} resolved to different inode after resize, expected {}, got {}",
-                path,
-                inode_id,
-                blocks.status.id
-            );
-        }
+        let blocks = {
```
szbr00, line 1696: Required: This response is built from the owned `inode` snapshot captured before `fs_dir.write()` was released. The new read guard does not refresh that snapshot, so a concurrent `complete_file`, `add_block`, or second resize between the unlock and this block is omitted from the returned `FileBlocks`; `FsWriterBase::resize` then replaces its entire local block map with that stale response. Re-resolve the captured inode ID under this read guard before building the response. For the `inode_id == None` path, also retain the previous path-to-inode identity check so delete/recreate does not return the old inode under the new path.

szbr00 (commented):
Required: The PR verification table still marks both `cargo test -p curvine-tests --test fs_test` and the existing path-only resize/assign-worker regression as Pending. This changes client/master RPC behavior on the write path, and the current CI result does not provide the claimed Linux cluster scenario. Please run and record those targeted checks before merge.

curvine-tests/tests/fs_test.rs
```diff
@@ -587,6 +590,50 @@ async fn open_file_rename_read_write(fs: &CurvineFileSystem) -> CommonResult<()>
...
+    fs.rename(&src, &dst).await?;
+    assert!(!fs.exists(&src).await?);
+    assert!(fs.exists(&dst).await?);
+
+    const HOLE_END: i64 = 4096;
+    const PREFIX: &[u8] = b"prefix";
+    const SUFFIX: &[u8] = b"suffix";
+
+    // Flushing this buffered write during the following seek triggers the
+    // rename-sensitive sparse resize and block assignment paths.
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Addressed: the inline comment now states the actual trigger - the write at an offset past EOF invokes FsWriterBase::write, which calls resize when pos > len and no block exists; there is no buffered write flushed by the preceding seek.

curvine-master/src/master/fs/master_filesystem.rs
bigbigxu: Thanks. Addressed: resize_by_id now holds the fs_dir write guard across the whole flow (resolve, resize_inode, remove_blocks, response build), so there is no window between the metadata update and building the returned FileBlocks - concurrent complete_file / add_block / second resize can no longer be omitted, and the client no longer replaces its local block map with a stale response. Since resolution and response build are now atomic under one write lock, the legacy (inode_id == None) path is also protected against the delete/recreate race that the previous path-to-inode identity check guarded against.

szbr00 (approved):

Issue #1673, opened 2026-09-08
[BUG]: Worker WriteHandler ignores flush frames - client flush()/hflush does not flush buffered data


bigbigxu:
## Describe the bug

In `WriteHandler::write` (curvine-worker), a `RequestStatus::Running` frame carrying `DataHeaderProto { flush: true }` does not actually flush the block writer.

When the client calls `BlockWriterRemote::flush()` (e.g. via FUSE fsync / hflush semantics), `BlockClient::write_flush` sends a Running frame with `flush: true` and no data. The worker's `write()` only uses the header for seek handling: when `flush == true` it skips `seek_to` and returns success **without ever calling `file.flush()`**. The real flush only happens later in `complete()`.

As a result:

1. Client `flush()` returns success but buffered data in the worker-side `BufWriter` is not pushed to the OS, so the hflush/fsync durability guarantee is not honored.
2. Short-circuit readers on the same node (which read the block file directly) may not see recently written bytes.
3. The batch write path is inconsistent: `BatchWriteHandler::flush_batch` correctly calls `file.flush()` on flush frames, but the single-block path does not.

## To Reproduce

1. Write data to a file through the remote (non-short-circuit) write path so that the worker holds a buffered `BlockWriteContext`.
2. Call writer `flush()` on the client before `complete()` (e.g. fsync from FUSE).
3. From another reader (short-circuit read on the same node), read the written range before the block is completed.
4. The reader may observe stale/missing data, although `flush()` returned success.

## Expected behavior

When the worker receives a Running write frame with `DataHeaderProto.flush == true`, it should call `file.flush()` (pushing the buffered data to the OS) before returning success, aligned with `BatchWriteHandler::flush_batch`.

The flush frame's `offset` is informational (current client write position) and must not be treated as a seek target: it can legitimately equal `block_size` when flushing a fully written block, which the seek validation would reject.

## OS Version

- Linux (worker/master deployment), Curvine main branch (commit 10e073533)
- Affected file: `curvine-worker/src/worker/handler/write_handler.rs`

## Additional context

Proposed fix (7 lines): mark `need_flush` when `header.flush` is set, and after the data-write section call `file.flush()?` so a flush failure is propagated to the client instead of being silently dropped. Write metrics (`write_bytes` / `write_time_us` / `write_count`) remain tied to the `write_region` operation itself.

vishanth14:
[...]

PR #1674, opened 2026-09-08
fix(worker): execute flush on flush frames in WriteHandler::write (#1673)

bigbigxu:
# Summary

This PR fixes `WriteHandler::write` in the worker so that a Running write frame carrying `DataHeaderProto { flush: true }` actually executes `file.flush()` instead of silently ignoring it.

# Issue Describe / Design

Closes #<ISSUE>

- **Root cause**: in `WriteHandler::write`, the `DataHeaderProto` header was only used for seek handling. When `flush == true` the code skipped `seek_to` and returned success without calling `file.flush()`; buffered data was only pushed to the OS later in `complete()`. Client `flush()`/hflush therefore returned success without making data durable/visible, and short-circuit readers could miss recently written bytes.
- **Reproduction**: write through the remote path, call writer `flush()` before `complete()`, then read the range via short-circuit read on the same node — data may be missing although `flush()` succeeded.
- **Fix approach**: set `need_flush` when `header.flush` is present, and execute `file.flush()?` after the data-write section, aligning the single-block path with the already-correct `BatchWriteHandler::flush_batch`. The flush frame's `offset` is informational (it may equal `block_size` for a fully written block) and is intentionally not treated as a seek target. Write metrics stay tied to `write_region`; a flush failure now propagates to the client via `?`.

# Changes

| `curvine-worker/src/worker/handler/write_handler.rs` | In `write()`: track `need_flush` from `DataHeaderProto.flush` and call `file.flush()?` after the data write | Behavior change: flush frames now actually push buffered data to the OS and propagate flush errors; normal write/seek frames are unchanged |

# Test verified

| Manual code review of write path vs `BatchWriteHandler::flush_batch` | PASS | Semantics aligned; flush after data write |
| To be run on Linux CI: `cargo test -p curvine-worker` | PASS |  |

# Dependencies

- Related PRs: None
- Related issues: Closes #<ISSUE>
- Blocks / blocked by: None

### DIFF
--- a/curvine-worker/src/worker/handler/write_handler.rs
+++ b/curvine-worker/src/worker/handler/write_handler.rs
@@ -231,6 +231,7 @@ impl WriteHandler {
         let context = try_option_mut!(self.context);
         Self::check_context(context, msg)?;
 
+        let mut need_flush = false;
         if msg.header_len() > 0 {
             let header: DataHeaderProto = msg.parse_header()?;
             if !header.flush {
@@ -242,6 +243,8 @@ impl WriteHandler {
                     );
                 }
                 file.seek_to(header.offset)?;
+            } else {
+                need_flush = true;
             }
         }
 
@@ -262,6 +265,10 @@ impl WriteHandler {
             self.metrics.write_bytes.inc_by(msg.data_len() as i64);
             self.metrics.write_time_us.inc_by(used as i64);
             self.metrics.write_count.inc();
+        }
+
+        if need_flush {
+            file.flush()?;
         }
 
         Ok(msg.success())


curvine-worker/src/worker/handler/write_handler.rs
```diff
@@ -264,6 +267,10 @@ impl WriteHandler {
             self.metrics.write_count.inc();
         }
 
+        if need_flush {
+            file.flush()?;
```
szbr00, line 271: The new behavior — flush frames actually executing `file.flush()` and propagating errors — isn't covered by any test that would fail on the previous code. Note that a plain integration test can't distinguish old vs. new behavior with the default file layout, because `LocalFile::flush()` on an unbuffered `fs::File` is a no-op; observability requires either a spy `BlockIO` device that records `flush()` calls, or a fault-injection point on the storage flush to assert that a flush error reaches the client. Could you add a regression test along one of those lines, or at minimum a `worker_test.rs` single-block analogue of the batch flush-frame test to lock in the frame contract?

curvine-worker/src/worker/handler/write_handler.rs
```diff
@@ -242,6 +243,8 @@ impl WriteHandler {
                     );
                 }
                 file.seek_to(header.offset)?;
+            } else {
+                need_flush = true;
```
szbr00, line 247: Ignoring `header.offset` on flush frames is a subtle protocol decision: the client sends its current write position, which can legitimately equal `block_size` for a fully written block and would be rejected by the seek validation below. A one-line comment here (e.g. "flush frame offset is informational; never seek") would prevent a future change from treating it as a seek target and breaking flush-at-EOF.

szbr00 (commented):
One note for future debugging: on the default file layout the write device is an unbuffered `fs::File` (`LocalFile::write_region` writes through directly, `LocalFile::flush` is a no-op), so this fix does not change data visibility or durability for file-backed blocks — same-node short-circuit readers already see page-cache writes immediately. The behavioral effect is real for the bdev/SPDK layout (actual device flush via `spdk_flush`) and for consistent error propagation. It may be worth softening the "short-circuit readers could miss recently written bytes" rationale in the description accordingly.

PR #1676, opened 2026-09-08
ci: run focused POSIX smoke in existing build job

gongxun0928:
# Summary

Run only the focused 53-case Curvine POSIX/filesystem-semantics suite for pull requests inside the existing Compile and Smoke Test build job. The test reuses the `curvine-server` and `curvine-fuse` debug binaries already produced by that job and does not upload artifacts.

# Issue Describe / Design

No linked issue.

- Reuse the existing build output instead of running a second full `make build`.
- Assemble only the runtime layout needed to start master, worker, and FUSE.
- Run in the published `curvine-tests:rocky9` image, which already contains pinned LTP 20250930.
- Pin cvtest to merge commit `5d7ea3052e7a812289d1a8e57aeddda0cb7af425`.
- Read cvtest's `suites/ci-fast.txt` manifest and require its only active entry to be `ltp:posix-cv-smoke`.
- Do not run generic `smoketest`, `fs-cv-smoke`, full `syscalls-cv`, or pressure suites in regular PR CI.
- Keep results only on the ephemeral runner and print failure diagnostics to the job log; no Actions artifact storage is used.

# Changes

| `.github/workflows/build.yml` | Add PR-only focused POSIX smoke after the existing build, with privileged FUSE access | Pull request builds validate the 53 cases from `ltp:posix-cv-smoke`; no separate LTP jobs are generated |
| `curvine-tests/README.md` | Document the pinned cvtest fast manifest and excluded suites | Documentation matches the single-job, focused POSIX design |

# Test verified

| `git diff --check` | PASS | No whitespace errors in the final squashed diff |
| `cargo fmt --check` | PASS | Formatting is clean |
| Workflow YAML parse | PASS | `.github/workflows/build.yml` parses successfully |
| Local cvtest validation | PASS | `ltp:posix-cv-smoke` passed 53/53 on Rocky Linux 9 + LTP 20250930 + Curvine FUSE in 16 seconds |
| [GitHub Actions PR run](https://github.com/CurvineIO/curvine/actions/runs/34337112339) | PASS | `ltp:posix-cv-smoke` passed 53/53 in 18 seconds; complete build job finished in 5m31s |

# Dependencies

- Related PRs: #1675, [CurvineIO/cvtest#2](https://github.com/CurvineIO/cvtest/pull/2)
- Related issues: None
- Blocks / blocked by: Requires `ghcr.io/curvineio/curvine-tests:rocky9` from #1675 and pinned cvtest commit `5d7ea3052e7a812289d1a8e57aeddda0cb7af425`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,9 @@ env:
   CARGO_TERM_COLOR: always
   # Tune build parallelism: 4 uses more CPU on standard 4-vCPU runner; set to 2 if OOM
   CARGO_BUILD_JOBS: 4
+  CVTEST_REPO: CurvineIO/cvtest
+  CVTEST_REF: 5d7ea3052e7a812289d1a8e57aeddda0cb7af425
+  MOUNT_PATH: /curvine-fuse
 
 jobs:
   # Speed note: slowness is from runner CPU/RAM (standard = 4 vCPU, ~16GB), not container size.
@@ -37,9 +40,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     needs: cleanup
-    # curvine-tests image extends curvine-compile:rocky9 and pre-installs Flask, werkzeug, pip, jq, fio (see curvine-tests/Dockerfile)
+    timeout-minutes: 90
+    # curvine-tests extends curvine-compile:rocky9 and includes the build tools,
+    # FUSE userspace packages, and pinned LTP used by the PR POSIX smoke test.
     container:
       image: ghcr.io/curvineio/curvine-tests:rocky9
+      options: --privileged --device /dev/fuse
     
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


gongxun0928 (commented):
- Portal `ltp_test.py` default suites previously included `fs-cv-smoke`, which only exists as a cvtest-curated runtest file and is not installed by the stock LTP tree

.github/workflows/e2e-ltp.yml
```diff
@@ -0,0 +1,284 @@
...
+          ./bin/curvine-master.sh restart
+          sleep 5
+          ./bin/curvine-worker.sh restart
+          sleep 5
+          sudo ./bin/curvine-fuse.sh restart -d
+          python3 - <<'EOF'
+          import os, time, sys
+          p = os.path.join(os.environ["MOUNT_PATH"], ".ci-ready")
+          payload = b"ci" * 8192
+          deadline = time.time() + 120
```
szbr00: Required: The FAST_SUITES env var is set to `ltp:smoketest,ltp:fs-cv-smoke` (comma-separated), but the shell for-loop iterates over whitespace-delimited words. The entire string `"ltp:smoketest,ltp:fs-cv-smoke"` is passed as a single `--suite` argument to `cvtest run`. If cvtest does not internally parse comma-separated suite names, `ltp:fs-cv-smoke` silently never runs.

The downstream `name="${suite/:/-}"` also produces a single artifact filename containing a comma rather than two separate files, which confirms the author expected iteration.

Fix options:
- Change `FAST_SUITES` to space-separated values (`ltp:smoketest ltp:fs-cv-smoke`) so the for-loop iterates correctly
- Or split on commas explicitly with `IFS=,`
- Or invoke `cvtest run` twice with one `--suite` each
- Or document that `cvtest run --suite` accepts comma-separated input if that is the intended design

.github/workflows/e2e-ltp.yml
```diff
@@ -0,0 +1,284 @@
...
+          rustup default stable
+          rustc --version
+
+      - name: Build curvine distribution
+        run: make build
+
+      - name: Install pinned LTP
+        run: ./cvtest/tools/ltp/install-ltp.sh /opt/ltp
+
+      - name: Start cluster and mount FUSE
```
szbr00: Consider: This workflow runs directly on ubuntu-latest, not inside the `curvine-tests:rocky9` container used by `build.yml` and `daily-ut-coverage.yml`. GitHub-hosted ubuntu runners do not load the FUSE kernel module or install FUSE userspace tools by default. Adding a step before "Start cluster and mount FUSE" to install FUSE would make the fast lane more likely to succeed:

```yaml
- name: Install FUSE
  run: sudo apt-get update && sudo apt-get install -y fuse3 && sudo modprobe fuse
```

(The PR body says this is not a required status check until proven on hosted runners, so this is advisory.)

szbr00 (commented):
Consider: The checkout-cvtest → setup-rust → build → install-LTP → start-cluster preamble is duplicated verbatim across all three jobs (~48 lines × 3). Extracting this into a composite action or a reusable workflow would reduce the maintenance surface as lanes grow or setup steps diverge (e.g., adding FUSE install, changing the build command). The duplication already introduced a subtle inconsistency: the "Run cvtest fast lane" step uses `${{ env.FAST_SUITES }}` (line 111) while the syscalls and pressure variants reference their own env vars — any future change to the cluster startup needs to be applied in three places.

.github/workflows/e2e-ltp.yml
gongxun0928: Thanks. Addressed in 1dfba31b: the standalone multi-lane workflow and comma-separated FAST_SUITES loop were removed. The existing build job now reads the pinned cvtest ci-fast.txt manifest, requires its only active entry to be ltp:posix-cv-smoke, and runs that suite once.

.github/workflows/e2e-ltp.yml
gongxun0928: Thanks. Addressed in 1dfba31b: the POSIX smoke now runs inside the existing build job using ghcr.io/curvineio/curvine-tests:rocky9 with --privileged and /dev/fuse exposed. The hosted CI run mounted Curvine FUSE successfully and passed all 53 cases: https://github.com/CurvineIO/curvine/actions/runs/34337112339

gongxun0928:
Review summary follow-up:

- The duplicated three-job setup is no longer present. Commit 1dfba31b removed the standalone multi-lane workflow and keeps a single POSIX smoke path in the existing build job.
- The final PR diff no longer changes curvine-tests/regression/tests/ltp_test.py. The existing Portal cv-fs default predates this PR and remains outside this focused CI change; PR CI installs and runs the curated posix-cv-smoke suite through pinned cvtest instead.
- Hosted CI passed: 53/53 POSIX cases and all required checks are green.

No further code change is needed for the review comments on this PR.

szbr00 (approved):

PR #1682, opened 2026-09-09
fix(worker): abort block on deterministic finalize prepare failure (#1681)

### LINKED ISSUE #1681: [BUG]: Deterministic finalize prepare failure strands block in Writing, leaking space and hanging resident clients
**Describe the bug**

`BlockStore::finalize_block` treats a deterministic prepare-stage failure the same as a transient publish-stage failure: both are rolled back to `Writing` via `rollback_file_finalize`.

A prepare failure happens when the committed length reported by the client cannot be materialized from staging - e.g. after a worker restart, a rewrite copies the larger committed generation into staging, then a resident client re-commits a shorter length. `prepare_finalize` is grow-only, so it rejects the shorter length with `length mismatch` at `crates/adapters/curvine-storage-local/src/layout/file_layout.rs:117`.

`rollback_file_finalize` only restores the block state to `Writing`; it does NOT discard the stale staging copy, does NOT clear `committed_rewrites`, and does NOT release the reserved space. Because `Writing` is active but not final, a later re-open skips `prepare_write`'s copy and reuses the stale staging, and re-finalize fails identically. The block is permanently stranded, and each stuck block leaks ~2x its reserved space (staging + committed). Once the worker runs out of space, every later write fails and the resident SDK appears to hang.

**To Reproduce**

1. Run a worker with the file layout (`FileLayout`, `preserves_committed_on_write = true`).
2. Finalize a block (e.g. 50 bytes committed).
3. Reopen the block for rewrite - staging is copied from the committed generation.
4. Re-commit a shorter length (e.g. 20 bytes) and call finalize.
5. Finalize fails with `length mismatch, expected: 20, actual: 50`; the block is rolled back to `Writing`; staging and its reserved space remain.
6. Re-open + re-finalize loops forever with no progress, and reserved space keeps leaking until the worker runs out of capacity.

Production symptom (after a worker restart), the resident Java RSS shuffle SDK hangs on `close`:

```
[curvine-worker] ERROR: Block 1595166976638978 length mismatch, expected: 12640, actual: 80347136 (crates/adapters/curvine-storage-local/src/layout/file_layout.rs:117)
```

**Expected behavior**

The `length mismatch` error is correct and must still be returned to the client - the worker is faithfully detecting an inconsistent committed length. But a deterministic prepare failure must not strand the block: the worker should discard the staging copy, restore the last committed generation (or drop a first-write block), and reclaim the reserved space, so the block returns to a clean, readable state with no space leak.

A publish-stage failure (transient, at the atomic rename) should keep rolling back to `Writing` so a retry reuses the same staging without re-copying.

**Screenshots**

N/A

 - OS: Linux (production worker)
 - Curvine version/branch: main
 - Layout: file-backed (`FileLayout`, `preserves_committed_on_write = true`)

**Additional context**

Fix: in `finalize_block`, on a prepare failure call `abort_block` (discard staging, restore the committed generation, reclaim space) before returning the original error; keep `rollback_file_finalize` for publish failures.

Regression test: `failed_rewrite_finalize_with_shorter_length_restores_committed_block`.

bigbigxu:
# Summary

On a deterministic finalize prepare failure (client re-commits a length that cannot be materialized from staging), `finalize_block` now aborts the block instead of rolling it back to `Writing`, preventing it from being permanently stranded with a leaked staging copy that exhausted worker capacity and hung resident clients.

# Issue Describe / Design

Closes #1681

**Root cause**

`finalize_block` treated prepare-stage and publish-stage failures identically, rolling both back to `Writing` via `rollback_file_finalize`. That rollback only restores the block state; it does not discard the stale staging copy, clear `committed_rewrites`, or release the reserved space.

A prepare failure is deterministic: after a worker restart a rewrite copies the larger committed generation into staging, then a resident client re-commits a shorter length, and grow-only `prepare_finalize` rejects it with `length mismatch` (`crates/adapters/curvine-storage-local/src/layout/file_layout.rs:117`). Because `Writing` is active but not final, a later re-open skips `prepare_write`'s copy and reuses the stale staging, and re-finalize fails identically - an infinite loop that leaks ~2x reserved space per stuck block until the worker runs out of capacity and every later write fails. The resident Java RSS SDK surfaces this as a hang on `close`.

**Reproduction**

Finalize a 50B block -> reopen for rewrite (staging copied from committed) -> re-commit 20B -> finalize fails at the prepare length check -> block rolled back to `Writing` with staging + reservation left behind -> re-open/re-finalize loops forever.

**Fix approach**

Distinguish the two stages:

- prepare failure (deterministic) -> `abort_block`: discard staging, restore the last committed generation (or drop a first-write block), reclaim the reserved space, then return the original `length mismatch` error. The error still propagates; only the stranded state is removed.
- publish failure (transient, at the atomic rename) -> keep `rollback_file_finalize` so a retry reuses the same staging without re-copying.

# Changes

| `curvine-worker/src/worker/block/block_store.rs` | `finalize_block`: on prepare failure call `state.abort_block(block)` and return the original error; keep `rollback_file_finalize` for publish failure. Add regression test. | Prepare failures now restore the committed generation and reclaim space instead of stranding the block in `Writing`. Publish-failure rollback path and the returned error are unchanged. |

# Test verified

| `cargo test -p curvine-worker finalize` (Linux) | PASS | New `failed_rewrite_finalize_with_shorter_length_restores_committed_block` and existing `failed_file_finalize_restores_writing_state` both pass |

Formatting/lint (`make format`) not run locally (Windows dev environment); changes follow the surrounding rustfmt style.

# Dependencies

- Related PRs: None
- Related issues: #1681
- Blocks / blocked by: None

### DIFF
--- a/curvine-worker/src/worker/block/block_store.rs
+++ b/curvine-worker/src/worker/block/block_store.rs
@@ -18,7 +18,7 @@ use crate::worker::storage::{
 };
 use crate::worker::Worker;
 use curvine_config::ClusterConf;
-use curvine_core_error::CommonResult;
+use curvine_core_error::{CommonError, CommonResult};
 use curvine_model::{ExtendedBlock, StorageInfo};
 use parking_lot::{Mutex, MutexGuard};
 use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
@@ -168,27 +168,64 @@ impl BlockStore {
         let started = Instant::now();
         let plan = reservation.prepare(block.len);
         self.observe_file_layout_operation("finalize", started.elapsed());
-        let result = match plan {
-            Ok(plan) => self.with_dataset_write("finalize", |state| {
-                state.publish_file_finalize(&reservation, plan)
-            }),
-            Err(error) => Err(error),
+        let plan = match plan {
+            Ok(plan) => plan,
+            Err(error) => {
+                if error.to_string().contains("length mismatch") {
+                    // Deterministic mismatch: retrying can never succeed, so
+                    // abort the block instead of looping forever.
+                    if let Err(abort) =
+                        self.with_dataset_write("finalize", |state| state.abort_block(block))
+                    {
+                        log::error!(
+                            "failed to abort block {} after prepare error {}: {}; rollback to Writing",
+                            block.id,
+                            error,
+                            abort
+                        );
+                        self.rollback_finalize_reservation(block.id, &abort, |state| {
+                            state.rollback_file_finalize(&reservation)
+                        });
+                    }
+                } else {
+                    // Transient failure: release the reservation so the block
+                    // returns to Writing and finalize can be retried.
+                    self.rollback_finalize_reservation(block.id, &error, |state| {
+                        state.rollback_file_finalize(&reservation)
+                    });
+                }
+                return Err(error);
+            }
         };
-        match result {
+
+        match self.with_dataset_write("finalize", |state| {
+            state.publish_file_finalize(&reservation, plan)
+        }) {
             Ok(meta) => Ok(meta),
             Err(error) => {
-                if let Err(rollback) = self.with_dataset_write("finalize", |state| {
+                self.rollback_finalize_reservation(block.id, &error, |state| {
                     state.rollback_file_finalize(&reservation)
-                }) {
-                    log::error!(
-                        "failed to roll back block {} finalization after {}: {}",
-                        block.id,
-                        error,
-                        rollback
-                    );
-                }
+                });
                 Err(error)
             }
+        }
+    }
+
+    /// Roll back a finalize reservation after a failed prepare or publish.
+    /// Failures are logged while the caller keeps the original error.
+    fn rollback_finalize_reservation(
+        &self,
+        block_id: i64,
+        reason: &CommonError,
+        rollback: impl FnOnce(&mut BlockDataset) -> CommonResult<()>,
+    ) {
+        if let Err(failed) = self.with_dataset_write("finalize", rollback) {
+            log::error!(
+                "failed to roll back block {} finalization after {}: {}",
+                block_id,
+                reason,
+                failed
+            );
         }
     }
 
@@ -562,6 +599,65 @@ mod tests {
 
         FileUtils::delete_path(&active_path, true)?;
         assert!(store.finalize_block(&block)?.is_final());
+        Ok(())
+    }
+
+    #[test]
+    fn failed_rewrite_finalize_with_shorter_length_restores_committed_block() -> CommonResult<()> {
+        let store = create_store_with_capacity("rewrite-shrink-abort", "16MB")?;
+        let mut block = ExtendedBlock::with_mem(1, "50B")?;
+        let finalized = finalize_block(&store, &block)?;
+        assert!(finalized.is_final());
+        let available = store.read()?.available();
+
+        let rewriting = store.open_block(&block)?;
+        assert_eq!(rewriting.state(), &BlockState::Writing);
+
+        block.len = 20;
+        let error = store.finalize_block(&block).unwrap_err();
+        assert!(
+            error.to_string().contains("length mismatch"),
+            "expected a length mismatch prepare failure, got: {error}"
+        );
+
+        let restored = store.get_block(block.id)?;
+        assert!(restored.is_final());
+        assert_eq!(restored.len(), 50);
+        let active_path = store
+            .short_circuit(&restored)?
+            .expect("file layout must expose a local path");
+        assert_eq!(std::fs::metadata(&active_path)?.len(), 50);
+
+        let (_, dir) = store.read()?.layout_for(&restored)?;
+        let staging = BlockMeta::new(block.id, block.len, &dir);
+        let staging_path = FileLayout::block_path(&dir, &staging)?;
+        assert!(!staging_path.exists());
+        assert_eq!(store.read()?.available(), available);
+        Ok(())
+    }
+
+    #[test]
+    fn failed_first_write_finalize_with_length_mismatch_aborts_block() -> CommonResult<()> {
+        let store = create_store_with_capacity("first-write-abort", "16MB")?;
+        let available = store.read()?.available();
+
+        let block = ExtendedBlock::with_mem(1, "50B")?;
+        let writing = store.open_block(&block)?;
+        assert_eq!(writing.state(), &BlockState::Writing);
+        let staging_path = store
+            .short_circuit(&writing)?
+            .expect("file layout must expose a local path");
+        write_block(&staging_path, 30)?;
+
+        let error = store.finalize_block(&block).unwrap_err();
+        assert!(
+            error.to_string().contains("length mismatch"),
+            "expected a length mismatch prepare failure, got: {error}"
+        );
+
+        assert!(store.get_block(block.id).is_err());
+        assert!(!std::path::Path::new(&staging_path).exists());
+        assert_eq!(store.read()?.available(), available);
         Ok(())
     }
 


curvine-worker/src/worker/block/block_store.rs
```diff
@@ -168,13 +168,40 @@ impl BlockStore {
         let started = Instant::now();
         let plan = reservation.prepare(block.len);
         self.observe_file_layout_operation("finalize", started.elapsed());
-        let result = match plan {
-            Ok(plan) => self.with_dataset_write("finalize", |state| {
-                state.publish_file_finalize(&reservation, plan)
-            }),
-            Err(error) => Err(error),
+        // A prepare failure is deterministic: the committed length reported by
```
szbr00: This rationale block runs 11 lines and largely restates the PR description and commit message; the rest of this file keeps comments to 2-4 lines. Consider condensing to the essential point: a prepare failure cannot be materialized from staging, rolling back to Writing would strand the block, so abort restores the committed generation and reclaims the reserved space.

Minor wording note: "A prepare failure is deterministic" is slightly broad. `prepare_finalize` can also fail on transient I/O (`metadata()` / `set_len`). The abort is still safe in that case because the committed generation is preserved, so the comment could avoid asserting determinism for every possible prepare error.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -565,6 +592,46 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn failed_rewrite_finalize_with_shorter_length_restores_committed_block() -> CommonResult<()> {
+        // After a worker restart a resident client reopens a committed block for
```
szbr00: This 6-line header comment exceeds the 3-line convention used elsewhere in this file. The scenario is worth a brief note, but the detailed root-cause narrative already lives in the PR description and the inline comment in `finalize_block`. Consider trimming it to roughly 3 lines.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -565,6 +592,46 @@ mod tests {
...
+        assert!(finalized.is_final());
+        let available = store.read()?.available();
+
+        // Reopen for rewrite: staging is copied from the 50B committed generation.
+        let rewriting = store.open_block(&block)?;
+        assert_eq!(rewriting.state(), &BlockState::Writing);
+
+        // Re-commit a shorter length; finalize fails at the prepare length check.
+        block.len = 20;
+        assert!(store.finalize_block(&block).is_err());
```
szbr00: Minor: this asserts only `is_err()`. Capturing the error and checking that it is the expected `length mismatch` would confirm the test exercises the deterministic prepare-stage path rather than failing for an unrelated reason.

bigbigxu: Thanks. Fixed in 853a47f20: the test now asserts the error message contains `length mismatch`, confirming it exercises the prepare-stage path rather than failing for an unrelated reason.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -168,13 +168,40 @@ impl BlockStore {
...
+                {
+                    log::error!(
+                        "failed to abort block {} after finalize prepare error {}: {}",
+                        block.id,
+                        error,
+                        abort
+                    );
+                }
+                return Err(error);
+            }
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 853a47f20: on the prepare-error path, if `abort_block` fails we now fall back to `rollback_file_finalize(&reservation)`, so the block is restored to `Writing` instead of being stranded in `Finalizing` (not active, cannot be reopened). Both the abort and rollback failures are logged.

curvine-worker/src/worker/block/block_store.rs
bigbigxu: Thanks. Fixed in 853a47f20: trimmed the test header comment to 3 lines; the detailed root-cause narrative stays in the PR description and the `finalize_block` comment.

curvine-worker/src/worker/block/block_store.rs
bigbigxu: Thanks. Fixed in 853a47f20: condensed the rationale to 4 lines and reworded it to avoid asserting every prepare failure is deterministic - it now notes the committed length may fail to materialize from staging, or staging I/O (`metadata()`/`set_len`) may fail, and abort stays safe either way because the committed generation is preserved.

curvine-worker/src/worker/block/block_store.rs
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in 853a47f20: the test now captures the error with `unwrap_err()` and asserts the message contains the expected length mismatch, so it verifies the intended prepare-stage failure instead of passing on an unrelated error.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -168,13 +168,40 @@ impl BlockStore {
...
+        // and rolling back to Writing would strand the block: re-open reuses the
+        // stale staging copy (Writing is active but not final, so prepare_write
+        // skips the copy) and re-finalize fails identically, leaking the reserved
+        // rewrite copy until the worker runs out of space and every later write
+        // fails. Abort instead: discard staging and restore the last committed
+        // generation (or drop a first-write block), so the block returns to a
+        // clean, readable state while the error still propagates to the client.
+        let plan = match plan {
+            Ok(plan) => plan,
+            Err(error) => {
```
szbr00, line 173: The abort now covers every prepare error, but only the length mismatch is truly deterministic. `prepare_finalize` can also fail on transient I/O — notably the `set_len` grow for sparse-tail rewrites can hit ENOSPC, and since the commit-time length is only bounded by the open-time `block_size` (not the reserved bytes), that grow can exceed the open-time reservation. On such a failure the entire rewrite staging (potentially GBs of newly written data) is discarded and the old generation restored, where the previous rollback-to-`Writing` kept staging for a cheap retry — the same rationale this PR uses to keep rollback for transient publish failures. Consider limiting the abort to the deterministic mismatch (e.g., by distinguishing the error) or documenting the accepted trade-off.

curvine-worker/src/worker/block/block_store.rs
```diff
@@ -565,6 +592,46 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn failed_rewrite_finalize_with_shorter_length_restores_committed_block() -> CommonResult<()> {
```
szbr00, line 606: This test covers the rewrite branch of the abort (restore committed generation), but the other half of the fix — dropping a first-write block via `remove_block_by_id` — changes behavior (previously rolled back to `Writing` and kept the block) and has no coverage. Consider a companion test: open a first-write block, write staging, finalize with a mismatched length, then assert the block is removed and `available()` is restored.

curvine-worker/src/worker/block/block_store.rs
bigbigxu: Good catch — fixed in 4f3ead54e by distinguishing the error rather than just documenting the trade-off.

The length check in `prepare_finalize` now returns a typed `FsError::BlockLengthMismatch` (errno 37) instead of a string box, mirroring how `NoAvailableWorker` avoids string matching. `finalize_block` downcasts the prepare error and routes on it:

- `BlockLengthMismatch` is the only truly deterministic failure, so it aborts (restoring the committed generation, or dropping a first write).
- Any other prepare error — the transient staging I/O you flagged, notably a `set_len` ENOSPC on the sparse-tail grow — falls through to `rollback_file_finalize`, which keeps the staging copy so a retry re-finalizes it cheaply. This is now symmetric with how transient publish failures are already handled.

So a transient grow failure no longer discards GBs of freshly written staging and restores the old generation. The typed variant also round-trips across RPC encode/decode without collapsing into `Common`.

curvine-worker/src/worker/block/block_store.rs
bigbigxu: Added in 4f3ead54e as `failed_first_write_finalize_with_length_mismatch_aborts_block`, the companion to the rewrite test.

It opens a first-write block (no committed generation), writes 30 bytes into staging against a committed length of 50, and finalizes. It then asserts:

- the returned error is the deterministic `length mismatch`, not an unrelated failure mode;
- `get_block` fails — the block was dropped via `remove_block_by_id`, not rolled back to `Writing`;
- the staging path no longer exists;
- `available()` is restored to the pre-open baseline, confirming the reserved space was reclaimed rather than leaked.

curvine-worker/src/worker/block/block_store.rs
szbr00: Heads-up: the current head (`9a3a9889`) does not contain the typed error described above — `FsError::BlockLengthMismatch` exists nowhere in this PR, `prepare_finalize` still returns an `err_box!` string, and only `block_store.rs` is modified. It looks like the error-crate / `file_layout.rs` part was lost in the force-push/squash (`4f3ead54e` is no longer in the history). What's on the branch now routes on `error.to_string().contains("length mismatch")` around line 174.

That is functionally correct today, but it dispatches behavior on display text produced in another crate — exactly what the typed variant was meant to avoid. If the message in `file_layout.rs` is ever reworded, the deterministic mismatch would silently route to the transient-rollback branch and revive the #1681 stranded-block loop (the new tests would catch a naive reword only because they assert the message too; the coupling is incidental, not structural). If those changes were lost, please re-push them; if they were dropped deliberately, it's still worth restoring them — adding the `curvine-error` dependency to `curvine-storage-local` is cycle-free.

szbr00 (approved):

PR #1685, opened 2026-09-11
fix(build): unblock workspace build on Rust 1.98 (#1684)

### LINKED ISSUE #1684: [BUG]: Workspace fails to build on Rust 1.98
## Describe the bug

Building the Curvine workspace with Rust 1.98 fails with clippy and query depth errors in CLI, bench, and SDK binding crates.

## To Reproduce

1. Use Rust 1.98.
2. Run cargo build --workspace or maturin release for Python SDK.

## Expected behavior

Clean build with -D warnings.

## OS Version

Linux; Curvine main; Rust 1.98.

bigbigxu:
# Summary

Restore a clean Rust 1.98 workspace build by updating the pinned toolchain and workspace deps, raising recursion limits where deep UnifiedFileSystem async types overflow layout queries, and fixing new clippy denials.

# Issue Describe / Design

Closes #1684

- Root cause: Rust 1.98 toolchain bump exposes stricter clippy lints and deeper async type layout instantiation in SDK/CLI crates.
- Fix: Align 
ust-toolchain.toml and key deps; add crate-level 
ecursion_limit; small targeted clippy fixes (no behavior change intended).

# Changes

| 
ust-toolchain.toml, Cargo.toml | Pin 1.98; bump tokio/bytes/libc | Build/lockfile alignment |
| curvine-cli, curvine-bench, SDK crates, curvine-libsdk | #![recursion_limit] | Compile-time only |
| load_task_runner.rs | if let Some for transfer report | None |
| write_cache_writer.rs, fuse tests, fs-api/client debug | clippy fixes | None |

# Test verified

| cargo check -p curvine-cli -p curvine-libsdk-python --features extension-module | PASS | local |
| Cherry-pick onto open/main | PASS | resolved load_task_runner conflict (keep 	ransfer_client) |

# Dependencies

- Related issues: #1684
- Related PRs: None

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,12 +158,12 @@ curvine-web = { path = "curvine-web" }
 
 # The most important library start
 
-libc = "0.2.189"
-tokio = { version = "1.53.1", features = ["full"] }
+libc = "0.2.167"
+tokio = { version = "1.42.0", features = ["full"] }
 futures = "0.3.31"
 async-stream = "0.3.6"
-tokio-util = { version = "0.7.19", features = ["codec"] }
-bytes = "1.12.0"
+tokio-util = { version = "0.7.12", features = ["codec"] }
+bytes = "1.9.0"
 serde = { version = "1.0.215", features = ["derive"] }
 rocksdb = { version = "0.22.0", default-features = false, features = ["lz4"] }
 raft = { version = "0.7.0", features = ["prost-codec"], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,12 +158,12 @@ curvine-web = { path = "curvine-web" }
 
 # The most important library start
 
-libc = "0.2.167"
-tokio = { version = "1.42.0", features = ["full"] }
+libc = "0.2.189"
+tokio = { version = "1.53.1", features = ["full"] }
 futures = "0.3.31"
 async-stream = "0.3.6"
-tokio-util = { version = "0.7.12", features = ["codec"] }
-bytes = "1.9.0"
+tokio-util = { version = "0.7.19", features = ["codec"] }
+bytes = "1.12.0"
 serde = { version = "1.0.215", features = ["derive"] }
 rocksdb = { version = "0.22.0", default-features = false, features = ["lz4"] }
 raft = { version = "0.7.0", features = ["prost-codec"], default-features = false }
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -1014,7 +1014,7 @@ async fn wait_transfer_complete0(
     loop {
         let status = match client.status(job_id).await {
             Ok(status) => status,
-            Err(_err @ FsError::JobNotFound(_)) if !fail_if_not_found => {
+            Err(FsError::JobNotFound(_)) if !fail_if_not_found => {
                 time::sleep(Duration::from_millis(
                     client_conf.sync_check_interval_min_ms,
                 ))
--- a/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
+++ b/crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
@@ -1014,7 +1014,7 @@ async fn wait_transfer_complete0(
     loop {
         let status = match client.status(job_id).await {
             Ok(status) => status,
-            Err(err @ FsError::JobNotFound(_)) if !fail_if_not_found => {
+            Err(_err @ FsError::JobNotFound(_)) if !fail_if_not_found => {
                 time::sleep(Duration::from_millis(
                     client_conf.sync_check_interval_min_ms,
                 ))
--- a/crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
+++ b/crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
@@ -61,7 +61,10 @@ pub struct WriteCacheWriter {
 }
 
 impl WriteCacheWriter {
-    #[allow(clippy::result_large_err)]
+    #[expect(
+        clippy::result_large_err,
+        reason = "returns UnifiedWriter on error so the caller can fall back to uncached write"
+    )]
     pub async fn new(
         primary: UnifiedWriter,
         cv: CurvineFileSystem,
--- a/crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
+++ b/crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
@@ -61,6 +61,7 @@ pub struct WriteCacheWriter {
 }
 
 impl WriteCacheWriter {
+    #[allow(clippy::result_large_err)]
     pub async fn new(
         primary: UnifiedWriter,
         cv: CurvineFileSystem,
--- a/crates/sdk/curvine-sdk-core/src/lib.rs
+++ b/crates/sdk/curvine-sdk-core/src/lib.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![recursion_limit = "512"]
+
 mod core;
 
 mod lib_filesystem;
--- a/crates/sdk/curvine-sdk-core/src/lib.rs
+++ b/crates/sdk/curvine-sdk-core/src/lib.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// LibFilesystem::create/open monomorphize deep UnifiedFileSystem futures here.
 #![recursion_limit = "512"]
 
 mod core;
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -16,5 +16,5 @@
 # under the License.
 
 [toolchain]
-channel = "1.92.0"
+channel = "1.98"
 components = ["rustfmt", "clippy", "rust-analyzer"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,7 +25,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.17",
+ "getrandom 0.2.16",
  "once_cell",
  "version_check",
 ]
@@ -38,7 +38,7 @@ checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
  "const-random",
- "getrandom 0.3.4",
+ "getrandom 0.3.3",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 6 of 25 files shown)


rust-toolchain.toml
```diff
@@ -16,5 +16,5 @@
 # under the License.
 
 [toolchain]
-channel = "1.92.0"
+channel = "1.98"
```
szbr00, line 19: This moves the repo to 1.98, but the compile images still bake in 1.92.0:

- `curvine-docker/compile/Dockerfile_rocky9:70,72`
- `curvine-docker/compile/Dockerfile_rocky9_cached:52,54`
- `curvine-docker/compile/Dockerfile_centos7:99,101`
- `curvine-docker/compile/Dockerfile_ubuntu22:75,77`
- `curvine-docker/compile/Dockerfile_ubuntu24:75,77`

`rust-toolchain.toml` overrides the `rustup default stable` that `build.yml` runs, so the pre-installed toolchain is no longer the one used. The CI log on this PR shows the result:

```
info: note that the toolchain '1.98-x86_64-unknown-linux-gnu' is currently in use (overridden by '/__w/curvine/curvine/rust-toolchain.toml')
warn: the missing active toolchain `1.98-x86_64-unknown-linux-gnu` has been auto-installed
warn: this might cause rustup commands to take longer time to finish than expected
```

Every build job now downloads and installs a full toolchain at runtime, and any of these images built without network access to the rustup mirror will fail outright. Bumping `--default-toolchain` / `rustup default` in those five Dockerfiles as part of this change keeps CI on the pre-baked toolchain.

Separately, the pin became less precise: `channel = "1.92.0"` selected an exact patch release, while `"1.98"` floats to the newest 1.98.x (CI resolved 1.98.1). `"1.98.0"` preserves the previous reproducibility semantics.

Cargo.toml
```diff
@@ -158,12 +158,12 @@ curvine-web = { path = "curvine-web" }
 
 # The most important library start
 
-libc = "0.2.167"
-tokio = { version = "1.42.0", features = ["full"] }
+libc = "0.2.189"
+tokio = { version = "1.53.1", features = ["full"] }
```
szbr00: These are workspace *minimum* versions, so raising them is a stronger commitment than the build fix requires. Refreshing `Cargo.lock` alone would let 1.98 resolve newer tokio/bytes/libc while leaving consumers free to stay on the old floors; raising the floors forces every crate depending on curvine onto tokio >= 1.53.1, bytes >= 1.12.0, libc >= 0.2.189.

Is each of these actually required by 1.98, or were they refreshed opportunistically? tokio 1.42 -> 1.53 spans eleven minor releases. Narrowing to the bumps that fix a concrete compile error, and noting the reason, keeps a future bisect able to attribute a regression to the right one.

The lockfile churn points the same way. It is ~2900 lines and includes changes unrelated to these four crates: the entire `windows` / `wit-bindgen` / `wasmparser` subtree is gone, while `darling`, `frunk_derives`, `derive_utils`, `jni-sys-macros`, `quick-xml`, `winnow` and `toml_edit` appear. That looks like a full `cargo update` rather than a targeted one. `cargo update -p tokio -p bytes -p libc -p tokio-util` would keep the diff to the crates this PR is about.

For the record the lock itself is consistent with the new floors (it resolves bytes 1.12.1, libc 0.2.189, tokio 1.53.1, tokio-util 0.7.19), so this is about the scope of the change rather than a stale lock.

crates/sdk/curvine-sdk-core/src/lib.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![recursion_limit = "512"]
```
szbr00, line 16: With this, six crates carry a crate-level `recursion_limit`: 256 in `curvine-cli`, `curvine-bench` and the pre-existing `curvine-fuse`, 512 here and in `curvine-libsdk`, `curvine-libsdk-java` and `curvine-libsdk-python`. The attribute is per-crate, so every new crate that monomorphises the deep `UnifiedFileSystem` async types will overflow the default 128 and need its own, and 256 vs 512 reads as arbitrary without a note on what triggered it.

Two suggestions:

- a one-line comment at each site naming the type that overflows, so nobody "cleans up" the attribute later and re-breaks the build;
- a follow-up issue to box the deep future at the `UnifiedFileSystem` boundary, which removes the need to replicate the limit in every downstream crate instead of working around it per crate.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
```diff
@@ -61,6 +61,7 @@ pub struct WriteCacheWriter {
 }
 
 impl WriteCacheWriter {
+    #[allow(clippy::result_large_err)]
```
szbr00: A bare `#[allow]` drops the reason the large error is deliberate: `new` returns `(UnifiedWriter, FsError)` so the caller can fall back to an uncached write instead of losing the writer. `#[expect]` (stable since 1.81) keeps that documented and warns if the suppression ever becomes unnecessary:

```rust
#[expect(
    clippy::result_large_err,
    reason = "returns the UnifiedWriter so the caller can fall back to an uncached write"
)]
```

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
```diff
@@ -1014,7 +1014,7 @@ async fn wait_transfer_complete0(
     loop {
         let status = match client.status(job_id).await {
             Ok(status) => status,
-            Err(err @ FsError::JobNotFound(_)) if !fail_if_not_found => {
+            Err(_err @ FsError::JobNotFound(_)) if !fail_if_not_found => {
```
szbr00: The `@` binding is now unused, so it can be dropped rather than renamed with an underscore prefix:

```rust
Err(FsError::JobNotFound(_)) if !fail_if_not_found => {
```

Leaving `_err` in place invites a future reader to wonder whether the error was meant to be logged here.

szbr00 (commented):
The `build` CI job is red on this PR, so the 1.98 build isn't unblocked yet.

Clippy aborts in `curvine-rocksdb` on a lint that is new in 1.98:

```
error: manual checked division
  --> crates/adapters/curvine-rocksdb/src/rocksdb/db_engine.rs:496:26
    = note: `-D clippy::manual-checked-ops` implied by `-D warnings`
```

The site is untouched by this PR:

```rust
let denom = hit.saturating_add(miss);
let ppm = if denom == 0 {
    0
} else {
    hit.saturating_mul(1_000_000) / denom
};
```

The lint wants `checked_div`, which preserves the current semantics exactly:

```rust
let ppm = hit
    .saturating_mul(1_000_000)
    .checked_div(denom)
    .unwrap_or(0);
```

One caveat on scope: clippy failed while compiling `curvine-rocksdb`, so every crate downstream of it (`curvine-store`, `curvine-master`, `curvine-worker`, ...) was never linted. More 1.98 denials may be hiding behind this one. Worth running `cargo clippy --all-targets -- -D warnings --allow clippy::uninlined-format-args` on 1.98 locally until it is fully clean before re-pushing.

Cargo.toml
```diff
@@ -158,12 +158,12 @@ curvine-web = { path = "curvine-web" }
...
 futures = "0.3.31"
 async-stream = "0.3.6"
-tokio-util = { version = "0.7.12", features = ["codec"] }
-bytes = "1.9.0"
+tokio-util = { version = "0.7.19", features = ["codec"] }
```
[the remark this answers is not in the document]
bigbigxu: Thanks. This PR commits an updated Cargo.lock aligned with the workspace minimums (bytes/tokio/libc/tokio-util resolve at or above the new floors), so locked builds on the branch head should succeed.

rust-toolchain.toml
bigbigxu: Thanks. Fixed in c3e87b406: all curvine-docker/compile Dockerfiles now install rustup default 1.98, matching rust-toolchain.toml channel = "1.98". CI should use the pre-baked toolchain instead of auto-installing at job start. We intentionally use channel 1.98 (not a patch pin like 1.92.0) so 1.98.x patch updates stay aligned with rustup.

Cargo.toml
bigbigxu: Thanks for the scope note. The lock on this branch is consistent with the raised workspace floors for the 1.98 migration. We have not narrowed the lock diff yet; happy to revert the workspace minimum bumps and refresh only tokio/bytes/libc/tokio-util via a targeted cargo update in a follow-up commit if you want a smaller diff.

crates/sdk/curvine-sdk-core/src/lib.rs
bigbigxu: Thanks. We are keeping crate-level recursion_limit for now. Boxing deep UnifiedFileSystem futures at the SDK boundary would add heap allocation on sync call paths; we prefer the compile-time limit over that runtime cost. We can add a one-line comment at each recursion_limit site in a small follow-up if that helps.

crates/client/curvine-unified-fs/src/unified/write_cache_writer.rs
bigbigxu: Thanks. Addressed in c3e87b406: replaced allow with expect(clippy::result_large_err, reason = "returns UnifiedWriter on error so the caller can fall back to uncached write").

crates/client/curvine-unified-fs/src/unified/unified_filesystem.rs
bigbigxu: Thanks. Addressed in c3e87b406: use Err(FsError::JobNotFound(_)) without a binding.

bigbigxu:
## Follow-up on review summary (build / clippy)

Thanks for the detailed review.

- **curvine-rocksdb manual_checked_ops**: fixed in 1cbfafe3f (checked_div + unwrap_or(0)).
- **curvine-fuse plock / dcache clippy**: fixed in 68294e630.
- **curvine-master useless_borrows_in_formatting**: fixed in c3e87b406.
- **Compile Dockerfiles vs rust-toolchain**: fixed in c3e87b406 (all compile images on rustup 1.98, matching channel 1.98 in rust-toolchain.toml).

Please re-run CI on the latest head (c3e87b406). If anything else fails under 1.98 clippy, we will address in this PR.

Cargo.toml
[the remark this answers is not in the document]
bigbigxu: Thanks. This PR commits an updated Cargo.lock aligned with the workspace minimums (bytes/tokio/libc/tokio-util resolve at or above the new floors), so locked builds on the branch head should succeed.

rust-toolchain.toml
bigbigxu: Thanks. Fixed in c3e87b406: all curvine-docker/compile Dockerfiles now install rustup default 1.98, matching rust-toolchain.toml channel = "1.98". CI should use the pre-baked toolchain instead of auto-installing at job start. We intentionally use channel 1.98 (not a patch pin like 1.92.0) so 1.98.x patch updates stay aligned with rustup.

Cargo.toml
bigbigxu: Thanks for the scope note. The lock on this branch is consistent with the raised workspace floors for the 1.98 migration. We have not narrowed the lock diff yet; happy to revert the workspace minimum bumps and refresh only tokio/bytes/libc/tokio-util via a targeted cargo update in a follow-up commit if you want a smaller diff.

szbr00 (approved):

Issue #1666, opened 2026-09-07
[BUG]: resize/assign_worker by inode id journal and return the caller-supplied path without validating it against the resolved inode


Novohudonossor:
## Description

#1663 added an optional `inode_id` to `FileResizeRequest` / `AssignWorkerRequest` so a client can address a file whose path may have changed since it was opened. In that branch the caller-supplied path is never validated against the inode that gets resolved, and it is then forwarded to the journal and to the response.

`MasterFilesystem::resolve_file_inode` resolves purely by id and uses `path` only to build the error message:

```rust
Some(v) if v > 0 => match fs_dir.store.get_inode(v, None)? {
    Some(view) => Ok(InodePtr::from_owned(view)),
    None => err_ext!(FsError::file_not_found(path).ctx(format!("inode_id={}", v))),
},
```

That string travels on as `audit_path`:

- `FsDir::resize_inode(audit_path, ..)` → `journal_writer.log_complete_file(self, audit_path, ..)`
- `FsDir::assign_worker_inode(audit_path, ..)` → `journal_writer.log_add_block(self, audit_path, ..)`
- `inode.to_file_status(path)` when the `FileBlocks` response is built

Two places where it has an effect beyond a log line:

**1. The UFS export task is submitted for the stale path.** `ufs_loader::complete_file` does `Path::from_str(&e.path)`, looks up the mount for that path and calls `submit_export_task(&path, &mnt)`. With a stale path that resolves the wrong mount, or none at all, for a journal entry which otherwise describes the correct inode.

**2. The returned `FileStatus` contradicts itself.** `to_file_status` sets `path: path.to_owned()` from the caller's string while `name: self.name().to_owned()` comes from the resolved inode, so the two disagree exactly in the case the new `inode_id` field was introduced to handle.

To be explicit about what does *not* break, since I checked both before writing this:

- Master replay is safe. `journal_loader::complete_file` re-resolves with `resolve_file_inode(&fs_dir, &entry.path, Some(entry.file.id))`, so it cannot land on a different file; the stale path only reaches its `warn!`.
- `get_block_locs(path, fs_dir, file)` works off the already-resolved `file`; its `path` argument is used for error text only.

The path-based entry points are unaffected: `resize` and `assign_worker` delegate with `inode_id: None` and go through `resolve_path`, so there `audit_path` is the resolved path.

[...]

PR #1686, opened 2026-09-11
fix(master): resolve path by inode id for status and UFS export (#1666)

bigbigxu:
# Summary

Fix path consistency for by-inode-id operations without rewriting paths on the write hot path. `GetFileStatus` accepts optional `inode_id`, and UFS export resolves the live path via inode id.

# Issue Describe / Design

Related to #1666

- Root cause: by-id resize/assign/complete journaled and returned the caller-supplied (possibly stale) path. UFS export used that path for mount lookup.
- Fix approach:
  - Keep request path on write RPCs (no hot-path path rebuild).
  - Add optional `inode_id` to `GetFileStatusRequest` for by-id status resolution.
  - UFS `complete_file` resolves export path with `get_inode_path(inode_id)`, falling back to journaled path if the inode is missing.
  - Share `FsDir::get_inode_path` with TTL executor.

# Changes

| `crates/common/curvine-proto/proto/master.proto` | Add optional `inode_id` to `GetFileStatusRequest` | Backward compatible |
| `curvine-master/.../master_filesystem.rs` | `file_status_by_id` + tests | By-id status uses request path; inode identity from resolved inode |
| `curvine-master/.../fs_dir.rs` | Add `get_inode_path` | Path reconstruction helper |
| `curvine-master/.../ufs_loader.rs` | Resolve export path by inode id | UFS export uses live path after rename |
| `curvine-master/.../ttl_executor.rs` | Reuse `get_inode_path` | Behavior unchanged |
| `crates/.../fs_client.rs` | Client `file_status_by_id` | New API surface |

# Test verified

| `cargo test -p curvine-master --lib -- file_status_by_id` | PASS | by-id status + rename |
| `cargo test -p curvine-master --lib -- resolve_complete_entry_path` | PASS | UFS path by inode + fallback |

# Dependencies

- Related PRs: None
- Related issues: #1666
- Blocks / blocked by: None

### DIFF
--- a/crates/client/curvine-client-core/src/file/fs_client.rs
+++ b/crates/client/curvine-client-core/src/file/fs_client.rs
@@ -165,8 +165,17 @@ impl FsClient {
     }
 
     pub async fn file_status(&self, path: &Path) -> FsResult<FileStatus> {
+        self.file_status0(path, None).await
+    }
+
+    pub async fn file_status_by_id(&self, path: &Path, inode_id: i64) -> FsResult<FileStatus> {
+        self.file_status0(path, Some(inode_id)).await
+    }
+
+    async fn file_status0(&self, path: &Path, inode_id: Option<i64>) -> FsResult<FileStatus> {
         let header = GetFileStatusRequest {
             path: path.encode(),
+            inode_id,
         };
 
         let rep_header: GetFileStatusResponse = self.rpc(RpcCode::FileStatus, header).await?;
@@ -177,6 +186,7 @@ impl FsClient {
     pub async fn file_status_bytes(&self, path: &Path) -> FsResult<BytesMut> {
         let header = GetFileStatusRequest {
             path: path.encode(),
+            inode_id: None,
         };
         self.rpc_bytes(RpcCode::FileStatus, header).await
     }
--- a/curvine-master/src/master/fs/master_filesystem.rs
+++ b/curvine-master/src/master/fs/master_filesystem.rs
@@ -479,10 +479,18 @@ impl MasterFilesystem {
     }
 
     pub fn file_status<T: AsRef<str>>(&self, path: T) -> FsResult<FileStatus> {
+        self.file_status_by_id(path, None)
+    }
+
+    pub fn file_status_by_id<T: AsRef<str>>(
+        &self,
+        path: T,
+        inode_id: Option<i64>,
+    ) -> FsResult<FileStatus> {
+        let path = path.as_ref();
         let fs_dir = self.fs_dir.read();
-        let inp = Self::resolve_path(&fs_dir, path.as_ref())?;
-        let status = fs_dir.file_status(&inp)?;
-        Ok(status)
+        let inode = Self::resolve_file_inode(&fs_dir, path, inode_id)?;
+        Ok(inode.to_file_status(path)?)
     }
 
     pub fn exists<T: AsRef<str>>(&self, path: T) -> FsResult<bool> {
@@ -1857,6 +1865,102 @@ mod tests {
             "expected path in error, got: {}",
             err
         );
+    }
+
+    #[test]
+    fn get_inode_path_follows_parent_links() {
+        let fs = test_fs("get-inode-path");
+        let status = fs.create("/dir/file.log", true).unwrap();
+        let sync_fs_dir = fs.fs_dir();
+        let fs_dir = sync_fs_dir.read();
+        assert_eq!(fs_dir.get_inode_path(status.id).unwrap(), "/dir/file.log");
+    }
+
+    #[test]
+    fn get_inode_path_after_rename_returns_new_path() {
+        let fs = test_fs("get-inode-path-rename");
+        let status = fs.create("/dir/old.log", true).unwrap();
+        fs.rename("/dir/old.log", "/dir/new.log", RenameFlags::empty())
+            .unwrap();
+
+        let sync_fs_dir = fs.fs_dir();
+        let fs_dir = sync_fs_dir.read();
+        assert_eq!(fs_dir.get_inode_path(status.id).unwrap(), "/dir/new.log");
+    }
+
+    #[test]
+    fn file_status_by_id_after_rename_resolves_inode_keeps_request_path() {
+        // By-id status resolves the inode (id/name) but keeps the request path.
+        let fs = test_fs("file-status-by-id-stale-path");
+        let created = fs.create("/a/old.log", true).unwrap();
+        fs.rename("/a/old.log", "/a/new.log", RenameFlags::empty())
+            .unwrap();
+
+        let status = fs
+            .file_status_by_id("/a/old.log", Some(created.id))
+            .unwrap();
+        assert_eq!(status.id, created.id);
+        assert_eq!(status.path, "/a/old.log");
+        assert_eq!(status.name, "new.log");
+
+        // Path-based status still requires the live path.
+        let by_path = fs.file_status("/a/new.log").unwrap();
+        assert_eq!(by_path.id, created.id);
+        assert_eq!(by_path.path, "/a/new.log");
+
+        // Write path still journals/returns the request path (no hot-path rebuild).
+        let blocks = fs
+            .resize_by_id(
+                "/a/old.log",
+                Some(created.id),
+                FileAllocOpts::with_truncate(128),
+            )
+            .unwrap();
+        assert_eq!(blocks.status.id, created.id);
+        assert_eq!(blocks.status.path, "/a/old.log");
+        assert_eq!(blocks.status.name, "new.log");
+        assert_eq!(blocks.status.len, 128);
+    }
+
+    #[test]
+    fn file_status_by_id_missing_inode_returns_file_not_found() {
+        let fs = test_fs("file-status-missing-inode");
+        let missing_id = 9_999_999_i64;
+        let err = fs
+            .file_status_by_id("/missing.log", Some(missing_id))
+            .unwrap_err();
+        assert_file_not_found_roundtrip(&err);
+        assert!(
+            err.to_string()
+                .contains(&format!("inode_id={}", missing_id)),
+            "expected inode_id in error context, got: {}",
+            err
+        );
+    }
+
+    #[test]
+    fn complete_file_by_id_keeps_request_path() {
+        let fs = test_fs("complete-by-id-stale-path");
+        let created = fs.create("/b/old.log", true).unwrap();
+        fs.rename("/b/old.log", "/b/new.log", RenameFlags::empty())
+            .unwrap();
+
+        let blocks = fs
+            .complete_file(
+                "/b/old.log",
+                Some(created.id),
+                64,
+                vec![],
+                "test-client",
+                true,
+                None,
+            )
+            .unwrap()
+            .expect("only_flush should return FileBlocks");
+
+        assert_eq!(blocks.status.id, created.id);
+        assert_eq!(blocks.status.path, "/b/old.log");
+        assert_eq!(blocks.status.name, "new.log");
     }
 
     fn worker_with_status(
--- a/curvine-master/src/master/journal/ufs_loader.rs
+++ b/curvine-master/src/master/journal/ufs_loader.rs
@@ -15,7 +15,7 @@
 use crate::master::journal::{
     CompleteFileEntry, DeleteEntry, JournalEntry, MkdirEntry, RenameEntry,
 };
-use crate::master::JobManager;
+use crate::master::{JobManager, SyncFsDir};
 use curvine_config::JournalConf;
 use curvine_core_error::{err_box, CommonResult};
 use curvine_error::FsError;
@@ -31,11 +31,12 @@ use std::time::Duration;
 #[derive(Clone)]
 pub struct UfsLoader {
     job_manager: Arc<JobManager>,
+    fs_dir: SyncFsDir,
     copy_timeout: Duration,
 }
 
 impl UfsLoader {
-    pub fn new(job_manager: Arc<JobManager>, conf: &JournalConf) -> Self {
+    pub fn new(job_manager: Arc<JobManager>, fs_dir: SyncFsDir, conf: &JournalConf) -> Self {
         let copy_timeout = match DurationUnit::from_str(&conf.ufs_copy_timeout) {
             Ok(unit) => unit.as_duration(),
             Err(e) => {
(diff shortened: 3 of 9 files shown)


curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -49,10 +50,29 @@ impl UfsLoader {
...
         }
     }
 
+    pub fn get_real_path(&self, inode_id: i64) -> FsResult<Path> {
+        let path = self.fs_dir.read().get_inode_path(inode_id)?;
+        Ok(Path::from_str(path)?)
+    }
+
+    pub(crate) fn resolve_complete_entry_path(&self, e: &CompleteFileEntry) -> FsResult<Path> {
+        match self.get_real_path(e.file.id) {
```
szbr00: `get_inode_path` can only ever return one path per inode, but Curvine supports hard links: `FsDir::link` (fs_dir.rs:1263) adds a `FileEntry` edge in the destination parent and bumps `nlink` without touching the original inode's `parent_id`/`name`. So for a linked file this walk always yields the canonical path, never the alias.

If a client opens, writes and closes a hard-link alias, `e.path` is the alias and is correct, but this replaces it with the canonical path. `get_mnt` then either resolves a different fs-mode mount (exporting to the wrong UFS) or returns `None`, in which case `complete_file` returns `Ok(())` and the alias mount silently never gets the export. That is a regression against the previous `Path::from_str(&e.path)` for the no-rename case.

A guard that fixes rename without breaking links: keep `e.path` when it still resolves to `e.file.id`, and only fall back to `get_inode_path` when it does not. `e.file.nlink() > 1` is also available here as a cheap pre-check.

curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -49,10 +50,29 @@ impl UfsLoader {
...
+            Ok(path) => Ok(path),
+            Err(err) => {
```
szbr00: This arm catches every error from path reconstruction, not just a missing inode. `get_inode_path` propagates RocksDB read failures via `?` and returns plain `err_box!` strings for the missing-ancestor and cycle cases, so the caller cannot distinguish them — all three land here, get a `warn!`, and are replaced with `e.path`.

The impact is that a transient storage error during UFS replay silently downgrades to exporting a possibly-stale path instead of failing the entry. That defeats the retry budget in `journal_loader.rs:477-509`, which would otherwise retry up to `max_retry_num` before aborting or skipping.

Suggest making the not-found case explicit — e.g. have `get_inode_path` return `FsResult<Option<String>>`, or a typed `FsError` variant — and only fall back on not-found while propagating I/O and consistency errors.

curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -479,10 +479,18 @@ impl MasterFilesystem {
...
+        path: T,
+        inode_id: Option<i64>,
+    ) -> FsResult<FileStatus> {
+        let path = path.as_ref();
         let fs_dir = self.fs_dir.read();
-        let inp = Self::resolve_path(&fs_dir, path.as_ref())?;
-        let status = fs_dir.file_status(&inp)?;
-        Ok(status)
+        let inode = Self::resolve_file_inode(&fs_dir, path, inode_id)?;
+        Ok(inode.to_file_status(path)?)
```
szbr00, line 493: When `inode_id` is set this still hands the caller's string to `to_file_status`, so `status.path` comes from the request while `status.name` and `status.id` come from the resolved inode. That is precisely the contradiction #1666 describes under "the returned `FileStatus` contradicts itself", and the issue's expected behavior is that the response path is derived from the resolved inode.

The PR's rationale for keeping the request path is avoiding a rebuild on the write hot path, but that does not apply here: `GetFileStatus` is a read RPC, and `fs_dir.get_inode_path()` is now available on this same code path. Deriving the path in the `Some(inode_id)` branch would close the second half of the issue without touching `resize`/`assign_worker`.

As written, a client calling `file_status_by_id` after a rename receives `path="/a/old.log"` with `name="new.log"` — any consumer that reconstructs the full path from `path.parent() + name` gets a different answer than `path` itself.

curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1859,6 +1867,102 @@ mod tests {
...
+        let fs = test_fs("file-status-by-id-stale-path");
+        let created = fs.create("/a/old.log", true).unwrap();
+        fs.rename("/a/old.log", "/a/new.log", RenameFlags::empty())
+            .unwrap();
+
+        let status = fs
+            .file_status_by_id("/a/old.log", Some(created.id))
+            .unwrap();
+        assert_eq!(status.id, created.id);
+        assert_eq!(status.path, "/a/old.log");
```
szbr00, line 1903: These assertions pin a self-contradictory `FileStatus`: `status.path` is the stale request path (`/a/old.log`) while `status.name` comes from the resolved inode (`new.log`). The same shape is asserted for `resize_by_id` at line 1920 and `complete_file` at line 1962.

Issue #1666 explicitly calls this out as the second half of the bug and expects the response path to be derived from the resolved inode. If keeping the request path is intentional here — for audit reasons, or to avoid a rebuild on the write hot path — that rationale is worth capturing in the test name and a comment. Otherwise these assertions will block the fix, since correcting the behavior means changing them.

curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -205,3 +225,90 @@ impl UfsLoader {
...
+        let created = fs.create("/y/file.log", true).unwrap();
+        let mut entry = complete_entry_with_stale_path(&fs, "/y/file.log", created.id);
+        entry.file.id = 9_999_999;
+
+        let resolved = resolve_entry_path_for_test(&fs, &entry).unwrap();
+        assert_eq!(resolved.full_path(), "/y/file.log");
+    }
+
+    /// Mirrors `UfsLoader::resolve_complete_entry_path` without needing JobManager.
+    fn resolve_entry_path_for_test(
```
szbr00: This helper reimplements the match from `resolve_complete_entry_path` rather than calling it, so both tests would still pass if the production method or its call site at line 162 were changed or removed — they cover the idea, not the code.

The `JobManager` dependency is what makes the real method awkward to reach here (`fs_only_for_test` returns a `MasterFilesystem`, which has no `job_manager`). Since `resolve_complete_entry_path` only touches `self.fs_dir`, extracting it into a free function over `(&FsDir, &CompleteFileEntry)` would let `UfsLoader` and the tests share one implementation and remove this mirror. `journal_test.rs:589` already builds a full `JournalSystem` if you would rather construct a real loader.

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -117,6 +117,57 @@ impl FsDir {
...
+            match &inode_view {
+                File(f) => {
+                    components.push(f.name.clone());
+                    current_id = f.parent_id();
+                }
+                Dir(d) => {
+                    components.push(d.name.clone());
+                    current_id = d.parent_id();
+                }
+                FileEntry(e) => {
```
szbr00, line 158: This arm pushes `e.name` and breaks, so the returned string omits every remaining ancestor — `/alias.log` instead of `/dir/alias.log`. The comment says it preserves the previous fallback, but there is no fallback here; the result is a silently truncated path rather than an error.

It looks unreachable in practice, since `store.get_inode(id, None)` reads a top-level inode record and `InodePath::resolve` (inode_path.rs:54) explicitly re-resolves `FileEntry` before it can reach a caller. But the helper is now `pub` and its output drives UFS mount selection, where a truncated path resolves to the wrong mount or none. Returning an error here, or noting the unreachability explicitly, would be safer than a plausible-looking wrong path.

crates/client/curvine-client-core/src/file/fs_client.rs
```diff
@@ -165,8 +165,17 @@ impl FsClient {
     }
 
     pub async fn file_status(&self, path: &Path) -> FsResult<FileStatus> {
+        self.file_status0(path, None).await
+    }
+
+    pub async fn file_status_by_id(&self, path: &Path, inode_id: i64) -> FsResult<FileStatus> {
```
szbr00, line 171: Nothing in the repo calls `file_status_by_id`, and no client ever populates the new `GetFileStatusRequest.inode_id` field — the only reader is `master_handler.rs:220`. So the by-id status path added here is unreachable in production, and the only effective behavior change in this PR is the UFS export resolution.

The call sites that actually produce stale paths are still unchanged: `fs_writer_base.rs:388` and `:533` pass `&self.path` captured at open time to `assign_worker_by_id` / `resize_by_id`. If this API is groundwork for a follow-up that wires it in, that is worth saying in the PR description; otherwise consider dropping it until there is a caller, so the field is not left as untested surface on a public RPC.

curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -49,10 +50,29 @@ impl UfsLoader {
...
+        let path = self.fs_dir.read().get_inode_path(inode_id)?;
+        Ok(Path::from_str(path)?)
+    }
+
+    pub(crate) fn resolve_complete_entry_path(&self, e: &CompleteFileEntry) -> FsResult<Path> {
+        match self.get_real_path(e.file.id) {
+            Ok(path) => Ok(path),
+            Err(err) => {
+                warn!(
+                    "complete_file: get_inode_path({}) failed: {}, fallback to entry.path={}",
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Fixed in a97fcdd24: fallback is now limited to FsError::FileNotFound; store/corruption errors propagate. Missing-ancestor in get_inode_path is also typed as FileNotFound.

curvine-master/src/master/journal/ufs_loader.rs
bigbigxu: Thanks. Fixed in a97fcdd24: only FileNotFound falls back to e.path; other errors from get_real_path / get_inode_path now fail the UFS entry so replay can retry.

curvine-master/src/master/meta/fs_dir.rs
bigbigxu: Thanks. Fixed in a97fcdd24: unexpected FileEntry during the parent walk now returns an error instead of a truncated name-only path.

curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -49,10 +50,29 @@ impl UfsLoader {
...
     }
 
+    pub fn get_real_path(&self, inode_id: i64) -> FsResult<Path> {
+        let path = self.fs_dir.read().get_inode_path(inode_id)?;
+        Ok(Path::from_str(path)?)
+    }
+
+    pub(crate) fn resolve_complete_entry_path(&self, e: &CompleteFileEntry) -> FsResult<Path> {
+        match self.get_real_path(e.file.id) {
+            Ok(path) => Ok(path),
```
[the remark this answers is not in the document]
bigbigxu: Thanks for the suggestion. Declined for this PR: hard links are not meaningful for UFS export, so any live path of the inode is acceptable. We keep get_inode_path to fix rename-after-open stale journal paths.

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -117,6 +117,57 @@ impl FsDir {
...
+                    );
+                }
+            };
+
+            match &inode_view {
+                File(f) => {
+                    components.push(f.name.clone());
+                    current_id = f.parent_id();
+                }
+                Dir(d) => {
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Declined for the same reason: UFS export does not need hard-link-aware alias selection; rebuilding via parent_id/name is enough for the rename-stale case this PR targets.

curvine-master/src/master/journal/ufs_loader.rs
bigbigxu: Thanks. Declined: for UFS export, hard-link aliases are not a concern — any path of the inode is fine. The rebuild is specifically for rename-after-open where e.path is stale.

curvine-master/src/master/fs/master_filesystem.rs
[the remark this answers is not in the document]
bigbigxu: Thanks. Declined: rebuilding the path on GetFileStatus by-id is intentionally avoided because of the cost. By-id still resolves the inode (id/name); the request path is kept in FileStatus.path. Write hot paths similarly keep the request path.

bigbigxu: Thanks. Declined intentionally: we do not rebuild path on the by-id status read path due to overhead. status.name/id come from the resolved inode; status.path stays as the request path. The UFS export path (rename-stale) is what this PR fixes.

curvine-master/src/master/fs/master_filesystem.rs
bigbigxu: Thanks. The assertions are intentional with the decision above: by-id / write RPCs keep the request path and do not rebuild. Issue #1666 is addressed on the UFS export side via get_inode_path in complete_file.

curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -205,3 +225,90 @@ impl UfsLoader {
...
+
+    /// Mirrors `UfsLoader::resolve_complete_entry_path` without needing JobManager.
+    fn resolve_entry_path_for_test(
+        fs: &crate::master::fs::MasterFilesystem,
+        e: &CompleteFileEntry,
+    ) -> FsResult<Path> {
+        match fs.fs_dir.read().get_inode_path(e.file.id) {
+            Ok(path) => Ok(Path::from_str(path)?),
+            Err(_) => Ok(Path::from_str(&e.path)?),
+        }
```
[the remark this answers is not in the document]
bigbigxu: Thanks. In a97fcdd24 we inlined the resolution into UfsLoader::complete_file and removed the helper/tests. Path rebuild coverage remains in get_inode_path unit tests (including after rename).

curvine-master/src/master/journal/ufs_loader.rs
bigbigxu: Thanks. Agreed the mirrored helper was weak coverage. In a97fcdd24 the free function/helper is gone; complete_file calls get_real_path directly, and get_inode_path unit tests cover the rename rebuild.

crates/client/curvine-client-core/src/file/fs_client.rs
bigbigxu: Thanks. Agreed the main behavior change is UFS export resolution. file_status_by_id / optional GetFileStatusRequest.inode_id are wired through master_handler as groundwork for by-id status; clients are not required to populate it yet. Write paths keep the request path by design (hot-path cost).

szbr00 (approved):

Issue #1690, opened 2026-09-12
[BUG]: cache-mode files stay uncached after a worker loses their only replicas


papapumpkin:
**Describe the bug**

We replaced a cache worker with fresh local storage and found that files cached on that worker never warmed back up. Their metadata was still present on the master, but reads kept returning `all replicas has been lost` / `GetBlockLocations err=Common` and falling back to S3.

The replacement joined normally. Nearly seven hours later, it still had only a few thousand blocks, compared with roughly 570,000 on the old worker. Two training jobs using the affected data went from about 0.3 seconds per step to about 0.5 seconds and stayed there.

The S3 objects are intact. The problem is getting them back into the cache when the file metadata already exists. In the `SubmitJob` auto-cache path, `overwrite` is left unset, which the load worker treats as `false`. We also tried `cv load --no-overwrite`; it eventually failed with `target already exists` on a chunk file.

**To Reproduce**

This is the sequence we hit in production; we have not repeated the worker replacement in a separate test cluster.

1. Set up an S3-backed `cache_mode` mount with auto-cache enabled and one cached replica per block, using the legacy `SubmitJob` load path.
2. Cache a file and confirm it can be read from Curvine.
3. Replace the worker holding its only cached replica with a worker using empty storage. Leave master metadata intact.
4. Keep reading the file through `UnifiedFileSystem`, with UFS reads enabled.
5. The file remains readable through S3, but Curvine keeps reporting lost replicas and does not restore the cached blocks.
6. Try an explicit load with overwrite disabled:

   ```sh
   cv load --no-overwrite -w /cache/dataset/example.zarr
   ```

7. The load task fails:

   ```text
   task <job>_<run>_task_1 failed: target already exists: /cache/dataset/example.zarr/c/0/0/0
   ```

**Actual behavior**

The master registered the worker loss immediately, then registered the replacement about 20 minutes later. Capacity came back, but the cached data did not:

| Measurement | Before replacement | Worker unavailable | Replacement joined |
|---|---:|---:|---:|
| Live cache workers | Full fleet | One fewer | Full fleet |
| Cache capacity | Full capacity | Down by one worker | Full capacity |
| Master block count | About 18.9 million | About 18.3 million | About 18.3 million |
| Blocks on affected worker | About 570,000 | Worker absent | 2 |

The old worker held about 5 TB of cached data. Almost seven hours later, its replacement had fewer than 3,000 blocks, and requests from both slow jobs were still failing on lost chunks.

```text
[...]

PR #1691, opened 2026-09-14
fix(master): reject obsolete blocks in worker reports (#1690)

HarryMWinters:
## Summary

A worker can retain old blocks after a file is overwritten or its cache is invalidated. The file's metadata record (inode) may still exist, but its current block list no longer contains those IDs. The previous report handler checked only that the inode existed and was a file, allowing a delayed report to restore obsolete block locations.

This PR checks reported IDs against the inode's current block list under the same filesystem write lock that updates locations. Obsolete IDs in full reports or `Finalized` entries are returned for worker-side deletion. Existing deletion acknowledgments and deferred handling of unknown incremental `Writing` reports are preserved.

This is the first of two stacked PRs. The [follow-up (fork PR #1)](https://github.com/HarryMWinters/curvine/pull/1) adds a bounded grace period after worker departures, including graceful exits, before cache-metadata cleanup. That change relies on these validation checks to prevent returning workers from restoring blocks discarded while offline.

Related to [CurvineIO/curvine#1690](https://github.com/CurvineIO/curvine/issues/1690), following [the cache invalidation fix in #1526](https://github.com/CurvineIO/curvine/pull/1526).

## Validation

- Membership, overwrite/invalidation, and lock-order tests passed. Two added inode-decoding-error tests also passed for full and incremental reports, covering failed-chunk atomicity, committed earlier chunks, repair/retry, and duplicate-free replay.
- The earlier broader run reported 1,464 Rust passes (three PostgreSQL cases returned early) and 28 Go/Python passes. Nine Rust cases were skipped. Compiler-depth failures in `fallback_read_test` and `write_cache_test`, a read-only-mount failure in `ufs_test`, and the HDFS JNI doctest failure all reproduced on the PR's base.
- External-service and platform coverage remains incomplete; one hardcoded external-fixture test was stopped. After the final test-only commit, only the two new tests were rerun: **2 passed, 115 filtered out**.


curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1318,98 +1307,40 @@ impl MasterFilesystem {
...
-                BlockReportStatus::Deleted => {
-                    batch.push((false, item.id, loc));
-                    wm.deleted_block(list.worker_id, item.id);
-                }
+            let deleted: Vec<_> = chunk
+                .iter()
+                .filter(|block| block.status == BlockReportStatus::Deleted)
+                .map(|block| block.id)
+                .collect();
+            let obsolete = self.fs_dir.write().apply_reported_blocks(
```
HarryMWinters, line 1325: **Lock duration:** Each write-locked chunk contains at most 4,096 report entries, but loads and hashes each distinct inode's entire block list. Large-file lock duration and concurrent metadata latency remain unmeasured. Is a representative performance check warranted?

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -1024,6 +1009,50 @@ impl FsDir {
...
+                batch.delete_location(block.id, worker_id)?;
+                continue;
+            }
+
+            let inode_id = InodeId::get_id(block.id);
+            let ids = match current_blocks.entry(inode_id) {
+                std::collections::hash_map::Entry::Occupied(entry) => entry.into_mut(),
+                std::collections::hash_map::Entry::Vacant(entry) => {
+                    let ids: HashSet<i64> = match self.store.get_inode(inode_id, None)? {
+                        Some(File(file)) => file.blocks.iter().map(|meta| meta.id).collect(),
```
HarryMWinters, line 1037: **Block ordering:** Existing inode lookups use binary search. Are block vectors guaranteed to remain sorted across mutations and restoration? If so, could we cache those vectors instead of building hash sets, and remove or revise the synthetic unsorted-vector test?

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -1024,6 +1009,50 @@ impl FsDir {
...
+            if block.status == BlockReportStatus::Deleted {
+                batch.delete_location(block.id, worker_id)?;
+                continue;
+            }
+
+            let inode_id = InodeId::get_id(block.id);
+            let ids = match current_blocks.entry(inode_id) {
+                std::collections::hash_map::Entry::Occupied(entry) => entry.into_mut(),
+                std::collections::hash_map::Entry::Vacant(entry) => {
+                    let ids: HashSet<i64> = match self.store.get_inode(inode_id, None)? {
```
HarryMWinters, line 1036: **Error policy:** Inode read or decoding errors now fail the report instead of being logged and skipped; missing inodes remain handled normally. Incremental reports requeue failures, but an initial full-report error can terminate worker startup. Is this the intended recovery policy, especially for persistent errors? The new master retry tests do not establish worker-process recovery.

curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1318,98 +1307,40 @@ impl MasterFilesystem {
...
+                list.worker_id,
+                list.full_report,
+                chunk,
```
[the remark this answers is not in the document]
HarryMWinters: Agreed: a delayed report can run between location removal and cache invalidation. That window also exists on this PR’s base; the membership check alone does not close it.

The [follow-up PR on my fork](https://github.com/HarryMWinters/curvine/pull/1) serializes report processing, cleanup, and heartbeat handling with a per-worker lifecycle lock. I propose keeping that coordination in the follow-up and adding a regression test that explicitly pauses cleanup in this window. This PR would remain focused on rejecting blocks already discarded from the inode.

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -1024,6 +1009,50 @@ impl FsDir {
...
+                        Some(File(file)) => file.blocks.iter().map(|meta| meta.id).collect(),
+                        _ => HashSet::new(),
+                    };
+                    entry.insert(ids)
+                }
+            };
+            if ids.contains(&block.id) {
+                batch.add_location(block.id, &BlockLocation::new(worker_id, block.storage_type))?;
+            } else if full_report || block.status == BlockReportStatus::Finalized {
```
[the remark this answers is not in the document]
HarryMWinters: This is the same cleanup/report interleaving discussed in the [other thread](https://github.com/CurvineIO/curvine/pull/1691#discussion_r4002928361). The ID is still current at this point, so membership validation cannot distinguish it from a valid report. The follow-up’s per-worker lifecycle synchronization addresses that ordering.

HarryMWinters:
Regarding the performance concerns in [Copilot’s review](https://github.com/CurvineIO/curvine/pull/1691#pullrequestreview-5194898413):

I’ll run a focused local comparison against the base, measuring report throughput, exclusive-lock duration, concurrent metadata latency, and peak memory with large-file and mixed-file inventories. That should give us evidence before choosing an optimization. Are there particular workload sizes or latency targets we should cover?

HarryMWinters:
I compared base `ae330604` with PR head `aa0aa165` locally on Linux aarch64 with a six-CPU quota: optimized Rust 1.98.1, real RocksDB metadata, warm replay, journaling disabled, one reporting stream, and concurrent metadata read/write probes. Three fresh-process runs per workload, with equal report counts in each base/head pair.

Median throughput, in thousands of reported block entries/s:

| Workload | Base | PR |
|---|---:|---:|
| One 1,000-block file; report all 1,000 | 56.8 | 901.7 |
| One 10,000-block file; report 1,000 | 7.2 | 900.1 |
| One 10,000-block file; report 8,192 in one RPC | 7.1 | 1,185.9 |
| 100 files × 100 blocks; report 10 per file | 328.0 | 627.4 |
| 1,000 files × 1,000 blocks; report one per file | 57.4 | 39.4 |
| Full 10,000-block inventory; ten 1,000-entry RPCs | 7.3 | 729.1 |

Two concerns showed up:

- **Sparse reports:** 31–32% lower throughput; median per-run exclusive-lock hold p95 increased from 1.82 to 26.19 ms, and sampled peak process RSS from 140.5 to 157.8 MiB. The throughput regression persisted in three probe-free pairs.
- **Continuous incremental replay:** unrelated metadata operations sometimes waited nearly the entire 0.17–2.14 s reporting burst, despite much shorter individual lock holds. Each probe completed only 1–3 operations per run, so these are blocking observations, not reliable latency percentiles. A separate PR-only control inserting a 100 µs sleep between RPCs restored regular progress, while sparse probe p95 remained about 26 ms.

The full-inventory benchmark explicitly waits for reconciliation between inventories, introducing scheduling gaps. The paced control is also artificial; neither establishes production latency. This is a local in-process test, with no network, Raft, or actual file payloads, and no matched-arrival-rate comparison. All benchmark correctness checks passed.

The direction I’m leaning toward is reducing membership-set work first: keep reading each touched inode once per chunk, but retain membership only for the block IDs being reported instead of building a set of every current block ID. Validation and location updates would stay under the same filesystem write lock. This should reduce sparse-report allocation and processing, but would not by itself guarantee fairness between successive lock acquisitions. I’d verify the change with focused correctness tests and repeat the benchmarks, including early, late, and missing IDs.

HarryMWinters:
Grouping reported IDs by inode helped sparse reports but added overhead in a dense-report case. Neither std-lock version resolved the observed metadata-reader stalls between successive report chunks.

I compared the original upstream base, this PR, and the local prototypes using identical RPC/RocksDB workloads, with three repetitions and interleaved ordering. For four workers reporting 512,000 blocks across 512 inodes:

| Version | Report time | Worst metadata read |
|---|---:|---:|
| Original upstream base | 3.18 s | 10 ms |
| Current PR | 6.44 s | 6,438 ms |
| PR + inode grouping | 3.79 s | 3,794 ms |
| PR + parking_lot | 5.84 s | 78 ms |
| PR + grouping + parking_lot | 4.35 s | 57 ms |

Values are medians across three runs; “worst metadata read” is the median of each run’s maximum latency.

The parking_lot prototype uses its default fairness policy for the filesystem metadata lock, preserving chunk boundaries and atomic validation/location updates. Both variants restored reader progress during reporting, but remained behind baseline in this concurrent workload.

Grouping plus parking_lot had approximately **27% lower report throughput than baseline**; the gap persisted at 31% with metadata and heartbeat probes disabled. Results depend on concurrency: with one reporting worker, the combined prototype completed the same inventory in **4.80 s versus baseline’s 10.14 s**.

Fair lock admission looks worth pursuing, but these results do not establish baseline performance parity. All 165 benchmark runs completed cleanly. These are local experiments with synthetic metadata, without Raft or file-data workloads; baseline also lacks the new membership protection. The prototypes remain unpushed, and a production lock change would still need to address panic handling and debug reentry checks.

HarryMWinters:
I compared this PR and local prototypes against base `ae330604` using real RPC/RocksDB on Linux. This matched workload has four workers reporting 512,000 blocks across 512 inodes in shuffled 1,000-entry full-report RPCs, alongside file-status and set-attribute requests.

| Implementation | Report time | Worst read | Worst write |
|---|---:|---:|---:|
| Original base | 3.11 s | 10.2 ms | 10.1 ms |
| Current PR | 6.46 s | 6,462 ms | 57.7 ms |
| PR + grouped reported IDs | 3.88 s | 3,876 ms | 34.0 ms |
| PR + exclusive `parking_lot` lock | 5.86 s | 79.3 ms | 52.7 ms |
| PR + grouping + exclusive `parking_lot` | 4.37 s | 57.7 ms | 37.4 ms |
| PR + upgradable `parking_lot` read lock | 6.55 s | 3.6 ms | 54.1 ms |
| PR + grouping + upgradable read lock | 4.98 s | 4.7 ms | 43.3 ms |

Values are medians of three fresh-process runs; “worst” is the median of per-run maximum RPC latency. The base lacks the new membership protection. Results depend on concurrency: with one worker reporting the same inventory, base took 10.12 s versus 6.92 s for upgrade and 5.29 s for grouping + upgrade.

The two PR variants using the current lock completed only one read during each reporting burst; upgrade variants completed thousands. Earlier membership tests showed grouping helped reports with few blocks per inode but added overhead in a dense case. `parking_lot` restored reader progress. Its upgradable read guard allowed reads during validation, then atomically promoted to a write guard for location updates, preserving validation/update ordering. Writers remained blocked throughout both phases.

The upgrade operation itself was inexpensive; further preparation/batching experiments did not recover the remaining throughput cost. Two attempts in a later local direct-batch experiment hit background runtime shutdown panics after reporting finished. They were excluded and repeated successfully; the cause is unconfirmed.

One path is to retain the current membership algorithm and adopt the upgradable lock, considering grouping separately. Smaller report chunks (below 1,000 entries here) could reduce writer waits at the cost of more commits; versioned validation or per-inode locking would add concurrency complexity. These optimizations have not been pushed to this PR. Focused correctness tests passed, but these synthetic-metadata experiments exclude Raft/file-data workloads and establish no production latency guarantees.

**Edit (2026-09-16): Added median, p90, and p99 metadata RPC latencies for the grouped + upgradable `parking_lot` prototype, with the original baseline for comparison. These come from the same saved traces; no benchmarks were rerun.**

| Metadata RPC | Implementation | Median | p90 | p99 | Samples per run |
|---|---|---:|---:|---:|---|
| File-status read | Original base | 0.183 ms | 1.931 ms | 2.851 ms | 2,210 / 2,297 / 2,285 |
| File-status read | PR + grouping + upgradable lock | 0.110 ms | 1.272 ms | 2.653 ms | 4,120 / 4,257 / 4,152 |
| Set-attribute write | Original base | 0.212 ms | 1.920 ms | 2.931 ms | 2,226 / 2,303 / 2,273 |
| Set-attribute write | PR + grouping + upgradable lock | 38.066 ms | 40.626 ms | 42.179 ms | 130 / 130 / 130 |

Each value is the median of the corresponding statistic from three runs, not a pooled percentile. These are dispatch-to-response RPC times, including responses that finish after reporting ends. Each probe allows one outstanding request and skips missed 1 ms scheduling slots. The grouped-upgradable variant has only 130 write samples per run, so its write p99 reflects the largest few observations. Compared with baseline, reads remain responsive while the write cost affects typical requests as well as the tail. These remain measurements of the earlier prototype, not a new benchmark of a later implementation.

**Edit (2026-09-16, chunking results): Added a fresh baseline comparison while varying master-side entries per lock acquisition in the grouped + upgradable implementation.**

Worker full-report RPCs stayed at **1,000 entries**. The master split each RPC into processing chunks, releasing the lock after validating and committing each chunk. At 250 entries, one RPC therefore takes four lock acquisitions and still receives one response. Only foreground report processing was tuned; cleanup and reconciliation limits stayed unchanged.

The table uses the same four-worker, 512,000-block workload described above. These are new runs against the updated local implementation and original base `ae330604`, separate from the earlier prototype measurements. Each number is the median of its corresponding statistic across three fresh-process runs; latency triplets are **median / p90 / p99**, in milliseconds.

| Implementation / master processing cap | Report time | Read median / p90 / p99 (ms) | Write median / p90 / p99 (ms) |
|---|---:|---:|---:|
| Original base | 3.376 s | 0.172 / 2.391 / 3.431 | 0.207 / 2.388 / 3.441 |
| Grouped + upgrade, 4,096 | 5.425 s | 0.115 / 1.654 / 3.217 | 42.038 / 43.950 / 45.369 |
| Grouped + upgrade, 1,000 | 5.383 s | 0.118 / 1.587 / 3.273 | 41.495 / 43.631 / 45.310 |
| Grouped + upgrade, 500 | 6.901 s | 0.121 / 1.079 / 1.825 | 26.265 / 27.625 / 29.655 |
| Grouped + upgrade, 250 | 8.077 s | 0.123 / 0.623 / 1.128 | 15.191 / 16.192 / 16.756 |
| Grouped + upgrade, 100 | 9.103 s | 0.129 / 0.354 / 0.601 | 6.416 / 7.212 / 7.829 |

Caps of 4,096 and 1,000 both process these RPCs in one chunk; their report times differed by less than 1%. Percent changes compare the displayed medians. Compared with the 4,096 control, **250 entries reduced write p99 by 63% and read p99 by 65%, but increased report duration by 49% (33% lower throughput)**. At 100 entries, write p99 fell further, with another throughput cost.

Most of the cost persisted without metadata probes: reporting took 5.387 s at 4,096 versus 7.919 s at 250. With 64 blocks per inode instead of 1,000, the 250-entry cap increased report duration by 17% (0.388 → 0.453 s). Smaller chunks repeat inode processing and commits, so there is no workload-independent optimum.

All 45 measured runs across the sweep and follow-up controls, plus a small pilot, passed RPC, inventory, and teardown checks. The 4,096-cap main runs had only 130 write samples each, versus 514–515 at 250; p99 remains descriptive, not a production guarantee. Probe pacing and latency accounting are as described above. The shorter controls, including oversized incremental RPCs, had too few writes for strong tail estimates. Changing the production cap would also need focused correctness tests at the new chunk boundaries.

A master processing limit would be separate from `master.block_report_limit`, which controls worker full-report RPC pagination. I would keep the initial tuning change focused on this limit; more elaborate batching or concurrency strategies could follow if required by the latency target. These optimizations have not been pushed to this PR.

Are there particular metadata read/write latency and report-throughput targets that should guide the processing cap? Is this tradeoff acceptable within this PR's scope?

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -1024,6 +1009,50 @@ impl FsDir {
...
+                batch.delete_location(block.id, worker_id)?;
+                delete_blocks.push(block.id);
```
szbr00, line 1047: The old code logged an aggregate warning when obsolete blocks were found:

```rust
warn!(
    "block_report found {} missing-inode and {} non-file-inode blocks for worker {}; scheduling worker deletion",
    missing_blocks, not_file_blocks, list.worker_id
);
```

The new code silently returns `delete_blocks` without master-side logging. This reduces observability for debugging stale cache issues or incorrect block IDs. Consider adding a warning when `delete_blocks` is non-empty:

```rust
if !delete_blocks.is_empty() {
    warn!(
        "block_report found {} obsolete blocks for worker {}; scheduling worker deletion",
        delete_blocks.len(), worker_id
    );
}
```

This preserves the debugging signal without the per-block noise of the old deferred-writing warnings.

szbr00:
> I compared this PR and local prototypes against base `ae330604` using real RPC/RocksDB on Linux. This matched workload has four workers reporting 512,000 blocks across 512 inodes in shuffled 1,000-entry full-report RPCs, alongside file-status and set-attribute requests.
> 
> | Implementation | Report time | Worst read | Worst write |
> |---|---:|---:|---:|
> | Original base | 3.11 s | 10.2 ms | 10.1 ms |
> | Current PR | 6.46 s | 6,462 ms | 57.7 ms |
> | PR + grouped reported IDs | 3.88 s | 3,876 ms | 34.0 ms |
> | PR + exclusive `parking_lot` lock | 5.86 s | 79.3 ms | 52.7 ms |
> | PR + grouping + exclusive `parking_lot` | 4.37 s | 57.7 ms | 37.4 ms |
> | PR + upgradable `parking_lot` read lock | 6.55 s | 3.6 ms | 54.1 ms |
> | PR + grouping + upgradable read lock | 4.98 s | 4.7 ms | 43.3 ms |
> 
> Values are medians of three fresh-process runs; “worst” is the median of per-run maximum RPC latency. The base lacks the new membership protection. Results depend on concurrency: with one worker reporting the same inventory, base took 10.12 s versus 6.92 s for upgrade and 5.29 s for grouping + upgrade.
> 
> The two PR variants using the current lock completed only one read during each reporting burst; upgrade variants completed thousands. Earlier membership tests showed grouping helped reports with few blocks per inode but added overhead in a dense case. `parking_lot` restored reader progress. Its upgradable read guard allowed reads during validation, then atomically promoted to a write guard for location updates, preserving validation/update ordering. Writers remained blocked throughout both phases.
> 
> The upgrade operation itself was inexpensive; further preparation/batching experiments did not recover the remaining throughput cost. Two attempts in a later local direct-batch experiment hit background runtime shutdown panics after reporting finished. They were excluded and repeated successfully; the cause is unconfirmed.
> 
> One path is to retain the current membership algorithm and adopt the upgradable lock, considering grouping separately. Smaller report chunks (below 1,000 entries here) could reduce writer waits at the cost of more commits; versioned validation or per-inode locking would add concurrency complexity. The prototypes remain unpushed. Focused correctness tests passed, but these synthetic-metadata experiments exclude Raft/file-data workloads and establish no production latency guarantees.
> 
> Are there particular metadata read/write latency and report-throughput targets we should cover? Is the upgradable-lock tradeoff acceptable for this PR, or should we pursue a more complex concurrency strategy first?
> 

Thank you for your consideration. We can discuss the performance optimization in the Slack channel.The invitation link is on README page, you are very welcome to join the Curvine Slack.

szbr00:
Could you please fix the compilation failure issue ？Thank you.

HarryMWinters:
@szbr00 I believe the compilation failure comes from a CI startup race (or general test runner flakiness). Would it be possible to manually re-run the failed step and see if it resolves?

Additionally I want to emphasize that this **branch is _not_ ready for final review or merge**. Once we decide on optimization strategy and acceptable latency and throughput tradeoffs I'll push whatever optimization pattern we've decided on.

lzjqsdd:
> @szbr00 I believe the compilation failure comes from a CI startup race (or general test runner flakiness). Would it be possible to manually re-run the failed step and see if it resolves?
> 
> Additionally I want to emphasize that this **branch is _not_ ready for final review or merge**. Once we decide on optimization strategy and acceptable latency and throughput tradeoffs I'll push whatever optimization pattern we've decided on.


> @szbr00 I believe the compilation failure comes from a CI startup race (or general test runner flakiness). Would it be possible to manually re-run the failed step and see if it resolves?
> 
> Additionally I want to emphasize that this **branch is _not_ ready for final review or merge**. Once we decide on optimization strategy and acceptable latency and throughput tradeoffs I'll push whatever optimization pattern we've decided on.

Trying fix smoke ci build in #1704

curvine-master/src/master/fs/master_filesystem.rs
gongxun0928, line 1325: The membership check and the location batch in apply_reported_blocks only need fs_dir.read(), not write().

What the write lock is buying: it stops an overwrite from clearing file.blocks between "ID is current" and "location row written". But that interleaving is harmless:

Block IDs are never reused (next_seq is only incremented; overwrite/free/invalidate_cache just blocks.clear()), so a stale row can never alias a current block.
Reads go file.blocks → point lookup in CF_BLOCK, never the other way, so a row whose ID is no longer in the inode is unreachable.
apply_overwrite_file already leaves location rows in place and waits for the worker's Deleted ack — the system tolerates "in CF_BLOCK but not in inode" by design. The worst case here is one more orphan row, cleaned up by the next reconcile.
What the read lock still buys: mutual exclusion with FsDir::restore (swaps the RocksDB instance) and delete_locations(worker) (two-step, non-atomic delete across CF_LOCATION/CF_BLOCK). Both run under write(), so a read() on the report path keeps them apart without blocking any reader.

Concurrent reports don't conflict: each writes its own (block, worker) keys, repeats are idempotent, and the WriteBatch is atomic on its own.

So: make apply_reported_blocks take &self, call it under fs_dir.read(), and the validation work stops stalling metadata traffic.

curvine-master/src/master/meta/fs_dir.rs
```diff
@@ -1024,6 +1009,50 @@ impl FsDir {
...
+    pub(crate) fn apply_reported_blocks(
+        &mut self,
+        worker_id: u32,
+        full_report: bool,
+        blocks: Vec<BlockReportInfo>,
+    ) -> FsResult<Vec<i64>> {
+        let mut batch = self.store.new_batch();
+        let mut delete_blocks = Vec::new();
+        // File block vectors are not guaranteed to be sorted by ID.
+        let mut current_blocks = HashMap::new();
```
gongxun0928, line 1025: Workers report from a HashMap, it's unsorted, so one inode's blocks are scattered across every RPC of a full report and the per-chunk inode cache reloads each inode hundreds of times.

Two ways to fix it:

Option A (recommended): sort. Sort all_blocks() by id on the worker before paginating — block id high bits are the inode id, so each inode becomes one contiguous run across RPCs. On the master, sort the RPC once as well (incremental reports and older workers are unordered), then replace the HashMap<inode, HashSet> with a single "current inode" cursor and check membership by merging against the already-sorted file.blocks. Each inode loads about once per full report, cache memory is O(1), and both sorts cost microseconds per RPC.

Option B: hoist the cache. Keep the current structure but move HashMap<inode, HashSet> from per-chunk to per-RPC. This removes the cross-chunk reloads within one RPC, but with unsorted input each inode is still reloaded once per RPC it appears in, so it helps far less on a large full report.

HarryMWinters:
<head></head><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><strong>Update (2026-09-18): Updated proposed optimizations and a new comparison against unsorted upstream baseline. These optimizations have not been pushed to this PR.</strong></p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Following the review suggestions and further experiments, I’m proposing:</p><ul style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><li><strong>Worker-side sorting:</strong><span class="Apple-converted-space"> </span>sort full inventories by block ID before RPC pagination, preserving incremental report ordering.</li><li><strong>Master-side inode grouping:</strong><span class="Apple-converted-space"> </span>reuse each inode lookup within a processing chunk.</li><li><strong>Ordinary shared locking:</strong><span class="Apple-converted-space"> </span>use a<span class="Apple-converted-space"> </span><code dir="ltr">parking_lot</code><span class="Apple-converted-space"> </span>read guard instead of an upgradable guard, retaining it through membership validation and the atomic location-batch commit. Inode mutations remain excluded; checkpoint creation needs exclusive protection.</li><li><strong>Cheaper reconciliation scans:</strong><span class="Apple-converted-space"> </span>retain one stable RocksDB iterator, borrow key/value slices, constrain iteration to the worker’s index range, and bound initial buffer reservation.</li><li><strong>Separate master processing limits:</strong><span class="Apple-converted-space"> </span>initially 250 reported entries per foreground lock acquisition and 500 stale-location deletions per cleanup batch. Worker RPCs remain at 1,000 entries.</li></ul><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">This comparison includes the benefit of worker sorting: baseline<span class="Apple-converted-space"> </span><code dir="ltr">d896034</code><span class="Apple-converted-space"> </span>sends unsorted inventories, while the proposed package sorts the same worker assignments. Baseline retains its 4,096-entry processing/deletion caps; the candidate uses 250/500 throughout. The caps remain benchmark settings, pending implementation and review.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">All cases use four workers, 1,000 blocks per inode, and 1,000 scheduled file-status reads plus 1,000 set-attribute writes per second. Inventory sizes are<span class="Apple-converted-space"> </span><strong>per worker</strong>. Values are medians of three per-run statistics; latency triplets are<span class="Apple-converted-space"> </span><strong>median / p90 / p99 in milliseconds</strong>. Completion includes inventory cloning, candidate sorting, report processing and reconciliation.</p><div style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><div><div>

Workload per worker | Version | Completion | Read median / p90 / p99 (ms) | Write median / p90 / p99 (ms)
-- | -- | -- | -- | --
128k blocks, no cleanup | Unsorted baseline | 3.030 s | 0.469 / 2.460 / 3.928 | 0.495 / 2.357 / 3.686
  | Proposed | 0.286 s | 0.388 / 0.723 / 4.026 | 0.575 / 0.803 / 4.426
128k reported + 128k stale locations | Unsorted baseline | 3.193 s | 0.462 / 3.403 / 87.007 | 0.500 / 3.244 / 77.110
  | Proposed | 0.478 s | 0.530 / 1.116 / 10.216 | 0.640 / 1.219 / 11.215
128k blocks, 12 consecutive cycles | Unsorted baseline | 42.662 s | 0.655 / 3.739 / 22.139 | 0.702 / 3.605 / 21.969
  | Proposed | 4.226 s | 0.565 / 14.207 / 41.442 | 0.684 / 14.933 / 42.489
1M blocks, reopened database | Unsorted baseline | 29.572 s | 0.411 / 3.739 / 45.104 | 0.436 / 3.575 / 45.916
  | Proposed | 2.342 s | 0.421 / 0.907 / 67.962 | 0.564 / 0.920 / 68.411

</div></div></div><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><strong>Completion was 6.7–12.6× faster, with substantially better heavy-cleanup latency. Metadata latency did not improve uniformly:</strong><span class="Apple-converted-space"> </span>repeated-report p99 was roughly twice baseline, and million-block p99 was about 50% higher. These percentiles cover each variant’s active reporting/reconciliation window; faster completion changes the proportion of time spent in each phase.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">These are fresh matched runs, with execution order alternated. All 24 measurements and two excluded pilots passed inventory, RPC, reconciliation and teardown checks. Measured runs had no RPC errors or connection-pool drops; four final scheduling slots were recorded as generator-late drops out of 514,984 offered metadata requests. Latency covers successful requests from scheduled arrival through response, including responses after reconciliation ends. This differs from the earlier closed-loop probe methodology, so comparisons should stay within this table.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The probe pool is 128 connections per operation, or 256 for the million-block case. Short candidate runs have only hundreds of samples; the million-block candidate’s read p99 ranged from 43–97 ms across the three runs. These are descriptive local RPC/RocksDB measurements without Raft or file-data workloads. “Reopened” does not mean a cold OS cache.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">I’d pursue this package first and leave executor tuning, scheduling changes and unchanged-write suppression for separate consideration; their benefits were less consistent across workloads. Are the remaining metadata latency tradeoffs acceptable for this PR, or are there particular workload sizes and latency targets we should meet before proceeding?</p>

Issue #1190, opened 2026-07-19
[FEATURE] cv fsck: deep block location inspector with per-worker storage type


thuongle2210:
**Is your feature request related to a problem? Please describe.**

Normal users and operators have no way to see the **actual** per-worker storage type of block replicas. The existing `cv blocks` command shows only the file-level default storage policy (e.g., `SpdkDisk`), not the real placement (e.g., one replica on `SpdkDisk`, another on `DISK` due to `choose_dir` fallback). This makes it impossible to diagnose storage tier distribution, verify data placement, or detect policy mismatches from the CLI or UI.

HDFS solves this with `hdfs fsck <path> -files -blocks -locations`, which shows per-replica storage type (SSD, DISK, ARCHIVE, etc.) for every block. Curvine has no equivalent.

**Describe the solution you'd like**

Add a `cv fsck <path>` command that inspects actual per-worker block locations by reading `BlockLocation` from the master's RocksDB (cold path, on-demand RPC).

### For files

```
$ cv fsck /data/file.gz

File: /data/file.gz
Size: 12.4 GB | Blocks: 93 | Policy: SpdkDisk

Block 1 (blk_1074365040): 134 MB
  worker-a:50010  SSD
  worker-b:50010  DISK
  worker-c:50010  DISK

Block 2 (blk_1074365041): 134 MB
  worker-a:50010  SSD
  worker-b:50010  SSD
  worker-c:50010  DISK

...

Status: OK
```

### For directories (summary by default)

```
$ cv fsck /data/

Directory: /data/
Files: 1,204 | Blocks: 3,847

--- Storage Distribution ---
  SSD:  2,100 blocks (54.6%)
  DISK: 1,747 blocks (45.4%)

--- Per-Worker Distribution ---
  worker-a: 1,280 blocks (33.3%)
  worker-b: 1,284 blocks (33.4%)
  worker-c: 1,283 blocks (33.3%)

--- Policy Mismatch ---
  247 blocks on DISK but policy is SpdkDisk

Status: WARNING
```

### For directories with --detail flag

```
$ cv fsck /data/ --detail

/data/file1.gz (12.4 GB, 93 blocks)
  Block 1: worker-a SSD, worker-b DISK, worker-c DISK
  Block 2: worker-a SSD, worker-b SSD, worker-c DISK
  ...

/data/file2.gz (8.2 GB, 64 blocks)
  Block 1: worker-a SSD, worker-b SSD, worker-c SSD
  ...

[...]

PR #1692, opened 2026-09-14
fix(client): persist actual storage type for single-file writers

thuongle2210:
## Summary

- This change persists the **actual storage type** selected by workers when writing blocks, ensuring that the metadata layer accurately reflects where data is stored across the multi-tier cache hierarchy. Previously, the code only recorded the requested storage type from the client policy, not the actual type assigned by the worker during allocation.
- This resolve part of [issues-1190](https://github.com/CurvineIO/curvine/issues/1190)

## Changes

| File | Change | Purpose |
|------|--------|---------|
| `block_writer.rs` | Added `to_block_location()` helper; collect actual types from each writer; new `actual_storage_type()` method on `WriterAdapter` | Use per-writer actual type instead of uniform block policy type |
| `block_writer_local.rs` | Added `actual_storage_type: StorageType` field; capture from `write_context` during init; expose via accessor | Preserve the storage tier assigned during local write |
| `block_writer_remote.rs` | Added `actual_storage_type: StorageType` field; capture from `write_context` during init; expose via accessor | Preserve the storage tier assigned during remote write |
| `curvine-tests/testing.rs` | Added `TestingBuilder::mutate_worker_conf()` | Allow per-worker storage configuration in tests |
| `curvine-tests/tests/actual_storage_type_test.rs` | **New file** — two test scenarios | Verify single writers and multi-worker replicas both preserve actual storage types |

## Impacts

- **Correctness**: Block locations now reflect true data placement, not just policy intent
- **Replica placement**: Multi-replica scenarios correctly attribute each replica to its actual storage backend
- **Backward compatible**: Existing code paths unaffected; only metadata accuracy improves

## Test Coverage

1. **Single writer persistence**: Verifies local and remote writers independently preserve their actual storage type (Mem vs Disk fallback)
2. **Multi-worker replicas**: Confirms that when a 2-replica write spans workers with different storage configs (one Mem, one Disk), each replica's actual type is recorded correctly

## Future Works
- Extend to batch files writer
- Foundation for storage tier observability and fsck API validation


thuongle2210:
Hi @szbr00 , @bigbigxu , I made a change to persist&reflect the actual storage type for single-file writers. Could you pls help me review this MR.

crates/client/curvine-client-core/src/block/block_writer.rs
```diff
@@ -341,11 +348,26 @@ impl BlockWriter {
...
+        let worker = WorkerAddress {
+            worker_id: 7,
+            ..Default::default()
+        };
+        let actual_storage_type = StorageType::Disk;
+        let location = to_block_location(&worker, actual_storage_type);
+
+        assert_eq!(location.worker_id, 7);
+        assert_eq!(location.storage_type, StorageType::Disk);
+        assert_ne!(location.storage_type, StorageType::Mem);
```
szbr00, line 368: The `assert_ne!(location.storage_type, StorageType::Mem)` is redundant — the preceding `assert_eq!(..., StorageType::Disk)` already implies it. More broadly, `to_block_location` is a thin pass-through to `BlockLocation::new`, so this unit test largely restates the constructor; the real behavior (worker-reported actual type being persisted) is already covered by `actual_storage_type_test.rs`. Consider dropping the redundant assertion to keep the test focused.

curvine-tests/tests/actual_storage_type_test.rs
```diff
@@ -0,0 +1,132 @@
...
+    let locations = master_fs
+        .fs_dir
+        .read()
+        .get_block_locations(blocks.block_locs[0].block.id)?;
+    assert_eq!(locations.len(), 1);
+    assert_eq!(
+        locations[0].worker_id,
+        blocks.block_locs[0].locs[0].worker_id
+    );
+    assert_eq!(locations[0].storage_type, StorageType::Disk);
```
szbr00, line 130: Requesting `StorageType::Mem` here but asserting the persisted actual type is `StorageType::Disk` depends on the default single-worker test cluster having no Mem tier, so the worker silently falls back to Disk. That assumption is not obvious from the test and would break (or change meaning) if a Mem tier were ever added to the default testing config. Consider a short comment documenting the fallback assumption, or configuring the worker storage explicitly so the expected type is self-evident.

thuongle2210: Thank you, I added the explicit creation.

crates/client/curvine-client-core/src/block/block_writer.rs
thuongle2210: Ok, I removed the thin function and a related assertion.

szbr00 (commented):
Optional: The single-file path now commits worker-reported types, but `BatchBlockWriter::to_commit_blocks` still uses `CommitBlock::from(&LocatedBlock)`, which stamps `located_block.storage_type` (the file policy) onto every replica. `create_files_batch` therefore still writes the same temporary RocksDB mismatch this PR fixes for `FsWriter`, until the next block report. If that path should stay out of this PR, please file a follow-up with an owner; otherwise thread the per-writer `write_block` storage_type through the batch adapters the same way.

thuongle2210:
> Optional: The single-file path now commits worker-reported types, but `BatchBlockWriter::to_commit_blocks` still uses `CommitBlock::from(&LocatedBlock)`, which stamps `located_block.storage_type` (the file policy) onto every replica. `create_files_batch` therefore still writes the same temporary RocksDB mismatch this PR fixes for `FsWriter`, until the next block report. If that path should stay out of this PR, please file a follow-up with an owner; otherwise thread the per-writer `write_block` storage_type through the batch adapters the same way.

Ah,Batch writer will be handle at the follow-up PR as I mentioned at PR description.

szbr00 (approved):

PR #1694, opened 2026-09-14
fix(worker): flush running write frames

### LINKED ISSUE #1673: [BUG]: Worker WriteHandler ignores flush frames - client flush()/hflush does not flush buffered data

vishanth14:
## Summary

Fixes an issue where `WriteHandler` ignored flush-only `Running` frames.

When a `Running` frame contains `DataHeaderProto { flush: true }` with no data, the worker now calls `file.flush()` instead of only skipping the seek operation.

This ensures buffered data is flushed before the request returns successfully and keeps the single-block write path consistent with `BatchWriteHandler::flush_batch`.

## Changes

- Call `file.flush()?` when `header.flush` is `true`.
- Preserve the existing seek behavior for non-flush frames.
- Add a regression test covering a flush-only `Running` frame without a subsequent `Complete` frame.

## Testing

`cargo test -p curvine-server --test worker_test test_worker_running_flush_only_no_complete -- --nocapture`

Result: **1 passed, 0 failed**

Fixes #1673


vishanth14:
Hi @bigbigxu , I’ve implemented the fix and added a regression test for this issue. The test passes successfully. I’ve opened this PR for review. Thank you.

curvine-worker/src/worker/handler/write_handler.rs
```diff
@@ -226,23 +226,35 @@ impl WriteHandler {
...
+            if header.offset < 0 || header.offset >= context.block_size {
+                return err_box!(
+                    "Invalid seek offset: {}, block length: {}",
+                    header.offset,
+                    context.block_size
+                );
+            }
+            file.seek_to(header.offset)?;
+        } else {
+            file.flush()?;
```
bigbigxu, line 244: Required: `handle_data_header` runs before the payload write at lines 260–263, so a `Running` message containing both data and `flush=true` flushes only earlier writes. Its own payload is written afterward and the RPC still succeeds. `Message` permits header and data together, and `write_data` already uses that combination for seek frames. Record `need_flush`, write the payload, and then call `file.flush()?`; alternatively, explicitly reject data-bearing flush frames and document that protocol restriction. Please add an ordering regression test as well.

curvine-server/tests/worker_test.rs
```diff
@@ -427,6 +427,138 @@ fn test_worker_batch_remote_write_complete_and_read_back() -> CommonResult<()> {
...
+        .request(RequestStatus::Running)
+        .build();
+    let response = client.rpc_check(read_data_msg)?;
+    let data = response.data.as_slice();
+
+    // The data written in step 3 must be present even without a Complete frame.
+    // If the fix is absent, this assertion will fail because flush() was never
+    // called for the flush-only frame, and the in-buffer data may not be
+    // persisted to the on-disk store.
+    assert!(
```
bigbigxu, line 534: Required: This read-back does not observe whether the flush frame actually invoked `file.flush()`. The configured Disk layout uses `LocalFile`, whose `write_region` writes directly through `std::fs::File`; another handle can read those bytes before `File::flush`. The worker unit test also calls `handle_data_header` directly, so it would still pass if `WriteHandler::write` stopped invoking that helper. Please create an observable/failing `BlockIO` and exercise the `WriteHandler::write` entry path, or use a genuinely buffered/device-backed integration path, so the assertion fails when the flush dispatch is removed.

bigbigxu (changes requested):
Required: The verification section only records the `curvine-server` integration test, while the current head also adds `flush_header_propagates_flush_error` under `curvine-worker`, and no CI result is currently reported for this head. Please run and record the targeted `curvine-worker` test together with the strengthened real-entry regression test.

curvine-server/tests/worker_test.rs
```diff
@@ -427,6 +427,138 @@ fn test_worker_batch_remote_write_complete_and_read_back() -> CommonResult<()> {
...
+        !data.is_empty(),
+        "flush-only Running frame did not persist data; data likely lost because file.flush() was not called"
+    );
+    assert_eq!(&data[..write_data.len()],write_data,"flushed data content mismatch");
+
+    Ok(())
+}
+
+fn start_worker_with_tempdir() -> ClusterConf {
```
szbr00, line 543: `start_worker_with_tempdir` creates a real directory under the system temp dir but never removes it, so each run leaves a `curvine-test-worker-*` directory behind. `let _ = create_dir_all(...)` also swallows the error, so a setup failure surfaces later as a confusing worker-side error rather than at the cause. The existing `start_worker()` uses the in-memory `[MEM:10MB]` layout and has no such leak — consider reusing that pattern here, or registering cleanup, and propagating the `create_dir_all` error.

curvine-server/tests/worker_test.rs
```diff
@@ -427,6 +427,138 @@ fn test_worker_batch_remote_write_complete_and_read_back() -> CommonResult<()> {
     Ok(())
 }
 
+#[test]
+fn test_worker_running_flush_only_no_complete() -> CommonResult<()> {
+    // Regression test for GH#1673.
```
szbr00, line 432: This header comment block runs ~15 lines and the numbered "Strategy" section mostly restates the code immediately below it. Per the project comment guidance, please trim it to a few lines that capture the why (the GH#1673 regression: a flush-only `Running` frame must reach `file.flush()`) and drop the step-by-step narration.

curvine-server/tests/worker_test.rs
```diff
@@ -427,6 +427,138 @@ fn test_worker_batch_remote_write_complete_and_read_back() -> CommonResult<()> {
...
+
+    // The data written in step 3 must be present even without a Complete frame.
+    // If the fix is absent, this assertion will fail because flush() was never
+    // called for the flush-only frame, and the in-buffer data may not be
+    // persisted to the on-disk store.
+    assert!(
+        !data.is_empty(),
+        "flush-only Running frame did not persist data; data likely lost because file.flush() was not called"
+    );
+    assert_eq!(&data[..write_data.len()],write_data,"flushed data content mismatch");
```
szbr00, line 538: `&data[..write_data.len()]` will panic with an out-of-bounds slice index if `data` is non-empty but shorter than `write_data.len()`; the preceding `!data.is_empty()` guard does not ensure sufficient length. Consider asserting `data.len() >= write_data.len()` first (or comparing `&data[..]` to the expected slice) so a short read produces a clear assertion message instead of a panic.

vishanth14:
Thanks for the review @szbr00 @bigbigxu, I’ve addressed all of the requested changes in commit `2c63dc61`.

- Fixed the `flush=true` handling so that data-bearing `Running` frames write the payload first and then call `file.flush()`.
- Added an ordering regression test to verify the `write → flush` sequence.
- Added observable and failing `BlockIO` tests that exercise the `WriteHandler::write` entry path and verify flush dispatch and error propagation.
- Updated the worker integration regression test.
- Fixed `start_worker_with_tempdir` to propagate directory-creation failures and clean up the temporary directory.
- Shortened the GH#1673 regression-test comment as requested.
- Replaced the potentially panicking `&data[..write_data.len()]` assertion with a safe `data.get(..write_data.len())` comparison.

### Verification

- `cargo fmt --check` — passed
- `git diff --check` — passed
- Targeted worker tests were attempted, but could not execute because the build is blocked during `librocksdb-sys v0.16.0+8.10.0` compilation with GCC 16.2.1, before the tests could run.

szbr00:
Thank you for your PR. We can discuss any issue about Curvine in the Slack channel.The invitation link is on README page, you are very welcome to join the Curvine Slack.

vishanth14:
Thanks @szbr00 , I’ve joined the Curvine Slack community. Looking forward to contributing more!

bigbigxu (approved):

PR #1696, opened 2026-09-15
refactor(fuse): unify FuseConf timeouts as DurationUnit (#1695)

### LINKED ISSUE #1695: [SUB] Unify FuseConf timeouts as DurationUnit
Parent: #1023

`FuseConf` still mixes timeout representations: kernel cache fields used to be millisecond integers (`entry_timeout_ms` / `attr_timeout_ms` / `negative_timeout_ms`), while userspace cache TTLs were duration strings (`meta_cache_timeout` / `node_cache_timeout`) parsed in `init()`. That is the remaining slice of #1023 §3 after earlier `*_ms` migrations.

Store every FuseConf timeout as `DurationUnit`. Canonical TOML/CLI names drop the `_ms` suffix (`entry_timeout`, `attr_timeout`, `negative_timeout`). Keep `*_ms` serde and clap aliases so existing configs and mount scripts keep working. `init()` copies `as_duration()` into the existing kernel/userspace `Duration` TTL fields.

- Keep `*_ms` as the canonical names: rejected; `DurationUnit` is unit-aware, so the suffix is misleading.
- Collapse skip-serialized `Duration` twins in this same change: deferred; FUSE reply paths already consume `Duration` (`as_secs` / `subsec_nanos` / `is_zero`).

**Additional context**
Bare integers remain milliseconds (`DurationUnit` default). Duration strings such as `"1s"` / `"500ms"` / `"1h"` are accepted. `std::str::FromStr` is implemented on `DurationUnit` so clap can parse CLI overrides.

# Goal

| Typed timeout fields | One parser for TOML integers, duration strings, and CLI flags |
| Drop `_ms` from canonical names | Names match `DurationUnit`; no false “milliseconds only” signal |
| Keep `*_ms` aliases | Existing TOML, Fluid options, and `--entry-timeout-ms` keep working |

# Not Goal

- Removing skip-serialized `attr_ttl` / `entry_ttl` / `negative_ttl` / `node_cache_ttl` / `meta_cache_ttl` in this change
- Changing default timeout values
- The rest of #1023 (dead params, `tasks_per_mnt` already landed, defaults review, `io_threads` rename)

# Test Plan

| `cargo test -p curvine-config fuse_conf` | Canonical DurationUnit TOML + `*_ms` aliases + init TTL copy | curvine-config |
| `cargo test -p curvine-runtime --lib common::duration_unit` | `FromStr` / serde duration parsing | curvine-runtime |
| Fluid `test_config_parse.py` mapping | `fuse.entry_timeout_ms` renders canonical `entry_timeout` | curvine-docker/fluid |

bigbigxu:
# Summary

Unify `FuseConf` kernel and userspace cache timeouts as `DurationUnit`, rename canonical keys off the `_ms` suffix, and keep `*_ms` serde/CLI aliases so existing TOML and mount scripts still load.

# Issue Describe / Design

Closes #1695
Related to #1023 §3 (collapse TTL representation onto `DurationUnit`)

Design decisions:
- Config fields are `DurationUnit`; skip-serialized `Duration` TTL fields remain for FUSE reply paths (`as_secs` / `subsec_nanos` / `is_zero`).
- Canonical names: `entry_timeout`, `attr_timeout`, `negative_timeout`, plus existing `meta_cache_timeout` / `node_cache_timeout`.
- Bare integers stay milliseconds; strings such as `1s` / `500ms` are accepted.
- `DurationUnit` implements `std::str::FromStr` so clap can parse CLI overrides.
- `ClientCliArgs` accepts `DurationUnit`; clap long names are set explicitly because serde aliases would otherwise become the primary flag names.

# Changes

| `crates/common/curvine-config/src/fuse_conf.rs` | Timeout fields are `DurationUnit`; `*_ms` serde/CLI aliases; CLI alias tests | Existing `*_ms` TOML still works; new canonical keys accepted |
| `crates/core/runtime/src/common/duration_unit.rs` | Implement `FromStr` | Enables clap parsing of duration strings |
| `crates/common/curvine-macros/src/client_cli.rs` | Support `DurationUnit` CLI fields | Generated fuse CLI overrides parse durations |
| `etc/curvine-cluster.toml`, docker example conf | Document canonical duration keys | Docs/examples only |
| `curvine-docker/fluid/*` | Map `*_ms` options to canonical names; cover legacy CLI keys | Fluid still accepts old option keys |
| `curvine-fuse/src/cli/mount_args.rs` | Assert canonical and legacy timeout CLI overrides plus TTLs | Test-only |
| `curvine-fuse/src/fuse_utils.rs` | Comment update | None |

# Test verified

| `cargo test -p curvine-runtime --lib common::duration_unit` | PASS | Includes `FromStr` unit test |
| `cargo test -p curvine-common-macros --test derive_client_cli` | PASS | Existing CLI derive coverage |
| `cargo test -p curvine-config fuse_conf` | PASS | Includes canonical/legacy timeout CLI alias + TTL assertions |
| `python curvine-docker/fluid/tests/test_config_parse.py` | PASS | Includes legacy `*-timeout-ms` mapping to canonical flags |
| `cargo test -p curvine-fuse --lib cli::mount_args` | N/A on Windows | Mount-path alias test added; fuse lib tests require Unix |

# Dependencies

- Related PRs: None
- Related issues: #1695, #1023
- Blocks / blocked by: None


crates/core/runtime/src/common/duration_unit.rs
```diff
@@ -106,6 +107,14 @@ impl DurationUnit {
     }
 }
 
+impl FromStr for DurationUnit {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
szbr00, line 113: Required: The new clap `FromStr` path accepts unitless decimals and negative strings, so `--entry-timeout 1.0` becomes 1ms and `--entry-timeout=-1s` becomes 0ms. Inherent `DurationUnit::from_str` takes the empty-unit float branch for `"1.0"` (`1.0 * MILLISECONDS`) and casts negatives to `u64` (`"-1s"` saturates to 0). CSI still documents and tests that value (`curvine-csi/examples/storage-class.yaml:44`, `fuse_cli_test.go` passthrough `--entry-timeout 1.0`); on current main that flag is unknown and fails loudly, but after this rename `validate-config`/mount will succeed with a 1ms kernel dentry TTL instead of the old 1s meaning. Reject unitless fractional and negative strings before the `f64` cast (numeric TOML already rejects `visit_i64 < 0`), and update the CSI example/fixtures to an explicit `1s` or `1000`.

crates/common/curvine-config/src/fuse_conf.rs
```diff
@@ -156,25 +159,30 @@ pub struct FuseConf {
...
-    #[client_cli]
-    pub negative_timeout_ms: u64,
-
-    // Cache time for file and directory attributes, in milliseconds.
-    // This means that after a file or directory attribute search, if the same attribute is requested again, the kernel will first check the cache.
-    // If the record in the cache is still valid (i.e. the timeout time has not exceeded), the cached result will be returned directly without making a request to the user space again
+    // `entry_timeout_ms` / `--entry-timeout-ms` kept so existing TOML and mount
+    // scripts do not silently fall back to the default.
+    #[serde(alias = "entry_timeout_ms")]
+    #[client_cli(long = "entry-timeout", alias = "entry-timeout-ms")]
```
szbr00, line 169: Required: The compatibility contract is that `--entry-timeout-ms` / `--attr-timeout-ms` / `--negative-timeout-ms` still apply the same `DurationUnit` (and derived TTL) as the canonical flags, but no test goes through `FuseMountArgs::try_parse_from`. `toml_legacy_timeout_ms_alias_preserved` only covers serde; `derive_client_cli` has no `#[client_cli(alias = ...)]` case; `mount_args` only asserts `--meta-cache-ttl`. A regression in the proc-macro clap alias wiring would still pass this suite. Add canonical and legacy CLI cases and assert the applied `DurationUnit` plus `entry_ttl` / `attr_ttl` / `negative_ttl` after `get_conf()`.

curvine-docker/fluid/config-parse.py
```diff
@@ -68,12 +68,12 @@ class ConfigParser:
...
-        ('entry-timeout', 'entry-timeout-ms'),
-        ('attr-timeout-ms', 'attr-timeout-ms'),
-        ('attr-timeout', 'attr-timeout-ms'),
-        ('negative-timeout-ms', 'negative-timeout-ms'),
-        ('negative-timeout', 'negative-timeout-ms'),
+        ('entry-timeout-ms', 'entry-timeout'),
+        ('entry-timeout', 'entry-timeout'),
+        ('attr-timeout-ms', 'attr-timeout'),
+        ('attr-timeout', 'attr-timeout'),
+        ('negative-timeout-ms', 'negative-timeout'),
```
szbr00, line 75: Optional: The Fluid mapping now rewrites both `entry-timeout-ms` and `entry-timeout` to `--entry-timeout`, but `test_advanced_client_and_fuse_toml_options_are_rendered` only supplies the canonical option key. A broken `entry-timeout-ms` (or attr/negative) branch would still pass. Add a case for the legacy CLI spellings and assert the generated script emits the canonical flags.

szbr00 (changes requested):
Required: FuseConf DurationUnit unification is sound for TOML aliases and defaults, but the new clap FromStr path silently mis-parses unitless decimals and negatives, and the *_ms CLI alias contract is not covered on the mount path. Inline comments have the concrete fixes.

Inspected: curvine-config FuseConf, curvine-macros ClientCliArgs (DurationUnit + alias), DurationUnit::from_str, FuseMountArgs get_conf, Fluid mappings, CSI passthrough (entry-timeout=1.0).

crates/common/curvine-config/src/fuse_conf.rs
bigbigxu: Thanks. Fixed in 7be33c678: added FuseConfCliOverrides and FuseMountArgs::try_parse_from cases for `--entry-timeout` / `--attr-timeout` / `--negative-timeout` and the `*-ms` aliases, asserting DurationUnit plus entry_ttl / attr_ttl / negative_ttl after apply/init and get_conf().

crates/core/runtime/src/common/duration_unit.rs
bigbigxu: Thanks for the suggestion. Kept the existing DurationUnit parser in this PR because it is shared config infrastructure, not specific to the FuseConf rename. Unitless integers remain milliseconds, and strings with units such as `1s` / `500ms` stay valid. Tightening unitless fractions or negatives would be a separate DurationUnit change.

curvine-docker/fluid/config-parse.py
bigbigxu: Thanks. Fixed in 7be33c678: added test_legacy_fuse_timeout_cli_options_emit_canonical_flags so entry/attr/negative `*-timeout-ms` options render the canonical CLI flags.

szbr00 (approved):

PR #1700, opened 2026-09-16
fix(master): reserve scheduled_bytes during worker placement (#1699)

### LINKED ISSUE #1699: [BUG]: Master concurrent block placement oversubscribes heartbeat available space
**Describe the bug**
Master worker placement uses the last heartbeat `available` value. Concurrent `add_block` / `choose_worker` calls on the same node all see that stale remaining space, so a worker with 1 GiB free and 128 MiB blocks can be chosen far more than 8 times before the next heartbeat. Heterogeneous disks make this worse: the worker-level sum can look large enough while no single directory can hold the block.

**To Reproduce**
1. Run a Master with one Live worker reporting about 1 GiB `available`.
2. Issue many concurrent writes (or `add_block` requests) whose block size is 128 MiB, before the worker heartbeat updates `available`.
3. Observe that far more than `available / block_size` blocks are placed on that worker.

**Expected behavior**
Between heartbeats, Master should reserve in-flight allocations (`scheduled_bytes`) so later chooses see remaining space as `available - scheduled_bytes`. Placement must require at least one non-failed directory that still fits the block after subtracting that reservation. Heartbeats should reclaim reservation from the drop in `available`, and abandoned reservations (client crash / timeout) should expire after `worker_lost_interval`.

 - OS: n/a (Master placement logic)
 - Curvine: `open/main`

**Additional context**
HDFS uses a similar `scheduledSize` on the datanode descriptor. This issue is Master block placement only: Transfer scheduler and worker-local directory policy are out of scope. Master-assigned `storage_id` is also out of scope.

# Goal

| Worker-level `scheduled_bytes` reserved under the worker-manager write lock on `choose_worker` | Concurrent chooses cannot oversubscribe heartbeat `available` |
| Per-directory check: each dir deducts the worker-wide reservation | Sum of disks looks large enough while no single disk fits the block |
| Reclaim by heartbeat `available` delta; never go below 0 | Reservations drop once writes show up in heartbeat space |
| Expire leftover `scheduled_bytes` after `worker_lost_interval` | Client crash / timeout does not leave the node permanently false-full |
| Prometheus gauge `scheduled_bytes` | Operators can see in-flight reservation |
| Unschedule on `acquire_new_block` / `assign_worker` metadata failure | Failed metadata updates do not leak reservation |

# Not Goal

- Transfer `scheduler.rs` admission / byte budget
- Worker-local dir choosing (`RobinChoosingPolicy`)
- Master assigning a specific `storage_id` on the wire
- HDFS dual-window (`curr`/`prev`) aging
- Decrementing reservation from `blockReceived` (Curvine reclaims via heartbeat `available` delta)

# Test Plan

| 1 GiB available / 128 MiB blocks stops after 8 chooses | Concurrent over-allocation | `curvine-master` worker_manager |
| Heartbeat with unchanged `available` keeps reservation | In-flight writes stay reserved | worker_map, worker_manager |
| Heartbeat `available` drop reclaims reservation | Successful writes free scheduled bytes | WorkerInfo, worker_manager |
| Reservation never goes negative | Clamp after over-reclaim | WorkerInfo |
| Timeout expire after `worker_lost_interval` | False-full recovery | WorkerInfo, worker_manager |
| Per-dir check with worker-wide deduct | Heterogeneous disks | WorkerInfo, policies |
| `can_allocate` rejects non-Live workers | Blacklist / decommission | WorkerInfo, policies |

bigbigxu:
# Summary

Prevent Master from placing more blocks on a worker than heartbeat `available` can hold when many `choose_worker` calls race before the next heartbeat. In-flight allocations are reserved as `scheduled_bytes` and reclaimed from the `available` delta (or expired after `worker_lost_interval`).

# Issue Describe / Design

Closes #1699

- Root cause: placement used stale heartbeat `available`; concurrent chooses all saw the same remaining space.
- Reproduction: 1 GiB free, 128 MiB blocks, many concurrent `add_block` calls before heartbeat — far more than 8 placements.
- Fix approach:
  - Reserve `block_size` under the worker-manager write lock after `choose_worker`.
  - Eligibility is Live plus at least one non-failed directory with `available - scheduled_bytes >= block_size`.
  - Heartbeat copies reservation and subtracts `max(0, old_available - new_available)`, clamped at 0.
  - Leftover reservation expires after `worker_lost_interval` to avoid permanent false-full on client crash.
  - Unschedule if `acquire_new_block` / `assign_worker` metadata fails.
  - Cluster gauge `scheduled_bytes` on Master scrape.

Out of scope: Transfer scheduler, worker-local dir policy, Master-assigned `storage_id`, HDFS dual-window aging.

# Changes

| `crates/common/curvine-model/src/worker_info.rs` | `scheduled_bytes`, `can_allocate`, reclaim, expire | Placement remaining space subtracts in-flight reservation |
| `crates/common/curvine-model/src/proto_utils.rs` | Decode `scheduled_bytes` as 0 | Master-only field, not on the wire |
| `curvine-master/src/master/fs/state/worker_map.rs` | Copy reservation across heartbeat; reclaim by available delta | Heartbeat no longer drops in-flight reservation |
| `curvine-master/src/master/fs/worker_manager.rs` | `choose_worker` schedules; expire on choose/heartbeat | Concurrent chooses stop after remaining space is reserved |
| `curvine-master/src/master/fs/policy/*.rs` | Use `can_allocate` / `allocatable_available` | Policies skip workers that cannot fit the block |
| `curvine-master/src/master/fs/master_filesystem.rs` | Write lock on choose; unschedule on metadata failure; allocatable uses reservation | Failed add_block does not leak reservation |
| `curvine-master/src/master/job/job_runner.rs` | Write lock around `choose_worker` | Load job pick uses the same reservation path |
| `curvine-master/src/master/master_metrics.rs` | Gauge `scheduled_bytes` | New Prometheus series |

# Test verified

| `cargo test -p curvine-model --lib` | PASS | 47 tests, including reclaim/expire/per-dir |
| `cargo test -p curvine-master --lib` | PASS | 121 passed, 1 ignored (`bench_create_tree_large`) |
| `make format` | SKIP | `make` not available on this Windows host; ran `cargo fmt --all` |

# Dependencies

- Related PRs: None
- Related issues: #1699
- Blocks / blocked by: None


crates/common/curvine-model/src/worker_info.rs
```diff
@@ -134,6 +140,61 @@ impl WorkerInfo {
...
+    pub fn allocatable_available(&self) -> i64 {
+        self.available.saturating_sub(self.scheduled_bytes)
+    }
+
+    pub fn can_allocate(&self, block_size: i64) -> bool {
+        if !self.is_live() {
+            return false;
+        }
+        if self.storage_map.is_empty() {
+            return self.allocatable_available() >= block_size;
```
szbr00, line 157: Critical: can_allocate(block_size == 0) treats remaining space >= 0 as eligible, so a Live worker with available == scheduled_bytes == 0 is accepted. WorkerPolicy::choose_workers builds ChooseContext with block_size = 0, and MasterReplicationManager::assign uses that path. Default policy is local → robin; previously robin required available > 0, so replication can now pick a completely full target and fail when the replica is written. Keep the no-block path requiring positive allocatable space (and at least one non-failed dir with remaining > 0). The new test allocatable_available_subtracts_scheduled_bytes currently asserts can_allocate(0) on a zero-remaining worker and would hide this regression.

curvine-master/src/master/job/job_runner.rs
```diff
@@ -81,7 +81,7 @@ impl LoadJobRunner {
 
     pub fn choose_worker(&self, block_size: i64) -> FsResult<WorkerAddress> {
         let ctx = ChooseContext::with_num(1, block_size, vec![]);
-        let worker_mgr = self.master_fs.worker_manager.read();
+        let mut worker_mgr = self.master_fs.worker_manager.write();
```
szbr00: Critical: choose_worker now always increments scheduled_bytes, but LoadJobRunner uses it to pick an executor, not to place a block. create_all_tasks calls this for every file before branching on CV→UFS vs UFS→CV, and several exits never unschedule: unsupported path combo, job_max_files, install_plan_if_current returning false, and the race-lost Occupied+running return. Export tasks never write to the chosen worker, so heartbeat available does not reclaim the reservation. Even a successful UFS→CV load reserves job.block_size per file (often 128 MiB) rather than status.len, so a directory of small files can false-fill the node until worker_lost_interval (default 10m) and block add_block. Use a non-reserving pick (choose_workers) for executor selection; if load-write pressure must be counted, reserve only UFS→CV with a drop guard and the file size.

curvine-master/src/master/fs/master_filesystem.rs
```diff
@@ -1735,13 +1744,22 @@ impl MasterFilesystem {
...
-            self.choose_worker_for_file(inode.as_file_ref()?, client_addr, exclude_workers)?;
+        let file = inode.as_file_ref()?;
+        let block_size = file.block_size as i64;
+        let choose_workers = self.choose_worker_for_file(file, client_addr, exclude_workers)?;
         let has_spdk = {
             let wm = self.worker_manager.read();
             wm.workers_have_spdk(&choose_workers)
         };
-        let block = fs_dir.assign_worker_inode(path, &mut inode, block.id, &choose_workers)?;
+        let block = match fs_dir.assign_worker_inode(path, &mut inode, block.id, &choose_workers) {
```
szbr00, line 1754: Required: unschedule runs only when assign_worker_inode returns Err. BlockMeta::assign_worker is a no-op (returns false) when locs are already set or alloc_opts is absent, and FsDir still returns Ok. An idempotent retry therefore keeps the reservation created by choose_worker_for_file until expiry, even though no new write will consume heartbeat available. Have the metadata helper report whether it mutated the block, and unschedule the successful no-op path (add a test that retries assign_worker and checks scheduled_bytes == 0).

crates/common/curvine-model/src/worker_info.rs
```diff
@@ -60,6 +60,10 @@ pub struct WorkerInfo {
     pub startup_time_ms: u64,
     pub capacity: i64,
     pub available: i64,
+    #[serde(default)]
+    pub scheduled_bytes: i64,
+    #[serde(default)]
+    pub scheduled_since_ms: u64,
```
szbr00: Required: scheduled_bytes and scheduled_since_ms are inserted between existing public WorkerInfo fields instead of being appended, and WorkerInfo is Serialize-d. /api/workers returns Vec<WorkerInfo> directly, so these Master-only values (zeroed in proto_utils, absent from WorkerInfoProto) become part of the HTTP JSON contract. Append them after component_info and skip serde (or keep the reservation map on WorkerManager) so the public model stays append-only and the REST/proto worker views stay aligned.

szbr00 (changes requested):
Inspected `curvine-model` WorkerInfo / proto_utils and `curvine-master` worker_manager, worker_map, policies, master_filesystem, job_runner, and metrics against live head 4955623.

The add-block reservation and heartbeat reclaim path matches #1699. Requesting changes because the reserving `choose_worker` is also used for replication (`block_size = 0`) and load/export executor pick, which can select a full replica target or leave workers false-full until `worker_lost_interval` (default 10m). See inline comments.

szbr00 (approved):

PR #1703, opened 2026-09-17
perf(master): skip inode lookup for non-mount complete_file

bigbigxu:
# Summary

`complete_file` now checks the fs_mode mount on the journal path first, so non-mount completes skip inode path resolution. Mounted files still rebuild the live path after open-then-rename and reuse the same mount for export. Cross-mount rename is not allowed, so mount selection does not need a second lookup on the live path.

# Issue Describe / Design

No linked issue.

Previously every complete journal entry resolved the inode path before `get_mnt`. Most completes are not under a UFS mount, so that lookup was wasted. This change:

- Early-return when `e.path` is not a fs_mode mount
- Keep inode-id path rebuild for rename-after-open within the same mount
- Call `get_mnt` once and reuse that mount for `submit_export_task`

# Changes

| `curvine-master/src/master/journal/ufs_loader.rs` | Check mount on journal path before inode lookup; single `get_mnt` | Non-mount completes skip inode resolution. Mounted files still export to the live inode path. |
| `curvine-server/tests/journal_test.rs` | Cover non-mount skip and in-mount rename-after-open | No behavior change; asserts skip and live-path export. |

# Test verified

| `cargo fmt -- curvine-server/tests/journal_test.rs` | PASS | `make format` unavailable on this Windows host |
| `cargo test -p curvine-server --test journal_test test_ufs_loader_complete_file` | PASS | `test_ufs_loader_complete_file_skips_non_mount`, `test_ufs_loader_complete_file_rename_within_mount_exports_live_path` |
| `cargo test -p curvine-master` | PASS | |

# Dependencies

- Related PRs: None
- Related issues: None
- Blocks / blocked by: None


curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -146,6 +146,11 @@ impl UfsLoader {
             return Ok(());
         }
 
+        let entry_path = Path::from_str(&e.path)?;
+        let Some((_, mnt)) = self.get_mnt(&entry_path)? else {
```
szbr00, line 150: Required: `get_mnt` is now decided from `e.path` (the complete RPC / journal path) and that `mnt` is reused for `submit_export_task(&path, &mnt)` after the inode rebuild. That is not equivalent to #1686, which selected the mount from the live path.

Two failures:

1. Skip: open `/tmp/f`, rename into `/mnt/f`, complete with the original path. `get_mnt(&entry_path)` is None, so this returns without exporting. `UfsLoader::rename` also no-ops when src is not under a mount, so this was the only export chance. fs_mode durability is lost.

2. Wrong UFS target: open `/mnt/f`, rename to `/other/f` (or another mount), complete with the original path. Early `get_mnt` hits the old mount, then `MountInfo::get_ufs_path` does `path.replacen(&self.cv_path, "", 1)` and writes under a bogus UFS location instead of skipping.

Keep the early `get_mnt(&entry_path)` as the inode-walk skip (saves `fs_dir.read()` plus O(depth) `get_inode` on the non-mount complete path). After `path` is rebuilt, call `get_mnt(&path)` again and export only with that mount. That extra lookup runs only on the already-mount-hit path.

szbr00 (dismissed):
Required: There is no test whose assertion fails if this skip or mount-reuse is wrong. `cargo test -p curvine-master` does not cover `UfsLoader::complete_file` (the existing UfsLoader case is `curvine-server/tests/journal_test.rs::test_ufs_loader_mkdir_recreates_missing_ufs_parent`). Add a test in that file that (1) completes a non-mount file and asserts no UFS object is created, and (2) creates under an fs_mode mount, renames within the mount, completes with the original journal path, and asserts the export lands at the live UFS path. Run that crate, not only `curvine-master`.

curvine-master/src/master/journal/ufs_loader.rs
```diff
@@ -146,6 +146,11 @@ impl UfsLoader {
...
+            return Ok(());
+        };
```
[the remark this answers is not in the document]
bigbigxu: Thanks. Declined: cross-mount rename is not allowed, so journal path and the inode-rebuilt live path stay on the same fs_mode mount. Early `get_mnt(&entry_path)` plus reusing that mount for `submit_export_task(&live_path, &mnt)` matches #1686 for legal rename-after-open. We still rebuild the live path so the export target is the post-rename UFS object.

curvine-master/src/master/journal/ufs_loader.rs
bigbigxu: Thanks. Declined the second `get_mnt(&path)` and the two failure cases: they assume rename across mount points (including non-mount into a mount), which is not allowed. Keeping the early skip on `e.path` and reusing that mount is equivalent for in-mount rename. Tests cover the legal cases: non-mount complete does not export, and create+rename within one fs_mode mount still exports to the live UFS path (`5002e0b7c`).

bigbigxu:
Thanks. Added coverage in `curvine-server/tests/journal_test.rs` (`5002e0b7c`):

1. `test_ufs_loader_complete_file_skips_non_mount` — complete a file outside the fs_mode mount and assert no UFS object / no export job.
2. `test_ufs_loader_complete_file_rename_within_mount_exports_live_path` — create under the mount, rename within the same mount, complete with the original journal path, and assert export source/target use the live path.

Ran `cargo test -p curvine-server --test journal_test test_ufs_loader_complete_file` (PASS). Cross-mount rename is out of contract, so those cases are not in the test.

lzjqsdd (approved):

